PR #99, opened 2021-06-14
Install Script

### LINKED ISSUE #80: Create install.sh
Create an install.sh script in the root of the repo for installing Microshift. The script shall perform the necessary installation and configuration of dependencies (from RPMs) on the host system, then install and (by default) start Microshift binary (not container on Podman).
Allow to install as systemd unit. Stretch goal: Allow running as container on podman managed by systemd (as generated by `podman generate systemd`).

oglok:
fix #80 
In order to fully test this PR, we'd need to:

- Open up the repo to the public so binaries (and checksums) can be downloaded from Github
- Agree on a common location for all OSes, or put conditions for OStree based OSes.


install.sh
```diff
@@ -0,0 +1,129 @@
...
+
+# If RHEL, use subscription manager to register
+register_subs() {
+    subscription-manager register --auto-attach
+    subscription-manager repos --enable=rhocp-4.7-for-rhel-8-x86_64-rpms
+}
+
+# Apply SElinux policies (Permissive mode at the moment)
+apply_selinux_policy() {
+    setenforce 0
```
rootfs, line 30: https://github.com/redhat-et/microshift/issues/85#issuecomment-860981460

oglok: Thanks @rootfs , I've updated the install script with these instructions.

install.sh
```diff
@@ -0,0 +1,129 @@
...
+
+# Start CRI-O
+verify_crio() {
+    systemctl enable crio
+    systemctl start crio
+
+}
+
+# Download and install kubectl
+get_kubectl() {
```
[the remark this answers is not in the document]
oglok: For the moment, I'd prefer to keep it just with kubectl, and see what's the best way to download the oc client according to the microshift version in the next patches.

install.sh
```diff
@@ -0,0 +1,129 @@
...
+    fi
+}
+
+
+# CRI-O config to match Microshift networking values
+crio_conf() {
+    sed -i 's/10.85.0.0\/16/10.42.0.0\/24/' /etc/cni/net.d/100-crio-bridge.conf
+
+     if [ "$DISTRO" == "rhel" ]; then
+        sed -i 's/\/usr\/libexec\/crio\/conmon/\/usr\/bin\/conmon/' /etc/crio/crio.conf
```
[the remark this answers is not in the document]
oglok: I'd like to keep it this way for just one line per file, especially when one of them it's just meant for RHEL.

install.sh
```diff
@@ -0,0 +1,147 @@
...
+
+# If RHEL, use subscription manager to register
+register_subs() {
+    sudo subscription-manager register --auto-attach
+    sudo subscription-manager repos --enable=rhocp-4.7-for-rhel-8-x86_64-rpms
+}
+
+# Apply SElinux policies
+apply_selinux_policy() {
+    sudo semanage fcontext -a -t container_runtime_exec_t /usr/local/bin
```
rootfs, line 32: `/usr/local/bin/microshift`

install.sh
```diff
@@ -0,0 +1,147 @@
...
+        sudo curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/devel:kubic:libcontainers:stable.repo
+        sudo curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable:cri-o:$CRIOVERSION.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable:cri-o:$CRIOVERSION/$OS/devel:kubic:libcontainers:stable:cri-o:$CRIOVERSION.repo
+        sudo dnf install -y cri-o cri-tools
+    fi
+}
+
+
+# CRI-O config to match Microshift networking values
+crio_conf() {
+    sudo sed -i 's/10.85.0.0\/16/10.42.0.0\/24/' /etc/cni/net.d/100-crio-bridge.conf
```
rootfs, line 65: this doesn't seem necessary on my (fresh fc 34) setup

install.sh
```diff
@@ -0,0 +1,147 @@
...
+
+# Usage:
+# ENV_VAR= ./install.sh
+#
+# Environment variables:
+#
+# - ROLES="controlplane,node"
+#   Option by default, but controlplane role can be run standalone.
+ROLES="controlplane,node"
+VERSION=v0.2
```
rootfs, line 12: need a new build :D

install.sh
```diff
@@ -0,0 +1,147 @@
...
+EOF
+
+    sudo systemctl enable microshift.service --now
+
+}
+
+# Locate kubeadmin configuration to default kubeconfig location
+prepare_kubeconfig() {
+    mkdir $HOME/.kube
+    sudo cp /var/lib/microshift/resources/kubeadmin/kubeconfig $HOME/.kube/config
```
copejon, line 123: Overwriting the user's kubeconfig would have the side effect of breaking their access to other clusters/contexts they would have had.  We should merge the configs instead.  From SO:
```
export KUBECONFIG=~/.kube/config:~/someotherconfig 
kubectl config view --flatten
```
You can then pipe that out to a new file if needed.

install.sh
```diff
@@ -0,0 +1,147 @@
...
+    sudo chown $USER:$USER $HOME/.kube/config
+}
+
+
+# Script execution
+get_distro
```
copejon, line 129: the script should end or throw a warning to stderr for unsupported distros

install.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/sh
+set -e
```
copejon, line 2: enabling `pipefail` would allow us to catch errors within pipes, which may otherwise produce garbage output and make debugging more difficult.

PR #108, opened 2021-06-15
BUG: resolve directory already exists failure

usrbinkat:
when running install.sh more than once to get a dev environment going, I encountered the following error causing the script to abort.

Error:
```sh
mkdir: cannot create directory ‘/var/lib/kubelet/’: File exists
```


rootfs:
thank you!

PR #111, opened 2021-06-16
Add basic README.md

fzdarsky:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Microshift
-Microshift (aka "µShift") is OpenShift<sup>1</sup> Kubernetes in a small form factor and optimized for edge computing.
+Microshift is OpenShift<sup>1</sup> Kubernetes in a small form factor and optimized for edge computing.
 
 Edge devices deployed out in the field pose very different operational, environmental, and business challenges from those of cloud computing. These motivate different engineering trade-offs for Kubernetes at the far edge than for cloud or near-edge scenarios. Microshift's design goals cater to this:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,57 @@
...
+
+## Using Microshift
+To give Microshift a try, simply install a recent test version (we don't provide stable releases yet) using:
+```
+curl -sfL https://raw.githubusercontent.com/redhat-et/microshift/main/install.sh | sh -
+```
+
+This will install Microshift's dependencies (CRI-O), install it as a systemd service and start it.
+
+For convenience, the script copies the kubeconfig to $HOME/.kube/config, too, so after installation you should be able to access your cluster directly:
```
copejon, line 26: Per my comment in https://github.com/redhat-et/microshift/pull/99#discussion_r652952649,  we shouldn't obliterate user's kubeconfigs.

README.md
```diff
@@ -1 +1,57 @@
...
+* build on and integrate cleanly with edge-optimized OSes like Fedora IoT and RHEL for Edge, while
+* providing a consistent development and management experience with standard OpenShift.
+
+We believe these properties should also make Microshift a great tool for other use cases such as Kubernetes applications development on resource-constrained systems, scale testing, and provisioning of lightweight Kubernetes control planes.
+
+**Note: Microshift is still early days and moving fast. Features are missing. Things break. But you can still help shape it, too.**
+
+<sup>1) more precisely [OKD](https://www.okd.io/), the Kubernetes distribution by the OpenShift community</sup>
+
+## Using Microshift
```
copejon, line 18: Since we only support fed/rhel/centos, we should call that out here.

README.md
```diff
@@ -1 +1,57 @@
-# microshift
+# Microshift
+Microshift (aka "µShift") is OpenShift<sup>1</sup> Kubernetes in a small form factor and optimized for edge computing.
```
copejon, line 2: We dropped all uShift usage from the code, but I'd argue that it's a µShift is the brand, and so should be used in documentation instead of "microshift", except for `µShift (aka "Microshift")`up front.

_OR_. we should not use µShift in docs at all if we only ever mention the branding spelling once.

copejon: And to pile on - we can't use `µShift` throughout the docs without it resulting in confusion when running `$ microshift`

copejon:
It's probably worth describing our versioning scheme as well

README.md
fzdarsky: Good points. Will remove µShift from text. Probably more something for a future logo :)

README.md
fzdarsky: Fixed.

README.md
fzdarsky: Updated this accoring to resolution for #99.

README.md
```diff
@@ -1 +1,62 @@
...
+### Building
+You can locally build Microshift using one of two methods, either directly on the host by installing the build-time dependencies:
+```
+sudo dnf install -y glibc-static
+make
+```
+
+or using a container build (running on Podman or Docker):
+```
+make microshift
```
copejon, line 51: Can we emphasize this option? The more contributors we have, the more important a standard build env will become.

README.md
```diff
@@ -1 +1,62 @@
...
+```
+
+### Running
+Use `install.sh` to set up your sytem and install run-time dependencies for Microshift, then simply:
```
copejon, line 55: install is going to download, create, and run a microshift release as a service.  That's not going to be ideal in a dev env where i want to run the binary i just built.  Maybe in the near future we can refactor the install script into a lib, to power 2 separate scripts: install.sh, and setup_dev_env.sh

fzdarsky: Agree on the use case. Could maybe be solved in install.sh itself with a flag to opt out (or in?) of installing + running the micrshift service.
Just not sure how to address this in this PR, so would suggest adding a new issue for it.

copejon:
lgtm

PR #113, opened 2021-06-17
add dockerfile

rootfs:
@copejon


copejon:
LGTM! In the near future, I'll transition this file into the build flow so make will build it for us

PR #122, opened 2021-06-22
enable containerized microshift on MacOS

rootfs:
I have built a container `docker.io/rootfs/ushift:macos` for testing
[asciicast](https://asciinema.org/a/MJZ69vE544HEUi2fpmwtcGpZe)


oglok:
Could we replace the CNI spec version to the latest?

```sudo sed -i 's/0.3.1/0.4.0/' /etc/cni/net.d/100-crio-bridge.conf```

Just to keep in sync with what the install script does.

rootfs:
> Could we replace the CNI spec version to the latest?
> 
> `sudo sed -i 's/0.3.1/0.4.0/' /etc/cni/net.d/100-crio-bridge.conf`
> 
> Just to keep in sync with what the install script does.

thanks, fixed

PR #125, opened 2021-06-23
design: arm support

rootfs:


fzdarsky:
@rootfs looks like this PR mixes the design proposal with removal of SDN-related code. I'd propose to separate out the latter into a new PR.

PR #127, opened 2021-06-25
this will allow for microshift to be rebooted

cooktheryan:
Right now if the server is rebooted the *setenforce 0* value no longer exists. Adding this change will allow for us to be able to reboot until proper selinux support is added


rootfs:
I am good with disabling selinux for now, pass baton to @oglok 
/assign @oglok

cooktheryan:
Not the proudest day for sure but it will allow people to continue until we get all the rules in place

Issue #86, opened 2021-06-07
Add firewalld rules


fzdarsky:

rootfs:
for both single and multi-node setup

copejon:
@fzdarsky @rootfs Is there more to add beyond #128 or can we close this issue?

rootfs:
cc @cooktheryan

cooktheryan:
@rootfs I know we are good with single node. Do we have multi-node established yet?

rootfs:
@cooktheryan let's solve the multi-node case when it is a thing :D

cooktheryan:
K. Then let's just close this as is then

PR #128, opened 2021-06-25
initial entries

cooktheryan:
This was an initial take at ports to allow. Once we get to the point of adding worker nodes it may need to massaged especially to allow the flannel coms. The testing I did on my side was to access the cluster from another instance within the same CIDR using the kubeconfig and then create an application using nodeport and verify I can access.


cooktheryan:
Assist #86 but does not close

PR #143, opened 2021-07-01
Use ubi8-minimal instead of regular ubi8

### LINKED ISSUE #141: [RFE] Switch to ubi-minimal
https://github.com/redhat-et/microshift/blob/7a42d43c9f3af4d066fa8aa9c71776bd2b08f13a/images/build/Dockerfile#L28

Not sure about the real size but I guess it worth trying. [More info about ubi-minimal](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index#con_understanding-the-ubi-minimal-images_assembly_types-of-container-images)
Note there is also ubi-micro but it doesn't include any package manager.

e-minguez:
It saves ~60 MB:

```
$ podman images | grep localhost
localhost/micro                              latest  548daf096454  35 seconds ago      335 MB
localhost/regular                            latest  c1d913e921fa  13 minutes ago      396 MB
```

Closes #141

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,11 @@
 # -v /run/containers:/run/containers \
 # -v /var/log:/var/log   microshift
 
-FROM  registry.access.redhat.com/ubi8/ubi
-RUN yum install -y \
-    policycoreutils-python-utils \
-    conntrack \
-    iptables-services || yum clean all -y
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.4
+RUN microdnf install -y \
+      policycoreutils-python-utils \
+      conntrack \
+      iptables-services && \
+    microdnf clean all
 ADD _output/bin/microshift /microshift
 ENTRYPOINT ["/microshift", "run"]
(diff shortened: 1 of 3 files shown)


e-minguez:
Not sure if there will be some other requisites, I only tested it doing a podman build and it was built successfully

rootfs:
@e-minguez thanks, can you also fix this one? https://github.com/redhat-et/microshift/blob/main/hack/all-in-one/Dockerfile#L1

e-minguez:
> @e-minguez thanks, can you also fix this one? https://github.com/redhat-et/microshift/blob/main/hack/all-in-one/Dockerfile#L1

Unfortunately cri-o is not available in the ubi8 repositories https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/x86_64/baseos/os/Packages/c/ https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/x86_64/appstream/os/Packages/c/

I can try to minimize the size of the all-in-one with some tweaks...

PR #153, opened 2021-07-09
Add framework for managing startup dependencies and shutdown

### LINKED ISSUE #152: Refactor embedded services, improve startup and shutdown behavior
1) Currently, the component services embedded into Microshift have their initialization, configuration, and execution logic spread over multiple places in the repo. That logic should be consolidated into one place per component and at the same time be standardized onto a common interface for managing component service lifecycle.

2) Standardizing the service interface also allows us to implement a generic logic for ensuring startup-ordering (waiting for dependencies to become ready before starting a dependent service) as well as cancellation.

3) For some component services (kube-apiserver, kube-scheduler, ...) we do not implement cancellation yet: Executing these services via their Cobra commands means they try to grab the system interrupts (SIGTERM et al.) for themselves and only stop on these interrupts rather than a stop channel.

4) While at it, implement signalling readiness of MicroShift (incl. its embedded services) to systemd and cleanly shut down on SIGTERM and SIGHUP signals.

fzdarsky:
Adds a "Service interface" that when implemented by embedded component services allows generic management of service startup (incl. waiting on dependencies to become ready), readiness, and cancellation.

Adds such a generic service manager that works based on expressing service dependencies as directed acyclic graph.
Implements signaling readiness to systemd and cancellation of embedded services after interrupts from the system.

Refactors the first two services (etcd, kube-apiserver) to illustrate the pattern, using a GenericService implementation to call the legacy code for bringing up non-ported services.

Closes #152


fzdarsky:
The PR starts all the embedded services, but some of them haven't been refactored yet and don't implement cancellation (they just execute forever until MicroShift's process terminates). As a result, the behavior (incl. when receiving interrupts) should still be the same.

To get a feeling for the UX after refactoring, comment out L80-91 and L95-110 of pkg/cmd/run.go and run.

You may need to wipe `/var/lib/microshift` before.

pkg/servicemanager/manager.go
```diff
@@ -0,0 +1,177 @@
...
+	for _, dependency := range s.Dependencies() {
+		// Enforce that services can only be added after adding their dependencies,
+		// i.e. they'll always remain topology sorted. Should we want to relax this
+		// constraint later, we can add topo sorting in the Run() step.
+		if _, exists := m.serviceMap[dependency]; !exists {
+			return fmt.Errorf("dependecy '%s' of service '%s' not yet defined", dependency, s.Name())
+		}
+	}
+
+	m.services = append(m.services, s)
```
rootfs, line 44: need a mutex here

rootfs:
this looks good, please finish it up and also run some benchmark to see if the new way accelerates the startup as well.

pkg/servicemanager/manager.go
fzdarsky: Thanks for the review, @rootfs. You mean for hedging against `AddService()` being called concurrently?
Note that at least currently, the order of adding services is significant (to enforce the invariant of dependencies being added before dependents), so `AddService()` shouldn't be called concurrently. Otherwise, we'd have to enable topo sorting. See also the comments in `AddService()` and `Run()`.

How should we proceed? Add mutex proactively or add note in method description?

fzdarsky:
> this looks good, please finish it up and also run some benchmark to see if the new way accelerates the startup as well.

As long as the services other than etcd and kube-apiserver (who have a startup dependency already) haven't been ported, I don't expect to see much diff (other than seeing a few less error messages in the log).
Would you propose I port over the other services and run the benchmark or shall we merge, then port, then benchmark for control?

rootfs:
> Would you propose I port over the other services and run the benchmark or shall we merge, then port, then benchmark for control?

let's merge this one first and please go ahead with another PR for the rest of the services.

PR #155, opened 2021-07-09
delete obsolete doc data

copejon:


rootfs:
what happens to the digest?

copejon:
Per out-of-band conversation, it was decided that using the latest z-stream release of the ubi-minimal:8.4 base image was acceptable.  This is performed here: https://github.com/redhat-et/microshift/blob/a3c8051cf1bbe5c70eede8dc9073984a96bcc077/Makefile#L88 when `maked` is called by `release.sh`.

Issue #144, opened 2021-07-01
Optionally start OpenShift components


rootfs:
In the current implementation, all components in the `pkg/assets` are started. We need to fine tune the process to make the startup configurable and only turn on those that are allowed in the configuration file.

guymguym:
Just to add on that I think that enabling components should also be possible after startup.
The reason I believe this is important to support on runtime is that the trigger to enable openshift components is when trying to deploy some workload that depends on it, and since workloads will not be bundled into microshift then this event will occur on runtime.

rootfs:
@guymguym yes, the configuration file is watched, changes will be picked up to start other components if they are enabled during runtime.

fzdarsky:
@rootfs could you provide an example use of this, please?
The set of components we deploy should anyway be limited to a bare minimum, with everything that's non-essential (say, a local registry) being something that users can deploy post cluster-up. Are you suggesting some of the components we deploy shouldn't belong to the essential minimal set? Or is the intention to make components replaceable?

guymguym:
@fzdarsky One example of a feature that I wanted to enable post cluster-up is metrics which is not enabled by default on microshift. For example I stumbled this -
```
$ kubectl top pod
error: Metrics API not available
```
I'm not sure if metrics should be included in a bare minimum deployment, but I definitely hear requirements for metrics on Edge that includes also prometheus, which might also be a custom deployment configuration.

copejon:
@guymguym 's use case either falls under pre-loading images (#162 ) or should be left to the user as a manual post-boot operation.

I think at this stage, we should remain opinionated about microshift's runtime services, lest we open a can of compatibility worms before we're ready to address them.  @rootfs ?

PR #158, opened 2021-07-13
component loader design

rootfs:
@fzdarsky @copejon @oglok 

this is to address #144

### DIFF
--- a/doc/design-component.md
+++ b/doc/design-component.md
@@ -1,0 +1,69 @@
+Component Loader Design
+=======================
+
+# Background
+MicroShift is designed to provide a turnkey solution that can readily serve the target workloads 
+with little intervention, probably in an airgapped environment. This requires MicroShift to
+be able to provides the most, if not all, infrastructure components during bootstrapping. 
+
+In its current form, MicroShift has an predefined asset repo and a sequence of 
+loading the manifests to the cluster. Although new assets can be added programmatically, 
+this workload clearly suffers from scalability and flexibilty. 
+
+Specifically, currently the core components are: Service-ca, OpenShift DNS, OpenShift Router, and HostPath provisioner.
+For non-interactive use cases, few of these components are required. While for Data Services oriented workloads 
+may prefer a more reliable and scale-out storage is beyond the capacity of the HostPath volume. 
+
+So it is clear that MicroShift must have a way to customize which components to load during bootstrapping and normal startup.
+
+# Existing Solutions
+
+## Operator
+Operator aims to manage the entire lifecycles of complicated components. Although Operator solves the issue
+of managing inter-dependecy, it is deemed heavier than the MicroShift use case. In many perceived occasions, the MicroShift clusters are deployed just once: they are either stay on offline mode for life (e.g. an Edge device deployed in isolated environment) or destroyed without care (e.g. a developer's test environment). There is no
+need to keep the Operator running to keep the API objects in the desired states. On the other hand, Operators' all-on pattern does not serve MicroShift's small footprint design goal.
+
+## Kubernetes Addon
+Kubernetes Addon mechanism loads cluster wide infrastructure after cluster initialization. Addons such as networking, storage, and management are curated in the Kubernetes repo. Such arrangement, however, does not fit 
+into the MicroShift use case:
+- MicroShift may need more than one components. For instance, Ceph bucket notification, a use case for processing
+Object Store objects on demand, involves Rook/Ceph and Knative (Eventing, Serving, and Networking). 
+- These components are often inter-dependent. Kubernetes Addons cannot ensure the components are created or started in the desired order.
+
+## Helm
+Helm packages multiple components in a chart form and deploy them in a declarative manner. Once deployed, Helm exists till the next it upgrades or removes the chart from the cluster. The non-interactive and low footprint features fit into MicroShift use well. Still the chart must be accessible by the MicroShift cluster when operating
+in an airgapped environment.
+
+# Design
+As discussed, MicroShift Component Loader must be able to:
+- Customizable for different workloads targeting varying use cases
+- Managed multiple, sometimes inter-dependent component
+- Allow MicroShift to invoke when the cluster is ready
+- Allow MicroShift to monitor readiness status
+- Low, preferrably zero footprint
+- Able to operate in offline mode
+
+## Component Loader Workflow
+
+When MicroShift starts up the cluster, it reads from the configuration file and finds the Container image that Component Load Job (i.e. running as a Kubernetes Job) uses, constructs a Kubernetes Job using this Container images, creates one ConfigMap for the Job to report progress, creates another ConfigMap for parameters required by the Job (if exists), then starts and watches the Job.
+
+When the Job starts, it reads its configuration from the parameter ConfigMap, creates components, reports readiness status by updating the predefined ConfigMap.
+
+Once the Job exits, MicroShift catches the status and checks the ConfigMap and gets the status.
+
+# Issues
+
+## Repair Components
+It is likely that after startup, some of the components, such as Pods and Deployments, degrade or are even killed. To cope with this case, the Component Loader can be started as a CronJob so it can start periodically to repair the components.
+
+## Offline Operation Support
+
+### Container Image
+The Container images of the Component Loader and the compnonents it created are packaged or pre-downloaded when integrated with MicroShift.
+
+### Reuse Helm Chart
+Similarly, if Helm Charts are used as or by a Component Loader, the Charts are also locally stored together with the airgapped Container images.
+
+
+
+


doc/design-component.md
```diff
@@ -0,0 +1,69 @@
+Component Loader Design
+=======================
+
+# Background
+MicroShift is designed to provide a turnkey solution that can readily serve the target workloads 
+with little intervention, probably in an airgapped environment. This requires MicroShift to
+be able to provides the most, if not all, infrastructure components during bootstrapping. 
+
+In its current form, MicroShift has an predefined asset repo and a sequence of
```
fzdarsky, line 9: "... asset repo that gets compiled into the MicroShift binary and loaded into the running cluster." ?

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+with little intervention, probably in an airgapped environment. This requires MicroShift to
+be able to provides the most, if not all, infrastructure components during bootstrapping. 
+
+In its current form, MicroShift has an predefined asset repo and a sequence of 
+loading the manifests to the cluster. Although new assets can be added programmatically, 
+this workload clearly suffers from scalability and flexibilty. 
+
+Specifically, currently the core components are: Service-ca, OpenShift DNS, OpenShift Router, and HostPath provisioner.
+For non-interactive use cases, few of these components are required. While for Data Services oriented workloads 
+may prefer a more reliable and scale-out storage is beyond the capacity of the HostPath volume.
```
fzdarsky, line 15: Another use case is adding an agent for remote management of the cluster.

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
+Component Loader Design
+=======================
+
+# Background
+MicroShift is designed to provide a turnkey solution that can readily serve the target workloads 
+with little intervention, probably in an airgapped environment. This requires MicroShift to
+be able to provides the most, if not all, infrastructure components during bootstrapping.
```
fzdarsky, line 7: "At the same time, one of MicroShift's design goals is to keep its fixed core functionality to a minimum while allowing to easily extend or swap out infrastructure components to match the use case."

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+may prefer a more reliable and scale-out storage is beyond the capacity of the HostPath volume. 
+
+So it is clear that MicroShift must have a way to customize which components to load during bootstrapping and normal startup.
+
+# Existing Solutions
+
+## Operator
+Operator aims to manage the entire lifecycles of complicated components. Although Operator solves the issue
+of managing inter-dependecy, it is deemed heavier than the MicroShift use case. In many perceived occasions, the MicroShift clusters are deployed just once: they are either stay on offline mode for life (e.g. an Edge device deployed in isolated environment) or destroyed without care (e.g. a developer's test environment). There is no
+need to keep the Operator running to keep the API objects in the desired states. On the other hand, Operators' all-on pattern does not serve MicroShift's small footprint design goal.
```
fzdarsky, line 24: I'm not sure what "all-on pattern" means? Do you mean that Operators are continuously running their control loops, which adds load on the cluster we'd like to avoid?
Of course, users are free to install whatever however, including using Operators. I agree, though, that it's not a good default mechanism for MicroShift.

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+Kubernetes Addon mechanism loads cluster wide infrastructure after cluster initialization. Addons such as networking, storage, and management are curated in the Kubernetes repo. Such arrangement, however, does not fit 
+into the MicroShift use case:
+- MicroShift may need more than one components. For instance, Ceph bucket notification, a use case for processing
+Object Store objects on demand, involves Rook/Ceph and Knative (Eventing, Serving, and Networking). 
+- These components are often inter-dependent. Kubernetes Addons cannot ensure the components are created or started in the desired order.
+
+## Helm
+Helm packages multiple components in a chart form and deploy them in a declarative manner. Once deployed, Helm exists till the next it upgrades or removes the chart from the cluster. The non-interactive and low footprint features fit into MicroShift use well. Still the chart must be accessible by the MicroShift cluster when operating
+in an airgapped environment.
+
```
fzdarsky, line 36: I propose to also add `kustomize` into this list.

To me, `kustomize` makes a lot of sense as - unlike Helm - it is K8s native, has mechanisms to apply resources in the right order, and supports remote resources and 'kustomizations', which is nice for a GitOps pattern of operations. It also means we don't need to install an additional dependency.
It's (small) disadvantage vs. Helm is that it doesn't have a default archive format (though it could always be tar-balled, of course).

Maybe we can even support both? E.g. using kustomize built-in and bootstrap a Helm extension with it?

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+- Customizable for different workloads targeting varying use cases
+- Managed multiple, sometimes inter-dependent component
+- Allow MicroShift to invoke when the cluster is ready
+- Allow MicroShift to monitor readiness status
+- Low, preferrably zero footprint
+- Able to operate in offline mode
+
+## Component Loader Workflow
+
+When MicroShift starts up the cluster, it reads from the configuration file and finds the Container image that Component Load Job (i.e. running as a Kubernetes Job) uses, constructs a Kubernetes Job using this Container images, creates one ConfigMap for the Job to report progress, creates another ConfigMap for parameters required by the Job (if exists), then starts and watches the Job.
```
fzdarsky, line 48: What would be the advantage of running this as a K8s Job and requiring a dedicated container image for this vs. for example adding a goroutine to MicroShift that simply monitors a given directory (say, `/var/lib/microshift/manifests` or a configurable list of directories) for `kustomization.yaml`s or Helm charts and then applies them at cluster init or when the directory changes?

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+
+When the Job starts, it reads its configuration from the parameter ConfigMap, creates components, reports readiness status by updating the predefined ConfigMap.
+
+Once the Job exits, MicroShift catches the status and checks the ConfigMap and gets the status.
+
+# Issues
+
+## Repair Components
+It is likely that after startup, some of the components, such as Pods and Deployments, degrade or are even killed. To cope with this case, the Component Loader can be started as a CronJob so it can start periodically to repair the components.
```
fzdarsky, line 57: Shouldn't that use K8s's native mechanisms to restart pods?

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Offline Operation Support
+
+### Container Image
+The Container images of the Component Loader and the compnonents it created are packaged or pre-downloaded when integrated with MicroShift.
```
fzdarsky, line 62: Yep, that's a good one. I'll create a dedicated issue for this.

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Reuse Helm Chart
+Similarly, if Helm Charts are used as or by a Component Loader, the Charts are also locally stored together with the airgapped Container images.
```
fzdarsky, line 65: Locally stored, but maybe in different directories? For example, `/var/lib/microshift/images` vs `/var/lib/microshift/charts`?

fzdarsky (changes requested):

doc/design-component.md
guymguym: @fzdarsky @rootfs Maybe this can just run the same loader script, like a "micro" reconcile loop for the resources it created? This will fuse the loader-Job with the repair-CronJob.

fzdarsky: @guymguym I'm not convinced yet we'd want to use a Job to deploy a component and "repair" it through a periodically running Job at all. Or make it default, thereby forcing users to learn it.

I feel this can be solved simpler, more flexible and using well-established mechanisms if MicroShift was to monitor a folder on disk that users (or MicroShift itself for defaults) can drop manifests / kustomization.yamls / Helm charts into and those get automatically applied. Repairing deployments, monitoring for readiness of a dependency before starting a service, etc. should be based on standard K8s mechanisms and practices IMO.

doc/design-component.md
mangelajo: That's what I understood. I also don't see operators as a fit for MicroShift components.

doc/design-component.md
mangelajo: I agree, Kustomize yamls sound like the lightest-weight solution we could implement. For helm I wonder what's the memory footprint of the helm processes and images (I have been looking into that, but couldn't find anything), of course those don't run continuously, but when they need to run they will replace other processes's pages or cached pages from memory which is harmful.

Would it make sense to start by something like kustomize, and then if we see gaps, we can look into supporting helm? what do you think?

doc/design-component.md
```diff
@@ -0,0 +1,69 @@
...
+- Managed multiple, sometimes inter-dependent component
+- Allow MicroShift to invoke when the cluster is ready
+- Allow MicroShift to monitor readiness status
+- Low, preferrably zero footprint
+- Able to operate in offline mode
+
+## Component Loader Workflow
+
+When MicroShift starts up the cluster, it reads from the configuration file and finds the Container image that Component Load Job (i.e. running as a Kubernetes Job) uses, constructs a Kubernetes Job using this Container images, creates one ConfigMap for the Job to report progress, creates another ConfigMap for parameters required by the Job (if exists), then starts and watches the Job.
+
```
mangelajo, line 49: Without becoming operators, have you considered using some simple form of CRD? it is possible to do free-form CRDs which have the benefit of being easier to observe via kubectl. The benefit of that is that the CRD could have a status field.

Some example of old free-form CRDs we used in submariner (you just define them, then you put whatever you like as fields inside).
https://github.com/submariner-io/submariner-charts/blob/v0.0.1/submariner-k8s-broker/templates/crd.yaml#L16

You can even tell kubectl what columns to print: 
https://github.com/submariner-io/submariner-charts/blob/v0.8.0/submariner-k8s-broker/crds/crd.yaml#L36

doc/design-component.md
mangelajo, line 62:
```suggestion
The Container images of the Component Loader and the components it created are packaged or pre-downloaded when integrated with MicroShift.
```

just a typo

cooktheryan:
@fzdarsky @rootfs Do we want the loader to run in-cluster? We can gain some leverage in functionality and rely on cron or systemd startup components to identify when to issue the loader

cooktheryan:
I think we still have the component loader launch pods for the creation of kubernetes resources as that would allow for us to use the in-cluster kubeconfig and not need to create a secret with a kubeconfig

Here is a rundown of the specific things I am seeing 

Image loading - Although path is specified here we could dictate to force type: image to only interact with items in images/ directory
```
Name: hostpath-Image
process: (add/remove)
type: image
path: images/00-hostpath.tar
```

kubectl - Similar to above but all yaml based items live within their own subdirectory of yaml
```
Name: deployment
process: (add/remove)
type: yaml
path:  yaml/MY_APP
```

kustomize
```
Name: kustomized-app
process: (add/remove)
type: kustomized-yaml
path:  kustomized-yaml/MY_APP
```

helm repo
```
Name: helm-repo
process: (add/remove)
type: helm-repo
path: helm-repo/volsync
```

helm chart
```
Name: helm-chart
process: (add/remove)
type: helm-chart
path: helm-chart/volsync
```

The add and remove key value is there to allow for some sort of management of the items to allow for us to clean up as we go. To manage this at Scale we provide directions to users on how to run a pod that would write to the specific directory replacing YAML components as we need. 

For the code pieces, we should be able to mount the kubectl, helm, and kustomize pieces into a container and use the in-cluster kubeconfig file that is generated for a service account allowing us to not need to manage or maintain a unique secret. For the image pieces we could either use the podman golang library or even go as hack to use OS within GOLang

fzdarsky:
> @fzdarsky @rootfs Do we want the loader to run in-cluster? We can gain some leverage in functionality and rely on cron or systemd startup components to identify when to issue the loader

Embedded into MicroShift, because MicroShift needs to emit the manifests for the components with the right parameters incl. the container image digests.

cooktheryan:
@fzdarsky embedded into Microshift bin or container but not specifically the microshift cluster itself correct?

fzdarsky:
> @fzdarsky embedded into Microshift bin or container but not specifically the microshift cluster itself correct?

I'm not sure I understood your question correctly, but would say the manifest templates we need to render and then apply to the cluster are embedded (via bindata) into the MIcroShift binary, once applied they'd run on the cluster.

cooktheryan:
got it!

fzdarsky:
It seems to me discussions are already going into the weeds of one specific component loader approach (the job-based one) while we still haven't concluded on the general approach, e.g. whether we actually need a super flexible solution or a simple and highly opinionated manifest applier would be sufficient.

fzdarsky:
Closing as we're staying with the original loader for now.

PR #159, opened 2021-07-14
[WIP] Component Loader

rootfs:
following the design doc #158


rootfs:
depend on #167

cooktheryan:
I'm wondering if we may want to try to approach the format of the items similar to how kuttl expects things.

Shell
```
apiVersion: kuttl.dev/v1beta1
kind: TestStep
commands:
  - timeout: 90
    command: ./05-create-user.sh
```


Straight YAML
```
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
---
apiVersion: v1
kind: Secret
metadata:
  name: secret-key
  namespace: test
type: Opaque
```

This way we can provide the Loader container of our own with only forced in items we expect and want people to use

fzdarsky:
Closing as we're staying with the original loader for now.

PR #165, opened 2021-07-18
Improve test coverage

fzdarsky:


rootfs:
@fzdarsky this is the error from CI
`pkg/servicemanager/manager_test.go:79:7: the cancel function returned by context.WithCancel should be called, not discarded, to avoid a context leak`

rootfs (approved):

PR #167, opened 2021-07-19
add kubeconfig for cluster and public access

rootfs:
Currently kubeadmin kubeconfig points to `127.0.0.1`. This IP doesn't work outside of the localhost.

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -160,6 +160,12 @@ func initKubeconfig(cfg *config.MicroshiftConfig) error {
 	if err := util.Kubeconfig(cfg.DataDir+"/resources/kubeadmin/kubeconfig", "system:admin", []string{"system:masters"}, cfg.Cluster.URL); err != nil {
 		return err
 	}
+	if err := util.Kubeconfig(cfg.DataDir+"/resources/kubeadmin/kubeconfig.public", "system:admin", []string{"system:masters"}, cfg.Cluster.PublicURL); err != nil {
+		return err
+	}
+	if err := util.Kubeconfig(cfg.DataDir+"/resources/kubeadmin/kubeconfig", "system:admin", []string{"system:masters"}, cfg.Cluster.URL); err != nil {
+		return err
+	}
 	if err := util.Kubeconfig(cfg.DataDir+"/resources/kube-apiserver/kubeconfig", "kube-apiserver", []string{"kube-apiserver", "system:kube-apiserver", "system:masters"}, cfg.Cluster.URL); err != nil {
 		return err
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -29,8 +29,10 @@ var (
 )
 
 type ClusterConfig struct {
+	// URL is api endpoint accessed by local host
 	URL string `yaml:"url"`
-
+	// PublicURL is api endpoint accessed external hosts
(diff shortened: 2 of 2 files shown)


fzdarsky:
ClusterConfig.URL **is** already the URL of the API server. Instead of introducing a new config parameter (which we should only do if there's no better alternative, as one of our design goals is to minimise params) we may want to default to binding to the host's primary IP (=hostIP) instead of loopback **or** (more future-proof for multi-node) allocate a VIP and bind to that.

Of course, the URL param can always be overridden explicitly. Also, note that by design we default to not having any network access from the internet (i.e. no public IP/URL or open network ports for SSH).

rootfs:
@fzdarsky Currently `ClusterConfig.URL` is currently `127.0.0.1`. This is for local access. The new public URL is for Pods or external access, which is set to the host IP by default. They serve different purposes. `127.0.0.1` works as long as the command runs locally, while public URL, in the long term, should be bind to the public facing IP (host IP or load balancer IP). So calls from outside the host or inside the Pods can access the API service.

We need two URLs for HA: when the public URL is only temporary (e.g. DHCP, load balancer IP, etc), the local access can be a reliable backup for local access.

fzdarsky:
@rootfs yes. `ClusterConfig.URL` is currently `127.0.0.1`, which is why I suggested changing _that_ to default to the `hostIP` (or auto-allocated VIP) instead. It'll be used for "internal" API traffic (as by default, there isn't any "public" access; if a user wants it, they can override the default URL param) between multiple nodes.
If running single node, using loopback instead of the hostIP is a relatively small optimization, but if users want that, they could still configure loopback explicitly.

rootfs:
@fzdarsky what if the host IP changes (e.g. acquired from DHCP)

fzdarsky:
> @fzdarsky what if the host IP changes (e.g. acquired from DHCP)

That's when we'll have to switch to using a DNS name instead of an IP anyway. Basing off the host IP wouldn't work in this case.

cooktheryan:
@fzdarsky @rootfs if we switch to DNS name I think this would fix #266 bc then I could put an elastic or float IP in front of a host then assign the DNS to that address. This would allow for us to have a different IP than the address of localhost and the system private IP address. Although, I dont know if this would cause a larger issue or not but we wouldn't want to force traffic out of the host private address -> elastic or floating IP -> back to the host itself.

fzdarsky:
> Although, I dont know if this would cause a larger issue or not but we wouldn't want to force traffic out of the host private address -> elastic or floating IP -> back to the host itself.

We cannot rely on the existence of an external load balancer. Instead, we'll probably need to allocate a VIP from DHCP and float it across control plane instances (e.g. assign VIP to leader) or use client-side load balancing. That also depends on how we want to handle DNS (e.g. whether to use mDNS or not). Probably needs some analysis / design discussions considering the multi-node case.

@mangelajo @oglok ^?

cooktheryan:
@fzdarsky I agree. I do think if we had some sort of way to provide alternate FQDNs we can solve a lot of problems regardless of the hosting provider. I don't think a config file would be the best solution but it could be an easy way for us to provide the option for additional hostname to validate the certificate with. I think a config file or even an after the fact parameter that we can issue against the microshift binary we could add the additional FQDN to the certificate or regenerate

PR #169, opened 2021-07-21
Fix high CPU load issue

fzdarsky:


guymguym (commented):
@fzdarsky The recursive imple of `And()` is a bit too delicate IMHO.
Wouldn't you prefer a more straightforward impl with sync.WaitGroup where every channel is waited in its own go routine?

rootfs:
The high cpu issue resolved, let's take this patch. @guymguym improvement welcome

PR #171, opened 2021-07-21
[WIP] acm-registration-scripts

copejon:


hack/acm-spoke.sh
```diff
@@ -0,0 +1,4 @@
+#! /usr/bin/env bash
```
rootfs, line 1: what's the file for?

hack/acm-hub.sh
```diff
@@ -0,0 +1,52 @@
...
+  searchCollector:
+    enabled: true
+  version: 2.2.0
+EOF
+
+oc new-project "$CLUSTER_NAME" 1>/dev/null
+oc label namespace "$CLUSTER_NAME" cluster.open-cluster-management.io/managedCluster="$CLUSTER_NAME"
+oc apply -f "$WORK_DIR"/managed-cluster.yaml
+oc apply -f "$WORK_DIR"/klusterlet-addon-config.yaml
+sleep 3
```
rootfs, line 49: why sleep? can we do wait and watch?

oglok:
Shouldn't this script go to the microshift-demo repo instead of here?

fzdarsky:
Agree with @oglok, it's not a core component of MicroShift (which we try to keep minimal), but it's an option of how to use/manage it. Therefore, I'd also put it into the microshift-demos repo.

hack/acm-hub.sh
copejon: We definitely can enhance these scripts. Given the time crunch, I wanted to keep it simple.

hack/acm-spoke.sh
copejon: Just to simplify the spoke registration step.

copejon:
Fair enough, will push other there

PR #190, opened 2021-08-04
Initialize also with pre-existing data dir

### LINKED ISSUE #189: [BUG] Certs are not initialised when /var/lib/microshift dir is present
#### What happened:
MicroShift doesn't initialise certs, configs, etc. if /var/lib/microshift dir exists prior to running MicroShift.

MicroShift should initialise certs when they are actually missing.

1. mkdir /var/lib/microshift
2. install & run microshift

Reasons that /var/lib/microshift dir would be present prior to "first run" is preloading of images or manifests.

fzdarsky:
Currently, Microshift checks for the non-existence of the data dir (/var/lib/microshift) to infer that
it runs for the first time and thus should initialize certs, kubeconfigs, etc. This creates a problem
on systems where /var/lib/microshift already exists (e.g. for pre-loading of images or manifests).

Until we have cleaned-up initialization as part of the multi-node work, this patch introduces a work-
around that makes initialization conditional on the cert dir being present instead, which is more
specific.

Closes #189

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


oglok:
PR tested on RHEL for Edge in the e2e deployment use case.

/lgtm

copejon:
LGTM

PR #195, opened 2021-08-04
fix release.sh bug

copejon:
Fixes a release script bug

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ cross-build: cross-build-linux-amd64 cross-build-linux-arm64
 ###############################
 _build_containerized:
 	echo BIN_TIMESTAMP==$(BIN_TIMESTAMP)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -83,7 +83,7 @@ cross-build: cross-build-linux-amd64 cross-build-linux-arm64
 ###############################
 _build_containerized:
 	echo BIN_TIMESTAMP==$(BIN_TIMESTAMP)
-	$(CTR_CMD) build -t $(IMAGE_REPO):$(RELEASE_PRE)-$(TIMESTAMP)-linux-$(ARCH) \
+	$(CTR_CMD) build -t $(IMAGE_REPO):$(SOURCE_GIT_TAG)-linux-$(ARCH) \
```
copejon, line 86: Should be using SOURCE_GIT_TAG rather than ignoring it.

PR #198, opened 2021-08-06
Disconnected script + release assets

cooktheryan:
This gets us closer to a full pre-loading image and pruning story

### DIFF
--- a/hack/disconnected.sh
+++ b/hack/disconnected.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 DESTINATION_PATH=$1
+if [[ -z $DESTINATION_PATH ]]; then
+   echo "A path must be specified before running this script. Example (sh hack/disconnected.sh /tmp/)"
+   exit 1
+fi
+
+# Set temporary location
 PODMAN_ROOT="/tmp/storage"
 rm -rf ${PODMAN_ROOT} || true
 mkdir -p ${PODMAN_ROOT}
@@ -12,7 +18,4 @@ for i in ${CONSTSANTS} ${BINDATA}; do
    podman --root ${PODMAN_ROOT} --runroot  ${PODMAN_ROOT}  pull  ${i}
 done
 
-if [[ -z $DESTINATION_PATH ]]; then
-   DESTINATION_PATH="./"
-fi
 sudo tar czf  ${DESTINATION_PATH}/microshift-images-amd64.tgz -C /tmp/ storage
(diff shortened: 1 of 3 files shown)


cooktheryan:
Example release https://github.com/cooktheryan/microshift/releases/tag/4.7.0-0.microshift-2021-08-06-145514

hack/disconnected.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+DESTINATION_PATH=$1
+PODMAN_ROOT="/tmp/storage"
```
rootfs, line 3: can you test if using other dir than `/var/lib/containers/storage` works with crio? I think crio may complain if the images are not originally saved in the same directory.

hack/disconnected.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+DESTINATION_PATH=$1
```
copejon, line 2: Can we have this fail if not specified?  E.g. `...=${1?"error message "}`

hack/disconnected.sh
cooktheryan: @rootfs it doesn't seem to care. This is output from a new box ```storage/overlay/777b2c648970480f50f5b4d0af8f9a8ea798eea43dbcf40ce4a8c7118736bdcf/diff/bin/arch
[fedora@ip-172-31-86-59 ~]$        sudo chown -R root:root /var/lib/containers/storage
[fedora@ip-172-31-86-59 ~]$     sudo systemctl enable crio
    sudo systemctl restart crio
Created symlink /etc/systemd/system/multi-user.target.wants/crio.service → /usr/lib/systemd/system/crio.service.

sudo crictl images
[fedora@ip-172-31-86-59 ~]$ 
[fedora@ip-172-31-86-59 ~]$ sudo crictl images
IMAGE                                  TAG                 IMAGE ID            SIZE
quay.io/coreos/flannel                 v0.14.0             8522d622299ca       68.9MB
quay.io/openshift/okd-content@sha256   <none>              60cd591c9b8ca       311MB
quay.io/openshift/okd-content@sha256   <none>              2fc332cf4cd6a       297MB
quay.io/openshift/okd-content@sha256   <none>              6fca754c16c9f       278MB
quay.io/openshift/okd-content@sha256   <none>              ebb673575b8b1       320MB
quay.io/openshift/okd-content@sha256   <none>              ec71cef32b1ed       291MB
quay.io/openshift/okd-content@sha256   <none>              cb6c8de4eac8f       436MB
```

hack/disconnected.sh
cooktheryan: I dont see why not. Right now its an optional var but we can force require it

copejon: Ah - I missed it being optional - nvm then.

cooktheryan:
closing this and will re-open. Need to fix my repo

PR #204, opened 2021-08-11
allow for pre-population of images

cooktheryan:
This is part of #198 but does not close


cooktheryan:
@rootfs i'm closing this request. I dont like the path I chose. I will sync up with you about how we can use the component loader with this

PR #205, opened 2021-08-11
Readme update

cooktheryan:
More information for the README


README.md
```diff
@@ -53,25 +53,43 @@ You can locally build Microshift using one of two methods, either using a contai
...
-`CONFIG_ENV_ONLY=true ./install.sh`
+```
+CONFIG_ENV_ONLY=true ./install.sh
+```
 
 ### Running
 
 ```
-sudo microshift run
+sudo ./microshift run
```
rootfs, line 79: after installed, microshift may not be on the current dir

README.md
```diff
@@ -53,25 +53,43 @@ You can locally build Microshift using one of two methods, either using a contai
...
 ```
 
 Microshift keeps all its state in its data-dir, which defaults to `/var/lib/microshift` when running Microshift as privileged user and `$HOME/.microshift` otherwise. Note that running Microshift unprivileged only works without node role at the moment (i.e. using `--roles=controlplane` instead of the default of `--roles=controlplane,node`).
 
-You can find the kubeadmin's kubeconfig under `$DATADIR/resources/kubeadmin/kubeconfig`.
+### Kubeconfig
```
rootfs, line 84: this is already done by `install.sh` here https://github.com/redhat-et/microshift/blob/main/install.sh#L167

rootfs: i see. cc @oglok @copejon

README.md
```diff
@@ -50,28 +50,33 @@ rm -rf /var/lib/microshift && rm -r $HOME/.microshift
...
 ```
-sudo microshift run
+CONFIG_ENV_ONLY=true ./install.sh
 ```
 
 Microshift keeps all its state in its data-dir, which defaults to `/var/lib/microshift` when running Microshift as privileged user and `$HOME/.microshift` otherwise. Note that running Microshift unprivileged only works without node role at the moment (i.e. using `--roles=controlplane` instead of the default of `--roles=controlplane,node`).
 
-You can find the kubeadmin's kubeconfig under `$DATADIR/resources/kubeadmin/kubeconfig`.
+### Kubeconfig
+The `install.sh` script should place the kubeconfig file for you. If you need it for another user or to use externally the kubeadmin's kubeconfig is placed `/var/lib/microshift/resources/kubeadmin/kubeconfig` during configuration.
```
fzdarsky, line 81: To be precise, it's placed there the first time `microshift run` is executed. IOW, after running `install.sh` with `CONFIG_ENV_ONLY` flag on a fresh dev env, kubeconfig will not be available in either location.

oglok: That is correct. 

CONFIG_ENV_ONLY=true env var was meant for CI to basically configure the dependencies on the host, but not to deploy microshift itself.

The kubeconfig files are generated once microshift runs, and CONFIG_ENV_ONLY stops right before fetching microshift and preparing the `.kube/config` file.

oglok (changes requested):

PR #217, opened 2021-08-19
Refactor Kubelet to conform to the internal service manager

### LINKED ISSUE #201: Refactor kubelet to conform to common service interface
#### Design Document Link

PR: #217 
Parent issue https://github.com/redhat-et/microshift/issues/161

husky-parul:
https://github.com/redhat-et/microshift/issues/201

Closes #201

### DIFF
--- a/pkg/node/kubelet.go
+++ b/pkg/node/kubelet.go
@@ -32,6 +32,8 @@ import (
 	kubelet "k8s.io/kubernetes/cmd/kubelet/app"
 	kubeletoptions "k8s.io/kubernetes/cmd/kubelet/app/options"
 	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
+
+	"github.com/openshift/microshift/pkg/util"
 )
 
 const (
@@ -40,9 +42,8 @@ const (
 )
 
 type Kubelet struct {
-	kubeletflags   *kubeletoptions.KubeletFlags
-	kubeconfig     *kubeletconfig.KubeletConfiguration
-	kubeconfigfile string
+	kubeletflags *kubeletoptions.KubeletFlags
+	kubeconfig   *kubeletconfig.KubeletConfiguration
 }
 
 func NewKubelet(cfg *config.MicroshiftConfig) (*Kubelet, error) {
@@ -107,7 +108,6 @@ func (s *Kubelet) configure(cfg *config.MicroshiftConfig) error {
 	}
 	s.kubeletflags = kubeletFlags
 	s.kubeconfig = kubeletConfig
-	s.kubeconfigfile = filepath.Join(cfg.DataDir, "resources", "kubelet", "kubeconfig")
 
 	logrus.Infof("Starting kubelet %s, args: %v", cfg.NodeIP, args)
 	return nil
@@ -115,6 +115,17 @@ func (s *Kubelet) configure(cfg *config.MicroshiftConfig) error {
 
 func (s *Kubelet) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
 	defer close(stopped)
+
+	// run readiness check
+	go func() {
+		healthcheckStatus := util.RetryInsecureHttpsGet("https://127.0.0.1:10248/healthz")
+		if healthcheckStatus != 200 {
+			logrus.Fatalf("Kube-controller-manager failed to starttttttt")
+		}
+
+		logrus.Infof("%s is readyyy", s.Name())
+		close(ready)
+	}()
 
 	// construct a KubeletServer from kubeletFlags and kubeletConfig
 	kubeletServer := &kubeletoptions.KubeletServer{
--- a/pkg/node/kubelet.go
+++ b/pkg/node/kubelet.go
@@ -17,6 +17,7 @@ package node
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
 	"strconv"
 
@@ -34,6 +35,8 @@ import (
 
 	kubeletoptions "k8s.io/kubernetes/cmd/kubelet/app/options"
 	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
+	"k8s.io/kubernetes/pkg/kubelet/kubeletconfig/configfiles"
+	utilfs "k8s.io/kubernetes/pkg/util/filesystem"
 )
 
 const (
@@ -64,15 +67,12 @@ func (s *KubeletServer) configure(cfg *config.MicroshiftConfig) error {
 	}
 	// Prepare commandline args
 	args := []string{
-		"--config=" + cfg.DataDir + "/resources/kubelet/config/config.yaml",
 		"--bootstrap-kubeconfig=" + cfg.DataDir + "/resources/kubelet/kubeconfig",
 		"--kubeconfig=" + cfg.DataDir + "/resources/kubelet/kubeconfig",
 		"--container-runtime=remote",
 		"--container-runtime-endpoint=unix:///var/run/crio/crio.sock",
 		"--runtime-cgroups=/system.slice/crio.service",
-		"--cgroup-driver=systemd",
 		"--node-ip=" + cfg.NodeIP,
-		"--volume-plugin-dir=" + cfg.DataDir + "/kubelet-plugins/volume/exec",
 		"--logtostderr=" + strconv.FormatBool(cfg.LogDir == "" || cfg.LogAlsotostderr),
 		"--alsologtostderr=" + strconv.FormatBool(cfg.LogAlsotostderr),
 		"--v=" + strconv.Itoa(cfg.LogVLevel),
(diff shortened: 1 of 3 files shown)


pkg/node/kubelet.go
```diff
@@ -47,17 +86,50 @@ func StartKubelet(cfg *config.MicroshiftConfig) error {
...
+	kubeletServer := &kubeletoptions.KubeletServer{
+		KubeletFlags:         *s.kubeletflags,
+		KubeletConfiguration: *s.kubeconfig,
+	}
+
+	kubeletDeps, err := kubelet.UnsecuredDependencies(kubeletServer, utilfeature.DefaultFeatureGate)
+	if err != nil {
+		logrus.Fatalf("Error in fetching depenedencies %v", err)
+	}
+	if err := kubelet.Run(ctx, kubeletServer, kubeletDeps, utilfeature.DefaultFeatureGate); err != nil {
```
rootfs, line 128: add a readiness probe (similar to kube-controller-manger or kube-scheduler), using kubelet healthz option `healthzPort` https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/

pkg/node/kubelet.go
```diff
@@ -47,17 +86,50 @@ func StartKubelet(cfg *config.MicroshiftConfig) error {
...
+	kubeletoptions.AddGlobalFlags(cleanFlagSet)
 
+	if err := cmd.ParseFlags(args); err != nil {
+		logrus.Fatalf("failed to parse flags:%v", err)
+		return err
+	}
+	s.kubeletflags = kubeletFlags
+	s.kubeconfig = kubeletConfig
+	s.kubeconfigfile = filepath.Join(cfg.DataDir, "resources", "kubelet", "kubeconfig")
+	logrus.Infof("Starting kubelet %s, args: %v", cfg.HostIP, args)
```
rootfs, line 110: HostIP -> NodeIP

husky-parul: It created a merge conflict. I have rebased the PR branch

pkg/node/kubelet.go
```diff
@@ -47,19 +87,51 @@ func StartKubelet(cfg *config.MicroshiftConfig) error {
...
+	kubeletoptions.AddKubeletConfigFlags(cleanFlagSet, kubeletConfig)
+	kubeletoptions.AddGlobalFlags(cleanFlagSet)
+
+	if err := cmd.ParseFlags(args); err != nil {
+		logrus.Fatalf("failed to parse flags:%v", err)
+		return err
+	}
+	s.kubeletflags = kubeletFlags
+	s.kubeconfig = kubeletConfig
+	s.kubeconfigfile = filepath.Join(cfg.DataDir, "resources", "kubelet", "kubeconfig")
```
rootfs, line 110: is `kubeconfigfile` used after assignment?

husky-parul: It's not.

pkg/node/kubelet.go
```diff
@@ -116,6 +116,17 @@ func (s *Kubelet) configure(cfg *config.MicroshiftConfig) error {
 func (s *Kubelet) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
 	defer close(stopped)
 
+	// run readiness check
+	go func() {
+		healthcheckStatus := util.RetryInsecureHttpsGet("https://127.0.0.1:10248/healthz")
+		if healthcheckStatus != 200 {
+			logrus.Fatalf("Kube-controller-manager failed to starttttttt")
```
oglok, line 123: Kubelet instead of kube-ctrl-mgr  and remove repeated "t"s :-)

pkg/node/kubelet.go
```diff
@@ -116,6 +116,17 @@ func (s *Kubelet) configure(cfg *config.MicroshiftConfig) error {
...
+		}
+
+		logrus.Infof("%s is readyyy", s.Name())
```
oglok, line 126: Remove repeated "y"s

pkg/node/kubelet.go
```diff
@@ -16,19 +16,59 @@ limitations under the License.
...
 	kubelet "k8s.io/kubernetes/cmd/kubelet/app"
+	kubeletoptions "k8s.io/kubernetes/cmd/kubelet/app/options"
+	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
 )
 
-func StartKubelet(cfg *config.MicroshiftConfig) error {
-	command := kubelet.NewKubeletCommand()
+const (
+	// Kubelet component name
+	componentKubelet = "kubelet"
```
oglok, line 41: For consistency, either we use the const in all places (like L57), or we just remove the const and use the string "kubelet".

pkg/node/kubelet.go
husky-parul: Sorry I added this for easy grep search in microshift logs for the sake of debugging. The e2e are failing and I was trying to debug it. I will remove it.

oglok:
Two comments on this PR.

- Testing your PR, It seems to be failing in my env:

`FATA[0011] Kubelet failed to start failed to run Kubelet: running with swap on is not supported, please disable swap! or set --fail-swap-on flag to false. /proc/swaps contained: [Filename                    Type             Size            Used            Priority /dev/nvme0n1p6                          partition      8202236         709804          -2] `

-  Also, Kustomizer was moved to a different place.

oglok:
> Two comments on this PR.
> 
> * Testing your PR, It seems to be failing in my env:
> 
> `FATA[0011] Kubelet failed to start failed to run Kubelet: running with swap on is not supported, please disable swap! or set --fail-swap-on flag to false. /proc/swaps contained: [Filename Type Size Used Priority /dev/nvme0n1p6 partition 8202236 709804 -2] `
> 
> * Also, Kustomizer was moved to a different place.

Disabling swapping gets it to work. However, I don't know if that's a requirement for Kubelet. wdyt @fzdarsky ?

I also noticed a couple of deprecation warnings that we might want to address now:

```
Flag --cgroup-driver has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.                                                                                                                                                                                         Flag --volume-plugin-dir has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.      
```

fzdarsky:
> Disabling swapping gets it to work. However, I don't know if that's a requirement for Kubelet. wdyt @fzdarsky ?

Current best practice is to ensure swap is off on the host, as swap may lead to various resource management issues in kubelet. However, I don't think we can / want to generally assume this, e.g. on a developer's machine. Also, I understand swap will be supported down the line (see https://kubernetes.io/blog/2021/08/09/run-nodes-with-swap-alpha/).

To not fail on swap being enabled, the `--fail-swap-on=false` flag can be used I believe.

fzdarsky:
> I also noticed a couple of deprecation warnings that we might want to address now:
> 
> ```
> Flag --cgroup-driver has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
> 
> Flag --volume-plugin-dir has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.      
> ```

I agree it'd be great to fix this while we're at it.

pkg/node/kubelet.go
mangelajo: Another option is to use constants once you have more than one reference to the same string, that's what we did @ submariner-

pkg/node/kubelet.go
```diff
@@ -16,26 +16,61 @@ limitations under the License.
...
+		logrus.Infof("Failed to create a new kubelet configuration: %v", err)
+		return err
+	}
+	// Prepare commandline args
 	args := []string{
 		"--config=" + cfg.DataDir + "/resources/kubelet/config/config.yaml",
 		"--bootstrap-kubeconfig=" + cfg.DataDir + "/resources/kubelet/kubeconfig",
 		"--kubeconfig=" + cfg.DataDir + "/resources/kubelet/kubeconfig",
 		"--container-runtime=remote",
 		"--container-runtime-endpoint=/var/run/crio/crio.sock",
```
rootfs, line 71: per recent fix #248 this is now "unix://..."

pkg/node/kubelet.go
```diff
@@ -16,26 +16,61 @@ limitations under the License.
...
 		"--runtime-cgroups=/system.slice/crio.service",
+		"--cgroup-driver=systemd",
```
rootfs, line 73: this is already set [here](https://github.com/redhat-et/microshift/blob/main/pkg/config/kubelet.go#L37)

pkg/node/kubelet.go
```diff
@@ -46,18 +81,67 @@ func StartKubelet(cfg *config.MicroshiftConfig) error {
...
+	cleanFlagSet := pflag.NewFlagSet(componentKubelet, pflag.ContinueOnError)
+	cleanFlagSet.SetNormalizeFunc(cliflag.WordSepNormalizeFunc)
 
-	if err := command.ParseFlags(args); err != nil {
-		logrus.Fatalf("failed to parse flags:%v", err)
+	kubeletFlags := kubeletoptions.NewKubeletFlags()
+	kubeletConfig, err := kubeletoptions.NewKubeletConfiguration()
+	// programmer error
+	if err != nil {
+		logrus.Fatalf("programmer error %v", err)
```
rootfs, line 91: maybe a more informative error message?

husky-parul:
> > Disabling swapping gets it to work. However, I don't know if that's a requirement for Kubelet. wdyt @fzdarsky ?
> 
> Current best practice is to ensure swap is off on the host, as swap may lead to various resource management issues in kubelet. However, I don't think we can / want to generally assume this, e.g. on a developer's machine. Also, I understand swap will be supported down the line (see https://kubernetes.io/blog/2021/08/09/run-nodes-with-swap-alpha/).
> 
> To not fail on swap being enabled, the `--fail-swap-on=false` flag can be used I believe.

It is set in [/pkg/config/kubelet.go/KubeletConfig]( https://github.com/redhat-et/microshift/blob/main/pkg/config/kubelet.go#L39)

PR #218, opened 2021-08-23
Add design documentation

fzdarsky:


docs/design/design.md
```diff
@@ -0,0 +1,116 @@
...
+### Overall Architecture
+* MicroShift is an application deployed onto a running OS, preferably as container on `podman`, managed through `systemd`. As such, it cannot assume any responsiblity or control over the device or OS it runs on, including OS software or configuration updates or typical device management tasks such as configuring host CA certs or host telemetry. 
+* MicroShift runs as a single binary embedding as goroutines only those services strictly necessary to bring up a *minimal Kubernetes/OpenShift control and data plane*. Motivation:
+  * Maximizes reproducibility; cluster will come up fully or not at all.
+  * Does not require external orchestration, for example through operators, and allows for very fast start-up/update times.
+  * Makes it simple to grok as workload for a a Linux admin persona, works well / easier to implement with systemd.
+  * Smaller resource footprint has _not_ been a motivation, it may be a welcome side-effect.
+* MicroShift provides a small, optional set of infrastructure services to support common use cases and reuses OpenShift's container images for these:
+  * openshift-dns, openshift-ingress, service-ca, local storage provider
```
rootfs, line 68: openshift-ingress -> openshift router

docs/design/design.md
```diff
@@ -0,0 +1,116 @@
...
+* MicroShift instances (processes) run directly on the host or containerized on Podman. They can take on the roles of Control Plane, Node, or both:
+  * Instances with Control Plane role run etcd and the Kubernetes and OpenShift control plane services. As these services don't require a kubelet, pure Control Plane instances are not nodes in the Kubernetes sense and require fewer system privileges.
+  * Instances with Node role run a kubelet (and thus register as node) and kubeproxy and interface with CRI-O for running workloads. They may thus require higher system privileges.
```
rootfs, line 71: kubeproxy -> kube-proxy

screeley44:
@fzdarsky - thanks for this, my initial observation is I think it would be beneficial to have a *use case/user story* section as well.

docs/design/design.md
```diff
@@ -0,0 +1,116 @@
...
+* We ensure the tip of our development branch is deployable and while MicroShift is still early days and experimental we expect developers (and users who want the "latest") to build & deploy from source.
+* Releases are mainly provided for convenience to useres that just want to give MicroShift a quick try without friction. They are cut irregularly, e.g. to make a new feature available.
+* When rebasing onto a new OKD version, we vendor that version's packages and update the container image digests of the infrastructure services MicroShift deploys, i.e. the "release metadata" is baked into the MicroShift binary.
+* Eventually, we expect there to be a "MicroShift Release Image" that is based on / derived from the OpenShift Release Image: It references the MicroShift container image plus the subset of container images shared with and published by OpenShift. Defining a release image should allow to reuse the proven OpenShift CI and release toolinng later.
```
jbpratt, line 116:
```suggestion
* Eventually, we expect there to be a "MicroShift Release Image" that is based on / derived from the OpenShift Release Image: It references the MicroShift container image plus the subset of container images shared with and published by OpenShift. Defining a release image should allow to reuse the proven OpenShift CI and release tooling later.
```
don't mind me :smile_cat: just browsing through

docs/design/design.md
fzdarsky: Fixed, thanks.

docs/design/design.md
fzdarsky: I'll happily take that :). Fixed, thanks.

fzdarsky:
> @fzdarsky - thanks for this, my initial observation is I think it would be beneficial to have a _use case/user story_ section as well.

That's a good idea @screeley44. I'll add those in a follow-up PR.

mangelajo:
@mangelajo please review %)

docs/design/design.md
```diff
@@ -0,0 +1,116 @@
...
+  * Maximizes reproducibility; cluster will come up fully or not at all.
+  * Does not require external orchestration, for example through operators, and allows for very fast start-up/update times.
+  * Makes it simple to grok as workload for a a Linux admin persona, works well / easier to implement with systemd.
+  * Smaller resource footprint has _not_ been a motivation, it may be a welcome side-effect.
+* MicroShift provides a small, optional set of infrastructure services to support common use cases and reuses OpenShift's container images for these:
+  * openshift-dns, openshift-router, service-ca, local storage provider
+* MicroShift instances (processes) run directly on the host or containerized on Podman. They can take on the roles of Control Plane, Node, or both:
+  * Instances with Control Plane role run etcd and the Kubernetes and OpenShift control plane services. As these services don't require a kubelet, pure Control Plane instances are not nodes in the Kubernetes sense and require fewer system privileges.
```
mangelajo, line 70: I'm not sure this bullet point is completely correct.

For example, in-cluster k8s-API access by pods. Some CNIs, (and submariner as another example), require running pods on the control nodes too. To make sure that pods can reach (via pod network) the control nodes too. Normally those nodes are tainted to avoid regular user workloads from crawling in.

I guess that as long as we can have the possibility of the control plane running kubelet it should be fine.

docs/design/design.md
```diff
@@ -0,0 +1,116 @@
...
+  * Instances with Node role run a kubelet (and thus register as node) and kube-proxy and interface with CRI-O for running workloads. They may thus require higher system privileges.
+* While it's possible to run a single MicroShift instance with both Control Plane and Node roles, there may be reasons to run two instances - one Control Plane and one Node - on the same host, e.g. to run the Control Plane with fewer privileges for security reasons. Implementation decisions should consider this.
```
mangelajo, line 72: +1 I see now, that makes sense. Security first.

docs/design/design.md
```diff
@@ -0,0 +1,116 @@
...
+* Open issues / questions:
+  * Lightweight CNI?
+  * API Load Balancing?
+  * Service Load Balancing?
+  * Ingress?
+
+### Storage
+* MicroShift defaults to local ephemeral storage (enough for basic use cases).
+* Open issues / questions:
+  * Provide escape hatch to add own CSI (which?).
```
mangelajo, line 108: I believe something like NFS is internally provided by k8s, but I will try to spin it with our current release of microshift.

mangelajo: https://github.com/kubernetes/examples/blob/master/staging/volumes/nfs/nfs-pv.yaml
https://github.com/kubernetes/examples/blob/master/staging/volumes/nfs/nfs-pvc.yaml
(at least via manual provisioning)

docs/design/design.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Production / Supply Chain / Release Management
+* MicroShift vendors OKD source code without modification. Where it deploys container images for additional services, it deploys OKD's published container images, not the OpenShift downstream's.
+* MicroShift's versioning scheme follows OKD's. This scheme signals the base OpenShift version (4.x) and order/age of builds, but intentionally avoids signaling patch level, backward compatibility (as SemVer, for example), or stability.
+* We ensure the tip of our development branch is deployable and while MicroShift is still early days and experimental we expect developers (and users who want the "latest") to build & deploy from source.
+* Releases are mainly provided for convenience to useres that just want to give MicroShift a quick try without friction. They are cut irregularly, e.g. to make a new feature available.
```
mangelajo, line 114:
```suggestion
* Releases are mainly provided for convenience to users that just want to give MicroShift a quick try without friction. They are cut irregularly, e.g. to make a new feature available.
```

mangelajo (changes requested):
Some comments, great write up.

docs/design/design.md
mangelajo: After reading the other points this requires no action :)

docs/design/design.md
mangelajo: I would need to check this works on microshift today without changes, then I can submit a follow up PR to comment on this may be?

fzdarsky: We don't need to solve all the questions now. This is just an initial PR to document the design decisions we've made earlier and open questions we are already aware of. I'd expect this to be a living document that we add to over time as our understanding increases.

PR #220, opened 2021-08-24
Use systemd resolv.conf in case it exists

carlosedp:
In hosts where resolv.conf is controlled by systemd-resolved, the file
/etc/resolv.conf contains only the localhost IP "127.0.0.53" that is
managed by systemd-resolved. The correct file with upstream servers
in this case is /run/systemd/resolve/resolv.conf that if exists
is appended to Kubelet config to be used on containers that use
"dnsPolicy: Default".

Signed-off-by: Carlos de Paula <me@carlosedp.com>

### DIFF
--- a/pkg/config/kubelet.go
+++ b/pkg/config/kubelet.go
@@ -16,6 +16,7 @@ limitations under the License.
 package config
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


pkg/config/kubelet.go
```diff
@@ -60,6 +61,9 @@ featureGates:
   ServiceNodeExclusion: true
   SupportPodPidsLimit: true
 serverTLSBootstrap: false #TODO`)
+	if _, err := os.Stat("/run/systemd/resolve/resolv.conf"); !errors.Is(err, os.ErrNotExist) {
```
rootfs, line 64: please add a comment and reference the coredns recommendation

carlosedp: Added! Thanks for pointing this out.

PR #221, opened 2021-08-24
Adds a CONTRIBUTING.md file 

osilkin98:
Adds a `CONTRIBUTING.md` file containing instructions on setting up a Vagrant environment, integrating it with VS Code, and building microshift locally.


CONTRIBUTING.md
```diff
@@ -0,0 +1,176 @@
...
+
+# Ubuntu
+sudo apt install \
+    git \
+    build-essential \   # provides Make
+    golang-go \
+    glibc
+```
+
+You will also need to have `kubectl` and `docker` installed, you can follow
```
fzdarsky, line 147: Our target container engine is actually `podman`. `docker` is not tested and so may or may not work.

CONTRIBUTING.md
```diff
@@ -0,0 +1,176 @@
...
+the install guides here to do so for your appropriate system:
+
+- [Docker](https://docs.docker.com/engine/install/fedora/)
+- [Kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux)
```
fzdarsky, line 151: `kubectl` is installed by the install.sh script referenced below, so it's probably not necessary to call it out here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,176 @@
...
+
+Now we can build Microshift:
+
+```sh
+make build
+CONFIG_ENV_ONLY=true ./install.sh
+sudo microshift run 
+```
+
+*Note*: If `./install.sh` fails, you will need to go into the file and comment
```
fzdarsky, line 169: Rather than providing a work-around, I'd prefer to fix the issue. I can't reproduce the problem on Linux, though, the current install.sh just WFM.

fzdarsky (commented):
@osilkin98 thanks for the great write-up!

One thing I'd like to include (maybe in follow-up PR) is expectation on the type and style of contributions, e.g. guidance on when an enhancement will need to have a design doc before a PR can be accepted, that each PR needs an issue, that we'll merge PRs using squash commits, etc.

PR #241, opened 2021-09-01
kube-proxy conforms to internal service manager

### LINKED ISSUE #229: Refactor kubeproxy to conform to the internal service manager
#### Design Document Link

PR: #<PR Number>
Parent issue #161

husky-parul:

**Which issue(s) this PR addresses**: https://github.com/redhat-et/microshift/issues/229

Closes #229

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -91,13 +91,11 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 			},
 		)))
 		util.Must(m.AddService(kustomize.NewKustomizer(cfg)))
-
 	}
 
 	if config.StringInList("node", cfg.Roles) {
 		util.Must(m.AddService(node.NewKubeletServer(cfg)))
 		util.Must(m.AddService(node.NewKubeProxyServer(cfg)))
-
 	}
 
 	logrus.Info("Starting Microshift")
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -96,19 +96,8 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 
 	if config.StringInList("node", cfg.Roles) {
 		util.Must(m.AddService(node.NewKubeletServer(cfg)))
-		util.Must(m.AddService(servicemanager.NewGenericService(
-			"other-node",
-			[]string{"kube-apiserver"},
-			func(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
-				defer close(stopped)
-				defer close(ready)
+		util.Must(m.AddService(node.NewKubeProxyServer(cfg)))
 
-				if err := node.StartKubeProxy(cfg); err != nil {
-					return err
-				}
-				return nil
-			},
-		)))
 	}
 
 	logrus.Info("Starting Microshift")
--- a/pkg/node/kubelet.go
+++ b/pkg/node/kubelet.go
@@ -189,7 +189,7 @@ func loadConfigFile(name string) (*kubeletconfig.KubeletConfiguration, error) {
 
 const (
 	// proxy component name
-	component = "kubeproxy"
+	COMPONENT_KUBEPROXY = "kube-proxy"
 )
 
 type ProxyOptions struct {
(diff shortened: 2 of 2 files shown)


pkg/node/kubelet.go
```diff
@@ -84,3 +87,74 @@ func StartKubeProxy(cfg *config.MicroshiftConfig) error {
...
+func (s *ProxyOptions) Name() string           { return component }
+func (s *ProxyOptions) Dependencies() []string { return []string{"kube-apiserver"} }
+
+func (s *ProxyOptions) configure(cfg *config.MicroshiftConfig) error {
+	if err := config.KubeProxyConfig(cfg); err != nil {
+		logrus.Infof("Failed to create a new kube-proxy configuration: %v", err)
+		return err
+	}
+	args := []string{
+		"--logtostderr=" + strconv.FormatBool(cfg.LogDir == "" || cfg.LogAlsotostderr),
```
husky-parul, line 115: @rootfs & @fzdarsky Do we need to add `--congfig` in `args`?  `kubeProxy.Options` has a mechanism to read a config file using 
```
opts := kubeproxy.NewOptions()
opts.ConfigFile = cfg.DataDir + "/resources/kube-proxy/config/config.yaml"
opts.Complete()
```
`opts.Complete()` loads configuration from a file.
Right now I am not injecting `KubeProxyConfiguration` using `--config` flag and it works.

pkg/cmd/run.go
```diff
@@ -105,12 +105,13 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 				if err := node.StartKubelet(cfg); err != nil {
 					return err
 				}
-				if err := node.StartKubeProxy(cfg); err != nil {
-					return err
-				}
+				// if err := node.StartKubeProxy(cfg); err != nil {
+				// 	return err
+				// }
```
oglok, line 110: This should be removed.

pkg/cmd/run.go
```diff
@@ -105,12 +105,13 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
...
 				return nil
 			},
 		)))
+		util.Must(m.AddService(kustomize.NewKustomizer(cfg)))
```
oglok, line 114: Why is kustomizer service moved from "only control plane" to all kinds of hosts?

pkg/node/kubelet.go
```diff
@@ -84,3 +87,74 @@ func StartKubeProxy(cfg *config.MicroshiftConfig) error {
...
+	logrus.Infof("starting kube-proxy, args: %v", args)
+	return nil
+}
+
+func (s *ProxyOptions) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+
+	defer close(stopped)
+	// run readiness check
+	// go func() {
+	// 	healthcheckStatus := util.RetryInsecureHttpsGet("https://127.0.0.1:10250/healthz")
```
oglok, line 148: The default port is 10256.

pkg/node/kubelet.go
```diff
@@ -84,3 +87,74 @@ func StartKubeProxy(cfg *config.MicroshiftConfig) error {
 
 	return nil
 }
+
+const (
+	// proxy component name
+	component = "kubeproxy"
+)
+
+type ProxyOptions struct {
```
oglok, line 96: Not that it matters in terms of functionality, but for consistency with other components, this struct should be called after the component itself.

oglok (changes requested):
Once the readiness probe is implemented, the rest are just minor comments. Thanks for the great work @husky-parul !

pkg/node/kubelet.go
husky-parul: Yeah I tested with 10256 too but the server is not up

pkg/cmd/run.go
husky-parul: Must have been a typing error. Will fix it.

pkg/node/kubelet.go
fzdarsky: I'm totally fine doing that.

pkg/node/kubelet.go
rootfs: it should work
```console
# curl http:netstat -anp |grep 10256
tcp6       0      0 :::10256                :::*                    LISTEN      3743888/_output/bin
# curl http://127.0.0.1:10256/healthz
{"lastUpdated": "2021-09-07 14:47:41.024941778 -0400 EDT m=+85.884851029","currentTime": "2021-09-07 14:47:41.024941778 -0400 EDT m=+85.884851029"}
```

husky-parul: Using `http` works

pkg/node/kubelet.go
husky-parul: @oglok I noticed we do not follow the golang naming conventions for constants. Is it ok if I use golang standard `COMPONENT_KUBEPROXY = "kube-proxy"` instead of `componentKubeProxy = "kubeproxy"`

PR #256, opened 2021-09-07
Tag latest

### LINKED ISSUE #230: Release.sh should append "latest" tag to manifests
The release.sh tool does not set the latest tag on the generated manifest.  In addition to the microshift tag, we should also set latest as an enhancement to UX.

The manifest tag is specified at https://github.com/redhat-et/microshift/blob/09579efcc7a65b1f4f4abe451c560df101fc4447/scripts/release.sh#L170-L185

copejon:

Closes #230

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -173,6 +173,7 @@ podman_create_manifest(){
     podman manifest add "$IMAGE_REPO:$VERSION" "docker://$ref"
   done
     podman manifest push "$IMAGE_REPO:$VERSION" "$IMAGE_REPO:$VERSION"
+    podman manifest push "$IMAGE_REPO:$VERSION" "$IMAGE_REPO:latest"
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -56,8 +56,8 @@ Inputs:
     --help, -h    Print this help text.
 Outputs:
 - A version, formatted as 4.7.0-0.microshift-YYYY-MM-DD-HHMMSS, is applied as a git tag and pushed to the repo
+- Multi-architecture container manifest, tagged as quay.io/microshift/microshift:$VERSION && latest
```
rootfs, line 59: how about highlight them? e.g. 
```text
`quay.io/microshift/microshift:$VERSION` and `latest`
```

copejon:
@rootfs ptal

PR #257, opened 2021-09-07
Adds SecurePort parameterization 

### LINKED ISSUE #142: CLI option to set API server port
Currently the API port `6443` is hardcoded. This becomes an issue when the port is taken (especially when multiple microshift clusters are running on the same host). Need to parameterize API server port.

osilkin98:
Introduces a `SecurePort int` field to the `MicroshiftConfig` struct which is used to set the secure port that the API server binds to. This port can either be set through the CLI using `--secure-port`, through the `MICROSHIFT_SECUREPORT` environment variable, or the `securePort` attribute in the config file. 

Additionally, since `ClusterConfig` contains the URL which the cluster runs on, a `ConfigureClusterURL` method is added to Microshift which inserts the provided `SecurePort` into the `ClusterConfig.URL` field. 

Closes #142


pkg/config/config.go
```diff
@@ -197,9 +204,23 @@ func (c *MicroshiftConfig) ReadAndValidate(flags *pflag.FlagSet) error {
...
 	}
 
+	c.ConfigureClusterURL()
 	return nil
 }
 
+// configures the Cluster URL to have the provided Secure Port
+func (c *MicroshiftConfig) ConfigureClusterURL() {
+	re := regexp.MustCompile(`:(\d+)`)
+	if re.MatchString(c.Cluster.URL) {
```
rootfs, line 214: we need to validate both ipv4 and ipv6 addresses, you can reuse `IsValidSocketAddr`

```go
import "k8s.io/apimachinery/pkg/util/validation"

validation.IsValidSocketAddr("1.2.3.4:5678")
```

pkg/cmd/run.go
```diff
@@ -42,12 +43,22 @@ func NewRunMicroshiftCommand() *cobra.Command {
...
 	cmd.MarkFlagFilename("config", "yaml", "yml")
 	// All other flags will be read after reading both config file and env vars.
+	flags.Int("secure-port", cfg.SecurePort, "Port to bind this Microshift cluster to.")
 	flags.String("data-dir", cfg.DataDir, "Directory for storing runtime data.")
 	flags.StringSlice("roles", cfg.Roles, "Roles of this Microshift instance.")
 
 	return cmd
 }
 
+func getJsonString(v interface{}) string {
```
rootfs, line 53: is this function used?

osilkin98: This was a leftover debug method, I'll remove it.

pkg/config/config.go
osilkin98: Do you mean validating the addresses after we inject the Secure Port into the URL? 
```golang
// configures the Cluster URL to have the provided Secure Port
func (c *MicroshiftConfig) ConfigureClusterURL() error {
	re := regexp.MustCompile(`:(\d+)`)
	var proposedUrl string
	if re.MatchString(c.Cluster.URL) {
		proposedUrl = re.ReplaceAllString(c.Cluster.URL, fmt.Sprintf(":%d", c.SecurePort))
	} else {
		// if the cluster url doesn't have a port then we must insert it
		re = regexp.MustCompile(`^(\w+)://([\w\.]+)(.*)$`)
		pieces := re.FindStringSubmatch(c.Cluster.URL)
		proposedUrl = pieces[1] + "://" + pieces[2] + ":" + strconv.Itoa(c.SecurePort) + pieces[3]
	}

	// there are errors with our socket address
	errs := validation.IsValidSocketAddr(proposedUrl)
	if 0 < len(errs) {
		return fmt.Errorf("config error: %s", strings.Join(errs, "; "))
	}

	c.Cluster.URL = proposedUrl
	return nil
}
```

osilkin98: The usage of `re.MatchString` here is to determine if there is a port within the URL, and if not then to inject the port binding. Although the regex pattern should probably be updated as well so as not to confuse the port number with address bits.

rootfs: I see, that is a good point. Here we should only support `https` (i.e. secure), so `pieces[1]` should be overriden if it is not `https`. Then `validation.IsValidSocketAddr(piece[2] + ":" + strconv.Itoa(c.SecurePort)`  to validate the address format is right. Do you  see a case where `pieces[3]` is not empty?

osilkin98: @rootfs Should we be doing the URL validation in this function, or would it make more sense to have a hook that checks the settings post-configuration? I know that this code isn't scalable but my intuition makes me think that we should assume what comes in will be a valid URL, and what we should end up with is a valid URL. 
Could we do those checks within some unit-tests?

pkg/cmd/run.go
```diff
@@ -42,6 +42,7 @@ func NewRunMicroshiftCommand() *cobra.Command {
 	flags.StringVar(&cfg.ConfigFile, "config", cfg.ConfigFile, "File to read configuration from.")
 	cmd.MarkFlagFilename("config", "yaml", "yml")
 	// All other flags will be read after reading both config file and env vars.
+	flags.Int("secure-port", cfg.SecurePort, "Port to bind this Microshift cluster to.")
```
rootfs, line 45: to make it more intuitive, `secure kube api port`

cooktheryan:
@rootfs @osilkin98 what do we want to do with this PR? Should we move forward or is it covered by Franks change

rootfs:
@cooktheryan discussed with @osilkin98 to convert to a doc PR, detailing how to use config to map out different URL and ports

husky-parul:
@osilkin98 Please have a look into this. We need to rebase and merge. Thanks

Issue #239, opened 2021-08-31
Refactor OCPControllerManager to conform to the internal service manager


husky-parul:
#### Design Document Link

PR: #
Parent issue #161

husky-parul:
@rootfs @fzdarsky 
What is the intent of `util.Must(m.AddService(controllers.NewOpenShiftPrepJob()))` in run.go ?

fzdarsky:
Yeah, it was a placeholder for a service that runs to completion doing the PrepareOCP work. Just wondering, though, whether it logically belongs into one of the other services.

PR #260, opened 2021-09-09
Refactor OCPControllerManager to conform to the internal service manager

husky-parul:
Refactor OCPControllerManager to conform to the internal service manager #239

Closes#239

### DIFF
--- a/hack/all-in-one/README.md
+++ b/hack/all-in-one/README.md
@@ -68,17 +68,17 @@ kubectl get pods -A
 ### Access the cluster on the host
 #### Linux
 ```bash
-export KUBECONFIG=$(podman volume inspect ushift-vol --format "{{.Mountpoint}}")/microshift/resources/kubeadmin/kubeconfig
+export KUBECONFIG=$(podman volume inspect microshift-data --format "{{.Mountpoint}}")/microshift/resources/kubeadmin/kubeconfig
 kubectl get pods -A -w
 ```
 #### MacOS
 ```bash
-docker cp ushift:/var/lib/microshift/resources/kubeadmin/kubeconfig ./kubeconfig
+docker cp microshift-aio:/var/lib/microshift/resources/kubeadmin/kubeconfig ./kubeconfig
 kubectl get pods -A -w --kubeconfig ./kubeconfig
 ```
 #### Windows
 ```bash
-docker.exe cp ushift:/var/lib/microshift/resources/kubeadmin/kubeconfig .\kubeconfig
+docker.exe cp microshift-aio:/var/lib/microshift/resources/kubeadmin/kubeconfig .\kubeconfig
 kubectl.exe get pods -A -w --kubeconfig .\kubeconfig
 ```
 ## Limitation
--- a/hack/all-in-one/README.md
+++ b/hack/all-in-one/README.md
@@ -68,17 +68,17 @@ kubectl get pods -A
 ### Access the cluster on the host
 #### Linux
 ```bash
-export KUBECONFIG=$(podman volume inspect microshift-data --format "{{.Mountpoint}}")/microshift/resources/kubeadmin/kubeconfig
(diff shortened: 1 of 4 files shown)


pkg/util/net.go
```diff
@@ -52,3 +53,21 @@ func RetryInsecureHttpsGet(url string) int {
 
 	return status
 }
+
+func RetryTCPConnection(host string, port string) int {
+	status := 0
+	err := wait.Poll(5*time.Second, 120*time.Second, func() (bool, error) {
+		timeout := 3 * time.Second
+		_, err := tcpnet.DialTimeout("tcp", tcpnet.JoinHostPort("127.0.0.1", "8445"), timeout)
```
rootfs, line 61: host and port

pkg/util/net.go
```diff
@@ -52,3 +53,21 @@ func RetryInsecureHttpsGet(url string) int {
...
+
+		if err == nil {
+			status = 200
```
rootfs, line 64: 200 is HTTP status. If this is for TCP port probe, returning true/false is good to go

pkg/util/net.go
```diff
@@ -52,3 +53,21 @@ func RetryInsecureHttpsGet(url string) int {
 
 	return status
 }
+
+func RetryTCPConnection(host string, port string) int {
+	status := 0
+	err := wait.Poll(5*time.Second, 120*time.Second, func() (bool, error) {
+		timeout := 3 * time.Second
```
rootfs, line 60: can you check what's the default timeout in `net/http`? 3 seconds could be too short

husky-parul: The default timeout is no timeout so it will keep trying infinitely. 
https://github.com/golang/go/issues/24138

Increasing it to `30 * time.Second`, I see it is the recommended timeout -> https://pkg.go.dev/net/http

hack/all-in-one/README.md
```diff
@@ -68,17 +68,17 @@ kubectl get pods -A
 ### Access the cluster on the host
 #### Linux
 ```bash
-export KUBECONFIG=$(podman volume inspect ushift-vol --format "{{.Mountpoint}}")/microshift/resources/kubeadmin/kubeconfig
+export KUBECONFIG=$(podman volume inspect microshift-data --format "{{.Mountpoint}}")/microshift/resources/kubeadmin/kubeconfig
```
rootfs, line 71: can you revert these changes?

pkg/util/net.go
```diff
@@ -52,3 +53,21 @@ func RetryInsecureHttpsGet(url string) int {
...
 }
+
+func RetryTCPConnection(host string, port string) bool {
+	status := true
+	err := wait.Poll(5*time.Second, 120*time.Second, func() (bool, error) {
+		timeout := 30 * time.Second
+		_, err := tcpnet.DialTimeout("tcp", tcpnet.JoinHostPort(host, port), timeout)
+
+		if err == nil {
+			status = false
```
rootfs, line 64: why set it false?

hack/all-in-one/README.md
husky-parul: I will. Can't get the tests to pass though. :thinking:

pkg/util/net.go
```diff
@@ -52,3 +53,21 @@ func RetryInsecureHttpsGet(url string) int {
...
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil && err == wait.ErrWaitTimeout {
+		logrus.Warningf("Endpoint is not returning any status code")
```
rootfs, line 70: if timeout, this should return false to tell the caller the connection is not ready

hack/all-in-one/README.md
husky-parul: > can you revert these changes?

Will the changes to volume name will go in a different PR or we don't need this change at all? All other places use `microshift-data` as the volume name instead of `ushift-vol`

rootfs:
@oglok @mangelajo please take a final review

PR #267, opened 2021-09-13
Update CONTRIBUTING.md

osilkin98:
Fixes the Microshift run command from `sudo microshift run` to `sudo ./microshift run`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,7 +182,7 @@ Now we can build Microshift:
 ```sh
 make build
 CONFIG_ENV_ONLY=true ./install.sh
-sudo microshift run 
+sudo ./microshift run 
 ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -182,7 +182,7 @@ Now we can build Microshift:
 ```sh
 make build
 CONFIG_ENV_ONLY=true ./install.sh
-sudo microshift run 
+sudo ./microshift run
```
rootfs, line 185: so install.sh just downloads microshift in current dir?

osilkin98: `install.sh` is only run here to configure the environment, the script exits before anything is actually downloaded; `./microshift` here is the binary generated from `make build`

PR #269, opened 2021-09-14
build and enable selinux

cooktheryan:
Completes the integration of Selinux

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -52,14 +52,13 @@ register_subs() {
 }
 
 # Apply SElinux policies
-apply_selinux_policy() {
-    # sudo semanage fcontext -a -t container_runtime_exec_t /usr/local/bin/microshift ||
-    #   sudo semanage fcontext -m -t container_runtime_exec_t /usr/local/bin/microshift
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -183,6 +182,14 @@ User=root
 WantedBy=multi-user.target
 EOF
 
+    sudo mkdir -p /var/run/flannel
+    sudo mkdir -p /var/run/kubelet
+    sudo mkdir -p /var/lib/kubelet/pods
+    sudo mkdir -p /var/run/secrets/kubernetes.io/serviceaccount
+    sudo mkdir -p /var/hpvolumes
+    sudo semodule -i /tmp/microshift.pp
```
rootfs, line 190: will /tmp survive?

cooktheryan: Once the pp file is loaded it is no longer needed. We could potentially add in a delete of the file once loaded

rootfs: as long as it is persisted, it is good

cooktheryan:
@rootfs i dont know if this needs manually merged.

rootfs:
@cooktheryan manual it is :D

Issue #263, opened 2021-09-13
[BUG] Release builds newer than v0.2 don't run with Ubuntu 20.04


tadayosi:
I understand Ubuntu is not currently the main target distro for Microshift, but the pre-built binaries don't run on Ubuntu. I'm trying to run Microshift on GH Actions but it requires Ubuntu distros at this moment. IMO it's beneficial for the Microshift community to support Ubuntu for wider usages.

#### What happened:

Downloading https://github.com/redhat-et/microshift/releases/download/4.7.0-0.microshift-2021-08-31-224727/microshift-linux-amd64 and running it on Ubuntu 20.04 causes the following error:

```
$ ./microshift-linux-amd64 version
microshift-linux-amd64: dl-call-libc-early-init.c:37: _dl_call_libc_early_init: Assertion `sym != NULL' failed.
SIGABRT: abort
PC=0x427644f m=0 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x4265630, 0xc00009add0, 0x47a36d)
	/usr/lib/golang/src/runtime/cgocall.go:133 +0x5b fp=0xc00009ada0 sp=0xc00009ad68 pc=0x40481b
crypto/internal/boring._Cfunc__goboringcrypto_DLOPEN_OPENSSL(0x0)
	_cgo_gotypes.go:287 +0x4a fp=0xc00009add0 sp=0xc00009ada0 pc=0x6d0cca
crypto/internal/boring.init.0()
	/usr/lib/golang/src/crypto/internal/boring/boring.go:46 +0x4d fp=0xc00009ae08 sp=0xc00009add0 pc=0x6d750d
runtime.doInit(0x7aa8020)
	/usr/lib/golang/src/runtime/proc.go:5652 +0x8a fp=0xc00009ae38 sp=0xc00009ae08 pc=0x448aca
runtime.doInit(0x7aa45a0)
	/usr/lib/golang/src/runtime/proc.go:5647 +0x57 fp=0xc00009ae68 sp=0xc00009ae38 pc=0x448a97
runtime.doInit(0x7ace1c0)
	/usr/lib/golang/src/runtime/proc.go:5647 +0x57 fp=0xc00009ae98 sp=0xc00009ae68 pc=0x448a97
runtime.doInit(0x7acfc80)
[...]

PR #271, opened 2021-09-15
Support Ubuntu for install.sh

tadayosi:
I tested this with Vagrant `ubuntu/focal64` box and it works. Note due to #263, however, at this moment you can only do this:
```
CONFIG_ENV_ONLY=true ./install.sh
```
and you need to install `microshift` yourself with a [local build for Ubuntu](https://github.com/tadayosi/microshift/releases/download/test-ubuntu/microshift-ubuntu.tar.xz).

Other notes on the fix:
- `set -e -o pipefail` is bash specific so it doesn't make sense to use `/bin/sh`. Also `/bin/sh` resolves to `dash` on Ubuntu by default
- SELinux is already disabled by default for Ubuntu

No issues related to the pull req.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 set -e -o pipefail
 
 # Usage:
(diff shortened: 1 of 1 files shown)


rootfs:
@tadayosi can you check if `AppArmor` has any impact on microshfit?

tadayosi:
@rootfs Thanks for merging it. I'm not an AppArmor nor Ubuntu expert, but so far I haven't seen issues like this with Ubuntu:
https://stackoverflow.com/questions/59463038/coredns-cant-start-when-using-crio-and-with-selinux-on

I'll do more extensive tests with Ubuntu and if I find any issues I could submit another pull that disables AppArmor.

PR #281, opened 2021-09-17
Remove CNI IPv6 configuration and use right conmon binary for AIO container images

oglok:


hack/all-in-one/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN export VERSION=$(curl -s https://api.github.com/repos/redhat-et/microshift/r
 
 COPY unit /usr/lib/systemd/system/microshift.service
 COPY kubelet-cgroups.conf /etc/systemd/system.conf.d/kubelet-cgroups.conf
+COPY crio-bridge.conf /etc/cni/net.d/100-crio-bridge.conf
```
rootfs, line 11: can you also add the [conmon change](https://github.com/redhat-et/microshift/blob/main/install.sh#L153) too? Since this is the ubi image

oglok: The conmon change was in crio.conf (not the CNI). Is that also not working in the AIO image?

Issue #278, opened 2021-09-16
[Enhancement]: Provide RPM for MicroShift


fzdarsky:
#### What would you like to be added:
Create SPEC file for packaging MicroShift into an RPM, build RPM, make it available on public repo. That RPM should contain dependencies on the SELinux policies, CRI-O, etc.

#### Why is this needed:
For running MicroShift on RHEL hosts non-containerised in production, we need RPMs that users can embed into rpm-ostrees.

cooktheryan:
@mangelajo same with this one

mangelajo:
I think we can close this, https://copr.fedorainfracloud.org/coprs/g/redhat-et/

Thank's for bug-scrubbing, I'm terrible at this.

PR #288, opened 2021-09-20
Create initial spec for RPM packaging

mangelajo:
This can be built with the following commands for a release (download the source from github)
```
$ spectool -g -R microshift.spec
$ rpmbuild -ba microshift.spec
```

And with the following commands for local development and changes:

```
# build an rpm for your local architecture and distro
$ make devel-rpm

# build an srpm which you can upload to a builder service
$ make devel-srpm
```
Related-Issue: #278

### DIFF
--- a/microshift.spec
+++ b/microshift.spec
@@ -1,0 +1,70 @@
+%{!?github_tag: %global github_tag 4.7.0-0.microshift-2021-08-31-224727}
+%{!?version: %global version 4.7.0}
+%{!?release: %global release 2021_08_31_224727}
+
+# golang specifics
+%global golang_version 1.15
+#debuginfo not supported with Go
+%global debug_package %{nil}
+# modifying the Go binaries breaks the DWARF debugging
+%global __os_install_post %{_rpmconfigdir}/brp-compress
+
+Name: microshift
+Version: %{version}
+Release: %{release}%{dist}
+# this can be %{timestamp}.git%{short_hash} later for continous main builds
+Summary: Microshift binary
+License: ASL 2.0
+URL: https://github.com/redhat-et/microshift
+
+Source0: https://github.com/redhat-et/microshift/archive/refs/tags/%{github_tag}.tar.gz
+
+%if 0%{?go_arches:1}
+ExclusiveArch: %{go_arches}
+%else
+ExclusiveArch: x86_64 aarch64 ppc64le s390x
+%endif
+
+BuildRequires: gcc
+BuildRequires: glibc-static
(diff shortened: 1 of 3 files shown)


microshift.spec
```diff
@@ -0,0 +1,70 @@
...
+%ifarch x86_64
+GOARCH=amd64
+%endif
+
+make _build_local GOOS=${GOOS} GOARCH=${GOARCH}
+cp ./_output/bin/${GOOS}_${GOARCH}/microshift ./_output/microshift
+
+%install
+install -d %{buildroot}%{_bindir}
+install -p -m 755 ./_output/microshift %{buildroot}%{_bindir}/microshift
```
rootfs, line 63: can you also add selinux modules here? or it goes to another rpm?

microshift.spec
```diff
@@ -0,0 +1,70 @@
...
+# golang specifics
+%global golang_version 1.15
+#debuginfo not supported with Go
+%global debug_package %{nil}
+# modifying the Go binaries breaks the DWARF debugging
+%global __os_install_post %{_rpmconfigdir}/brp-compress
+
+Name: microshift
+Version: %{version}
+Release: %{release}%{dist}
```
rootfs, line 14: where is `dist` defined?

microshift.spec
mangelajo: Yes, that was my plan as per conversation in https://github.com/redhat-et/microshift/issues/277#issuecomment-922700554

I can do that if @cooktheryan is ok :)

microshift.spec
mangelajo: It comes from rpmbuild, normally the distro where you are building, or, a mock envionment created by the builder that you request for a specific distro.

I passed the output SRPM from this, and https://copr.fedorainfracloud.org/coprs/mangelajo/microshift/build/2829853/


I was thinking that we could automate that for every tag, and then have one for every new git commit in the form of "devel" version and "%{timestamp}.git%{commit-sha}" release.

mangelajo:
Thanks for the eyes @rootfs

microshift.spec
rootfs: i see, thanks for the education.

rootfs:
/lgtm
(good for me, waiting for @cooktheryan for selinux and @copejon on build decisions)

mangelajo:
Probably we should also add dependencies (cri-o cri-tools)

cooktheryan:
@mangelajo feel free to package them together. I can get you the required steps to generate the .pp file if you need them. Most items you should be able to derive from the install.sh

PR #292, opened 2021-09-21
add fully aio systemd unit file and script

sallyom:
Signed-off-by: Parul Singh <parsingh@redhat.com> 

This PR partially resolves https://github.com/redhat-et/microshift/issues/276

This PR:
Adds a systemd unit file to run microshift all-in-one with podman
The service:
-  creates `microshift-vol` podman volume if it does not exist
-  starts microshift via podman run command
-  writes location of KUBECONFIG to /etc/microshift-aio/microshift-aio.conf for easy `source /etc/microshift-aio/microshift-aio.conf` upon service startup
-  included the auto-update podman annotation, but this needs testing to be sure it works correctly (ie, not sure which tags, if it changes the tags)

* With `systemctl stop microshift-aoi`
    * microshift pod is stopped/deleted
    * `microshift-vol` is _not_ deleted

### DIFF
--- a/hack/all-in-one/microshift-aio-run
+++ b/hack/all-in-one/microshift-aio-run
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+setsebool -P container_manage_cgroup true
+
+if ! /usr/bin/podman volume exists microshift-vol
+then
+  /usr/bin/podman volume create microshift-vol
+fi
+
+[[ -d /etc/microshift-aio ]] || mkdir /etc/microshift-aio
+
+/usr/bin/podman run -d --rm \
+  --name microshift --privileged \
+  -v /lib/modules:/lib/modules \
+  -v microshift-vol:/var/lib \
+  --hostname microshift \
+  --label "io.containers.autoupdate=registry" \
+  -p 6443:6443 quay.io/microshift/microshift:4.7.0-0.microshift-2021-08-31-224727-aio
+
+cat <<EOF > /etc/microshift-aio/microshift-aio.conf
+export KUBECONFIG=$(/usr/bin/podman volume inspect microshift-vol --format "{{.Mountpoint}}")/microshift/resources/kubeadmin/kubeconfig
+EOF
--- a/hack/all-in-one/microshift-aio.service
+++ b/hack/all-in-one/microshift-aio.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=microshift all-in-one
+Wants=network-online.target
+After=network-online.target
+RequiresMountsFor=/run/containers/storage
+
+[Service]
+Restart=on-failure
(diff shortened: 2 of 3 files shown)


hack/all-in-one/microshift-aio-run
```diff
@@ -0,0 +1,24 @@
...
+set -euxo pipefail
+
+setsebool -P container_manage_cgroup true
+
+if ! /usr/bin/podman volume exists microshift-vol
+then
+  /usr/bin/podman volume create microshift-vol
+fi
+
+[[ -d /etc/microshift-aio ]] || mkdir /etc/microshift-aio
```
rootfs, line 12: what is this for?

hack/all-in-one/microshift-aio.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=microshift all-in-one
+Wants=network-online.target
+After=network-online.target
```
rootfs, line 4: wait for crio

hack/all-in-one/microshift-aio.service
```diff
@@ -0,0 +1,16 @@
...
+RequiresMountsFor=/run/containers/storage
+
+[Service]
+Restart=on-failure
+TimeoutStopSec=70
+ExecStart=/bin/bash /usr/bin/microshift-aio-run
+ExecStop=/usr/bin/podman stop -t 20 microshift 
+ExecStopPost=/usr/bin/podman rm microshift
```
rootfs, line 12: this probably doesn't need, since the container starts with `--rm`

hack/all-in-one/microshift-aio-run
sallyom: I wanted a file with the location of the KUBECONFIG like so:
```
cat <<EOF > /etc/microshift-aio/microshift-aio.conf
export KUBECONFIG=$(/usr/bin/podman volume inspect microshift-vol --format "{{.Mountpoint}}")/microshift/resources/kubeadmin/kubeconfig
EOF
```
I could not see a better way of getting that path from the service, but that is definitely extra and does not need to be included in the service. I figured it saves users from having to look up the path.

sallyom: I also could not figure out how to run `ExecStartPost=source /etc/microshift-aio/microshift-aio.conf` 
I tried `ExecStartPost=source`, `ExecStartPost=/bin/bash source` etc, nothing worked but there is def something that does, I gave up and pushed the commit but will keep looking

sallyom: @husky-parul if you find something that works or if we decide it's not necessary, you can remove it, feel free to squash/amend, too

rootfs: i see, that makes sense

PR #295, opened 2021-09-22
add sudo to podman to avoid rootless

husky-parul:

To be able to run microshift via podman we need to run podman as sudo otherwise the cluster never gets up and running as nodes are stuck in NotReady state.


fzdarsky (changes requested):
Generally good with this. Could we rename `ushift` —> `microshift` while we’re at it, though? We deprecated the use of the former as it was confusing to users. Also, it may make sense to rename the volume to something more descriptive lime `microshift-vol` —> `microshift-data`.

husky-parul:
> Generally good with this. Could we rename `ushift` —> `microshift` while we’re at it, though? We deprecated the use of the former as it was confusing to users. Also, it may make sense to rename the volume to something more descriptive lime `microshift-vol` —> `microshift-data`.

Instead of  `ushift` —> `microshift` I am using ` `ushift` —> `microshift-aio``

PR #301, opened 2021-09-23
Workaround rpmbuild with no --build-in-place support

mangelajo:
Embeds the logic of creating an rpm (and downloading sources
if necessary) in rpm/make-rpm.sh

Make targets have been renamed to rpm and srpm, as the same
command will be able to build both devel and git-tagged based rpms.


mangelajo:
moving to draft for now @rootfs I just realized that the srpm generated is not fully functional in copr, I need to do one extra fix.

PR #305, opened 2021-09-24
Embed git tags, commit ID and tree status in srpm

mangelajo:
Embedding those details, and then injecting then back during
srpm building (instead of looking for the git details which
doesn't exist anymore for an srpm build -it's a tarball-)
allows for microshift being properly compiled with those
details in the binary.

Closes-Issue: #476


mangelajo:
@fzdarsky  ^

mangelajo:
Re-tested manually the srpm / rpm building, this is good to go

PR #308, opened 2021-09-24
[WIP] mount host cri-o on microshift-aio

### LINKED ISSUE #276: [Enhancement]: Add systemd units for production deployments on Podman
`systemd` units for running MicroShift on Podman on RHEL in production. This includes creation of the Podman volumes for keeping MicroShift state, least privileges (seccomp et al.), ensuring MicroShift restarts upon failure, etc. We should also suggest an auto-update strategy. That production deployment should use an external CRI-O instance, so MicroShift container restarts/redeployments don't disrupt workloads (however, we may want to also provide an example for the "all-in-one" MicroShift container with CRI-O).

Potentially useful: [systemd for Podman](https://www.redhat.com/sysadmin/podman-shareable-systemd-services), [auto-updates](https://www.redhat.com/sysadmin/improved-systemd-podman), and [seccomp for containers](https://podman.io/blogs/2019/10/15/generate-seccomp-profiles.html).

#### Why is this needed:
We need a prescriptive way how to do this, so we a) maximise chances users successfully deploy without requiring lots of upfront reading and b) we can provide better test coverage for this scenario.

husky-parul:

This PR partially resolves #276

This PR
- copies cri-o systemd unit file on the microshift-aio 
- mounts host's cri-o as a volume


sallyom:
@husky-parul I have a PR to update the generated unit file, you'll want to update this PR too. Otherwise upon machine restart the service fails due to `container already exists with that name` error. The `--new` flag was missing when we generated the unit file before. 
Also, since `podman generate systemd --new` sets the ExecStart|Stop with the podman cmds, I modified the /usr/bin/microshift-script to only include the setsebool cmd,  recording of the kubeconfig path, & conditional volume create and it should run with the ExecPreStart. Take a look!

PR #315, opened 2021-09-28
update all-in-one systemd, use podman generated ExecStart|Stop

sallyom:

follow-up for https://github.com/redhat-et/microshift/issues/276

This PR updates the systemd ExecStart|Stop commands to be in line with `podman generate systemd --new --name ctrID`. The `--new` flag creates and uses a --cid-file & --pid-file to ensure container is removed upon stop so that when service is enabled and starts with machine start, the podman cmd will not fail due to `container with name microshift-aio already exists` 

Note: The unit file is generated with `Podman 3.3.1` and adds
```
Type=notify
NotifyAccess=all
```
also adds the `--sdnotify=conmon --cgroups=no-conmon` flags to the `podman run` cmd. 

older versions of podman generate this: 
```
Type=forking
```


sallyom:
/assign @fzdarsky 

I confirmed with VM restart that there are no issues with service startup. I also asked in slack w/ container-engines people if the `Type=notify` is only with the most recent podman version. Last time (last week) I used podman generate systemd it produces a unit file with `Type=forking` :shrug:

PR #316, opened 2021-09-29
[WIP] cluster policy controller 

### LINKED ISSUE #303: [Enhancement]: Add cluster-policy-controller
Add the [cluster-policy-controller](https://github.com/openshift/cluster-policy-controller) as an additional embedded service into the MicroShift binary.

#### Why is this needed:
The `cluster-policy-clontroller` allocates UIDs and SELinux labels for namespaces, allowing for example services like the service-ca as non-root user with randomised UID. It also manages cluster quota usage. Without it, manifests would need to be changed between OpenShift and Microshift, while the goal is to allow for workload portability.

sallyom:
This PR:
 - adds cluster-policy-controller as deployment

Closes #303

### DIFF
--- a/assets/apps/0000_60_service-ca_05_deploy.yaml
+++ b/assets/apps/0000_60_service-ca_05_deploy.yaml
@@ -23,7 +23,6 @@ spec:
         app: service-ca
         service-ca: "true"
     spec:
-      securityContext: {}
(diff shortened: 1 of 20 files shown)


sallyom:
/closing for https://github.com/redhat-et/microshift/pull/458

Issue #219, opened 2021-08-24
[BUG] runtime.sigpanic while running F34 on aarch64


jbpratt:
#### What happened:

Upon running `microshift`, I'm met with a large cgo trace, with what looks like an issue loading the boringssl DSO.

#### What you expected to happen:

A nice help command would be presented :smiley_cat: 

#### How to reproduce it (as minimally and precisely as possible):

1. Install FCOS34 on a Raspberry Pi (details below)
2. Once booted, follow the `install.sh` script (mainly install cri-{o,tools})
3. Download a build of microshift
4. execute it!

#### Anything else we need to know?:

Booting FCOS on an RPi4 can be done roughly like so:
1. Download RPi firemware
2. Download aarch64 FCOS image (https://builds.coreos.fedoraproject.org/browser?stream=testing-devel&arch=aarch64)
3. Create ignition file and generate image
```
podman run --pull=always \
	--privileged --rm \
	-v /dev:/dev \
	-v /run/udev:/run/udev \
	-v .:/data \
	-w /data quay.io/coreos/coreos-installer:release install ${DISK} -i ${IGN} -f ${IMAGE} --offline --insecure
```
4. Mount the second partition on the disk and unzip firmware
5. Boot! There is a 3GB memory limit that will need to be removed on restart.

#### Environment:
- Microshift version (use `microshift version`): https://github.com/redhat-et/microshift/releases/download/4.7.0-0.microshift-2021-08-10-163759/microshift-linux-arm64
- Hardware configuration: Raspberry Pi 4 (8GB)
- OS (e.g: `cat /etc/os-release`):

```
[core@kitkat ~]$ cat /etc/os-release
NAME=Fedora
VERSION="34.20210810.dev.0 (CoreOS)"
ID=fedora
VERSION_ID=34
VERSION_CODENAME=""
PLATFORM_ID="platform:f34"
PRETTY_NAME="Fedora CoreOS 34.20210810.dev.0"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:34"
HOME_URL="https://getfedora.org/coreos/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora-coreos/"
SUPPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
BUG_REPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
[...]

PR #318, opened 2021-09-30
drop static-build flags

### LINKED ISSUE #209: Investigate Dynamic Compiling
Static builds of microshift produce confusing, hard to debug warnings.  Since releases of microshift are intended to be as rpms or container images, the point of statically compiling the code is moot, as system dependencies can be baked into the installation.   We should investigate whether dynamically built microshift releases are a viable while also ensuring a consistent build env for developers is maintained.  Ensuring that developer, build time, and consumer environments contain identical versions of the system dependencies should be trivial as they are all defined in the same build image (./build/images/Dockerfile).  Issues arise when developers build on their host machines (`make`), where version differences may come into play.  Host builds should be strongly discouraged in documentation.

TODOs
- [ ] Prove a dynamic microshift binary can be run in a container runtime with system deps installed at container build time

copejon:

Closes #209

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,6 @@ CROSS_BUILD_BINDIR :=$(OUTPUT_DIR)/bin
 # restrict included verify-* targets to only process project files
 GO_PACKAGES=$(go list ./cmd/... ./pkg/...)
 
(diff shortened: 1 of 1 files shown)


jbpratt:
Closes #219

PR #336, opened 2021-10-14
Add OAuth Service to Microshift binary

### LINKED ISSUE #240: How to create a new project?
Hi, i am trying to use microshift to deploy a sample application and wanted to create a new project for this, but got the following error. from the readme i read that microshift is not going to enable everything openshift, so the question is, is this the expected behavior.

```
[anath@maximus100 oc]$ ./oc new-project test
error: unable to default to a user name: the server could not find the requested resource (get users.user.openshift.io ~)
```
Note: I got the similar error when trying out some other openshift specific resources such as with the imagestreams

oglok:

Closes #240


jiridanek:
The OAuth Server has been subsequently removed again in

* https://github.com/openshift/microshift/pull/721

ggiguash:
@jiridanek , if we need OAuth service API in MicroShift, we need a justification for it.
Please, work with @DanielFroehlich on the use-case and justification.

PR #346, opened 2021-10-19
creation of node port router svc

### LINKED ISSUE #334: [Enhancement]: Expose the route service
I would like to be able to access exposed the router services

#### Why is this needed:
We should embrace an experience similar to OpenShift routes so that usage between the two platforms is somewhat similar and we are not forcing users to open node ports per application unless desired

cooktheryan:
This creates another svc in the openshift-ingress namespace to allow for a node port to be opened. If a user is running the install.sh this will also open a firewall rule. In docs we will need to tell users to either place some sort of LB or enter the firewall rules to modify the traffic.

This was tested on AWS creating a wildcard entry pointing the public IP. Then when creating the route explicitly providing a route to be used for the host field. 

Closes #334


rootfs:
@cooktheryan cool!

cooktheryan:
@rootfs ill need another look through when you get a chance

PR #358, opened 2021-10-21
revert back to v4 from v2

### LINKED ISSUE #356: [BUG] serivce-ca crash looping
#### What happened:
```
openshift-service-ca            service-ca-84b44986cb-wsqxs           0/1     CrashLoopBackOff   5          16m
```

Service ca stays online

1. Deploy from install.sh or build from main
2. kubectl get po -wA

#### Environment:
- Microshift version (use `microshift version`): All
- Hardware configuration: amd64
- OS (e.g: `cat /etc/os-release`): 8.4 rhel and fedora
- Kernel (e.g. `uname -a`): 4.18.0-305.el8.x86_64
- Others:

cooktheryan:
Closes #356

### DIFF
--- a/assets/apps/0000_60_service-ca_05_deploy.yaml
+++ b/assets/apps/0000_60_service-ca_05_deploy.yaml
@@ -30,7 +30,7 @@ spec:
         imagePullPolicy: IfNotPresent
         command: ["service-ca-operator", "controller"]
         args:
-        - -v=2
+        - -v=4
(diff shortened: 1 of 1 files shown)


cooktheryan:
@fzdarsky looking back through history I see that we changed from v4-> v2 it seems like OCP uses v2 but since we have switched to v2 the container seems to crashloop after x amount of time

PR #398, opened 2021-10-29
update README.md on how to add pull secret

jianzzha:
**Document on how to pull images from private repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,27 @@ MicroShift keeps all its state in its data-dir, which defaults to `/var/lib/micr
 
 When starting the MicroShift for the first time the Kubeconfig file is created. If you need it for another user or to use externally the kubeadmin's kubeconfig is placed at `/var/lib/microshift/resources/kubeadmin/kubeconfig`.
 
+### Pulling Container Image From Private Registries
+
+Microshift may not have the pull secret for the registry that you are trying to use. For example, Microshift does not have the pull secret for registry.redhat.io. In order to use this registry, there are several approaches. The first approach is to use podman login,
+```sh
+podman login registry.redhat.io
+```
+
+Once the podman login is complete, Microshift will be able to pull images from this registry. This approach works across name spaces.
+
+The second approach is to create a pull secret, then let the service account to use this pull secret. This approach works within a name space. For example, if the pull secret is stored in a json formatted file "secret.json",
+```sh
+# First create the secret in a name space
+kubectl create secret generic my_pull_secret \
+    --from-file=secret.json \
+    --type=kubernetes.io/dockerconfigjson
+# Then attach the secret to a service account in the name space
+kubectl secrets link default my_pull_secret --for=pull
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -126,6 +126,29 @@ MicroShift keeps all its state in its data-dir, which defaults to `/var/lib/micr
...
+
+MicroShift may not have the pull secret for the registry that you are trying to use. For example, MicroShift does not have the pull secret for registry.redhat.io. In order to use this registry, there are several approaches. The first approach is to use podman login,
+```sh
+podman login registry.redhat.io
+```
+
+Once the podman login is complete, MicroShift will be able to pull images from this registry. This approach works across name spaces.
+
+This approach assumes podman is installed. This might not be true for all MicroShift environments. For example, if MicroShift is installed through RPM, CRI-O will be installed as dependency, but no podman. In this case, one can choose to install podman separately, or use other approaches described below. 
+
```
sallyom, line 139: * edit `as dependency` to `as a dependency` and  `but no podman` to `but not podman` 
* wdyt, is this more clear? `In this case, one can choose to install podman separate from the rpm and then login as above, or, to allow CRI-O access to private registries, create a pull secret as described below`

README.md
```diff
@@ -126,6 +126,29 @@ MicroShift keeps all its state in its data-dir, which defaults to `/var/lib/micr
...
+The second approach is to create a pull secret, then let the service account to use this pull secret. This approach works within a name space. For example, if the pull secret is stored in a json formatted file "secret.json",
+```sh
```
sallyom, line 141: edit `then link the secret to a service account in the namespace` (or, is it always `the default service account`? if so, `then link the secret to the default service account in the namespace`)

README.md
```diff
@@ -126,6 +126,29 @@ MicroShift keeps all its state in its data-dir, which defaults to `/var/lib/micr
...
 
+### Pulling Container Image From Private Registries
+
+MicroShift may not have the pull secret for the registry that you are trying to use. For example, MicroShift does not have the pull secret for registry.redhat.io. In order to use this registry, there are several approaches. The first approach is to use podman login,
+```sh
+podman login registry.redhat.io
+```
+
+Once the podman login is complete, MicroShift will be able to pull images from this registry. This approach works across name spaces.
+
```
sallyom, line 137: name spaces / namespaces

sallyom (commented):
a few nits, this is good information!

sallyom:
> > > Can you add this as well to docs/index.md so that it's reflected on microshift.io?
> > 
> > 
> > We should probably add this type of content on the website part, and link from the readme.md here to the website to avoid maintaining stuff in two places.
> 
> Sure. Any pointer on how to update the content on the website?

hm, we shouldn't have that content in 2 places. I think we need to link the docs in the README.md, but for now, add your information to the docs index.md. The docs are new, and now that they are there that's where we'll direct people rather than the README (ie, I think we'll update the README to have a brief intro then link to the docs)

iranzo:
Handled on https://github.com/redhat-et/microshift-documentation/pull/9 respecting authorship of commit

cooktheryan:
closed as it was merged in the docs repo

PR #404, opened 2021-10-29
update docs Makefile

sallyom:

This works for me, I'm running on fedora. Anyone else? 

The serve output suggests viewing at `0.0.0.0:4000` however I view at `127.0.0.1:4000`


cooktheryan:
@sallyom should we specify to potentially disable selinux or provide a context when launching?

Everything is good to go otherwise

sallyom:
> @sallyom should we specify to potentially disable selinux or provide a context when launching?
> 
> Everything is good to go otherwise

Let's hold this until we discuss next week, there's question whether the docs targets work on various systems (fedora, mac, other) - also, should we move the docs targets to the root Makefile (i pushed that change here)? Or, should we keep a separate docs/Makefile?

mangelajo (approved):
/lgtm

Thank you!! :)

mangelajo:
> > @sallyom should we specify to potentially disable selinux or provide a context when launching?
> > Everything is good to go otherwise
> 
> Let's hold this until we discuss next week, there's question whether the docs targets work on various systems (fedora, mac, other) - also, should we move the docs targets to the root Makefile (i pushed that change here)? Or, should we keep a separate docs/Makefile?

It should probably work on various systems as we ar using docker/podman.

PR #410, opened 2021-11-02
rename service file to match service purpose

copejon:
Service name should indicate it's intent as a lifecycle manager for containerized microshift and to match the naming convention in the directory (script -> script.service).


mangelajo:
I had the same thought when I tried it.

PR #415, opened 2021-11-03
Pre-checks into install.sh

alknopfler:

Closes #<Issue Number>

This PR contains:
- Pre-checks into the install.sh script to verify previously the CPU, Mem and disk

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -41,20 +41,13 @@ get_os_version() {
 
 # Function to check system prerequisites
 pre-check-installation(){
-    cpu_threshold='90'
     mem_threshold='1024'
-    disk_threshold='200'
+    disk_threshold='2048'
     numCPU_threshold='2'
     
     numCPU=$(nproc --all)
     if [ $numCPU -lt $numCPU_threshold ]; then
         echo "Warning: Pre-Install check number of CPUs cores less than recommended number: $numCPU_threshold"
-        #uncomment this line to exit on error. By now informative only
-        #exit 1
-    fi
-    cpu_used=$(grep 'cpu ' /proc/stat | awk '{usage=($2+$4)*100/($2+$4+$5)} END {print usage "%"}' | cut -f 1 -d ".")
-    if [ $cpu_used -gt $cpu_threshold ]; then
-        echo "Warning: Pre-Install check CPUs usage less than recommended number: $cpu_threshold"
         #uncomment this line to exit on error. By now informative only
         #exit 1
     fi
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,46 @@ get_os_version() {
     OS_VERSION=$(egrep '^(VERSION_ID)=' /etc/os-release | sed 's/"//g' | cut -f2 -d"=")
 }
 
+# Function to check system prerequisites
+pre-check-installation(){
+    cpu_threshold='90'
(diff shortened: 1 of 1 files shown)


alknopfler:
By now, the checks are only informational. Not sure if you want to exit if some checks are not passing

install.sh
```diff
@@ -39,6 +39,46 @@ get_os_version() {
...
+    #---disk
+    disk_free=$(df -m | grep /$ | grep -v -E '(tmp|boot)' | awk '{print $4}')
+    if [ $disk_free -lt $disk_threshold ]
+    then
+        echo "Error in prerequisites: DISK usage warning!!!"
+        #uncomment this line to exit on error. By now informative only
+        #exit 1
+    fi
+}
+
```
sallyom, line 81: If we do want to only warn with this check, we should change the wording in the echo to be something like
`Warning: Pre-Install check CPUs less than recommended number` 
`Warning: Pre-Install check available CPUs less than recommended` 
...
wdyt?

sallyom (commented):
we'll need a consensus on whether to error with the checks (probably not) and the wording of the warning msgs, but lg!

install.sh
alknopfler: yeah makes sense...I'll change it

install.sh
```diff
@@ -78,10 +115,17 @@ install_dependencies() {
             conntrack \
             firewalld
     else
-        sudo dnf install -y \
-            policycoreutils-python-utils \
-            conntrack \
-            firewalld
+        if [ "$DISTRO" = "centos" ] && [ "$VERSION" = "7" ]; then
```
mangelajo, line 118: we aren't targetting centos 7, centos 7 uses iptables-legacy (instead of nft based) and adds a set of complications we don't have time to look at now.

install.sh
```diff
@@ -39,6 +39,43 @@ get_os_version() {
     OS_VERSION=$(egrep '^(VERSION_ID)=' /etc/os-release | sed 's/"//g' | cut -f2 -d"=")
 }
 
+# Function to check system prerequisites
+pre-check-installation(){
+    cpu_threshold='90'
```
mangelajo, line 44: I wouldn't check CPU usage, could generate random issues in VMs.

install.sh
```diff
@@ -39,6 +39,43 @@ get_os_version() {
...
+    mem_threshold='1024'
+    disk_threshold='200'
```
mangelajo, line 46: The disk threshold should be much higher, is it MB?  the images, etc will account for more MBs

install.sh
alknopfler: removed!

install.sh
alknopfler: yes, 200Mb free after OS installation. I mean, you have your VM with 500mb, ubuntu takes 250, you have 250 free which is > than this threadhold -> you can install it

If not which should be the minimum threshold are you thinking?

alknopfler: after talking about it, let's change to 2048 mb (2gb)

mangelajo:
/lgtm

Although we're considering to remove install.sh process and document how to do it with dnf in a very easy way, it's fine for now, may be at that time the checks can be converted in some sort of pre-flight checks for the RPM, or a script inside the rpm

PR #420, opened 2021-11-04
Improve and pack the cleanup script in the rpms

mangelajo:
This will allow running `cleanup-all-microshift-data` which will
ask for confirmation. This is something which should be useful
during development phase and which could potentially be removed
for a final product.


oglok (commented):
/approve

PR #414, opened 2021-11-02
nightly image builds

cooktheryan:
Closes: #413

PR #425, opened 2021-11-04
create nightly aio

### LINKED ISSUE #413: [Enhancement]: Nightly AIO image
Create a nightly AIO image

#### Why is this needed:
Currently we create binaries every night we should do the same with AIO

alknopfler:

Closes #413


cooktheryan:
going to cancel for #414

PR #429, opened 2021-11-05
mDNS responder for microshift

mangelajo:
This commit adds support in microshift do do mDNS announcement on the LAN.

1) For the host name
For example if your hostname is `microshift.locaL`, other hosts will be able to
resolve microshift.local, because microshift will respond to the name with the IPv4 and IPv6 addresses.

2) For the hosts in the routes
for example, if you have a route with the host: `nginx.cluster.local`, and another `mqtt.cluster.local` ,
microshift would respond for both with the openshift-router IP addresses, so
other services in the lan can find those services.

This is a proposal of functionality which can be tested at this point, if we believe this is useful to the project 
I can start writing tests and documenting.

### DIFF
--- a/pkg/mdns/routes.go
+++ b/pkg/mdns/routes.go
@@ -1,0 +1,161 @@
+package mdns
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/openshift/microshift/pkg/mdns/server"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+const defaultResyncTime = time.Hour * 1
+
+func (c *MicroShiftmDNSController) restConfig() (*rest.Config, error) {
+	return clientcmd.BuildConfigFromFlags("", c.KubeConfig)
+}
+
+func (c *MicroShiftmDNSController) startRouteInformer(stopCh chan struct{}) error {
+	klog.InfoS("Starting MicroShift mDNS route watcher")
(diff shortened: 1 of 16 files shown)


oglok (commented):
Hey! o/

Is this PR ready for review? If so, please remove the WIP tag.

At a first glance, it looks ok (I need to do a deeper review). There is something different from how other services are structured though.

I see you have created a new service in the controllers pkg, that implements the interface given by the service manager framework. The Run() function from that controller, calls another new pkg called mDNS. Couldn't we get rid of that new controller, and implement the service interface directly from the mDNS pkg? (look at the kustomize pkg for example) or is there any other reason that I'm missing?

Thanks Ajo!

pkg/mdns/routes.go
```diff
@@ -0,0 +1,144 @@
...
+	dc, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "error creating dynamic informer")
+	}
+
+
+	informerFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dc, defaultResyncTime, v1.NamespaceAll, nil)
+	routersGVR, _ := schema.ParseResourceArg("routes.v1.route.openshift.io")
+
+	fmt.Printf("MDNS: gvr: %+v\n", routersGVR)
```
mangelajo, line 46: remove printfs...

mangelajo:
> Hey! o/
> 
> Is this PR ready for review? If so, please remove the WIP tag.

It is, but no testing, before merging I would like to write testing for the feature, even if it's at least unit tests. Now even though the general code review are useful, I don't feel like we should merge it.

> 
> At a first glance, it looks ok (I need to do a deeper review). There is something different from how other services are structured though.
> 
> I see you have created a new service in the controllers pkg, that implements the interface given by the service manager framework. The Run() function from that controller, calls another new pkg called mDNS. Couldn't we get rid of that new controller, and implement the service interface directly from the mDNS pkg? (look at the kustomize pkg for example) or is there any other reason that I'm missing?

Makes total sense. I didn't see the kustomize example (internal service) and I tried to following the other services structure, but what you propose makes perfect sense. :+1

> 
> Thanks Ajo!

Thank you! :)

oglok:
Please, write the unit tests you mentioned in the comments above. I think this PR is ready for the next level. :-)

mangelajo:
Working on it Ricardo, thanks for your patience ':D

mangelajo:
Tests written, ready to review.

mangelajo:
@oglok @rootfs this shall be ready.

oglok (commented):
/approve

oglok:
Awesome work @mangelajo

mangelajo:
Thanks @oglok !! :D let's multinode!

PR #436, opened 2021-11-09
restore docs/design

sallyom:

Restoring the docs/design folder in same place as it was pre-docs-migration.

TODO: give it a permanent place (`microshift/design`, probably) and update the design docs in `microshift-documentation` to point to this location.


sallyom:
we'll re-commit this documentation in a new PR, to restore the original commiters

PR #441, opened 2021-11-10
Update microshift.spec

cooktheryan:
Missing package required for using Selinux module


sallyom:
/lgtm  
@mangelajo lgt-you?

packaging/rpm/microshift.spec
```diff
@@ -117,6 +117,7 @@ Things break. But you can still help shape it, too.
 Summary: SELinux policies for MicroShift
 BuildRequires: selinux-policy
 BuildRequires: selinux-policy-devel
+BuildRequires: container-selinux
```
mangelajo, line 120:
```suggestion
Requires: container-selinux
```

If we require it at runtime (and not build)

sallyom:
/lgtm 
thanks!

PR #461, opened 2021-11-17
[WIP] Change kubeconfigs to point to the host IP

oglok:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -86,7 +86,7 @@ func NewMicroshiftConfig() *MicroshiftConfig {
 		NodeName:        nodeName,
 		NodeIP:          nodeIP,
 		Cluster: ClusterConfig{
-			URL:         "https://127.0.0.1:6443",
+			URL:         "https://" + nodeIP + ":6443",
 			ClusterCIDR: "10.42.0.0/16",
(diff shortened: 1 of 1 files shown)


oglok:
I'm now thinking about the case where the device is restarted and the IP address changes due to a dhcp renewal. MicroShift will keep the host IP in the kubeconfigs and it'll crash. 

Would that work with microshift.cluster.local using mDNS? @mangelajo

cooktheryan:
what about using a name in the kubeconfigs? That way if the DHCP address changes the name should get patched as well. I think it would net us more possibilities for customers in the event that the facility in which microshift was running has some sort of VPN back to a home base a users could then load the kubeconfig into Argo and app of apps if they wanted

mangelajo:
> I'm now thinking about the case where the device is restarted and the IP address changes due to a dhcp renewal. MicroShift will keep the host IP in the kubeconfigs and it'll crash.
> 
> Would that work with microshift.cluster.local using mDNS? @mangelajo

yes, mDNS would handle this, via microshift.local or hostname.local etc .. :)

mangelajo:
included in https://github.com/redhat-et/microshift/pull/471

PR #479, opened 2021-12-02
release.sh supports nightly releases

copejon:

Partially Satisfies #413 

The nightly build flow will:
- release an AIO manifest with  tag:  `4.8.0-0.microshift-DATE_TIME-nightly`
- release AIO images with tag: `4.8.0-0.microshift-DATE_TIME-nightly-linux-nft-ARCH`
- release a standalone manifest with  tag:  `4.8.0-0.microshift-DATE_TIME-nightly`
- release standalone images with tag: `4.8.0-0.microshift-DATE_TIME-nightly-linux-ARCH`
- Alias the most recent AIO and standalone manifests with a `nightly` tag.  Users can then reference the most recent build with `quay.io/microshift/microshift(-aio):nightly`. 

Will **not:**
- Publish a github release (meaning no binaries are published)
- Create a new github tag
- Re-alias the `lastest` tag.  This should always point to the latest stable release.

The changes here alter the `scriptes/release.sh` script to allow for the behavior listed above.  This is trigger with the new CLI flag `--nightly`.  The make target `release-nightly` will drive the script, and provide a `VERSION`, as it does for the `release` target.


Makefile
```diff
@@ -163,7 +163,7 @@ _build_containerized:
 _build_containerized_aio:
 	@if [ -z '$(CTR_CMD)' ] ; then echo '!! ERROR: containerized builds require podman||docker CLI, none found $$PATH' >&2 && exit 1; fi
 	echo BIN_TIMESTAMP==$(BIN_TIMESTAMP)
-	$(CTR_CMD) build -t $(IMAGE_REPO_AIO):$(SOURCE_GIT_TAG)-linux-$(IPTABLES)-$(ARCH) \
+	$(CTR_CMD) build -t $(IMAGE_REPO_AIO):$(SOURCE_GIT_TAG)-linux-$(ARCH) \
```
mangelajo, line 166: Hmm, without this we cannot support older systems like the ubuntu on jetson where they run on iptables kernel.

This is also important for older rhel7 testers.

I'd keep/build the different ones for now.

mangelajo (changes requested):
a comment about the "$(IPTABLES)" aliases for AIO

Makefile
copejon: The iptables variant is only specified for arm64, for an internal edge case.  I'm okay with leaving it for now.  IMO though using `nft` to reflect the common deployment path is confusing.

Issue #483, opened 2021-12-07
[Enhancement]: Remove install.sh from CI jobs


cooktheryan:
#### What would you like to be added:
CI steps should not use install.sh anymore

#### Why is this needed:
We will be deprecating install.sh in the near future

PR #487, opened 2021-12-08
Deprecate install.sh

fzdarsky:
Move `install.sh` into `/hack` as precursor to removing it completely in favour of the .rpm- and Podman-based deployments.


fzdarsky:
Depends on #483

mangelajo:
+1 from me once we have switched CI.

PR #490, opened 2021-12-08
scc,clusterquota CRD remove 'nullable: true' fields from required and add shortNames

### LINKED ISSUE #480: [BUG] MicroShift OpenAPIV3Schema validation errors
#### What happened:

If I try to create an SCC (probably any resource?) that doesn't explicitly include a required but nullable: true field in the yaml def, like, `priority: 0` or `defaultAddCapabilities: null` microshift errors. However, the same oc with a 4.8 OCP does not error, it happily creates the resource and sets the nullable: true fields to their empty value (0, null).

Same behavior as OpenShift

I discovered this while trying to deploy volsync in MicroShift, hence the volsync-mover SCC.
Try to apply this file with both MicroShift and an OCP 4.8 cluster:
(I have a very new `oc` 4.9.8) 

```
apiVersion: security.openshift.io/v1
kind: SecurityContextConstraints
metadata:
  name: volsync-mover
  labels:
allowHostDirVolumePlugin: false
allowHostIPC: false
allowHostNetwork: false
allowHostPID: false
allowHostPorts: false
allowPrivilegeEscalation: true
allowPrivilegedContainer: false
allowedCapabilities:
  - AUDIT_WRITE
  - SYS_CHROOT
fsGroup:
  type: RunAsAny
readOnlyRootFilesystem: false
requiredDropCapabilities:
  - MKNOD
runAsUser:
  type: RunAsAny
seLinuxContext:
  type: MustRunAs
supplementalGroups:
  type: RunAsAny
volumes:
  - configMap
  - downwardAPI
  - emptyDir
  - persistentVolumeClaim
  - projected
  - secret
```

SCC definition in MicroShift from vendored openshift/api
If you add these to the above yaml, MicroShift will succeed in creating the SCC:

```
priority: 0
defaultAddCapabilities: null
```

#### Environment:
- Microshift version (use `microshift version`): latest from main branch
- Hardware configuration:
- OS (e.g: `cat /etc/os-release`): Centos 8 Stream, podman running MicroShift (not AIO)

#### Relevant Logs 

```
The SecurityContextConstraints "volsync-mover" is invalid: 
* defaultAddCapabilities: Required value
* priority: Required value
```

sallyom:

SCC and ClusterResourceQuota validation behavior in OpenShift doesn't match MicroShift - ie, MicroShift validation is more strict, because it is kube-openapi validation while OpenShift uses custom validation. This PR remedies, so any SCC that can be created in OCP can be created in MicroShift. Also, SecurityContextConstraints shortNames `scc`  and ClusterResourceQuotas `clusterquota` rely on a hack in OpenShift, where these CRs  are _registered_ with openshift-apiserver, but _served_ with kube-apiserver. MicroShift doesn't require this, so instead, add the shortNames field to the CRD yaml. 

(same applies to ClusterResourceQuotas) 

Closes #480

### DIFF
--- a/assets/crd/0000_03_quota-openshift_01_clusterresourcequota.crd.yaml
+++ b/assets/crd/0000_03_quota-openshift_01_clusterresourcequota.crd.yaml
@@ -1,3 +1,5 @@
+# copied from https://github.com/openshift/api/blob/release-4.8/quota/v1/0000_03_quota-openshift_01_clusterresourcequota.crd.yaml
+# commented where modified from original
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -14,6 +16,8 @@ spec:
     listKind: ClusterResourceQuotaList
     plural: clusterresourcequotas
     singular: clusterresourcequota
+    shortNames:      # modified from openshift/api, oapi registers with openshift-apiserver, and shorNames()
+    - clusterquota   # same shortNames as OCP but with CRD instead of openshift-apiserver
   scope: Cluster
   versions:
   - name: v1
--- a/assets/crd/0000_03_security-openshift_01_scc.crd.yaml
+++ b/assets/crd/0000_03_security-openshift_01_scc.crd.yaml
@@ -1,3 +1,5 @@
+# copied from https://github.com/openshift/api/blob/release-4.8/security/v1/0000_03_security-openshift_01_scc.crd.yaml
+# commented where modified from original
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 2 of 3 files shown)


mangelajo:
@fzdarsky  to keep this change, may we need to add a patch mechanism (or just a patch, not sure if patching is implemented) in our rebase script?

mangelajo:
@fzdarsky  ^

fzdarsky:
> @fzdarsky to keep this change, may we need to add a patch mechanism (or just a patch, not sure if patching is implemented) in our rebase script?

@mangelajo the rebasing script isn't fully hands-off anyway: It does quite a bit of the work, but some things (like the resource manifests) don't apply 100% anyway and need manual review+tweaking. I'll see if I can add such patching, though, for the 4.9 rebase.

fzdarsky:
Tested and works like a charm for a new SCC.

/lgtm

Issue #493, opened 2021-12-10
[BUG] As soon as one component starts logging to a file, all klog will sink into that file


mangelajo:
#### What happened

All the klog based logs don't make it to the standard output.

#### What you expected to happen:

All the non-file redirected logs to make it to the standard output.

#### How to reproduce it (as minimally and precisely as possible):

Run microshift, and look for `grep -i mdns` output... it won't show up. But it's being printed.

#### Anything else we need to know?:

klog is a singleton library that we use from multiple places, as soon as one place modifies the configuration of klog,
the other users will be affected.

We need to stop changing klog settings from multiple places, and probably we need a klog replacement/extended version which will
map callers to modules or something like that.

#### Environment:
- Microshift version (use `microshift version`):
- Hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

PR #494, opened 2021-12-10
Do not modify default logging parameters

mangelajo:
klog is a singleton library, and since we use that library from all
our services in a single process setting different log files won't
work, and will syphon all logs into the last log file we add.

Keep the simple strategy of letting all output to stderr for now.

Related-Issue: #493

### DIFF
--- a/pkg/controllers/openshift-apiserver.go
+++ b/pkg/controllers/openshift-apiserver.go
@@ -62,10 +62,8 @@ func (s *OCPAPIServer) configure(cfg *config.MicroshiftConfig) error {
 	}
 	args := []string{
 		"--config=" + configFilePath,
-		"--alsologtostderr=" + strconv.FormatBool(cfg.LogAlsotostderr),
 		"--v=" + strconv.Itoa(cfg.LogVLevel),
 		"--vmodule=" + cfg.LogVModule,
-		"--logtostderr=" + strconv.FormatBool(cfg.LogDir == "" || cfg.LogAlsotostderr),
 	}
 
 	options := openshift_apiserver.OpenShiftAPIServer{
--- a/pkg/controllers/openshift-apiserver.go
+++ b/pkg/controllers/openshift-apiserver.go
@@ -225,14 +225,17 @@ apiServerArguments:
   - Scope
   - SystemMasters
   - RBAC
-  - Node
+  - Node`)
+
+	if cfg.AuditLogDir != "" {
+		data = append(data, `
 auditConfig:
-  auditFilePath: "` + cfg.AuditLogDir + `/openshift-apiserver/audit.log"
+  auditFilePath: `+cfg.AuditLogDir+`openshift-apiserver-audit.log
   enabled: true
   logFormat: json
   maximumFileSizeMegabytes: 100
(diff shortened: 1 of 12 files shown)


mangelajo:
@oglok @fzdarsky  ^

fzdarsky:
@mangelajo this makes sense to me.

Would you recommend we also remove those from MicroShift itself for the time being?

mangelajo:
Unit testing needs some looking: https://github.com/redhat-et/microshift/runs/4598372160?check_suite_focus=true#step:6:8

@oglok ^ can you have a look? , I'm looking into the cni/flannel rabbit-hole

mangelajo:
other than that it LGTM! :)

mangelajo:
/lgtm

If it doesn't do what we expect we can always fix it

pkg/controllers/openshift-apiserver.go
```diff
@@ -230,14 +225,17 @@ apiServerArguments:
...
   - SystemMasters
   - RBAC
-  - Node
+  - Node`)
+
+	if cfg.AuditLogDir != "" {
+		data = append(data, `
 auditConfig:
-  auditFilePath: "` + cfg.LogDir + `/openshift-apiserver/audit.log"
+  auditFilePath: `+cfg.AuditLogDir+`openshift-apiserver-audit.log
```
mangelajo, line 233:
```suggestion
  auditFilePath: "`+cfg.AuditLogDir+`openshift-apiserver-audit.log"
```

should we keep the commas?

pkg/node/kube-proxy.go
```diff
@@ -53,21 +52,14 @@ func (s *ProxyOptions) configure(cfg *config.MicroshiftConfig) {
...
 		logrus.Fatalf("Failed to write kube-proxy config: %v", err)
 	}
-	args := []string{
-		"--logtostderr=" + strconv.FormatBool(cfg.LogDir == "" || cfg.LogAlsotostderr),
-		"--alsologtostderr=" + strconv.FormatBool(cfg.LogAlsotostderr),
-		"--v=" + strconv.Itoa(cfg.LogVLevel),
-		"--vmodule=" + cfg.LogVModule,
-	}
+	// Keeping the args in case something must be added in the future
+	args := []string{""}
```
mangelajo, line 56:
```suggestion
	// Keeping the args in case something must be added in the future
	args := []string{}
```

mangelajo: I wonder if this makes kubeproxy fail to start, empty arg in the arg list?

Locally we would not see it, because kubeproxy is internal, and already has configured all iptables, etc.

test/config.yaml
```diff
@@ -1,6 +1,6 @@
 --- 
 dataDir: /tmp/microshift/data
-logDir: /tmp/microshift/logs
+auditLogDir: /tmp/microshift/logs
 logVLevel: 4
 logVModule: microshift=4
 logAlsotostderr: true
```
mangelajo, line 6: shall we remove those two?
```suggestion
```

mangelajo:
> 

xinx!!! :sparkle:

Issue #460, opened 2021-11-17
[Enhancement]: Multi Node Enablement


oglok:
#### Design Document Link

This enhacement is meant to discuss design and architecture decisions to implement a new feature called “Multi-node”. 

Currently, MicroShift provides a single node cluster with Kubernetes and OpenShift control plane and node components embedded into a single binary, and some extra services running as scheduled pods. Multi-node should allow users to run MicroShift in multiple devices and work as a single cluster.

Multi-node should be approached from two perspectives:
Multiple worker nodes: provides the ability to add new worker nodes to an existing cluster
Highly available control plane: provides the ability to add new control plane nodes providing control plane workload balancing and high availability. 

#### What would you like to be added:
The ability for MicroShift to create a single cluster out of several devices (multiple worker nodes, and high available control plane)

#### Why is this needed:
High availability or expanding capacity might be requirements in edge computing scenarios.

--------------------------------------------------------------------------------------------------------------

This enhacement should be the parent of several tasks that will enable different aspects of the feature.

oglok:
#461

oglok:
#471

PR #498, opened 2021-12-13
[RFE] Multi-node Request for Enhacement

oglok:
This commit only describes the addition of new  compute nodes to an existing
  MicroShift cluster. Highly available control plane will be described in later
  PRs.

This Enhacement proposal addresses part of the #460 epic.

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -114,3 +114,13 @@ When deciding between different design options, we follow the following principl
 * Releases are mainly provided for convenience to users that just want to give MicroShift a quick try without friction. They are cut irregularly, e.g. to make a new feature available.
 * When rebasing onto a new OKD version, we vendor that version's packages and update the container image digests of the infrastructure services MicroShift deploys, i.e. the "release metadata" is baked into the MicroShift binary.
 * Eventually, we expect there to be a "MicroShift Release Image" that is based on / derived from the OpenShift Release Image: It references the MicroShift container image plus the subset of container images shared with and published by OpenShift. Defining a release image should allow to reuse the proven OpenShift CI and release tooling later.
+
+### Multi-Node Configuration
+* MicroShift provides the ability to add a n-number of compute nodes and work as a single cluster. This feature allows to scale out a MicroShift deployment when more computing power is needed.
+* MicroShift makes use of the [TLS bootstraping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/) workflow to allow new nodes to join the cluster in a secure manner.
+* MicroShift generates a new asset that is crucial for bootstrapping new nodes:
+  * `/var/lib/microshift/resources/kubelet/bootstrap-kubeconfig`
+* The `bootstrap-kubeconfig` asset must be placed in the new nodes to allow them to join the MicroShift cluster.
+* New nodes will only contain the role `node` using the existing flag `--roles`. (Currently, MicroShift only supports one control plane entity and multiple nodes.)
+* MicroShift will handle certificate rotation for security reasons on the new nodes.
+* As a summary, the Kubelet from a new node that tries to join a MicroShift cluster, will use the bootstrap kubeconfig file to get limited access to the Kube API server. The Kubelet will then create and retrieve a CSR (Certificate Signing Request). MicroShift's controller manager is configured to automatically approve this new CSR and a new set of assets will be created by Kubelet (certs, kubeconfig).


cgwalters:
There's a lot going on here.  Something I am struggling to wrap my head around is that to me, a key point of "OpenShift 4" is that the cluster manages the OS.  That isn't handled by MicroShift today (right?).

Now one thing I was asked to comment on here is the relationship to OCP node join.  What I would say is basically all of that logic lives in https://github.com/openshift/cluster-machine-approver

OK I just did https://github.com/openshift/cluster-machine-approver/pull/150 - I hope that's helpful.

docs/design/design.md
```diff
@@ -114,3 +114,13 @@ When deciding between different design options, we follow the following principl
...
+* MicroShift makes use of the [TLS bootstraping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/) workflow to allow new nodes to join the cluster in a secure manner.
+* MicroShift generates a new asset that is crucial for bootstrapping new nodes:
+  * `/var/lib/microshift/resources/kubelet/bootstrap-kubeconfig`
+* The `bootstrap-kubeconfig` asset must be placed in the new nodes to allow them to join the MicroShift cluster.
+* New nodes will only contain the role `node` using the existing flag `--roles`. (Currently, MicroShift only supports one control plane entity and multiple nodes.)
+* MicroShift will handle certificate rotation for security reasons on the new nodes.
+* As a summary, the Kubelet from a new node that tries to join a MicroShift cluster, will use the bootstrap kubeconfig file to get limited access to the Kube API server. The Kubelet will then create and retrieve a CSR (Certificate Signing Request). MicroShift's controller manager is configured to automatically approve this new CSR and a new set of assets will be created by Kubelet (certs, kubeconfig).
```
stlaz, line 126: What makes my CSR applicable for an automatic approval?

mangelajo:
@cgwalters @stlaz thanks for looking at this, I'm having a look to the cluster-machine-approver

mangelajo:
and yes, @cgwalters MicroShift doesn't manage the OS today, not sure if we would need to look at that at some point.

mangelajo:
@fzdarsky

mangelajo:
hmm @cgwalters so, it looks like we were proposing the same mechanism used in OpenShift, but in openshift we have the cluster-machine-approver which makes additional checks based on the open shift machine API (which we wouldn't have on MicroShift), and other Node details.

Probably it makes sense to use the simpler kube-controller-manager to start with, and then in a future it could make sense to have something to extend the CSR made by kubelet with TPM details (for example), and then have an specific MicroShift approver that also can check the CSR based on a CA or again via the TPM hardware module on the masters.

cgwalters:
Assuming the bootstrap token is transferred and maintained securely, then I don't think any additional checks against a CA add much value.  TPMs on the other hand can be quite powerful, but require investment and aren't available everywhere.

One problem in OCP today is we transfer that token via served Ignition - xref https://github.com/openshift/machine-config-operator/pull/736
(Long story short, I think we want to move all secrets into the bootstrap ignition which is part of the cloud metadata)

dhellmann:
/kind design
/close

We've decided not to support multi-node.

giannisalinetti:
Hi team, with full respect I think that this decision is far from being strategic.
As a Red Hat solution architect I work with customers who are just comparing MicroShift with other already existing similar solutions. Some of them give the chance of setting up a multinode cluster, like for example k3s. 

Giving the option of setting up a minimal multi-node cluster (for example adding a sedond HA master or multiple workers) can be a true game changer, even for edge scenarios where application high-availability cannot be managed using existing tools and application HA features to support failover.

For this reason I suggest to reconsider this RFE.

derekwaynecarr:
@giannisalinetti can you provide more detail on which scenarios cannot be managed using existing tools and application HA features to support failover?  if you desire two control hosts to provide HA, i believe you will have less overall reliability than two single node solutions.  if you scale out to additional workers/control hosts, it also demands workflow during upgrade scenarios, and increases significantly the resource consumption that begins to look like traditional standalone openshift.  at the moment, we are focusing on ensuring we can meet minimal resource budget for single node scenarios.

Any additional detail you can provide for where microshift would be a fit, but standalone openshift would not, its always appreciated.  If its entirely around resource consumption, keep in mind that as you grow clusters, consumption increases at all levels (control hosts, and networking sdn).

dhellmann:
@giannisalinetti I would be interested in the customer's feedback to @derekwaynecarr's answer. Maybe we can do that privately (email, chat, etc.) so we can have all of the details.

giannisalinetti:
@dhellmann @derekwaynecarr sure!
We can discuss it by email if you agree: gsalinet@redhat.com

PR #499, opened 2021-12-13
Reload existing CA from disk on restart

mangelajo:
Without reloading any later operation will result in a panic
trying to access nil rootCA or nil rootKey.


cooktheryan:
@mangelajo we will most likely need to resubmit this PR. Sorry :/

PR #502, opened 2021-12-13
create this file so that it can be placed

cooktheryan:
Allow this file to be pulled from the repo to help with CI testing


cooktheryan:
file exists here https://github.com/redhat-et/microshift/blob/main/packaging/images/microshift-aio/crio-bridge.conf

PR #505, opened 2021-12-15
service-ca pod run as non-root

sallyom:

**NOTE** In order for `service-ca` to run with `non-root`, have to modify the signing-bundle and TLS crt,key volumes to be a configmap & secret, otherwise the non-root uid in the pod can't access the `service-ca.crt, tls.crt, tls.key` files it needs. Also, the OCP cluster-policy-controller must be running.

This PR depends on 
https://github.com/redhat-et/microshift/pull/504 (service-ca volumes as CA configmap, TLS secret)
https://github.com/redhat-et/microshift/pull/478 (cluster-policy-controller)
Closes #<Issue Number>

### DIFF
--- a/assets/apps/0000_60_service-ca_05_deploy.yaml
+++ b/assets/apps/0000_60_service-ca_05_deploy.yaml
@@ -23,7 +23,10 @@ spec:
         app: service-ca
         service-ca: "true"
     spec:
-      securityContext: {}
+      securityContext:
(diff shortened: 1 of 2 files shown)


sallyom:
this will fail until https://github.com/redhat-et/microshift/pull/478 merges

sallyom:
closing, no longer needed

PR #507, opened 2021-12-15
Remove stray log messages

fzdarsky:


fzdarsky:
Just noticed this confusing log message (saying "service started" when it's just being configured) exists in other components, too. Need to either remove them there, too, or - if their output is useful for development, rephrase them and drop them to Debug log level.

PR #510, opened 2021-12-16
Disable ubuntu ci job

cooktheryan:
Now that testing is more mature we can potentially turn down this test. The compiling may be worth keeping

### DIFF
--- a/.github/workflows/ubuntu.yaml
+++ b/.github/workflows/ubuntu.yaml
@@ -34,34 +34,34 @@ jobs:
           name: microshift
           path: ./microshift
   
-  deploy-microshift:
-    name: Deploy-Microshift
-    needs: Build-Microshift
-    runs-on: ubuntu-20.04
-    env:
-      KUBECONFIG: /tmp/kubeconfig
-    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ubuntu.yaml
```diff
@@ -34,34 +34,34 @@ jobs:
...
-      - name: start microshift
-        run: sudo /tmp/microshift run &
-
-      - name: sleep then run kuttl
-        run: |
-           sudo -i sh -c 'until [ -f /var/lib/microshift/resources/kubeadmin/kubeconfig ]; do sudo sleep 5; done'
-           sudo cp /var/lib/microshift/resources/kubeadmin/kubeconfig /tmp/kubeconfig
-           sudo chown `whoami`: /tmp/kubeconfig
-           make validate-cluster 
+#  deploy-microshift:
```
mangelajo, line 37: I'd propose to simply remove it, git history will always keep it for us :-)

cooktheryan: @mangelajo done!

PR #524, opened 2021-12-22
Revert "Revert "Reload existing CA from disk on restart""

mangelajo:
We need this back in, it was unrelated to those failures, this is only for re-loading the microshift CA from disk to memory on
microshift service reboot, so the bootstrap for multinode would work on later patches.
Reverts redhat-et/microshift#521

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -21,7 +21,6 @@ import (
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/controllers"
 	"github.com/openshift/microshift/pkg/util"
-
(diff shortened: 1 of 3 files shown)


mangelajo:
@cooktheryan may be I misunderstood the failures? 

```

See "systemctl status microshift.service" and "journalctl -xe" for details.
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2021-12-22T12:57:32Z"}
error: failed to execute wrapped command: exit status 1

```

why is that happening, is there a way to see the logs?

PR #457, opened 2021-11-15
selinux configs and  volume for microshift-containerized

husky-parul:
Signed-off-by: Parul Singh <parsingh@redhat.com>

**Which issue(s) this PR addresses**:
For podman deployment:

- systemd unit file for starting and managing microshift-containerized.

Closes #434, #433, #432

PR #527, opened 2021-12-22
missing persistent volume mount

cooktheryan:
We are not mounting the location of persistent volume mounts in the containerized version of microshift. If that container is replaced all persistent data will be lost


fzdarsky:
@cooktheryan if we merge #457 plus change to Podman volumes, this'll look more like
```
sudo podman run --network=host --privileged --rm --name microshift \
  -v /var/run/crio/crio.sock:/var/run/crio/crio.sock:rw,rshared \
  -v microshift-data:/var/lib/microshift \
  -v /var/lib/kubelet:/var/lib/kubelet:rw,rshared \
  -v /var/log:/var/log:rw,rshared \
  quay.io/microshift/microshift:latest
```
Would we still need this then?

PR #531, opened 2021-12-23
directory is required to exist

cooktheryan:
If the /var/hpvolumes does not exist MicroShift cannot start


mangelajo (approved):
/lgtm

mangelajo:
Normally the rpm would create the dir, but if the .service is installed manually as we recommend in some cases, this is necessary.

Issue #528, opened 2021-12-23
[BUG] failed to pull container image from quay


lucamaf:
#### What happened:
After installing the prereq on fedora 35 I'm trying to enable microshift but it fails to start.
When I explore the journal I can see the following log:

```
dic 23 08:57:22 fedora-home podman[28481]: Trying to pull quay.io/microshift/microshift:latest...
dic 23 08:57:23 fedora-home podman[28481]: Error: choosing an image from manifest list docker://quay.io/microshift/microshift:latest: no image found in manifest list for architecture amd64,>
dic 23 08:57:23 fedora-home systemd[1]: microshift.service: Main process exited, code=exited, status=125/n/a

```

#### What you expected to happen:
microshift starting up correctly as container on podman

#### How to reproduce it (as minimally and precisely as possible):
just following the provided documentation around running microshift

#### Environment:
- Microshift version (use `microshift version`): latest (4.8)
- Hardware configuration: 
- OS (e.g: `cat /etc/os-release`): fedora workstation 35
- Kernel (e.g. `uname -a`): Linux fedora-home 5.15.10-200.fc35.x86_64
- Others:

patrickstrick:
I see this too for the arm64 build. The problem appears to be that the current "latest" tag (also tagged 4.8.0-0.microshift-2021-12-22-195815) doesn't have the arm64 and amd64 builds as child manifests. Until this is fixed (assuming the next image pushed to Quay.io will be correct), you should be able to get it working by doing the following (it worked for me at least).
1. Editing `/etc/systemd/system/microshift.service` and replacing `quay.io/microshift/microshift:latest` with `quay.io/microshift/microshift:4.8.0-0.microshift-2021-12-22-195815-linux-arm64` or `quay.io/microshift/microshift:4.8.0-0.microshift-2021-12-22-195815-linux-amd64` depending on your system's architecture.
2. Run `sudo systemctl daemon-reload` to have systemd read the new version
3. Run `sudo systemctl restart microshift.service` to have the service restarted

cooktheryan:
This is fixed now. Apologies for the issue

PR #534, opened 2022-01-04
hack to temp fix releases

cooktheryan:
This is a workaround for #528


sallyom:
/approve
/lgtm

* this is temporary, we'll pinpoint what is broken after we have a good :latest image pushed, then remove this change.

PR #543, opened 2022-01-06
matching service name with docs

cooktheryan:
In the docs we use microshift as the service name for containerized but when installed via RPM this is different its names microshift-containerized


cooktheryan:
@sallyom based on our conversations today

PR #546, opened 2022-01-06
fix for microshift-containerized with volume mount

husky-parul:

This PR creates `microshift-data` volume for podman deployment.

Closes #<Issue Number>


packaging/systemd/microshift-containerized
```diff
@@ -2,7 +2,9 @@
 
 set -euxo pipefail
 
-[[ -d /etc/microshift-containerized ]] || mkdir -p /etc/microshift-containerized
-cat <<EOF > /etc/microshift-containerized/microshift-containerized.conf
-export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig
-EOF
+setsebool -P container_manage_cgroup true
+
```
sallyom, line 6: At the community mtg, we agreed this setsebool should be in the documentation rather than in the service. I'm also removing this file from the AIO service, here https://github.com/redhat-et/microshift/pull/554. We can now have the single file, for both containerized and AIO, and now only for the podman volume check/creation.

sallyom (changes requested):
Remove that setsebool line, but keep the `podman volume` check. Then, let's merge this!

packaging/systemd/microshift-containerized
sallyom: from the [podman run docs](https://docs.podman.io/en/latest/markdown/podman-run.1.html): 
```
If no such named volume exists, Podman will create one
```
(fyi I was the one who originally added this, so this is on me)

PR #554, opened 2022-01-12
fix microshift-aio unit file

sallyom:

This updates the unit file for microshift-aio, to use the correct image `quay.io/microshift/microshift-aio:latest`
Also, removes the microshift-aio script (the podman volume check script) as it is not needed. When passed a volume flag, podman auto-creates the volume with `podman run` if it doesn't exist. If it exists, podman will use it. 

/cc @husky-parul 
/cc @cooktheryan


sallyom:
@husky-parul we can name that systemd script something like `microshift-podman-volume-check` instead of `microshift-containerized` wdyt? If you agree, you can rename it in #546 (and then, if this is merged, update the aio-service file w/ the new name too)

packaging/systemd/microshift-aio.service
```diff
@@ -14,8 +14,8 @@ Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=on-failure
 TimeoutStopSec=70
 ExecStartPre=/bin/rm -f %t/%n.ctr-id
-ExecStartPre=/bin/bash /usr/bin/microshift-aio # not autogenerated
-ExecStart=/usr/bin/podman run --cidfile=%t/%n.ctr-id --sdnotify=conmon --cgroups=no-conmon --rm --replace -d --name microshift-aio --privileged -v /lib/modules:/lib/modules -v microshift-data:/var/lib --label io.containers.autoupdate=registry -p 6443:6443 quay.io/microshift/microshift:4.7.0-0.microshift-2021-08-31-224727-aio
+ExecStartPre=/bin/bash /usr/bin/microshift-containerized # not autogenerated
```
sallyom, line 17: @husky-parul here, rename this to `microshift-podman-volume-check` (also in #546)?

packaging/systemd/microshift-aio.service
```diff
@@ -14,8 +14,7 @@ Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=on-failure
 TimeoutStopSec=70
 ExecStartPre=/bin/rm -f %t/%n.ctr-id
-ExecStartPre=/bin/bash /usr/bin/microshift-aio # not autogenerated
-ExecStart=/usr/bin/podman run --cidfile=%t/%n.ctr-id --sdnotify=conmon --cgroups=no-conmon --rm --replace -d --name microshift-aio --privileged -v /lib/modules:/lib/modules -v microshift-data:/var/lib --label io.containers.autoupdate=registry -p 6443:6443 quay.io/microshift/microshift:4.7.0-0.microshift-2021-08-31-224727-aio
+ExecStart=/usr/bin/podman run --cidfile=%t/%n.ctr-id --sdnotify=conmon --cgroups=no-conmon --rm --replace -d --name microshift --privileged -v /lib/modules:/lib/modules -v microshift-data:/var/lib --label io.containers.autoupdate=registry -p 6443:6443 quay.io/microshift/microshift-aio:latest
```
husky-parul, line 17: Why do we need `/lib/modules:/lib/modules`?

sallyom: we don't, I was just going in to remove it! doing that now thanks :)

sallyom:
@husky-parul ci is probably failing looking for the removed files, :eyes:

packaging/systemd/microshift-aio.service
sallyom: updated, removed!

sallyom:
@cooktheryan this should merge, right? is it stuck? thnx

packaging/systemd/microshift-aio.service
```diff
@@ -14,8 +14,7 @@ Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=on-failure
 TimeoutStopSec=70
 ExecStartPre=/bin/rm -f %t/%n.ctr-id
-ExecStartPre=/bin/bash /usr/bin/microshift-aio # not autogenerated
```
larsks, line 17: While the volume check was unnecessary, the `microshift-aio` script also took care of ensuring that the `container_manage_cgroup` boolean was set. If you're dropping this script, should that functionality be rolled into the unit as another `ExecStart` directive, or is that simply going to become a documentation issue?

sallyom: we considered that, yes, and decided it would be a documentation issue. we can reconsider that, though - the thought was it's not the norm to manage such things (selinux booleans) in a unit file. definitely up for discussion, though!

PR #555, opened 2022-01-14
Wait for crd established

### LINKED ISSUE #516: [BUG] Missing readiness check in openshift-crd-manager
#### What happened:
`openshift-apiserver` depends on `openshift-crd-manager` to be ready:
https://github.com/redhat-et/microshift/blob/main/pkg/controllers/openshift-apiserver.go#L53
`openshift-crd-manager` currently only applies the CRDs, but doesn't wait for them to become active before signalling readiness:
https://github.com/redhat-et/microshift/blob/main/pkg/controllers/openshift-crd-manager.go#L44-L48

We may want to add a readiness check for CRDs to become available before starting up openshift-apiserver and -cm.

copejon:

Closes #516 

At present, the openshift-crd-manager performs a simple Get() on the applied CRDs to determine readiness.  A more precise (and [recommended method](https://v1-20.docs.kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/)) would be to check for the `status.conditions.status.type["Established"]` == `"true"`.  This guards against a case where the CRD was applied, the naming controller sets the `NameAccepted: true` condition, but the CRD is not yet "Established".  This can result in a race window where CRs will be rejected with a "CRD doesn't exist" type error, even though the apiserver is actively serving the API endpoint (i.e. Get() succeeds).


mangelajo (commented):
/lgtm

mangelajo:
@oglok  ^

pkg/assets/crd.go
```diff
@@ -44,57 +42,82 @@ var (
...
-	return nil
+	return false, nil
 }
 
-func waitForCRD(client apiextclientv1beta1.CustomResourceDefinitionsGetter, resource *apiextv1beta1.CustomResourceDefinition) error {
-	var lastErr error
-	if err := wait.Poll(customResourceReadyInterval, customResourceReadyTimeout, func() (bool, error) {
-		crd, err := client.CustomResourceDefinitions().Get(context.TODO(), resource.Name, metav1.GetOptions{})
+func WaitForCrdsEstablished(cfg *config.MicroshiftConfig) error {
+	lock.Lock()
```
rootfs, line 87: what is the lock for?

pkg/assets/crd.go
```diff
@@ -44,57 +42,82 @@ var (
...
 		}
 	case "apiextensions.k8s.io/v1beta1":
 		v1beta1Obj := obj.(*apiextv1beta1.CustomResourceDefinition)
-		_, err := clientset.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.TODO(), v1beta1Obj.Name, metav1.GetOptions{})
+		crd, err := clientset.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.TODO(), v1beta1Obj.Name, metav1.GetOptions{})
 		if err != nil {
 			lastErr := fmt.Errorf("error getting CustomResourceDefinition %s: %v", v1beta1Obj.Name, err)
 			logrus.Infof("getting openshift CRD status %v", lastErr)
-			return lastErr
+			if apierrors.IsNotFound(err) {
```
rootfs, line 70: maybe some refactoring here to check err and `crd.Status.Conditions` in one block

pkg/controllers/openshift-crd-manager.go
```diff
@@ -47,5 +47,10 @@ func (s *OpenShiftCRDManager) Run(ctx context.Context, ready chan<- struct{}, st
 	}
 	logrus.Infof("%s applied default CRDs", s.Name())
 
+	logrus.Infof("%s waiting for CRDs acceptance before proceeding", s.Name())
+	if err := assets.WaitForCrdsEstablished(s.cfg); err != nil {
+		logrus.Errorf("%s unable to confirm all CRDs are ready: %v", s.Name(), err)
+	}
+	logrus.Infof("%s all CRDs are ready", s.Name())
```
rootfs, line 54: should it exit here on error?

copejon: I was scratching my head a bit about the intent in the original code, which just logs the error, signals the service is "ready" unconditionally after it returns.  I opted to keep the PR scope focused on the wait logic.

https://github.com/redhat-et/microshift/blob/c006b98aa6003d14f13ba68d5341004b85c85e39/pkg/controllers/openshift-crd-manager.go#L42-L51

pkg/assets/crd.go
copejon: Unneeded I think, will drop.  Looks like the lock exists to flow-control calls to the api server for `readerApplier`'s.  The crd manager doesn't implement that interface though, so maybe it's not needed at all?

pkg/assets/crd.go
copejon: Yeah, it's not super readable. Will clean up.

PR #559, opened 2022-01-19
add script to run dlv with the microshift binary

mangelajo:
This is helpful to run the delve debugger in a remote host (or locally)
and then connect from your favorite IDE or delve instance for
remote debugging.

### DIFF
--- a/hack/runDlv.sh
+++ b/hack/runDlv.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+DLV=$HOME/go/bin/dlv
+[[ -x $DLV ]] || \
+  sudo dnf install -y golang &&
+  go install github.com/go-delve/delve/cmd/dlv@latest
+
+sudo systemctl kill microshift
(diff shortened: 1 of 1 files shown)


hack/runDlv.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+DLV=$HOME/go/bin/dlv
+[[ -x $DLV ]] || \
+  sudo dnf install -y golang &&
+  go install github.com/go-delve/delve/cmd/dlv@latest
+
+sudo systemctl kill microshift
+sudo systemctl disable --now microshift
+sudo systemctl kill microshift
```
copejon, line 9: Why 2 kills?

mangelajo: That's a good question let's kill one kill :-)

mangelajo, line 9:
```suggestion
```

Issue #558, opened 2022-01-18
[BUG] HTTPS_PROXY or HTTP_PROXY configured on the system breaks openshift-router


mangelajo:
#### What happened:

If the system needs an HTTPS/HTTP proxy to connect outside of the local network and this is configured
on the system the openshift services will also try to use this proxy for contacting the board API server,
for example k8s api connects to OpenShift api to expose the OpenShift resources, this will fail.

listing `oc get apiservices` the router api will be offline.

#### What you expected to happen:

OpenShift router working normally, `oc get apiservices` to all resources online even when HTTP/HTTPS proxies are configured.

#### Environment:
- Microshift version (use `microshift version`):
- Hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

#### Relevant Logs 

```
I0112 18:14:21.226488       1 template.go:437] router "msg"="starting router"  "version"="majorFromGit: \nminorFromGit: \ncommitFromGit: 9c9f9f42\nversionFromGit: 4.0.0-312-g9c9f9f42\ngitTreeState: clean\nbuildDate: 2021-10-18T17:15:38Z\n"
I0112 18:14:21.231297       1 metrics.go:155] metrics "msg"="router health and metrics port listening on HTTP and HTTPS"  "address"="0.0.0.0:1936"
I0112 18:14:21.265919       1 router.go:191] template "msg"="creating a new template router"  "writeDir"="/var/lib/haproxy"
I0112 18:14:21.266966       1 router.go:270] template "msg"="router will coalesce reloads within an interval of each other"  "interval"="5s"
I0112 18:14:21.280596       1 router.go:332] template "msg"="watching for changes"  "path"="/etc/pki/tls/private"
I0112 18:14:21.281687       1 router.go:262] router "msg"="router is including routes in all namespaces"  
E0112 18:14:21.314274       1 reflector.go:138] pkg/router/controller/factory/factory.go:125: Failed to watch *v1.Route: failed to list *v1.Route: the server is currently unable to handle the request (get routes.route.openshift.io)
E0112 18:14:22.808063       1 reflector.go:138] pkg/router/controller/factory/factory.go:125: Failed to watch *v1.Route: failed to list *v1.Route: the server is currently unable to handle the request (get routes.route.openshift.io)
[...]

PR #561, opened 2022-01-19
Populate the the NO_PROXY environment

mangelajo:
Fixes-Issue: #558

### DIFF
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -20,6 +20,7 @@ import (
 	tcpnet "net"
 	"net/http"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/util/net.go
```diff
@@ -81,3 +85,42 @@ func CreateLocalhostListenerOnPort(port int) (tcpnet.Listener, error) {
...
+}
+
+func mapKeys(entries map[string]struct{}) []string {
+	keys := make([]string, 0, len(entries))
+	for k := range entries {
+		keys = append(keys, k)
+	}
+
+	// sort keys to avoid issues with map key ordering in go future versions on the unit-test side
+	sort.Strings(keys)
```
mangelajo, line 114: I know the sort is unnecessary on production, but guarantees key ordering during unit testing, which is not a problem in current go, but future go versions will change this.

mangelajo:
Sorry for the new after LGTM, but it came without batteries included (testing :-P)

mangelajo:
ok, I verified it works on a reproduced environment.

Issue #556, opened 2022-01-17
[BUG] If the interface IP changes Microshift stops working


mangelajo:
#### What happened:

The openshift-api api endpoints are created based on the existing IP address when microshift starts,
if the IP changes, the kubernetes controller can't connect to the endpoint, and starts failing.

#### What you expected to happen:
Microshift should have recognized the IP change and updated the endpoints, or used a VIP for the API purposes.

#### How to reproduce it (as minimally and precisely as possible):

1. Assign an IP address to the MicroShift host
2. Start MicroShift
3. Change the IP address
4. MicroShift stops working
5. restart MicroShift
6. MicroShift still not working

DanielFroehlich:
workaround:
1. list endpoints: `oc get endpoints -A`
2. identify endpoints with the wrong (old) IP adress
3. `oc edit`  the endpoints and change to new ip adress
4.  `systemctl restart microshift`

PR #564, opened 2022-01-21
WIP: Add ipwatch controller to handle IP address changes

mangelajo:
The IPWatchController will restart MicroShift when IP address
changes have been detected.

This is an initial implementation which will trigger the initialization
of all components with the new IP address and relies on systemd or
podman restarting the process.

Related-Issue: #556


fzdarsky:
There exists already a mechanism for shutting down MicroShift and its embedded services, which is to cancel the context, either directly or via a SIGTERM sent to MicroShift. I believe we should reuse that mechanism rather than implementing a new + parallel mechanism.

oglok:
Closing this PR in favour of #650

Issue #162, opened 2021-07-15
Pre-loading container images into CRI-O


fzdarsky:
MicroShift needs a built-in mechanism to pre-load container images into nodes' CRI-O instances so that it can bootstrap its component services a) without network connectivity and b) without local image registry.

The mechanism should monitor `/var/lib/microshift/images` for oci-archives or multi-image docker-archives as produced by `podman save` and pre-load them into CRI-O on cluster initialization or whenever that directory changes.

Requirements:
* MUST NOT change image digests, so the images can be referenced from K8s manifests by digest, too.
* MUST be compatible with multi-arch images and having multiple tarballs/versions of images (e.g. for updates)
* SHOULD unzip tarballs if necessary.
* SHOULD support deleting pre-loaded images from `/var/lib/microshift/images` and subsequently pruning from CRI-O.

cooktheryan:
@fzdarsky do you have a minimum list of images we would want to ship?

rootfs:
@cooktheryan You can start with `kubect get pods -A |grep image` to filter out the actively running images.

cooktheryan:
@fzdarsky @rootfs I wanted to get your input on this before writing the code in go. This is what I am thinking (in bash) 
```
for i in `sudo ls /var/lib/microshift/images`; do sudo lsof $i && sudo podman load $i && sudo rm -rf $i ; done
```

We start the process on a predefined schedule and log to the microshift log file. We want to do lsof or something similar to make sure the file isn't being written to at the time of us attempting to load. Once loading we would remove the file.

rootfs:
@cooktheryan what does `podman load` for? do you mean running `podman load` while starting microshift?

cooktheryan:
The for loop was just the example of dynamically listing the items within the directory. For the loading before starting microshift we would need to make sure the deployments did not attempt to begin before the images were loaded

PR #568, opened 2022-01-25
Pre-loading container images into CRI-O

husky-parul:

Pre-loading container images into CRI-O 

Closes #<Issue Number>
#162

Should be merged after #593

### DIFF
--- a/packaging/rpm/make-microshift-app-images-rpm.sh
+++ b/packaging/rpm/make-microshift-app-images-rpm.sh
@@ -36,7 +36,7 @@ Name: microshift-app-images
 Version: 1
 Release: 1
 Summary: Creates RO container storage for user applications
-License: ASL 2.0
+License: Apache License 2.0
 URL: https://github.com/redhat-et/microshift
 
 BuildRequires:  podman
--- a/packaging/rpm/make-microshift-app-images-rpm.sh
+++ b/packaging/rpm/make-microshift-app-images-rpm.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+
+# First arg: file path containing user images per line
+# Second arg: container storage dir path
+# Third arg:  RPMBUILD_DIR
+
+RPMBUILD_DIR=$3
+_img_dir_=$2
+
+declare -a ARRAY
+
+#link filedescriptor 10 with stdin (standard input)
+exec 10<&0
+
+#stdin replaced with a file supplied as a first argument
+exec < $1
+let count=0
+
+#read user images into ARRAY
+while read LINE; do
+   ARRAY[$count]=$LINE
+   ((count++))
+done
+
+#restore stdin from file descriptor 10 then close filedescriptor 10
+exec 0<&10 10<&-
+
+#Generate microshift-app-images.spec 
+touch ./microshift-app-images.spec
+cat >./microshift-app-images.spec <<EOF
+%global _img_dir $_img_dir_
+%global imageStore %{buildroot}%{_img_dir}
+%global _IMAGES_ ${ARRAY[@]}
+
+Name: microshift-app-images
+Version: 1
+Release: 1
+Summary: Creates RO container storage for user applications
+License: ASL 2.0
+URL: https://github.com/redhat-et/microshift
+
+BuildRequires:  podman
+BuildRequires: crio
+
+
+%description
+This rpm creates a RO container storage for user applications, pull the app images and add the path to additional container image stores on target machine.
+
+%prep
+
+if [ -d  %{imageStore} ] 
+then
+   sudo rm -rf  %{imageStore}
+fi
+
+%build
+
+%install
+
+mkdir -p %{imageStore}
+
+declare -a ListOfImages=(%{_IMAGES_})
+
+
+for val in \${ListOfImages[@]}; do
+   sudo podman pull --root %{imageStore} \$val
+done
+sudo chmod -R a+rx  %{imageStore}
+
+
+%post
+sudo sed -i '/^additionalimagestores =*/a "$_img_dir_",' /etc/containers/storage.conf
+# if crio was already started, restart it so it read from new imagestore
+systemctl is-active --quiet crio && systemctl restart --quiet crio
+
+
+%files
+%{_img_dir}/*
+
+EOF
+cp ./microshift-app-images.spec $RPMBUILD_DIR/SPECS/microshift-app-images.spec
+
+QA_RPATHS=$(( 0x0001|0x0010 )) rpmbuild -bb $RPMBUILD_DIR/SPECS/microshift-app-images.spec
--- a/packaging/rpm/make-microshift-app-images-rpm.sh
+++ b/packaging/rpm/make-microshift-app-images-rpm.sh
@@ -80,4 +80,4 @@ systemctl is-active --quiet crio && systemctl restart --quiet crio
 EOF
 cp ./microshift-app-images.spec $RPMBUILD_DIR/SPECS/microshift-app-images.spec
 
-QA_RPATHS=$(( 0x0001|0x0010 )) rpmbuild -bb $RPMBUILD_DIR/SPECS/microshift-app-images.spec
+rpmbuild -bb --define "_topdir ${RPMBUILD_DIR}" $RPMBUILD_DIR/SPECS/microshift-app-images.spec
--- a/packaging/rpm/microshift-images.spec
+++ b/packaging/rpm/microshift-images.spec
@@ -1,5 +1,8 @@
 %global imageStore %{buildroot}/var/lib/microshift/images
 
+%define __arch_install_post QA_SKIP_RPATHS=2 /usr/lib/rpm/check-rpaths
+%define __brp_check_rpaths QA_SKIP_RPATHS=2 /usr/lib/rpm/check-rpaths
+
 Name: microshift-images
 Version: 1
 Release: 1
--- a/packaging/rpm/microshift-images.spec
+++ b/packaging/rpm/microshift-images.spec
@@ -1,65 +1,114 @@
-%global imageStore %{buildroot}/var/lib/microshift/images
-
-%define __arch_install_post QA_SKIP_RPATHS=2 /usr/lib/rpm/check-rpaths
-%define __brp_check_rpaths QA_SKIP_RPATHS=2 /usr/lib/rpm/check-rpaths
-
 Name: microshift-images
-Version: 1
-Release: 1
-# this can be %{timestamp}.git%{short_hash} later for continous main builds
-Summary: Create custom container storage
+
+# disable dynamic rpmbuild checks
+%global __os_install_post /bin/true
+%global __arch_install_post /bin/true
+AutoReqProv: no
+
+# where do we want the images to be stored on the final system
+%global imageStore /opt/microshift/images
+%global imageStoreSed %(echo %{imageStore} | sed 's/\//\\\//g')
+
+%define version %(echo %{baseVersion} | sed s/-/_/g)
+
+# to-be-improved:
+# avoid warnings for container layers:
+#   - warning: absolute symlink:
+#   - warning: Duplicate build-ids
+
+Version: %{version}
+Release: 2
+
+Summary: MicroShift related container images
 License: ASL 2.0
 URL: https://github.com/redhat-et/microshift
 
-BuildRequires:  podman
+BuildRequires: podman
 BuildRequires: crio
 
 
 %description
-This rpm creates a custom RO container storage and pull images and add path to additional container image stores.
+This rpm creates a custom RO container storage for the MicroShift container images
+and pull images and add path to additional container image stores.
 
 %prep
 
-if [ -d  %{imageStore} ] 
-then
-   sudo rm -rf  %{imageStore}
-fi
 
+if [ -d  %{buildroot}%{imageStore} ]
+then
+   sudo rm -rf  %{buildroot}%{imageStore}
+fi
 
 %build
 
 
 %install
 
-mkdir -p %{imageStore}
+mkdir -p %{buildroot}%{imageStore}
 
-declare -a ListOfImages=("quay.io/openshift/okd-content@sha256:27f7918b5f0444e278118b2ee054f5b6fadfc4005cf91cb78106c3f5e1833edd" \
-"quay.io/openshift/okd-content@sha256:bcdefdbcee8af1e634e68a850c52fe1e9cb31364525e30f5b20ee4eacb93c3e8" \
-"quay.io/openshift/okd-content@sha256:01cfbbfdc11e2cbb8856f31a65c83acc7cfbd1986c1309f58c255840efcc0b64" \
-"quay.io/coreos/flannel:v0.14.0" \
-"quay.io/microshift/flannel-cni:4.8.0-0.okd-2021-10-10-030117" \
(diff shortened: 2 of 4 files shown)


pkg/controllers/image-loader-controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+	{"/var/lib/microshift/tars/cli.tar", false, 1, []string{ "quay.io/microshift/cli" }},
+	{"/var/lib/microshift/tars/coredns.tar", false, 1, []string{"quay.io/microshift/coredns"}},
+	{"/var/lib/microshift/tars/haproxy_router.tar", false, 1, []string{"quay.io/microshift/haproxy-router"}},
+	{"/var/lib/microshift/tars/kube_flannel.tar", false, 1, []string{"quay.io/microshift/flannel"}},
+	{"/var/lib/microshift/tars/kube_flannel_cni.tar", false, 1, []string{"quay.io/microshift/flannel-cni"}},
+	{"/var/lib/microshift/tars/kube_rbac_proxy.tar", false, 1, []string{ "quay.io/microshift/kube-rbac-proxy"}},
+	{"/var/lib/microshift/tars/kubevirt_hostpath_provisioner.tar", false, 1, []string{"quay.io/microshift/hostpath-provisioner"}},
+	{"/var/lib/microshift/tars/pause.tar", false, 1, []string{"quay.io/microshift/pause:" + Base}},
+	{"/var/lib/microshift/tars/service_ca_operator.tar", false, 1, []string{"quay.io/microshift/service-ca-operator"}},
```
copejon, line 67: Syntax error here - the slice of struct literals should be wrapped in an outer set of curly braces, e.g.

```
var _ = []struct{
    foo string
    bar string
}{ <---
{...},
{...},
{...},
} <---

File: pkg/controllers/image-loader-controller.go [hunk 2]
```diff
@@ -0,0 +1,116 @@
...
+
+	{
+		loadedImages, err := runtime.Load(ctx, listOfImages.input, loadOptions)
+		if err == nil {
+			klog.Fatalf("Error loading image %s",listOfImages.input,err)
+			return err
```
copejon, line 73: `klog.Fatalf()` panics, so the return is unreachable.

pkg/controllers/image-loader-controller.go
```diff
@@ -0,0 +1,116 @@
...
+limitations under the License.
+*/
+package controllers
+
+import (
+	lb "github.com/containers/common/libimage"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/storage"
+	"github.com/openshift/microshift/pkg/config"
+	"k8s.io/klog/v2"
```
copejon, line 23: Missing imports:

```
openshift_apiserver "github.com/openshift/openshift-apiserver/pkg/cmd/openshift-apiserver"
"context"
"os"
```

pkg/controllers/image-loader-controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+	systemContext := &types.SystemContext{
+		SystemRegistriesConfPath:    "/etc/containers/registries.conf",
+		SystemRegistriesConfDirPath: "/dev/null",
+	}
+
+	runtime, err := lb.RuntimeFromStoreOptions(&lb.RuntimeOptions{SystemContext: systemContext}, storeOptions)
+	if err == nil {
+		klog.Errorf("Failed acquiring runtime %v", err)
+		return nil, err
```
copejon, line 106: function doesn't return an error.

pkg/controllers/image-loader-controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+	}
+
+	cleanup = func() {
```
copejon, line 110: Can skip variable assignment and define the anonymous func literal in the return statement.

pkg/controllers/image-loader-controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+	{
+		loadedImages, err := runtime.Load(ctx, listOfImages.input, loadOptions)
+		if err == nil {
+			klog.Fatalf("Error loading image %s",listOfImages.input,err)
+			return err
+		}
+		// Check if all the images are loaded!
+		for _, name := range loadedImages {
+			image, resolvedName, err := runtime.LookupImage(name, nil)
```
copejon, line 77: Variable `image` is unused, will produce a compile time error.

pkg/controllers/image-loader-controller.go
```diff
@@ -0,0 +1,116 @@
...
+			if err !=nil && name == resolvedName{
+					klog.Infof("%s loaded",name)
+			}
+			else{
```
copejon, line 81: `else` should follow the `}` on the same line.

pkg/controllers/image-loader-controller.go
```diff
@@ -0,0 +1,116 @@
...
+				klog.Warningf("One or more images were not loade in container storage",err)
```
copejon, line 82: typo `loade`.

I think too if we're returning the error, it should be logged as an error rather than a warning.

copejon (changes requested):

pkg/controllers/image-loader-controller.go
copejon: Is it safe to ignore `runtime.Shutdown()` errors?

packaging/rpm/make-airgap-rpm.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+imageStrore="/var/lib/microshift/airgap-images"
```
mangelajo, line 3: nit :-) 
```suggestion
imageStore="/var/lib/microshift/airgap-images"
```

packaging/rpm/microshift-airgap.spec
```diff
@@ -0,0 +1,31 @@
...
+This rpm creates a custom RO container storage and pull images for disconnected scenarios.
+
+%prep
+
+%build
+
+
+%install
+mkdir -p /var/lib/microshift/airgap-images
+sudo sed -i '/^additionalimagestores =*/a "/var/lib/microshift/airgap-images",' /etc/containers/storage.conf
```
mangelajo, line 19:
```suggestion
sudo sed -i '/^additionalimagestores =*/a "/var/lib/microshift/airgap-images",' /etc/containers/storage.conf
```
this bit should go in %post ^ right before the systemctl restart

packaging/rpm/microshift-airgap.spec
```diff
@@ -0,0 +1,31 @@
...
+%description
+This rpm creates a custom RO container storage and pull images for disconnected scenarios.
+
+%prep
+
+%build
+
+
+%install
+mkdir -p /var/lib/microshift/airgap-images
```
mangelajo, line 18:
```suggestion
mkdir -p /var/lib/microshift/container-images
```

mangelajo, line 18:
```suggestion
mkdir -p %{buildroot}/var/lib/microshift/airgap-images
```

mangelajo: After that you can call Podman, I wonder if we could include you ssh as a source, but we need to use --root %{buildroot}/var/lib/microshift/container-images

packaging/rpm/microshift-airgap.spec
```diff
@@ -0,0 +1,31 @@
...
+
+
+%post
+# if crio was already started, restart it so it read from new imagestore
+systemctl is-active --quiet crio && systemctl restart --quiet crio
+
+
+
+%files
+
```
mangelajo, line 29:
```suggestion
/var/lib/microshift/container-images/
```

packaging/rpm/microshift-airgap.spec
```diff
@@ -0,0 +1,31 @@
+Name: microshift
```
mangelajo, line 1: @fzdarsky was asking me if we could call this microshift-container-images.spec instead.

packaging/rpm/make-airgap-rpm.sh
fzdarsky: It'd be great to remove all reference to "airgap(ped)" as this this implies more than just disconnected from network (Faraday cages, separate power grid, etc.).

Simply `/var/lib/microshift/images`?

packaging/rpm/microshift-airgap.spec
fzdarsky: Just `/var/lib/microshift/images`?

packaging/rpm/microshift-airgap.spec
fzdarsky: Resp. just "microshift-images.{spec,rpm}", yes. So users would either just `dnf install microshift` and pull images over the network or `dnf install microshift microshift-images` and have images ready on the system when microshift starts.

fzdarsky:
How would this work for updates of MicroShift / MicroShift container images?

How would this work when users want to add their own container images (e.g. klusterlet agent, user applications)?

husky-parul:
@fzdarsky please TAL

packaging/rpm/microshift-images.spec
```diff
@@ -0,0 +1,62 @@
+%global imageStore %{buildroot}/var/lib/microshift/images
+
+Name: microshift-images
+Version: 1
+Release: 1
+# this can be %{timestamp}.git%{short_hash} later for continous main builds
+Summary: Create custom container storage
+License: ASL 2.0
```
fzdarsky, line 8: Is ASL the correct abbreviation (as it's Apache License not Apache _Software_ Licence or such?

packaging/rpm/microshift-images.spec
```diff
@@ -0,0 +1,62 @@
...
+if [ -d  %{imageStore} ] 
+then
+   sudo rm -rf  %{imageStore}
+fi
+
+
+%build
+
+
+%install
```
fzdarsky, line 29: Shouldn't this belong into the %build section, such that the RPM _contains_ the images (rather than pulling them when installed)? Not sure the ImageBuilder can actually run stuff on podman when building ostrees.

packaging/rpm/microshift-images.spec
```diff
@@ -0,0 +1,62 @@
...
+
+mkdir -p %{imageStore}
+
+declare -a ListOfImages=("quay.io/openshift/okd-content@sha256:27f7918b5f0444e278118b2ee054f5b6fadfc4005cf91cb78106c3f5e1833edd" \
```
fzdarsky, line 33: Instead of hardcoding image references here, could we extract them from the release metadata, e.g. [ ](https://github.com/redhat-et/microshift/blob/main/pkg/release/release_amd64.go)?
This is so it's a) guaranteed to be in sync when we rebase and b) it also works for non-amd64?

packaging/rpm/microshift-images.spec
```diff
@@ -0,0 +1,62 @@
...
+"quay.io/openshift/okd-content@sha256:dd1cd4d7b1f2d097eaa965bc5e2fe7ebfe333d6cbaeabc7879283af1a88dbf4e")
+
+for val in ${ListOfImages[@]}; do
+   sudo podman pull  --root %{imageStore} $val
+done
+sudo chmod -R a+rx  %{imageStore}
+
+
+%post
+sudo sed -i '/^additionalimagestores =*/a "/var/lib/microshift/images",' /etc/containers/storage.conf
```
fzdarsky, line 50: This is not idempotent, so would, for example, create duplicate entries on a `dnf reinstall`. Not certain this is a problem in practice.

packaging/rpm/microshift-images.spec
husky-parul: ASL is Apache Software License. Going to change it to AL 2.0

packaging/rpm/microshift-images.spec
husky-parul: It doesn't support non-amd64. Since I tested the rpmbuild on F35 (only amd64 devices), `podman pull` command implicitly pulls images for the respective processor type.

packaging/rpm/microshift-images.spec
husky-parul: As per the doc, https://rpm-packaging-guide.github.io/#what-is-a-spec-file
```
%build: Command or series of commands for actually building the software into machine code (for compiled languages) or byte code (for some interpreted languages).
```

```
%Install: Command or series of commands for copying the desired build artifacts from the %builddir (where the build happens) to the %buildroot directory (which contains the directory structure with the files to be packaged). This usually means copying files from ~/rpmbuild/BUILD to ~/rpmbuild/BUILDROOT and creating the necessary directories in ~/rpmbuild/BUILDROOT. This is only run when creating a package, not when the end-user installs the package. See [Working with SPEC files](https://rpm-packaging-guide.github.io/#working-with-spec-files) for details.
```

Since we are not building s/w and only using rpm to package files, I put it under `%install`

husky-parul:
@mangelajo @fzdarsky 
I understand make-microshift-images-rpm.sh will not create the .rpm for users images. Are we planning to keep two specs, one for microshift-images and one for users images?

packaging/rpm/microshift-images.spec
mangelajo: Latest version only does it on install, and also does uninstall.

packaging/rpm/microshift-images.spec
mangelajo: it's .spec files just being confusing,

What you expect for install is %post

%install is the phase that takes the %build, and put's it into the fake root filesystem,
from where %file will gather stuff.

packaging/rpm/microshift-images.spec
```diff
@@ -0,0 +1,114 @@
...
+   sed -i '/^additionalimagestores =*/a "%{imageStore}",' /etc/containers/storage.conf
+   # if crio was already started, restart it so it read from new imagestore
+   systemctl is-active --quiet crio && systemctl restart --quiet crio
+fi
+
+%postun
+
+# only on uninstall (0), not on upgrades(1)
+if [ $1 -eq 0 ];
+  sed -i '/"${imageStoreSed}",/d" /etc/containers/storage.conf
```
mangelajo, line 103:
```suggestion
  sed -i '/"${imageStoreSed}",/d" /etc/containers/storage.conf
  systemctl is-active --quiet crio && systemctl restart --quiet crio
```

I forgot about this bit ;-)

packaging/rpm/make-microshift-app-images-rpm.sh
```diff
@@ -0,0 +1,83 @@
...
+   ARRAY[$count]=$LINE
+   ((count++))
+done
+
+#restore stdin from file descriptor 10 then close filedescriptor 10
+exec 0<&10 10<&-
+
+#Generate microshift-app-images.spec 
+touch ./microshift-app-images.spec
+cat >./microshift-app-images.spec <<EOF
```
mangelajo, line 30: We should probably propagate the changes from the other spec here. 

But I propose we merge, and then we keep working on the userland side.

mangelajo:
https://github.com/redhat-et/microshift/pull/611
this will be necessary but can be independent of this patch.

packaging/rpm/make-microshift-app-images-rpm.sh
fzdarsky: @mangelajo good idea. Could you open an issue for that please so we don't forget?

PR #574, opened 2022-01-27
Make kube-proxy Service stoppable

mangelajo:
The kube-proxy k8s code does not support an stop channel, but
at least we can accept cancel on the context and move on, so
MicroShift can be stopped properly when necessary.

Related-Issue: #556

### DIFF
--- a/pkg/node/kube-proxy.go
+++ b/pkg/node/kube-proxy.go
@@ -95,20 +95,34 @@ featureGates:
 }
 
 func (s *ProxyOptions) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+	proxyStopErr := make(chan error, 1)
 
 	defer close(stopped)
+
 	// run readiness check
 	go func() {
 		healthcheckStatus := util.RetryInsecureHttpsGet("http://127.0.0.1:10256/healthz")
 		if healthcheckStatus != 200 {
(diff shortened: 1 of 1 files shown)


pkg/node/kube-proxy.go
```diff
@@ -106,9 +108,19 @@ func (s *ProxyOptions) Run(ctx context.Context, ready chan<- struct{}, stopped c
 		klog.Infof("%s is ready", s.Name())
 		close(ready)
 	}()
-	if err := s.options.Run(); err != nil {
-		klog.Fatalf("%s failed to start", s.Name(), err)
```
mangelajo, line 110: returning an error to stop other stuff depends on:

https://github.com/redhat-et/microshift/pull/575

We shouldn't exit so harshly without letting etcd at least shut down properly

mangelajo: oh, you're right @copejon thanks for the review. I'll fix this

mangelajo:
@fzdarsky @copejon @oglok ^ %)

pkg/node/kube-proxy.go
mangelajo: ehhm, my changes didnd't go in... let me fix this, sorry

mangelajo: ok, now. yes...

mangelajo:
Being handled on another PR

PR #575, opened 2022-01-27
If a service stops with error request all the other services to stop

mangelajo:
Related-issue: #556

### DIFF
--- a/pkg/servicemanager/manager.go
+++ b/pkg/servicemanager/manager.go
@@ -3,6 +3,7 @@ package servicemanager
 import (
 	"context"
 	"fmt"
+	"syscall"
 
 	"github.com/openshift/microshift/pkg/util/sigchannel"
 	"k8s.io/klog/v2"
@@ -97,9 +98,21 @@ func (m *ServiceManager) Run(ctx context.Context, ready chan<- struct{}, stopped
 func (m *ServiceManager) asyncRun(ctx context.Context, service Service) (<-chan struct{}, <-chan struct{}) {
 	ready, stopped := make(chan struct{}), make(chan struct{})
 	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				klog.Errorf("%s panicked: %s", service.Name(), r)
+				klog.Error("Stopping MicroShift")
+				syscall.Kill(syscall.Getpid(), syscall.SIGTERM)
+				if !sigchannel.IsClosed(stopped) {
+					close(stopped)
+				}
+			}
+		}()
+
 		klog.Infof("Starting %s", service.Name())
 		if err := service.Run(ctx, ready, stopped); err != nil {
-			klog.Infof("%s stopped: %s", service.Name(), err)
+			klog.Errorf("service %s exited with error: %s, stopping MicroShift", service.Name(), err)
+			syscall.Kill(syscall.Getpid(), syscall.SIGTERM)
 		} else {
 			klog.Infof("%s completed", service.Name())
 		}
--- a/pkg/servicemanager/manager_test.go
+++ b/pkg/servicemanager/manager_test.go
@@ -3,7 +3,10 @@ package servicemanager
 import (
 	"context"
 	"errors"
+	"os"
+	"os/signal"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/servicemanager/manager.go
```diff
@@ -94,12 +94,13 @@ func (m *ServiceManager) Run(ctx context.Context, ready chan<- struct{}, stopped
 	return ctx.Err()
 }
 
-func (m *ServiceManager) asyncRun(ctx context.Context, service Service) (<-chan struct{}, <-chan struct{}) {
+func (m *ServiceManager) asyncRun(ctx context.Context, cancel context.CancelFunc, service Service) (<-chan struct{}, <-chan struct{}) {
 	ready, stopped := make(chan struct{}), make(chan struct{})
 	go func() {
 		klog.Infof("Starting %s", service.Name())
 		if err := service.Run(ctx, ready, stopped); err != nil {
```
copejon, line 101: If the `Run()` call panics (either due to a panic called by our code, or by vendored dependent service) the cancel() won't occur.  It may be cleaner to exec the cancel in a `defer func(){  if r := recover(); r != nil { cancel(); //log the r string } }()` in a parent caller.  This should allow us to report fatal errors as fatal, shut down the services gracefully, and hopefully curb some of those 10k line stack traces.

mangelajo: I like that. @fzdarsky was commenting on something similar but simpler that other projects are doing.

mangelajo: Done, and verified with testing:

``` service-manager: context canceled
I0128 14:43:51.473440  197278 manager.go:115] Starting foo
I0128 14:43:51.473470  197278 manager.go:115] Starting bar-panic
E0128 14:43:52.474336  197278 manager.go:103] bar-panic crashed in panic: I'm in panic, stopping MicroShift
I0128 14:43:52.474368  197278 manager.go:120] foo completed ```

File: pkg/cmd/run.go [hunk 2]
```diff
@@ -119,8 +119,16 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	ready, stopped := make(chan struct{}), make(chan struct{})
 	go func() {
+		defer func() {
```
mangelajo, line 122: This shouldn't go here

pkg/servicemanager/manager.go
```diff
@@ -94,12 +95,28 @@ func (m *ServiceManager) Run(ctx context.Context, ready chan<- struct{}, stopped
 	return ctx.Err()
 }
 
-func (m *ServiceManager) asyncRun(ctx context.Context, service Service) (<-chan struct{}, <-chan struct{}) {
+func (m *ServiceManager) asyncRun(ctx context.Context, cancel context.CancelFunc, service Service) (<-chan struct{}, <-chan struct{}) {
 	ready, stopped := make(chan struct{}), make(chan struct{})
 	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				klog.Errorf("%s crashed in panic: %s", service.Name(), r)
```
fzdarsky, line 103: "crashed in panic" --> "panicked"?

pkg/servicemanager/manager.go
```diff
@@ -94,12 +95,28 @@ func (m *ServiceManager) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+				klog.Error("Stopping MicroShift")
+				cancel()
```
fzdarsky, line 105: Calling `cancel()` here stops the ServiceManager and its child services, but the main RunMicroShift() routine would still be stuck waiting for a SIGTERM, no?

It seems quite messy calling `cancel()` from multiple locations. Is there a better way handling this, e.g. sending SIGTERM to self, so we'll keep a single flow of control?

fzdarsky (changes requested):

pkg/servicemanager/manager.go
```diff
@@ -49,7 +50,7 @@ func (m *ServiceManager) AddService(s Service) error {
 	return nil
 }
 
-func (m *ServiceManager) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+func (m *ServiceManager) Run(ctx context.Context, cancel context.CancelFunc, ready chan<- struct{}, stopped chan<- struct{}) error {
```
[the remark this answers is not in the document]
mangelajo: true, but is it that important? did we plan to nest them?

We could put it in the interface, but is there any reason for the managed services to have access to cancel?

pkg/servicemanager/manager.go
mangelajo: Makes sense, I believe that would work.

pkg/servicemanager/manager.go
```diff
@@ -94,12 +95,28 @@ func (m *ServiceManager) Run(ctx context.Context, ready chan<- struct{}, stopped
...
 	ready, stopped := make(chan struct{}), make(chan struct{})
 	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				klog.Errorf("%s crashed in panic: %s", service.Name(), r)
+				klog.Error("Stopping MicroShift")
+				cancel()
+				// If the channel was not closed by the service, we close
+				// it after 100ms to make sure the service manager moves on
+				select {
```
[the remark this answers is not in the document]
mangelajo: close(stopped) if it was closed already in the service before panicking, would crash.

We could add another defer / recover before trying to close.

I'm not trying to add timeout, but trying to close the channel (never twice, twice == panic)

pkg/servicemanager/manager.go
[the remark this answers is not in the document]
mangelajo: Done.

pkg/servicemanager/manager.go
mangelajo: Done.

pkg/servicemanager/manager.go
mangelajo: Done.

pkg/servicemanager/manager.go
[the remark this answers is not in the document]
mangelajo: writing this on a different way :-)

fzdarsky (approved):

PR #577, opened 2022-01-27
Safely shutdown MicroShift if kustomization fails

mangelajo:
Depends-On: #575


mangelajo:
Being handled on another PR

PR #580, opened 2022-01-27
Make ocp-api-server stoppable by manager

mangelajo:


pkg/controllers/openshift-apiserver.go
```diff
@@ -119,12 +119,23 @@ func (s *OCPAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped c
 		return err
 	}
 
+	errorChannel := make(chan error, 1)
```
copejon, line 122: The error channel could also be used to capture error output from the preceding go routine.  As it is the return values from the routine are lost.

mangelajo:
handled on a separate pr

PR #581, opened 2022-01-27
OpenShift CRD Manager service must signal stop when completed

mangelajo:
Otherwise the service manager can't track completion of the service
when a stop is requested.


pkg/controllers/openshift-crd-manager.go
```diff
@@ -39,10 +39,11 @@ func (s *OpenShiftCRDManager) Dependencies() []string {
 }
 
 func (s *OpenShiftCRDManager) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
-	defer close(ready)
-	// To-DO add readiness check
```
mangelajo, line 43: the check is already there (waitForCrdEstablishment)

PR #582, opened 2022-01-27
Let the OpenshiftControllerManager service to be stopped

mangelajo:

### DIFF
--- a/pkg/controllers/openshift-controller-manager.go
+++ b/pkg/controllers/openshift-controller-manager.go
@@ -17,6 +17,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -102,12 +103,14 @@ servingInfo:
 
 func (s *OCPControllerManager) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
 	defer close(stopped)
+	errorChannel := make(chan error, 1)
 
 	// run readiness check
 	go func() {
 		healthcheckStatus := util.RetryTCPConnection("127.0.0.1", "8445")
 		if !healthcheckStatus {
-			klog.Fatalf(s.Name(), fmt.Errorf("healthcheck status"), "%s failed to start")
+			klog.Errorf(s.Name(), fmt.Errorf("healthcheck status"), "%s failed to start")
+			errorChannel <- errors.New("openshift-controller-manager healthcheck status failed")
 		}
 		klog.Infof("%s is ready", s.Name())
 		close(ready)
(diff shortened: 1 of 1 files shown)


pkg/controllers/openshift-controller-manager.go
```diff
@@ -102,12 +103,14 @@ servingInfo:
...
 func (s *OCPControllerManager) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
 	defer close(stopped)
+	errorChannel := make(chan error, 1)
 
 	// run readiness check
 	go func() {
 		healthcheckStatus := util.RetryTCPConnection("127.0.0.1", "8445")
 		if !healthcheckStatus {
-			klog.Fatalf(s.Name(), fmt.Errorf("healthcheck status"), "%s failed to start")
+			klog.Errorf(s.Name(), fmt.Errorf("healthcheck status"), "%s failed to start")
```
copejon, line 112: This log format doesn't look right.  Errorf() takes a format string first, which here is passed as the last arg.

pkg/controllers/openshift-controller-manager.go
```diff
@@ -116,13 +119,24 @@ func (s *OCPControllerManager) Run(ctx context.Context, ready chan<- struct{}, s
 	if err := assets.ApplyNamespaces([]string{
 		"assets/core/0000_50_cluster-openshift-controller-manager_00_namespace.yaml",
 	}, s.kubeconfig); err != nil {
-		klog.Warningf("failed to apply openshift namespaces %v", err)
+		klog.Errorf("failed to apply openshift namespaces %v", err)
```
copejon, line 122: Assuming RetryTCPConnection writes an error to the chan, what's the expected sequence of events?  It looks as though the err will be placed in the chan buffer, but ApplyNamespaces will be called.  If ApplyNamespaces errors, then the error written to the chan is lost at the return. 

Or:
If RetryTCPConnection sends an error to the chan buff, but ApplyNamespaces returns nil, then the 2nd go func is run.  Then the select will read from the chan buff and not block since there's a value to read, causing the function to return.  The concurrent StartControllerManager() call will continue running however.  

Question is: is it safe to assume the StartControllerManager routine will die gracefully and quietly (without spamming logs) if RetryTCPConnection returns and error?

pkg/controllers/openshift-controller-manager.go
mangelajo: oh right, I was just changing severity, but you're right.

mangelajo:
@copejon please see https://github.com/redhat-et/microshift/pull/637 for more details, I've moved work here to that PR and I believe I handled your concerns.

We also decided not to change the behaviour of services into soft-failures (@fzdarsky had a stronger opinion on this and didn't want that to be modified). So errors will be Fatalf which forces the process into exit in case of a failure.

PR #584, opened 2022-01-27
Make kube-scheduler Service return errors for healthcheck

mangelajo:

### DIFF
--- a/pkg/controllers/kube-scheduler.go
+++ b/pkg/controllers/kube-scheduler.go
@@ -17,6 +17,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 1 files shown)


fzdarsky:
See #585. Need to either panic on error here or globally.

mangelajo:
> See #585. Need to either panic on error here or globally.

>> Just noticed that returning an error instead of panicing means the ServiceManager just logs the service as stopped (with error) but MicroShift continues running in an undefined state. Need to either panic on error here or globally.


Not after https://github.com/redhat-et/microshift/pull/575 , I think it's ok now to move back to LGTM :-)

PR #585, opened 2022-01-27
Make kube-api-server to return errors from healthcheck

mangelajo:


fzdarsky:
Had to wrap my head around the various conditions on the error channel, but that makes much sense.

/lgtm

fzdarsky:
Just noticed that returning an error instead of panicing means the ServiceManager just logs the service as stopped (with error) but MicroShift continues running in an undefined state. Need to either panic on error here or globabally.

mangelajo:
> 

Not after https://github.com/redhat-et/microshift/pull/575 , I think it's ok now to move back to LGTM :-)

PR #592, opened 2022-02-01
Add after-release step triggering srpm/rpm build

mangelajo:


.github/workflows/release_to_copr.yaml
```diff
@@ -0,0 +1,35 @@
+name: release_to_copr
+
+# only execute a release when triggered manually
```
mangelajo, line 3:
```suggestion
```

Issue #591, opened 2022-02-01
Podman now looks like a dependency in the RPM installation


nerdalert:
#### What happened:

Install is suddenly failing when the Podman binary is not found, it was working 24 hours ago. There has been only one commit merged since then and it's possible it may be related to [If a service stops with error request all the other services to stop](https://github.com/redhat-et/microshift/commit/f7d423d1445b4af34deb911f5faab2795d2df679) but I'm not even sure that would be reflected in the RPM install yet. 

I'm installing from [nerdalert/microshift-installer](https://github.com/nerdalert/microshift-installer/blob/main/microshift-install.sh) using the RPM install instructions. 

- Below are the logs:

```shell
Jan 31 01:53:49 cluster-b systemd[2096]: microshift.service: Failed at step EXEC spawning /usr/bin/podman: No such file or directory
Jan 31 01:53:49 cluster-b systemd[1]: microshift.service: Main process exited, code=exited, status=203/EXEC
Jan 31 01:53:49 cluster-b systemd[2097]: microshift.service: Failed to locate executable /usr/bin/podman: No such file or directory
Jan 31 01:53:49 cluster-b systemd[2097]: microshift.service: Failed at step EXEC spawning /usr/bin/podman: No such file or directory
Jan 31 01:53:49 cluster-b systemd[1]: microshift.service: Control process exited, code=exited, status=203/EXEC
Jan 31 01:53:49 cluster-b systemd[1]: microshift.service: Failed with result 'exit-code'.
Jan 31 01:53:49 cluster-b systemd[1]: Failed to start MicroShift Containerized.
Jan 31 01:53:50 cluster-b systemd[1]: microshift.service: Scheduled restart job, restart counter is at 5.
Jan 31 01:53:50 cluster-b systemd[1]: Stopped MicroShift Containerized.
Jan 31 01:53:50 cluster-b systemd[1]: microshift.service: Start request repeated too quickly.
Jan 31 01:53:50 cluster-b systemd[1]: microshift.service: Failed with result 'exit-code'.
Jan 31 01:53:50 cluster-b systemd[1]: Failed to start MicroShift Containerized.
```

Ive only been digging into Microshift the past couple of days and haven't looked into if there is any e2e spun up yet. Would like to contribute if I can be of assistance getting some rolling if there isn't any.

#### How to reproduce it (as minimally and precisely as possible):

1. running the script linked and commenting out the Podman install (I have added Podman as a workaround)
2. or using the RPM instructions on the website

[...]

PR #595, opened 2022-02-02
Remove the microshift-containerized subpackage

mangelajo:
Our docs explain how to download the .service file [1] instead,
and it has proven problematic to package this.

Fix the microshift.service being overwritten by microshift-containerized,
even when the non-containerized package only is installed.

Resolves-Issue: #591


mangelajo:
@fzdarsky it compiled properly now: 

```
+ /usr/bin/rm -rf /home/majopela/GolandProjects/microshift/packaging/rpm/_rpmbuild/BUILDROOT/microshift-4.8.0-2022_01_06_210147_20_g22f61dca.fc35.x86_64
+ RPM_EC=0
++ jobs -p
+ exit 0
```

PR #620, opened 2022-03-07
Enable kubelet cgroup management

### LINKED ISSUE #617: [BUG] MicroShift is not parenting user workload in expected part of cgroup tree
#### What happened:
MicroShift has configured `kubelet` to launch containers under `system.slice` rather than `kubepods.slice`.

I expected `kubelet` to launch with the following parameters:

```
--cgroups-per-qos=true
--cgroup-driver=systemd
--cgroup-root=kubepods
```

This ensures that all cluster workload is isolated from `system.slice` and ensures that cfs_shares are evaluated local to other pods properly, as well as makes it easier to understand the `memory_usage` for all pods running on the platform.

Launch system using vagrant 34 VM using RPM based install method for MicroShift.

derekwaynecarr:
Closes #617

Properly configure kubelet cgroup management.

All end-user workloads are partitioned under /kubepods.slice.

Global cpu/memory accounting should be enabled via systemd host configuration as part of host setup.
Maybe we can ensure this is done in the RPM as a follow-up?

### DIFF
--- a/pkg/node/kubelet.go
+++ b/pkg/node/kubelet.go
@@ -118,7 +118,6 @@ authentication:
 tlsCertFile: ` + cfg.DataDir + `/resources/kubelet/secrets/kubelet-client/tls.crt
 tlsPrivateKeyFile: ` + cfg.DataDir + `/resources/kubelet/secrets/kubelet-client/tls.key
 cgroupDriver: "systemd"
-cgroupRoot: /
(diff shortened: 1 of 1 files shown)


fzdarsky:
I'm addressing the second part in a separate PR #697.

fzdarsky:
Partially addresses USHIFT-25.

PR #621, opened 2022-03-07
Use default GOMAXPROCS behavior

derekwaynecarr:
Stop setting GOMAXPROCS as more recent versions of Golang already defaults to NumCPU.
Aligns with OpenShift Standalone.

see: https://pkg.go.dev/runtime#GOMAXPROCS


mangelajo (commented):
/lgtm

PR #637, opened 2022-03-25
Equalise service stop or failures

mangelajo:
This PR handles service stop/failure to make them all behave equally, and also to be able
to be stopped on request, please see individual commits.

- Make kube-proxy Service stoppable
- Exit on kustomization failure
- Make oauth-api-server stoppable and return errors
- Make ocp-api-server stoppable by manager
- Let the OpenshiftControllerManager service to be stopped

### DIFF
--- a/pkg/controllers/openshift-apiserver.go
+++ b/pkg/controllers/openshift-apiserver.go
@@ -119,11 +119,9 @@ func (s *OCPAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped c
 		return err
 	}
 
-	stopCh := make(chan struct{})
-	if err := s.options.RunAPIServer(stopCh); err != nil {
+	if err := s.options.RunAPIServer(ctx.Done()); err != nil {
 		klog.Fatalf("Failed to start ocp-apiserver %v", err)
 	}
-
 	return ctx.Err()
 }
 
--- a/pkg/controllers/openshift-oauth-server.go
+++ b/pkg/controllers/openshift-oauth-server.go
@@ -102,7 +102,6 @@ func (s *OpenShiftOAuth) configure(cfg *config.MicroshiftConfig) {
 
 func (s *OpenShiftOAuth) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
 	defer close(stopped)
-	stopCh := make(chan struct{})
 
 	// run readiness check
 	go func() {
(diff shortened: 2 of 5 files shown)


pkg/controllers/openshift-oauth-server.go
```diff
@@ -114,9 +115,12 @@ func (s *OpenShiftOAuth) Run(ctx context.Context, ready chan<- struct{}, stopped
 		close(ready)
 	}()
 
-	if err := oauth_apiserver.RunOAuthAPIServer(s.options, stopCh); err != nil {
-		return err
-	}
+	go func() {
+		if err := oauth_apiserver.RunOAuthAPIServer(s.options, stopCh); err != nil {
```
fzdarsky, line 119: Why create and pass in a new stop channel here, shouldn't this be `ctx.Done()`?

pkg/controllers/openshift-apiserver.go
```diff
@@ -120,10 +120,15 @@ func (s *OCPAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped c
 	}
 
 	stopCh := make(chan struct{})
-	if err := s.options.RunAPIServer(stopCh); err != nil {
-		klog.Fatalf("Failed to start ocp-apiserver %v", err)
-	}
+	defer close(stopCh)
+
+	go func() {
+		if err := s.options.RunAPIServer(stopCh); err != nil {
```
fzdarsky, line 126: Why create and pass in a new stop channel here, shouldn't this be `ctx.Done()`?

mangelajo: As much as I looked I didn0t find the openshift-apiserver receiving the context in any form, so we can't cancel it that way. They use a stop channel. Hopefully this would change in a future.

mangelajo: That's why I wait for the context, outside, and propagate down to the stop channel.

pkg/controllers/openshift-oauth-server.go
mangelajo: Same reason, this api server does not receive any context anywhere,

pkg/controllers/openshift-apiserver.go
mangelajo: As we talked OOB I'm extracting the stop channel with ctx.Done() for simpler code.

pkg/controllers/openshift-oauth-server.go
mangelajo: Sorry, I didn't fully understood your comment, moving to ctx.Done() which I remember doing and not working (channel directionality(, but it works.

mangelajo:
This one should be good now.

PR #645, opened 2022-04-01
Build container rpms from docker repositories

### LINKED ISSUE #162: Pre-loading container images into CRI-O

mangelajo:
This will ease integration on build systems like copr, those
build systems build as a non privileged user where podman pull --root
won't work.

Also, this packs a tarball file in the rpm which is extracted during
the post phase to avoid additional limitations of the build systems
removing the tarball uid/gids of the files. Also cpio storage
in the rpm will save all as root/root unless explicitly told otherwise
via %files directives.

Closes #162 

Co-Authored-By:	Brent Salisbury <bsalisbu@redhat.com>


packaging/rpm/make-microshift-images-rpm.sh
```diff
@@ -23,26 +26,33 @@ ARCHITECTURES=${ARCHITECTURES:-"x86_64 arm64 arm ppc64le riscv64"}
 build() {
   cat >"${RPMBUILD_DIR}"SPECS/microshift-images.spec <<EOF
 %global baseVersion ${BASE_VERSION}
-
+%global github_tag ${GITHUB_TAG}
 EOF
```
mangelajo, line 30: WIP: this all needs to be changed into generating a .yaml file, and calling paack.py

mangelajo:
rebased on main

packaging/rpm/paack.py
```diff
@@ -0,0 +1,345 @@
...
+                                    '%s-%s-%s.src.rpm' % (package['name'], package['version'], package['release']))
+
+            # if we are in no-cleanup mode the existing srpm is reused instead
+            if not (os.path.exists(out_srpm) and self._no_cleanup):
+                spec = SpecFile(package['name'], os.path.join(self._rpmbuild_dir, 'SPECS',
+                    '%s-%s-%s.spec' % (package['name'], package['version'], package['release'])))
+                for arch_info in package['arch']:
+                    arch = arch_info['name']
+                    tar_path = self._create_tarball(package, arch_info)
+                    print("scanning %s to generate files section for arch %s" %(tar_path, arch))
```
oglok, line 240:
```suggestion
                    print("scanning %s to generate files section for arch %s" % (tar_path, arch))
```
nit: space for style consistency

oglok (commented):
/lgtm

Just a nit

PR #659, opened 2022-04-12
Allow multiple kustomization paths

mangelajo:
This extends the single path /var/lib/microshift/manifests
default into:

   - /usr/lib/microshift/manifests
   - /etc/microshift/manifests
   - /var/lib/microshift/manifests

which will be evaluated in order. While the first one could be fully
managed in ostree (core components), the second one could be managed
by configuration management, the third one is maintained for backwards
compatibility purposes.

In addition the MicroShift configuration file supports customization
of this via the manifestsDir entry.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"path/filepath"
+	"reflect"
 	"strconv"
 
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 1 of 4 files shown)


mangelajo:
Last changes include @copejon and @fzdarsky request to maintain the original default manifest directory too for backwards compatibility

oglok:
Thanks for the PR! I think we should test a specific config.yaml pointing at a set of manifests.

Code wise looks good.

mangelajo:
```
[root@fedora majopela]# cat /etc/microshift/config.yaml
---
manifests:
  - /kustom/
```

```
[root@fedora microshift]# microshift run 2>&1 | grep kust
I0421 14:37:33.524394 2568041 manager.go:112] Starting kustomizer
I0421 14:37:33.524699 2568041 apply.go:57] Applying kustomization at /kustom/kustomization.yaml```

mangelajo:
It seems to work as expected :-)

Issue #613, opened 2022-03-04
[Enhancement]: Make Audit Log Retention Settings Configurable


adambkaplan:
#### Design Document Link

PR: #<PR Number>

#### What would you like to be added:

The kube-apiserver has several parameters which allow audit log retention to be tuned. This is exposed by the following flags:

- `--audit-log-maxage` defines the maximum number of days to retain old audit log files
- `--audit-log-maxbackup` defines the maximum number of audit log files to retain
- `--audit-log-maxsize` defines the maximum size in megabytes of the audit log file before it gets rotated

See https://kubernetes.io/docs/tasks/debug-application-cluster/audit/

#### Why is this needed:

With default settings, the audit logs on a minimal cluster can grow quite large. I deployed microshift on a very old laptop with what is by today's standards absurdly small storage (15GB). My cluster ultimately died because the kube-apiserver audit logs generated over the course of ~20 days exceeded 5GB, exhausting the space on the drive.

adambkaplan:
An alternative approach, in the spirit of "minimal configuration": if the audit log directory is provided, audit logs are retained for 7 days?

fzdarsky:
Thank you @adambkaplan, I agree this needs addressing. My gut feeling is to default to a minimal level of audit logging (e.g. like you proposed) or maybe even disable audit logging, then provide a generic mechanism for users to pass parameters like these to kube-apiserver et al. I opened #618 for that.

PR #672, opened 2022-04-29
[WIP] add flag to specify a custom kube-scheduler config file

copejon:

Closes #<Issue Number>

### DIFF
--- a/pkg/controllers/kube-scheduler.go
+++ b/pkg/controllers/kube-scheduler.go
@@ -25,8 +25,8 @@ import (
 
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util"
-	"github.com/spf13/cobra"
 
+	"github.com/spf13/cobra"
 	klog "k8s.io/klog/v2"
 	kubescheduler "k8s.io/kubernetes/cmd/kube-scheduler/app"
 	schedulerOptions "k8s.io/kubernetes/cmd/kube-scheduler/app/options"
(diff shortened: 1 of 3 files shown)


pkg/controllers/kube-scheduler.go
```diff
@@ -60,8 +60,10 @@ func (s *KubeScheduler) configure(cfg *config.MicroshiftConfig) {
 		klog.Fatalf("initialization error command options: %v", err)
 	}
 
+	klog.Infof("kube-scheduler config: %s", cfg.ControlPlane.KubeSchedulerCustomConfigPath)
+
 	args := []string{
-		"--config=" + cfg.DataDir + "/resources/kube-scheduler/config/config.yaml",
+		"--config=" + cfg.ControlPlane.KubeSchedulerCustomConfigPath,
```
mangelajo, line 66: I think it's probably good enough for now, I wonder if we may need to merge what we configure in that config before passing it down to kube-scheduler (for example, to inject the kubeconfig file or setup leader election etc...)

fzdarsky:
1. We had discussed solving this through a generic mechanism to add extra parameters to embedded components (see #613) instead of introducing specific MicroShift configurations (we know we'll need more of such overrides anyway (e.g. for kubelet) and we'll just blow up our config code otherwise.

Maybe we want to keep this to w.i.p. until there's been a decision whether to switch to TopoLVM or not.

pkg/controllers/kube-scheduler.go
copejon: Maybe.  I agree that we should provide a more intelligent "override" path.  Whether MicroShift should do that by default, or via a discrete config/flag I'm not sure yet.

PR #681, opened 2022-05-09
Add openshift-apiserver as dependency for OCM

### LINKED ISSUE #515: [BUG] openshift-controller-manager should depend on openshift-apiserver
#### What happened:
`openshift-controller-manager` starts directly after `kube-apiserver` is ready without waiting for `openshift-apiserver` to be ready, too:
https://github.com/redhat-et/microshift/blob/main/pkg/controllers/openshift-controller-manager.go#L52

Wait for `openshift-apiserver to become ready`.

oglok:

Closes #515

### DIFF
--- a/pkg/controllers/openshift-apiserver.go
+++ b/pkg/controllers/openshift-apiserver.go
@@ -45,10 +45,8 @@ func NewOpenShiftAPIServer(cfg *config.MicroshiftConfig) *OCPAPIServer {
 	return s
 }
 
-func (s *OCPAPIServer) Name() string { return "ocp-apiserver" }
-func (s *OCPAPIServer) Dependencies() []string {
-	return []string{"kube-apiserver", "openshift-crd-manager"}
-}
+func (s *OCPAPIServer) Name() string           { return "ocp-apiserver" }
+func (s *OCPAPIServer) Dependencies() []string { return []string{"kube-apiserver"} }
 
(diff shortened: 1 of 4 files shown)


copejon:
The error in the e2e-rpm-install test is a happening because this patch's kubelet version (1.21) does not support (or have the FeatureGate enabled for) `SizeMemoryBackedVolumes`.  The latest vendored version is 1.23, where the feature is now beta and enabled by default.

ggiguash:
USHIFT-21

pkg/controllers/openshift-apiserver.go
```diff
@@ -45,10 +45,8 @@ func NewOpenShiftAPIServer(cfg *config.MicroshiftConfig) *OCPAPIServer {
 	return s
 }
 
-func (s *OCPAPIServer) Name() string { return "ocp-apiserver" }
-func (s *OCPAPIServer) Dependencies() []string {
-	return []string{"kube-apiserver", "openshift-crd-manager"}
-}
+func (s *OCPAPIServer) Name() string           { return "ocp-apiserver" }
```
fzdarsky, line 48: Nitpicking here, but shall we rename this `openshift-apiserver` to correspond with the component + repo name?

copejon: +1 improving readability

oglok: I'll address these changes in a following PR.

PR #687, opened 2022-05-17
Do not create .build-ids in rpms for containers (paack)

mangelajo:
rpmbuild macros otherwise create a /usr/lib/.build-id reference
to every elf file inside the containers, and those clash with
other system rpms.


oglok (commented):
/LGTM

PR #689, opened 2022-05-17
WIP: Rebase script - update to 4.11.nightly 

sallyom:
@fzdarsky

Updated the rebase script so now one can 

```
./scripts/rebase.sh nightly
```
latest passing nightly image will be downloaded, then, so long as the current patches apply cleanly (which they did, here) , it's good to go.

EDIT: This isn't as straightforward, since first we have to bump from 4.10 to 4.11 - that requires a kube bump to 1.24


sallyom:
updating to 4.11 will require updating to kube 1.24.0, this will require more work

sallyom:
https://github.com/openshift/microshift/pull/729 will take care of the update to 4.11 nightly, closing this

PR #692, opened 2022-05-24
kuttl should be minimally used to ensure the cluster infra has deployed

copejon:
The kuttl tool should only be used to verify that MicroShift's cluster-infra has deployed and is running.  This PR removes additional tests which attempt to prove cluster stability before executing the e2e test suites.  Early on this was useful as a sanity check.  Now the app deployments used to validate the cluster are introducing bugs, caused in part by cluster dependencies that MicroShift does not provide by default.

Jira: [USHIFT-12](https://issues.redhat.com//browse/USHIFT-12)


mangelajo:
/lgtm
as long as we can get equivalent testing from standard openshift CI

mangelajo:
oh ok, so /lgtm here means the bot will merge it, @oglok @fzdarsky you may want to have an eye to this PR too

mangelajo:
can we configure for more than one lgtm/approver?

fzdarsky:
@copejon to test whether MicroShift is up and running, we could/should IMHO use `oc wait --for condition=Ready --all -A pods --timeout=180s` or similar.

If that test is the only remaining reason to use kuttl, then let's remove it.

PR #707, opened 2022-05-31
image-builder: add missing dependencies

dhellmann:
yum-utils is needed for reposync

selinux-policy-devel is needed for the RPM build


fzdarsky:
Ah, good find, thanks.

/lgtm

PR #711, opened 2022-06-03
Remove Flannel in favor of Bridge CNI plugin

oglok:
This PR disables Flannel in favor of the default CNI plugin for our container runtime of choice, CRI-O. The default CNI plugin for CRI-O is the Bridge CNI.

Closes: https://issues.redhat.com/browse/USHIFT-119


oglok:
https://issues.redhat.com/browse/USHIFT-119

packaging/crio.conf.d/microshift.conf
```diff
@@ -5,10 +5,6 @@ conmon_cgroup = "pod"
 cgroup_manager = "systemd"
 
 [crio.network]
-# cbr0 is the name configured by flannel in /etc/cni/net.d/ config file
-# by declaring this crio will wait until that network is configured.
-cni_default_network = "cbr0"
```
dhellmann, line 10: Do we need to pick a different default, or is that configured some other way?

oglok: We faced an issue with pods starting being attached to two different bridges (CRI-O bridge and Flannel bridge) due to a race condition. We decided to fix it by letting CRI-O wait for the Flannel bridge to be created and attach all pods there. Now that Flannel is gone, we can just rely on the default bridge from CRI-O. 

Short answer, I don't think we need to add any default there.

dhellmann: Thanks, that makes sense.

dhellmann (approved):

PR #715, opened 2022-06-08
Remove embedded CLI calls

dhellmann:
The previous conversion from using the cobra interface to run embedded components left some cruft behind in places where there is a public options type. This PR removes that unneeded code to make the logic easier to follow.


dhellmann:
See #717, #718, #719, #720

PR #717, opened 2022-06-09
USHIFT-92: kube-apiserver: remove dead configuration code

dhellmann:
kube-apiserver, since we can construct an options struct directly.


oglok:
I've tested the PR and it works fine. Thanks for doing a cleanup based on what we discussed. 

/lgtm

PR #718, opened 2022-06-09
USHIFT-93: kube-scheduler: remove dead configuration code

dhellmann:
kube-scheduler, since we can construct an options struct directly.


oglok:
It has been successfully tested in my local machine.

PR #719, opened 2022-06-09
USHIFT-94: openshift-apiserver: remove dead configuration code

dhellmann:
openshift-apiserver, since we can construct an options struct
directly.


oglok:
Successfully tested in my dev env.

/lgtm

PR #721, opened 2022-06-09
USHIFT-97: Remove oauth

dhellmann:
We no longer want to support the APIs provided by the oauth server.


ggiguash:
USHIFT-97

oglok:
Hey @dhellmann , 

The PR looks good. I have one question regarding unwanted APIs. For example, currently, we cannot NOT initialize the project.openshift.io api service, or the openshift-apiserver won't even start. That leaves us with the project api being exposed, but completely unusable because the oauth-apiserver will be gone. Does this make sense?

dhellmann:
> Hey @dhellmann ,
> 
> The PR looks good. I have one question regarding unwanted APIs. For example, currently, we cannot NOT initialize the project.openshift.io api service, or the openshift-apiserver won't even start. That leaves us with the project api being exposed, but completely unusable because the oauth-apiserver will be gone. Does this make sense?

It's not ideal, but I think it's OK in the short term. We need to work with the API server team to make it possible to not bring in the APIs we don't want to expose.

fzdarsky:
> [...] That leaves us with the project api being exposed, but completely unusable because the oauth-apiserver will be gone.

IIRC, `Project` resource still works, it's the `ProjectRequest` (for `oc new-project`) that doesn't, or?

oglok:
Ok, let's remove it for now until we figure out what to do with the exposed APIs. The PR works fine.

/lgtm

PR #725, opened 2022-06-10
docs: update description of support deployment topology

dhellmann:


docs/design.md
```diff
@@ -17,7 +17,7 @@ MicroShift aims at meeting all of the following design goals:
   <sup>1) when used in combination with an edge-optimized OS like RHEL 4 Edge or Fedora IoT</sup>
 
 * **Production-grade:**
-  * MicroShift supports deployments with 1 or 3 control plane and 0..N worker instances.
+  * MicroShift supports deployments with 1 node acting as control plane and worker. Orchestration of multi-node control plane configurations introduces unwanted complexity into deployment and upgrade processes. A single-node control plane becomes a single point of failure when worker nodes are attached, which eliminates the high availability benefits of running workloads on kubernetes.
```
derekwaynecarr, line 20: Maybe add the following:

For application availability, we recommend running two single node clusters that deploy a common application in active/active or active/passive mode and use existing tools to support failover between those states when either host is unable to provide availability.

PR #726, opened 2022-06-10
docs: update deployment statements

dhellmann:
We've decided to focus on the RPM deployment approach because we want
to ensure that we don't introduce security concerns by having
something running in a container have permissions needed to control
the container runtime.


derekwaynecarr:
There is no upstream commitment to support running the kubelet in a container.  The code embedded in the kubelet to support secrets, configmaps, host path volumes, and other storage volumes has in the past been subject to CVE vulnerabilities when run in a container.  As a result of the difficulty in person-years to resolve those issues, we do not recommend running kubelet in a container or container-like environment at this time in any production use case.  Any execution of kubelet in a container is a non-production system and is unsupported beyond developer hacking.

dhellmann:
Thanks, @derekwaynecarr , I stole those words and added them to the commit message.

PR #733, opened 2022-06-15
rebase 4.10.18

sallyom:

The 4.10 rebase was missing patches, and instead of vendoring each openshift-apiserver-branch, should instead vendor openshift/kubernetes as a whole and patch from there. 

Also, this PR brings in OCP 4.10.18


sallyom:
For this PR, I've vendored `openshift/etcd openshift-4.10` branch, it appears as though that is v 3.5.1? However, what is running in OCP 4.10.18 is 3.5.3. Not sure if this is an issue.

copejon (approved):

PR #735, opened 2022-06-17
Adds a licensecheck make target that runs lichen

mangelajo:
Some exceptions have been added to .lichen.yaml for dual-licensed
go modules, as well as for some OpenShift components which
didn't specify licenses anywhere in their repositories.

Related-Issue: https://issues.redhat.com/browse/USHIFT-111

Signed-off-by: Miguel Angel Ajo Pelayo <majopela@redhat.com>

### DIFF
--- a/.lichen.yaml
+++ b/.lichen.yaml
@@ -1,0 +1,36 @@
+---
+# Licenses other than Apache-2.0 are governed by
+# https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md#approved-licenses-for-allowlist
+# Note that Allowlist also requires that projects were created
+# on GitHub at least 12 months prior and have at least 10 stars
+# or 10 forks
+allow:
+  - "Apache-2.0"
+  - "BSD-2-Clause"
+  - "BSD-2-Clause-FreeBSD"
+  - "BSD-3-Clause"
+  - "MIT"
+  - "ISC"
+  - "Python-2.0"
+  - "PostgreSQL"
+  - "X11"
+  - "Zlib"
+
+exceptions:
+  # The following references belong to OpenShift, and lichen is unable to find any referece
+  # to the type of license, it comes from RedHat, so we assume it's safe to import.
+  unresolvableLicense:
+  - path: "github.com/openshift/openshift-apiserver"
+  - path: "github.com/openshift/openshift-controller-manager"
+  # The CNFC maintains a list of project license exceptions which are safe or Apache-2
+  # compliant, those exceptions are maintained here:
+  # https://github.com/cncf/foundation/tree/main/license-exceptions
+  licenseNotPermitted:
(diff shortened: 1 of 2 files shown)


.lichen.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - "BSD-2-Clause-FreeBSD"
+  - "BSD-3-Clause"
+  - "MIT"
+  - "ISC"
+  - "Python-2.0"
+  - "PostgreSQL"
+  - "X11"
+  - "Zlib"
+
+# all from https://github.com/cncf/foundation/tree/main/license-exceptions
```
ggiguash, line 19: Can we expand on what this comment means?

.lichen.yaml
```diff
@@ -0,0 +1,32 @@
...
+exceptions:
+  unresolvableLicense:
+  - path: "github.com/openshift/openshift-apiserver"
```
ggiguash, line 22: Can we add a comment why these are here?

Makefile
```diff
@@ -253,3 +253,11 @@ release-nightly:
 component-images:
 	cd packaging/images/components && ./build.sh
 .PHONY: component-images
+
+licensecheck: DEBUG=true
```
ggiguash, line 257: Ce we remove the DEBUG statement? Otherwise, it generates too much output.

Makefile
```diff
@@ -79,8 +79,7 @@ debug:
 GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo'
 
 # targets "all:" and "build:" defined in vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk
-microshift: build-containerized-cross-build-linux-amd64
-.PHONY: microshift
+microshift: build
```
mangelajo, line 82: microshift was a phony target (no binary output associated) but we really have a ./microshift output which is handled by `make`,  now make microshift can handle that too, and other actions can depend on that if the binary doesn't exist in ./

Makefile
mangelajo: That was really to force debug info on the microshift build, as lichen analyzes the binary and not the source code ( a bit weird if you ask me).  

I checked and it works without that DEBUG, so I will remove it for simplicity.

mangelajo:
Thank's for testing @ggiguash I applied your comments and fixed an issue with the Makefile dependencies to get ./microshift properly built on demand (so lichen can explore it)

.lichen.yaml
mangelajo: Done

ggiguash (approved):

mangelajo:
@ggiguash @fzdarsky @copejon  this needs another LGTM

dhellmann (approved):
/lgtm

oglok (approved):

PR #737, opened 2022-06-21
Update rebase script

fzdarsky:
Steps to improving the `rebase.sh` script:
- Change heuristic for rebasing go.mod to be more predictable (make explicit which dep comes from where instead of using "latest wins").
- Improve extracting of assets from release image.
- Copy in more assets.
- Remove unused code.

### DIFF
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -129,7 +129,7 @@ download_release() {
     mkdir -p release-manifests
     pushd release-manifests >/dev/null
     content=$(oc adm release info ${authentication} --contents "${release_image}")
-    echo "${content}" | awk '/^# 0000_[A-Za-z0-9._-]*.yaml/{filename = $2;}{if (filename != "") print >filename;}'
+    echo "${content}" | awk '{ if ($0 ~ /^# [A-Za-z0-9._-]+.yaml$/ || $0 ~ /^# image-references$/ || $0 ~ /^# release-metadata$/) filename = $2; else print >filename;}'
     popd >/dev/null
 
     title "# Cloning ${release_image} component repos"
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -323,6 +323,12 @@ update_manifests() {
     pushd "${STAGING_DIR}" >/dev/null
 
     title "Rebasing manifests"
+    for crd in ${REPOROOT}/assets/crd/*.yaml; do
+        cp "${STAGING_DIR}"/release-manifests/$(basename ${crd}) "${REPOROOT}"/assets/crd || true
+    done
+    rm -f "${REPOROOT}"/assets/scc/*.yaml
+    cp "${STAGING_DIR}"/release-manifests/0000_20_kube-apiserver-operator_00_scc-*.yaml "${REPOROOT}"/assets/scc || true
+
     rm -f "${REPOROOT}"/assets/components/openshift-dns/dns/*
     cp "${STAGING_DIR}"/cluster-dns-operator/assets/dns/* "${REPOROOT}"/assets/components/openshift-dns/dns 2>/dev/null || true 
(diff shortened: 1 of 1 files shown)


scripts/rebase.sh
```diff
@@ -155,45 +88,119 @@ download_release() {
...
+
+valid_component_or_exit() {
+    local component=$1
+    if [[ ! " etcd kubernetes openshift-apiserver openshift-controller-manager " =~ " ${component} " ]]; then
+        echo "error: release reference must be one of [etcd kubernetes openshift-apiserver openshift-controller-manager], have ${component}"
+        exit 1
+    fi
+}
+
+update_go_mod() {
```
sallyom, line 172: When I run `./scripts/rebase.sh go.mod` I get this all the way down: 
```
$ ./scripts/rebase.sh go.mod
# Updating go.mod
skipping modulepath bazil.org/fuse: no or unknown command []
skipping modulepath bitbucket.org/bertimus9/systemstat: no or unknown command []
skipping modulepath cloud.google.com/go: no or unknown command []
skipping modulepath cloud.google.com/go/bigquery: no or unknown command []
---
```
and this returns empty from command line, also:
`replaced_modulepaths=$(go mod edit -json | jq -r '.Replace // []' | jq -r '.[].Old.Path' | xargs)` 

The main branch rebase.sh works as expected.

scripts/rebase.sh
```diff
@@ -38,73 +38,6 @@ title() {
...
-            version=$(printf '%s\n%s\n' "${base_version}" "${update_version}" | sort --version-sort | tail -n 1)
-        fi
-
-        echo "${mod} ${version}"
-    done < "${base_file}"
-}
-
-# Returns the list of release image names from a release_${arch}.go file
-get_release_images() {
-    file=$1
```
sallyom, line 103: `get_release_images` is still called on [#L 230](https://github.com/openshift/microshift/pull/737/files#diff-f5410281afaafce8468ac3b84a6bb200a4e1e334f659a4422c07066780a1d0aaR230)

sallyom: the fn works as/is, only need to add it back

sallyom (commented):
there are 2 issues
* `rebase.sh go.mod` needs a tweak
* `get_release_images` fn needs to be added back
other than that this lgtm!

Something to note:
I rebased against 4.10.20 and the _only_ git diff from rebasing was the 2 lines in release/release.go changing 4.10.18 to 4.10.20 - this bodes well for automating nightly updates via CI, and when that job fails is when we'll know patches, kube versions, or something manual needs updating. I believe that will not happen all that often/regularly.

scripts/rebase.sh
fzdarsky: Ooops, was a bit over-zealous. Thank you for catching this!

fzdarsky: Fixed.

scripts/rebase.sh
fzdarsky: > When I run `./scripts/rebase.sh go.mod` I get this all the way down:
This is expected as you're likely using a go.mod file that isn't annotated yet (what PR would address).
Basically, if we want more predictability when resolving conflicts between replace directives of different embedded components, we need a way to tell the rebase script how to. The annotations do that explicitly and that hopefully is more stable than "choose latest", which may change during a release.

> and this returns empty from command line, also: `replaced_modulepaths=$(go mod edit -json | jq -r '.Replace // []' | jq -r '.[].Old.Path' | xargs)`
That's weird. I'm not seeing that on the command line. And in the script at least it should work for you, too, otherwise you'd not be seeing all these "skipping..." messages.

sallyom: and, after a laptop restart all is well.

fzdarsky:
USHIFT-176

PR #738, opened 2022-06-22
Document Load Balancer for TCP Traffic using MetalLB

ggiguash:
Document Load Balancer for TCP Traffic using MetalLB
Inspired by https://microshift.io/docs/user-documentation/how-tos/example-usage/
[USHIFT-34](https://issues.redhat.com//browse/USHIFT-34)


mangelajo (approved):
/lgtm

dhellmann (approved):
This looks good. In the future if we document how to pre-load images, we could update this page or provide a separate version of the instructions for that approach.

PR #739, opened 2022-06-22
Remove Security API service and use local server for SCC objects

oglok:
Closes [USHIFT-98
](https://issues.redhat.com/browse/USHIFT-98)

### DIFF
--- a/assets/crd/0000_03_security-openshift_01_scc.crd.yaml
+++ b/assets/crd/0000_03_security-openshift_01_scc.crd.yaml
@@ -14,6 +14,8 @@ spec:
     listKind: SecurityContextConstraintsList
     plural: securitycontextconstraints
     singular: securitycontextconstraints
+    shortNames:
+      - scc
   scope: Cluster
   versions:
     - additionalPrinterColumns:
--- a/pkg/assets/bindata.go
+++ b/pkg/assets/bindata.go
@@ -142,7 +142,7 @@ func assetsComponentsFlannelClusterroleYaml() (*asset, error) {
 		return nil, err
 	}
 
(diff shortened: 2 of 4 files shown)


oglok:
Looking at OCP, there are two existing API services for SecurityContextConstraints:

```
oc get apiservice | grep security
v1.security.internal.openshift.io             Local                                                        True        7m58s
v1.security.openshift.io                      openshift-apiserver/api                                      True        22m
```

Per my conversation with the authteam, I got the following response about this:

_This was done because of the bootstrap process.
SCCs used to be served by the openshift-apiserver, which also contains all kinds of validations for this type.
However, openshift-apiserver is not present during cluster startup at install time but the SCC admission that needs SCCs is already running. And so, we had to turn SCCs into CRDs so that we can install them early in the cluster so that kube-apiserver can admit other pods (like openshift-apiserver, for example) to run.
But! There would be issue with the API discovery. And from here it's a little bit of black magic, but eventually you end up with two API groups because of the sins from the past._

In MicroShift, we do not have bootstrap phase as in OCP, so we don't need to have two different API services.

dhellmann (approved):

oglok:
In the 4.8 MicroShift release, we had only the local apiservice:

```
[vagrant@fedora ~]$ oc get apiservice | grep securi
v1.security.openshift.io                 Local                               True        14m
[vagrant@fedora ~]$ microshift version
MicroShift Version: 4.8.0-0.microshift-2022-04-20-141053
Base OKD Version: 4.8.0-0.okd-2021-10-10-030117
```

sallyom (commented):
@oglok I noticed this early on in ushift days, when I was looking at adding ClusterPolicyController addition & things were moving very fast - good find! 
lgtm but will let @fzdarsky tag it

oglok:
This will bring us to the 4.8 state where SCCs requests were served by the local kube apiserver, and not OpenShift apiserver.

copejon (approved):

PR #741, opened 2022-06-22
Document supported OpenShift APIs

copejon:

[USHIFT-47](https://issues.redhat.com//browse/USHIFT-47)


sallyom (approved):
lgtm

oglok:
```
podsecuritypolicyreviews                           security.openshift.io/v1               true         PodSecurityPolicyReview
podsecuritypolicyselfsubjectreviews                security.openshift.io/v1               true         PodSecurityPolicySelfSubjectReview
podsecuritypolicysubjectreviews                    security.openshift.io/v1               true         PodSecurityPolicySubjectReview
rangeallocations                                   security.openshift.io/v1               false        RangeAllocation

```

These APIs should not be exposed in MicroShift, and indeed, they were not in 4.8. It's all related to the inclusion of an API service for the security API group, that also causes the SCCs to fail. Related to this PR: #739

dhellmann (approved):

PR #742, opened 2022-06-23
Document rebase script

fzdarsky:
This is the documentation of the rebase process at the state after #737 merges.

Closes [USHIFT-65](https://issues.redhat.com//browse/USHIFT-65)

### DIFF
--- a/docs/rebase.md
+++ b/docs/rebase.md
@@ -1,0 +1,125 @@
+# Rebasing MicroShift
+## Overview
+
+MicroShift repackages a minimal core of OpenShift: A given MicroShift release is built from the same content as the corresponding OpenShift release, adding a small amount of "glue logic".
+
+Therefore, on a high level a rebase of MicroShift onto a newer OpenShift release starts from an given OpenShift release image and involves the following steps:
+
+* vendoring the source code of the embedded OpenShift components at the same commit as used in OpenShift,
+* embedding (as bindata) the resource manifests of the hosted OpenShift components, and 
+* updating the image references (digests) of the hosted OpenShift components.
+
+This process is supported by the `rebase.sh` script, which generates the necessary changes, but currently still requires human review and post-processing. The goal is to eventually automate generation, testing, and merging PRs at least for nightlies and z-stream updates.
+
+The following describes the current rebase process in more detail.
+
+# Process
+## Prerequisites
+
+On the machine used for the rebase,
+
+* install `git`, `golang` (>= 1.17), `oc`, and `jq`,
+* add a pull secret into `~/.docker/config.json`, and
+* git clone your personal fork of microshift and `cd` into it.
+
+## Downloading the target OpenShift release
+
+Run the following to download the OpenShift release to rebase to, specifying the target release image, e.g.:
+
+```
+./scripts/rebase.sh download registry.ci.openshift.org/ocp/release:4.11.0-0.nightly-2022-06-06-201913
+```
+
+This will create a directory `_output/staging`, download the specified release image's metadata (`release.txt`) and manifests, git-clone (only) the repos of the embedded components and the operators of the loaded components, and check out the commit used by that OpenShift release.
+
+## Rebasing the go.mod file and vendoring
+
+In MicroShift's `go.mod` file, we only explicitly add the `require` directives needed by MicroShift itself (marked with the comment `// microshift`) whereas we let `go mod tidy` figure out the direct and indirect requirements for the embedded components. For the rebase, the focus is therefore on the `replace` directives.
+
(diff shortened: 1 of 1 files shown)


docs/rebase.md
```diff
@@ -0,0 +1,125 @@
...
+git apply scripts/rebase_patches/*
+script/rebase.sh generated-apis
+git add vendor
+git commit -m "update vendoring"
+```
+
+These patches are produced by identifying the missing patches [TODO: describe process] and transforming those patches to apply against `/vendor` instead of o/k's staging dir [TODO: describe process]. This process requires some manual work but should be necessary rarely (likely after OpenShift rebases onto a upstream version).
+
+## Rebasing `Makefile`, `Dockerfiles`, and `.spec` file
+When updating to a new minor version of OpenShift, update the `Makefile` [here](https://github.com/openshift/microshift/blob/main/Makefile#L16) and [here](https://github.com/openshift/microshift/blob/main/Makefile#L55-L66) and also check whether the Dockerfiles need a new base image (e.g. [this](https://github.com/openshift/microshift/blob/main/packaging/images/openshift-ci/Dockerfile)) and the RPM `.spec` file needs updates (e.g. [this](https://github.com/openshift/microshift/blob/main/packaging/rpm/microshift.spec))
```
ggiguash, line 73: The lines in Makefile are likely to go out of sync quickly. Let's use a description of those places?

docs/rebase.md
```diff
@@ -0,0 +1,125 @@
...
+### Component images
+To update the component image references of the MicroShift release, run:
+
+```
+./scripts/rebase.sh images
+git add pkg/release
+git commit -m "update component images"
+```
+
+Note that as long as we don't use 100% OpenShift-built images, you may need to bump the `pause` and network and storage provider images manually.
```
ggiguash, line 92: It's not clear (to me at least) what such a "bump" would require

docs/rebase.md
```diff
@@ -0,0 +1,125 @@
...
+Embed the assets into the binary and commit the result:
+
+```
+./scripts/bindata.sh
+git add assets pkg/assets
+git commit -m "update manifests"
+```
+
+### Component configs
+The last step isn't automated at all yet, which is to compare whether the config parameters of embedded component changed.
```
ggiguash, line 125: Can we add at least an example of what needs to be compared to what?

ggiguash (changes requested):
Overall, looking great! Some minor comments, mostly for adding a bit more information.

docs/rebase.md
fzdarsky: Good point. Fixed this.

docs/rebase.md
fzdarsky: Removed that note as our change to OpenShift images will make this irrelevant anyway.

docs/rebase.md
fzdarsky: Added this.

ggiguash (approved):

sallyom (commented):
This has all the necessary details and will make rebases a much kinder & simpler ordeal. I'm going through the steps now to ensure everything is working as intended.

sallyom:
Are we oppposed to adding a section (or separate .md file) for building a community version, something like:


# Build MicroShift From a Community OKD Release

The rebase script can also be used to build a community version of MicroShift that does not require access to internal image registries. To build the community version, rebase against an [OKD image found here](https://origin-release.ci.openshift.org/#4-stable) by following the same rebase script, and downloading the OKD release image [like so](#downloading-the-target-openshift-release) instead of an OpenShift release. This build is not supported or maintained and there is no guarantee that OKD releases will continue to mirror OpenShift releases.

fzdarsky:
> Are we oppposed to adding a section (or separate .md file) for building a community version, something like:

That's a good idea, let me include that.

docs/rebase.md
```diff
@@ -0,0 +1,123 @@
+# Rebasing MicroShift
+## Overview
+
+MicroShift repackages a minimal core of OpenShift: A given MicroShift release is built from the same content as the corresponding OpenShift release, adding a small amount of "glue logic".
+
+Therefore, on a high level a rebase of MicroShift onto a newer OpenShift release starts from an given OpenShift release image and involves the following steps:
```
dhellmann, line 6:
```suggestion
Therefore, on a high level a rebase of MicroShift onto a newer OpenShift release starts from a given OpenShift release image and involves the following steps:
```

dhellmann (approved):
Very clear instructions. I found one typo, but that can be fixed in another PR if there's no other reason to update this one.

docs/rebase.md
fzdarsky: Thank you, included that in the update.

sallyom (approved):
https://github.com/openshift/microshift/pull/737 is approved & tagged, tagging this as well! 

/lgtm

PR #744, opened 2022-06-24
Implement a procedure for embedding container images inside edge ISO (Canceled) 

ggiguash:
Implement a procedure for embedding container images inside edge ISO
Update container RPM creation procedure to support pull secrets and add command line to make-microshift-images-rpm.sh script
[USHIFT-112](https://issues.redhat.com//browse/USHIFT-112)


ggiguash:
I will split it and open 2 new PRs.

PR #745, opened 2022-06-24
Fix paack issue with podman4 where overlay2 storage is used

mangelajo:
overlay2 format includes a block device inside the overlay directory
but tar can't unpack that during the install phase.

We exclude that file and move it to %post

Closes [USHIFT-122](https://issues.redhat.com/browse/USHIFT-122)

### DIFF
--- a/packaging/rpm/paack.py
+++ b/packaging/rpm/paack.py
@@ -89,6 +89,9 @@ find ./ -perm 000 -exec chmod 400 {} +
 
 %%_POST_%%
 
+# rpm-ostree seems to be unhappy about this, but crio storage (read-only) works without it
+# mknod -m 600 "%{imageStore}/overlay/backingFsBlockDev" b 253 0
+
 %postun
 
 %%_POSTUN_%%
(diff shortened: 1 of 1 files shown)


ggiguash (approved):

packaging/rpm/paack.py
```diff
@@ -89,6 +89,9 @@
 
 %%_POST_%%
 
+# rpm-ostree seems to be unhappy about this, but crio storage (read-only) works without it
+# mknod -m 600 "%{imageStore}/overlay/backingFsBlockDev" b 253 0
```
copejon, line 93: should this be commented out?

copejon (approved):

packaging/rpm/paack.py
copejon: From side-bar with @ggiguash 
>Yes, that's the error which [@majopela](https://coreos.slack.com/team/UEFFLKJAJ) tried to fix.

sallyom:
not super-familiar with rpm building specifics, but I see rpm build & install CI passed, I don't see any issue with it.

/lgtm

sallyom (approved):

ggiguash:
USHIFT-122

packaging/rpm/paack.py
mangelajo: Yes, it's fine, rpm install didn't work with rpm-ostree if I added it in post.

mangelajo: Thanks for reviewing and checking around.

PR #746, opened 2022-06-24
Hack scripts to help us debug scc

mangelajo:
Creates an scc-test namespace, and a service account with
privileges on such namespace, but no admin privileges.

Then it creates an sa.kubeconfig file which can be used
to install a privileged nginx deployment which should
be forbidden.

Closes #<Issue Number>


dhellmann (approved):

mangelajo:
@oglok can we give a lgtm to this one so we keep it for later ? :)

PR #747, opened 2022-06-24
Refactor go.mod

fzdarsky:
Refactors go.mod to allow better automation and more predictability of the rebase script.

The changes do not result in changes to vendored packages, except for updating the versions of dns and go-homedir, which had inconsistent require and replace directives.


fzdarsky:
> It looks like the main change is separating the direct and indirect dependencies into 2 `requires` blocks, is that right?

That part is due to the switch to go1.17, which formats differently. The main part is the annotations of replace directives that the rebase.sh would then use to determine how to resolve conflicting replace directives between components.

oglok (approved):
I ran `go mod tidy` and `go mod vendor`, reapplied the current patches from `scripts/rebase_patches` and tested it. It works fine and the go.mod is more clear now.

/lgtm

PR #751, opened 2022-06-28
USHIFT-123: use release creation time as bindata timestamp

dhellmann:
Update the rebase script to extract the creation date of the release
to use as the timestamp for the assets compiled into bindata.

Update bindata.sh to use the timestamp, if it is present.

Set the bindata_timestamp.txt content to the value for the current
release, 4.10.18.

Update bindata.go with the correct timestamp.

Closes USHIFT-123


oglok (commented):
I'm wondering if extracting this timestamp only makes sense in the rebase script or possibly as a `Makefile` target to execute it everytime we build?
For example, the SCC CRD does not contain the `shortName: scc` field, and we patch it manually. That would require a different timestamp from the original pulled from the release.txt.

oglok:
> > I'm wondering if extracting this timestamp only makes sense in the rebase script or possibly as a `Makefile` target to execute it everytime we build?
> > For example, the SCC CRD does not contain the `shortName: scc` field, and we patch it manually. That would require a different timestamp from the original pulled from the release.txt.
> 
> The idea is to have a static value that updates when the majority of the content changes so that we can have a CI job verify that the bindata always matches the assets (USHIFT-124). Creating a new timestamp every time we run make would not let us do that, right?

That's correct. My question is, how do we update the `bindata_timestamp.txt` content If I patch an asset with some specific change for MicroShift?

dhellmann:
> > > I'm wondering if extracting this timestamp only makes sense in the rebase script or possibly as a `Makefile` target to execute it everytime we build?
> > > For example, the SCC CRD does not contain the `shortName: scc` field, and we patch it manually. That would require a different timestamp from the original pulled from the release.txt.
> > 
> > 
> > The idea is to have a static value that updates when the majority of the content changes so that we can have a CI job verify that the bindata always matches the assets (USHIFT-124). Creating a new timestamp every time we run make would not let us do that, right?
> 
> That's correct. My question is, how do we update the `bindata_timestamp.txt` content If I patch an asset with some specific change for MicroShift?

Do we need to do that?

Now that I've seen the work @sallyom did in #752, I think I prefer that approach better than this one. I suggest that we close this PR and approve @sallyom's.

sallyom (approved):
I prefer the simpler https://github.com/openshift/microshift/pull/752 to offload bindata verify, but `build-machinery-go` adds `--nometadata` while metadata is desired here.

/lgtm
/approve

dhellmann:
This also closes USHIFT-124

PR #752, opened 2022-06-29
add bindata targets from openshift/build-machinery-go

sallyom:
@dhellmann @fzdarsky 

This adds the[ openshift/build-machinery-go bindata targets](https://github.com/openshift/build-machinery-go/blob/master/make/targets/openshift/bindata.mk) 

It adds the required targets to the Makefile, for example see [openshift/cluster-authentication-operator](https://github.com/openshift/cluster-authentication-operator/blob/master/Makefile#L31-#L32)

With this, `verify-bindata` will now run with `make verify` and will fail if bindata hasn't been updated.

The scripts/bindata.sh can be removed - we can now rely on build-machinery-go

### DIFF
--- a/pkg/assets/bindata.go
+++ b/pkg/assets/bindata.go
@@ -131,7 +131,7 @@ func assetsComponentsHostpathProvisionerClusterroleYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "assets/components/hostpath-provisioner/clusterrole.yaml", size: 609, mode: os.FileMode(436), modTime: time.Unix(1655736360, 0)}
+	info := bindataFileInfo{name: "assets/components/hostpath-provisioner/clusterrole.yaml", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
 	a := &asset{bytes: bytes, info: info}
 	return a, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/assets/bindata.go
```diff
@@ -131,7 +131,7 @@ func assetsComponentsHostpathProvisionerClusterroleYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "assets/components/hostpath-provisioner/clusterrole.yaml", size: 609, mode: os.FileMode(436), modTime: time.Unix(1655736360, 0)}
+	info := bindataFileInfo{name: "assets/components/hostpath-provisioner/clusterrole.yaml", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
```
sallyom, line 134: looking at openshift repos, this is consistent, but I'm not sure why the build-machinery-go bindata.mk results in this: 
* https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator/assets/bindata.go#L196 
* https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/v311_00_assets/bindata.go#L126

fzdarsky: I had removed the -nometadata back then to conserve file permissions when writing bindata assets to disk.

The use case was to eventually replace the highly redundant and complex "manifest applier" code for the loaded components with a simple aeevice to render component manifests as kustomization to disk, where the currently existing kustomizer service would find and automatically apply it. You can still see that code in the in the `component-kustomizer-poc` branch.

dhellmann: @fzdarsky, what's the use case for customizing the embedded manifests?

sallyom:
> We discussed this in the team meeting today, and decided that we do want to keep the metadata in order to eventually be able to implement the feature Frank describes in his comment which would let us remove some of the custom code we have for loading the manifests today. We agreed to go ahead with #751 for now, and revisit the source of the timestamp in the future if we decide the release timestamp is not the best value to use.

sgtm! /closing

PR #755, opened 2022-07-01
More accurate pseudoversions

fzdarsky:
Uses a combination of `go mod edit` and `go mod tidy` to generate pseudo versions instead of "manually" constructing them. This handles all format types correctly according to spec and is more robust, but unfortunately also quite a bit slower.

### DIFF
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -39,14 +39,6 @@ title() {
 }
 
 
-# Returns the list of release image names from a release_${arch}.go file
-get_release_images() {
-    file=$1
-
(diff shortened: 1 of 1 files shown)


scripts/rebase.sh
```diff
@@ -241,6 +229,13 @@ regenerate_openapi() {
 }
 
 
+# Returns the list of release image names from a release_${arch}.go file
```
dhellmann, line 232: What do you think about converting the image lists to YAML (or some other text format) that we put in the assets directory? Then this script could use yq, for example.

dhellmann (commented):
/lgtm

PR #757, opened 2022-07-01
Add an iso target to the Makefile.

jogeo:
This PR adds make targets, which call the existing image-builder scripts. Adding the 'iso' target provides a convenient single step, for example when building the install ISO for use in a test environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,13 @@ IMAGE_REPO :=quay.io/microshift/microshift
 IMAGE_REPO_AIO :=quay.io/microshift/microshift-aio
 OUTPUT_DIR :=_output
 RPM_BUILD_DIR :=packaging/rpm/_rpmbuild
+ISO_DIR :=scripts/image-builder/_builds
 CROSS_BUILD_BINDIR :=$(OUTPUT_DIR)/bin
 FROM_SOURCE :=false
 CTR_CMD :=$(or $(shell which podman 2>/dev/null), $(shell which docker 2>/dev/null))
 ARCH :=$(shell uname -m |sed -e "s/x86_64/amd64/" |sed -e "s/aarch64/arm64/")
 IPTABLES :=nft
+PULLSECRET :=~/pullsecret
 
 # restrict included verify-* targets to only process project files
 GO_PACKAGES=$(go list ./cmd/... ./pkg/...)
@@ -156,6 +158,17 @@ srpm:
 	RELEASE_PRE=${RELEASE_PRE} ./packaging/rpm/make-rpm.sh local
 .PHONY: srpm
 
+image-build-configure:
+	./scripts/image-builder/configure.sh
+.PHONY: image-build-configure
+
+image-build-iso: rpm 
+	./scripts/image-builder/build.sh -pull_secret_file $(PULLSECRET)
+.PHONY: image-build-iso
+
+iso: image-build-configure image-build-iso
+.PHONY: iso
+
 ###############################
 # containerized build targets #
 ###############################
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -156,6 +158,17 @@ srpm:
 	RELEASE_PRE=${RELEASE_PRE} ./packaging/rpm/make-rpm.sh local
 .PHONY: srpm
 
+image-build-configure:
+	./scripts/image-builder/configure.sh
+.PHONY: image-build-configure
+
+image-build-iso: rpm
```
dhellmann, line 165: I didn't choose the same filename for $(PULLSECRET) as the value on line 40, and the build didn't fail when that file was not present. We should make build.sh report that error.

dhellmann:
/approve

It would be good to have this linked to a Jira ticket. Mentioning the ticket ID in a comment should be enough to do that.

/test e2e-openshift-conformance-sig-storage

Makefile
jogeo: Thanks @dhellmann, we collided on adding a check for the pull secret file. I'll back out my change.

jogeo:
The iso make target helps support https://issues.redhat.com/browse/USHIFT-104

dhellmann:
/test e2e-rpm-install

I can't make sense of the error in the failed job. It looks like something else is simultaneously trying to update the RPM database.

Makefile
```diff
@@ -31,11 +31,13 @@ IMAGE_REPO :=quay.io/microshift/microshift
 IMAGE_REPO_AIO :=quay.io/microshift/microshift-aio
 OUTPUT_DIR :=_output
 RPM_BUILD_DIR :=packaging/rpm/_rpmbuild
+ISO_DIR :=scripts/image-builder/_builds
 CROSS_BUILD_BINDIR :=$(OUTPUT_DIR)/bin
 FROM_SOURCE :=false
 CTR_CMD :=$(or $(shell which podman 2>/dev/null), $(shell which docker 2>/dev/null))
 ARCH :=$(shell uname -m |sed -e "s/x86_64/amd64/" |sed -e "s/aarch64/arm64/")
 IPTABLES :=nft
+PULLSECRET :=~/pullsecret
```
fzdarsky, line 40: We're using pull secrets (or better: registry credentials) in other places, too, e.g. the rebase script.
Would it make sense to align on a common location for these, e.g. `$HOME/.docker/config.json` that `oc adm`, podman, etc. default to?

dhellmann: Good point.

ggiguash, line 40: I think we should be consistent with the documentation where pull secret files are placed in ~/pull-secret.txt

See [Install MicroShift on RHEL for Edge](https://github.com/openshift/microshift/blob/main/docs/rhel4edge_iso.md#prerequisites)

ggiguash: I commented below, asking to align the location with the one documented in [Install MicroShift on RHEL for Edge](https://github.com/openshift/microshift/blob/main/docs/rhel4edge_iso.md#prerequisites).

I think we should change the docs if it does elsewhere. I used ~/pull-secret.txt location - keeping the file name downloaded from console.redhat.com

ggiguash (changes requested):

Makefile
ggiguash: After an offline discussion with @fzdarsky, we recommend to use `$HOME/.pull-secret.json` location for the secret files. This is not specific to docker/container and also a hidden file.

@jogeo, can you also make the changes in the docs?
```
$ grep -r pull-secret.txt docs
docs/rhel4edge_iso.md:Download the OpenShift pull secret from the https://console.redhat.com/openshift/downloads#tool-pull-secret page and save it into the `~microshift/pull-secret.txt` file. 
docs/rhel4edge_iso.md:./scripts/image-builder/build.sh -pull_secret_file ~/pull-secret.txt
docs/rhel4edge_iso.md:./packaging/rpm/make-microshift-images-rpm.sh rpm ~/pull-secret.txt x86_64:amd64 rhel-8-x86_64
docs/rhel4edge_iso.md:./scripts/image-builder/build.sh -pull_secret_file ~/pull-secret.txt -custom_rpms ~/microshift/packaging/rpm/paack-result/microshift-containers-4.10.18-1.x86_64.rpm
```

ggiguash (approved):

PR #758, opened 2022-07-05
Implementation of h/w clock change watcher

ggiguash:
Used the existing ipwatch controller to add the h/w clock change monitor. 
The controller was renamed to SysConfigWatchController. In the future, we may add more checks to the same controller.

[USHIFT-33](https://issues.redhat.com//browse/USHIFT-33) & [USHIFT-116](https://issues.redhat.com//browse/USHIFT-116)


pkg/sysconfwatch/sysconfwatch.go
```diff
@@ -0,0 +1,100 @@
...
+
+const sysConfigCheckInterval = time.Second * 5
+
+type SysConfWatchController struct {
+	NodeIP  string
+	timerFd int
+}
+
+func NewSysConfWatchController(cfg *config.MicroshiftConfig) *SysConfWatchController {
+	// Create a realtime clock timer with asynchronous read support
```
ggiguash, line 39: New code to initialize the realtime clock timer

pkg/sysconfwatch/sysconfwatch.go
```diff
@@ -0,0 +1,100 @@
...
+		case <-ticker.C:
+			// Check the IP change
+			currentIP, _ := util.GetHostIP()
+			if c.NodeIP != currentIP {
+				klog.Warningf("IP address has changed from %q to %q, restarting MicroShift", c.NodeIP, currentIP)
+				os.Exit(0)
+				return nil
+			}
+
+			// Check the clock change
```
ggiguash, line 86: New code to check the timer status

pkg/sysconfwatch/sysconfwatch.go
```diff
@@ -0,0 +1,100 @@
...
+			// Initiate an asynchronous read operation on the timer object
+			// When the clock is reset, the read operation returns with the ECANCELED error code
+			_, err := unix.Read(c.timerFd, buf)
+			if err == unix.ECANCELED {
+				klog.Warningf("realtime clock change detected, restarting MicroShift")
```
mangelajo, line 91: I believe we should validate the clock change to see how much it was.

For example, an ntp clock sync would trigger this, but the change would be in the range of <1s, which may not trigger a restart of MicroShift. (Otherwise we get a restart every time ntp sync happens).

I suggest we:
1)  store the clock time at start, 
2) If the event happens: calculate the diff, if <"1-5s"  we log, and ignore, also store the new time for next check
3) If the diff from step 2 is >1-5s then we log, and trigger the restart.

Does it sound reasonable @ggiguash ?

mangelajo (changes requested):
see comments about NTP

Bit looks good, and I like the re-use into sysconfwatch.go

pkg/sysconfwatch/sysconfwatch.go
ggiguash: Yes, we need to avoid that situation when NTP fires. The only thing to check is whether OCP modules may "break" on smaller 1-5s time update.

dhellmann:
I thought we only needed to ensure the clock was synced on startup. Let's talk about it before taking this change.

pkg/sysconfwatch/sysconfwatch.go
ggiguash: @mangelajo, I looked into reliable calculation of the time difference between two samples. While in principle your suggestion works, it does not seem to be fail-proof. Because we are comparing 2 time samples, we need to guarantee that the clock update happened after the first sample was taken. I do not think there is a way to achieve this reliably. 

Working sequence
1. Take time sample 1
2. Sleep 5 secs
<Time is adjusted>
3. Time change event received: Take time sample 2
4. Check the time diff and decide if to exit
5. Next iteration

Problematic sequence is when the time adjustment occurs on step 5 for example. Then, the sample 1 contains the adjusted time, event is received, but sample 2 will give a valid range of time difference.

ggiguash: @mangelajo, I uploaded an implementation of time drift calculation, taking into account changes between system and monotonic clock offsets. The current allowed time drift is set to +/- 10 seconds.
If you approve of this code, resolve the conversation and I will update the docs.

mangelajo (changes requested):

pkg/sysconfwatch/sysconfwatch.go
```diff
@@ -0,0 +1,132 @@
...
+	// Take a snapshot of the system and monototic clock difference as a base reference
+	var smtDiffBase = sysMonTimeDiff()
+	for {
+		select {
+		case <-ticker.C:
+			// Check the IP change
+			currentIP, _ := util.GetHostIP()
+			if c.NodeIP != currentIP {
+				klog.Warningf("IP address has changed from %q to %q, restarting MicroShift", c.NodeIP, currentIP)
+				os.Exit(0)
```
dhellmann, line 101: Does this trigger a clean shutdown, or do we need to do something to shutdown the subcomponents in a particular order? (If we need that, we can do it in a separate PR.)

go.mod
```diff
@@ -35,6 +35,8 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require golang.org/x/sys v0.0.0-20210831042530-f4d43177bf5e
```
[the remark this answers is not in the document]
ggiguash: "go mod tidy" put it separately. I moved it manually to the appropriate block of direct dependencies.

pkg/sysconfwatch/sysconfwatch.go
```diff
@@ -0,0 +1,132 @@
...
+	// System-wide clock that measures real (i.e. wall-clock) time
+	// This clock is affected by discontinuous jumps in the system time (e.g. if the system administrator manually changes the clock)
+	// and by the incremental adjustments performed by adjtime and NTP
+	unix.ClockGettime(unix.CLOCK_REALTIME, &stm)
+	// Clock that cannot be set and represents monotonic time since some unspecified starting point
+	// It provides access to a raw hardware-based time that is not subject to NTP adjustments
+	// or the incremental adjustments performed by adjtime
+	unix.ClockGettime(unix.CLOCK_MONOTONIC_RAW, &mtm)
+
+	return stm.Sec - mtm.Sec
```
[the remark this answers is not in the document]
ggiguash: The difference between monotonic (continuously increasing for the process lifetime) and wall (susceptible to adjustments) should remain the same as long as wall time does not change. 

What you are suggesting will also work, but it is a bit more complicated check. Are you concerned that we are not comparing "apples-to-apples"? I can change it if you think it's a better way to code this.

pkg/sysconfwatch/sysconfwatch.go
ggiguash: This code is a carry over from a previously approved PR. I do not know if this is the right way. I put it in our meeting agenda to discuss.

pkg/sysconfwatch/sysconfwatch.go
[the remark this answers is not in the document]
dhellmann: Yes, I'm worried that the 2 clocks are different enough in nature that relying on them to consistently have the same value would lead to a lot of false-positives and restarts. Watching for a difference in the way they each measure their own notion of elapsed time shouldn't have that effect.

mangelajo (commented):
/lgtm

PR #759, opened 2022-07-05
USHIFT-133: Improve image build script error handling

dhellmann:
Report an error when the pull secret file specified on the command line does not exist or is empty.

Also adds general error message support to the usage function in the script so that it is clear what problem triggered a failure.


mangelajo (commented):
/lgtm
nice improvement

PR #765, opened 2022-07-12
ovn-kubernetes as default CNI plugin

zshi-redhat:


hack/cleanup.sh
```diff
@@ -32,11 +32,15 @@ sudo bash -c '
     echo "Killing conmon, pause processes"
     pkill -9 conmon
     pkill -9 pause
-
-
```
mangelajo, line 36: may be we should start using the crio wipe here?

zshi-redhat: fixed.

assets/components/ovn/configmap.yaml
```diff
@@ -0,0 +1,36 @@
...
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: ovnkube-config
+  namespace: openshift-ovn-kubernetes
+data:
+  ovnkube.conf:   |-
+    [default]
+    mtu="1400"
+    cluster-subnets="10.128.0.0/14/23"
```
dhellmann, line 12: Should this come from the microshift config file?

assets/components/ovn/configmap.yaml
```diff
@@ -0,0 +1,36 @@
...
+    encap-port="6081"
+    enable-lflow-cache=false
+    lflow-cache-limit-kb=870
+
+    [kubernetes]
+    service-cidrs="172.30.0.0/16"
```
dhellmann, line 18: Should this come from the microshift config file?

assets/components/ovn/configmap.yaml
```diff
@@ -0,0 +1,36 @@
...
+    ovn-config-namespace="openshift-ovn-kubernetes"
+    kubeconfig="/var/lib/microshift/resources/kubeadmin/kubeconfig"
```
dhellmann, line 20: This path is definitely set internally in microshift based on an option that sets the parent directory, so we should have the value substituted before this config is written out.

assets/components/ovn/master/daemonset.yaml
```diff
@@ -0,0 +1,571 @@
...
+      # volumes in all containers:
+      # (container) -> (host)
+      # /etc/openvswitch -> /var/lib/ovn/etc - ovsdb data
+      # /var/lib/openvswitch -> /var/lib/ovn/data - ovsdb pki state
+      # /run/openvswitch -> tmpfs - sockets
+      # /env -> configmap env-overrides - debug overrides
+      containers:
+      # ovn-northd: convert network objects in nbdb to flows in sbdb
+      - name: northd
+        image: "quay.io/zshi/ovn-daemonset:microshift"
```
dhellmann, line 50: What's the plan for updating the image references to point to images from the OCP release payload?

pkg/components/networking.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+	if err := assets.ApplyDaemonSets(ds, renderReleaseImage, nil, kubeconfigPath); err != nil {
+		klog.Warningf("Failed to apply daemonSet %v %v", ds, err)
+		return err
+	}
+	return nil
+}
+
+func startFlannel(kubeconfigPath string) error {
```
dhellmann, line 74: Do we need this function?

assets/components/ovn/master/daemonset.yaml
zshi-redhat: The plan is to use what's available in latest OCP release artifacts (oc adm release) for ovnk image as a starting point so that we can have a running ovnk cluster to evolve. I was waiting for [this ovnk PR](https://github.com/openshift/ovn-kubernetes/pull/1188) to merge and the work to get OCP artifacts to be ready.

pkg/components/networking.go
zshi-redhat: not needed any more, will remove.

zshi-redhat: removed.

assets/components/ovn/configmap.yaml
zshi-redhat: fixed by passing the kubeconfig path param when applying ovnk configmap.

assets/components/ovn/configmap.yaml
zshi-redhat: fixed to use service cidr from microshift.conf.

assets/components/ovn/configmap.yaml
zshi-redhat: fixed to use cluster cidr from microshift.conf.

assets/components/ovn/master/daemonset.yaml
dhellmann: Sounds good, thanks!

zshi-redhat:
4.12.0-0.nightly-2022-07-14-072429 contains the sock fix for ovn-kubernetes, which can be used for microshift.
I tried `rebase.sh images`, it can update release_amd64.go to use ovn-k image url from 4.12.0-0.nightly-2022-07-14-072429.

zshi-redhat:
Rebased against master to take the latest changes of ovs configure script and microshift rpm

zshi-redhat:
> @mangelajo It seems the manifests change is not used in CI tests which fails to find ovnk daemonsets, any ideas?

I intentionally add the e2e validation for ovnk daemonset and it failed. If I remove the validation commit for ovnk, most of the CI jobs passed. I wonder if CI is taking the latest PR as input to run.

PR #769, opened 2022-07-13
tighten permissions on configuration data files

sallyom:
also, small fix to the apiserver-audit-log path

Closes [USHIFT-84](https://issues.redhat.com/browse/USHIFT-84)

### DIFF
--- a/pkg/controllers/openshift-apiserver.go
+++ b/pkg/controllers/openshift-apiserver.go
@@ -196,7 +196,7 @@ apiServerArguments:
 	if cfg.AuditLogDir != "" {
 		data = append(data, `
 auditConfig:
-  auditFilePath: `+cfg.AuditLogDir+`openshift-apiserver-audit.log
+  auditFilePath: "`+cfg.AuditLogDir+`/openshift-apiserver-audit.log"
   enabled: true
   logFormat: json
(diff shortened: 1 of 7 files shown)


ggiguash (changes requested):
Since we're specifying a mask/mode for directory creation, it must be at least 0700 to have x bit set for the owner. Otherwise, cd would not work into that directory

ggiguash (approved):

pkg/controllers/openshift-apiserver.go
```diff
@@ -196,7 +196,7 @@ apiServerArguments:
 	if cfg.AuditLogDir != "" {
 		data = append(data, `
 auditConfig:
-  auditFilePath: `+cfg.AuditLogDir+`openshift-apiserver-audit.log
+  auditFilePath: "`+cfg.AuditLogDir+`/openshift-apiserver-audit.log"
```
dhellmann, line 199: Will the need for that slash depend on whether the AuditLogDir variable ends in one? Should we use filepath.Join()?

dhellmann (commented):
I had one question that can be addressed in another PR if needed.

/lgtm

ggiguash:
USHIFT-84

PR #774, opened 2022-07-14
USHIFT-132: Disable Admission Plugins for non-used APIs

oglok:

Closes [USHIFT-132](https://issues.redhat.com/browse/USHIFT-132)

### DIFF
--- a/pkg/controllers/openshift-apiserver.go
+++ b/pkg/controllers/openshift-apiserver.go
@@ -172,6 +172,14 @@ aggregatorConfig:
   - X-Remote-User
 kubeClientConfig:
   kubeConfig:  ` + cfg.DataDir + `/resources/kubeadmin/kubeconfig
+admission:
+  disabledPlugins:
+  - "project.openshift.io/ProjectRequestLimit"
+  - "build.openshift.io/BuildConfigSecretInjector"
+  - "build.openshift.io/BuildByStrategy"
+  - "image.openshift.io/ImageLimitRange"
+  - "image.openshift.io/ImagePolicy"
+  - "quota.openshift.io/ClusterResourceQuota"
 apiServerArguments:
   minimal-shutdown-duration:
   - 30s


oglok:
We need to enhance the CRD file to add the `wildcardPolicy: None` value as a default, and OpenAPI schema validation.

assets/crd/0000_04_routes-crd.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
fzdarsky, line 2: Where in the OpenShift repos does this CRD come from?

fzdarsky:
Before removing the openshift-apiserver, are we sure that we get the exact same behaviour with those CRDS + controllers as with the openshift-apiserver? I'm thinking not only about existing apps that need to run on both OpenShift and MicroShift without modification, but also, for example, if we need to add code from the policy controller for handling SCCs, will this depend on code in openshift-apiserver?

mangelajo:
Wow, can we really remove the openshift-apiserver? what about the SCCs/etc?

assets/crd/0000_04_routes-crd.yaml
oglok: https://github.com/openshift/router/blob/master/deploy/route_crd.yaml

oglok:
> 

This is something we discussed in the last team meeting. Actuaylly, the first version of the PR was basically removing the API service and adding the CRD without getting rid of the openshift apiserver. Without basic CI, it's very hard to tell.

I really want this week to get another set of eyes (@benluddy ) to help me analyse what else OpenShift API server does in the context of MicroShift. If we replace all aggregated APIs by CRDs, can we get rid of the OPenShift API server? Is there anything else regarding SCCs that the OpenShift API server is doing?

oglok:
> Before removing the openshift-apiserver, are we sure that we get the exact same behaviour with those CRDS + controllers as with the openshift-apiserver? I'm thinking not only about existing apps that need to run on both OpenShift and MicroShift without modification, but also, for example, if we need to add code from the policy controller for handling SCCs, will this depend on code in openshift-apiserver?

That's why we need to perform a deep analysis of what the OpenShift API server does in a MicroShift context once we get rid of the aggregated APIs.

oglok:
@Miciah has pushed this PR related to the Route CRD to OpenShift:

https://github.com/openshift/api/pull/1228

PR #776, opened 2022-07-15
USHIFT-180: add version config map

dhellmann:
Add ConfigMap called `microshift-version` to the `kube-public` namespace with version details of the MicroShift service so that API consumers can tell that they are talking to a MicroShift instance.

Closes USHIFT-180


dhellmann:
/hold

This isn't the only way to solve the problem, and we should discuss the security implications of making the version details freely available.

pkg/controllers/version.go
```diff
@@ -0,0 +1,63 @@
+/*
+Copyright © 2022 Microshift Contributors
```
ggiguash, line 2: MicroShift?

pkg/controllers/version.go
```diff
@@ -0,0 +1,63 @@
...
+	defer close(ready)
+
+	versionInfo := version.Get()
+	var data = map[string]string{
+		"major":   versionInfo.Major,
+		"minor":   versionInfo.Minor,
+		"version": versionInfo.String(),
+	}
+
+	kubeConfigPath := s.cfg.DataDir + "/resources/kubeadmin/kubeconfig"
```
ggiguash, line 56: filepath.Join?

pkg/assets/bindata.go
```diff
@@ -2684,7 +2685,34 @@ func assetsScc0000_20_kubeApiserverOperator_00_sccRestrictedYaml() (*asset, erro
...
+	return _assetsVersionMicroshiftVersionYaml, nil
+}
+
+func assetsVersionMicroshiftVersionYaml() (*asset, error) {
+	bytes, err := assetsVersionMicroshiftVersionYamlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "assets/version/microshift-version.yaml", size: 196, mode: os.FileMode(436), modTime: time.Unix(1654679854, 0)}
```
ggiguash, line 2715: mode 436 -> 420?

pkg/assets/bindata.go
```diff
@@ -150,7 +151,7 @@ func assetsComponentsHostpathProvisionerClusterroleYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "assets/components/hostpath-provisioner/clusterrole.yaml", size: 609, mode: os.FileMode(436), modTime: time.Unix(1654679854, 0)}
+	info := bindataFileInfo{name: "assets/components/hostpath-provisioner/clusterrole.yaml", size: 609, mode: os.FileMode(420), modTime: time.Unix(1654679854, 0)}
```
ggiguash, line 154: If FileMode and ModTime are common to all the bindataFileInfo calls, should we define them as constants in one place and use consistently?

ggiguash (commented):
I added few minor comments. Good to have a version identifier for MicroShift!

pkg/assets/bindata.go
dhellmann: This file is auto-generated by `go-bindata` (see `scripts/bindata.sh`). I'm not sure why the file mode values have all changed. Perhaps that's a difference in a new version of the tool?

pkg/controllers/version.go
dhellmann: This format was consistent with all of the other path manipulation, but you're right. I've updated this spot and filed #777 to fix it elsewhere in the code base.

pkg/controllers/version.go
dhellmann: Fixed

pkg/controllers/version.go
ggiguash: Thank you, I approved the #777 PR.

pkg/controllers/version.go
```diff
@@ -0,0 +1,64 @@
...
+
+func NewVersionManager(cfg *config.MicroshiftConfig) *VersionManager {
+	s := &VersionManager{}
+	s.cfg = cfg
+	return s
+}
+
+func (s *VersionManager) Name() string { return "version-manager" }
+func (s *VersionManager) Dependencies() []string {
+	return []string{"kube-apiserver", "openshift-apiserver"}
```
mangelajo, line 40:
```suggestion
	return []string{"kube-apiserver"}
```

mangelajo: I believe for config map we only need to wait for the kube-apiserver

mangelajo (changes requested):

pkg/controllers/version.go
dhellmann: Fixed in #785

berenss:
would you like to get a review from a RHACM server foundation engineer to ensure that the bits involved will do what we need?  I get that this is a 'generic' goal, but lets ensure it fits also the immediate need.

dhellmann:
@berenss sure, we'd be happy to have any specific feedback or requirements. This solves the immediate need to have a way to tell what version is in place from a workload, but it's not necessarily the only info that would be useful

PR #777, opened 2022-07-16
Use filepath join

dhellmann:
Use filepath.Join() to combine strings to build filenames instead of relying on string concatenation. This is more reliable if the configured directory name ends in a path separator character.

[USHIFT-180](https://issues.redhat.com//browse/USHIFT-180)

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"net"
+	"path/filepath"
 
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/controllers"
@@ -42,7 +43,7 @@ func initAll(cfg *config.MicroshiftConfig) error {
 }
 
 func loadCA(cfg *config.MicroshiftConfig) error {
-	return util.LoadRootCA(cfg.DataDir+"/certs/ca-bundle", "ca-bundle.crt", "ca-bundle.key")
+	return util.LoadRootCA(filepath.Join(cfg.DataDir, "/certs/ca-bundle"), "ca-bundle.crt", "ca-bundle.key")
 }
 
 func initCerts(cfg *config.MicroshiftConfig) error {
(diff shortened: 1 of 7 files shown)


pkg/cmd/init.go
```diff
@@ -141,23 +142,28 @@ func initServerConfig(cfg *config.MicroshiftConfig) error {
...
+		"kube-apiserver", []string{"kube-apiserver", "system:kube-apiserver", "system:masters"}, cfg.Cluster.URL); err != nil {
 		return err
 	}
-	if err := util.Kubeconfig(cfg.DataDir+"/resources/kube-controller-manager/kubeconfig", "system:kube-controller-manager", []string{"system:kube-controller-manager"}, cfg.Cluster.URL); err != nil {
+	if err := util.Kubeconfig(filepath.Join(cfg.DataDir, "/resources/kube-controller-manager/kubeconfig"),
+		"system:kube-controller-manager", []string{"system:kube-controller-manager"}, cfg.Cluster.URL); err != nil {
 		return err
-	}
-	if err := util.Kubeconfig(cfg.DataDir+"/resources/kube-scheduler/kubeconfig", "system:kube-scheduler", []string{"system:kube-scheduler"}, cfg.Cluster.URL); err != nil {
+	} if err := util.Kubeconfig(filepath.Join(cfg.DataDir, "/resources/kube-scheduler/kubeconfig"),
```
ggiguash, line 156: A nit: this probably should be on the next line.

ggiguash (approved):

dhellmann:
/hold

There's no point continually testing.

pkg/cmd/init.go
mangelajo:
```suggestion
	} 
	if err := util.Kubeconfig(filepath.Join(cfg.DataDir, "/resources/kube-scheduler/kubeconfig"),
```

mangelajo (commented):
/lgtm with a little comment from @ggiguash 

CI is upside down

pkg/cmd/init.go
dhellmann: That's so weird. This file does not look like this on my local system. There's a comment on that line instead. I've added a commit to adjust the whitespace to see if that helps with the rendering.

dhellmann: And of course I was on the wrong branch. This is fixes.

ggiguash (approved):

PR #781, opened 2022-07-19
Add microshift-ovs-init script as oneshot during boot

mangelajo:
This script comes from the machine config operator [1] and configures
br-ex for use with OVNKubernetes, making sure that the main interface
is connected to br-ex.

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml

Closes [USHIFT-145](https://issues.redhat.com/browse/USHIFT-145)

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -167,6 +167,11 @@ install -p -m644 packaging/systemd/microshift-openvswitch.service %{buildroot}%{
 install -p -m644 packaging/systemd/microshift-ovs-vswitchd.service %{buildroot}%{_unitdir}/microshift-ovs-vswitchd.service
 install -p -m644 packaging/systemd/microshift-ovsdb-server.service %{buildroot}%{_unitdir}/microshift-ovsdb-server.service
 
+# this script and systemd service configures openvswitch to properly operate with OVN
+install -p -m644 packaging/systemd/microshift-ovs-init.service %{buildroot}%{_unitdir}/microshift-ovs-init.service
+install -p -m755 packaging/systemd/configure-ovs.sh %{buildroot}%{_bindir}/configure-ovs.sh
+install -p -m755 packaging/systemd/configure-ovs-microshift.sh %{buildroot}%{_bindir}/configure-ovs-microshift.sh
+
 mkdir -p -m755 %{buildroot}/var/run/kubelet
 mkdir -p -m755 %{buildroot}/var/lib/kubelet/pods
 mkdir -p -m755 %{buildroot}/var/run/secrets/kubernetes.io/serviceaccount
@@ -207,11 +212,13 @@ sed -i -n -e '/^OPTIONS=/!p' -e '$aOPTIONS="--no-mlockall"' /etc/sysconfig/openv
 %systemd_post microshift-openvswitch.service
 %systemd_post microshift-ovs-vswitchd.service
 %systemd_post microshift-ovsdb-server.service
+%systemd_post microshift-ovs-init.service
 
 %preun networking
 %systemd_preun microshift-openvswitch.service
 %systemd_preun microshift-ovs-vswitchd.service
 %systemd_preun microshift-ovsdb-server.service
+%systemd_preun microshift-ovs-init.service
 
 %preun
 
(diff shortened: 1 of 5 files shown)


packaging/rpm/microshift.spec
```diff
@@ -242,7 +248,15 @@ sed -i -n -e '/^OPTIONS=/!p' -e '$aOPTIONS="--no-mlockall"' /etc/sysconfig/openv
 %{_unitdir}/microshift-ovs-vswitchd.service
 %{_unitdir}/microshift-ovsdb-server.service
 
+# ovn/ovs oneshot configuration script
```
zshi-redhat, line 251: nit: remove 'ovn'

zshi-redhat (commented):
@mangelajo I remembered you mentioned the missing of network manager ovs-bridge plugin, Shall we add NetworkManager-ovs as required package for microshift rpm?

Do you think it is worth updating the configure-ovs.sh on every automatic rebase? Like how image and manifests are pulled from openshift repo?

zshi-redhat:
> > @mangelajo I remembered you mentioned the missing of network manager ovs-bridge plugin, Shall we add NetworkManager-ovs as required package for microshift rpm?
> > Do you think it is worth updating the configure-ovs.sh on every automatic rebase? Like how image and manifests are pulled from openshift repo?
> 
> It's already set (see line 107). Probably would have made more sense to add it in this patch, but I added it when I created the microhisft-networking package

oh, I missed that, thanks for the pointer!

packaging/rpm/microshift.spec
zshi-redhat: It is host ovs configuration for ovn-k8s gateway setup.

mangelajo:
Ok all is working on my side, fresh R3 4E make iso boot tested and everything is running on br-ex

mangelajo:
I will include the other elements here: 
```
$ ovs-vsctl add-br br-ex -- set bridge br-ex other-config:hwaddr=\"<mac-address-of-eno1>\"
$ ovs-vsctl add-port br-ex eno1; ip addr flush dev eno1; pkill dhclient; dhclient -v br-ex
$ ovs-vsctl set interface eno1 type=system
$ ovs-vsctl set open . external-ids:ovn-remote="unix:/var/run/ovn/ovnsb_db.sock"
$ ovs-vsctl set open . other_config:n-handler-threads=1
$ ovs-vsctl set open . other_config:n-revalidator-threads=1
```
as part of this initialization

zshi-redhat:
> I will include the other elements here:
> 
> ```
> $ ovs-vsctl add-br br-ex -- set bridge br-ex other-config:hwaddr=\"<mac-address-of-eno1>\"

Ideally, this will be configured by ovn-k8s when starts up: https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/util/nicstobridge.go#L237

> $ ovs-vsctl add-port br-ex eno1; ip addr flush dev eno1; pkill dhclient; dhclient -v br-ex
> $ ovs-vsctl set interface eno1 type=system
> $ ovs-vsctl set open . external-ids:ovn-remote="unix:/var/run/ovn/ovnsb_db.sock"

By ovn-k8s: https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L2150

> $ ovs-vsctl set open . other_config:n-handler-threads=1

This cannot be set in openshift2.17+, it will be automatically configured.

> $ ovs-vsctl set open . other_config:n-revalidator-threads=1
> ```
> 
> as part of this initialization

packaging/systemd/configure-ovs-microshift.sh
```diff
@@ -0,0 +1,18 @@
...
+OVS_HANDLER_THREADS=${OVS_HANDLER_THREADS:-1}
+OVS_REVALIDATOR_THREADS=${OVS_REVALIDATOR_THREADS:-1}
+
+# let ovn-controller know we are using an unix socket for SBDB communication instead of TLS/TCP
+ovs-vsctl set open . external-ids:ovn-remote="unix:/var/run/ovn/ovnsb_db.sock"
+
+# limit the number of threads in ovs-vswitchd to limit the memory consumption
+
+if [[ "${OVS_HANDLER_THREADS}" != "auto" ]]; then
+    ovs-vsctl set open . other_config:n-handler-threads=${OVS_HANDLER_THREADS}
```
zshi-redhat, line 13: This won't be applied with openvswitch2.16 version.

packaging/systemd/configure-ovs-microshift.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+set -eux
+
+OVS_HANDLER_THREADS=${OVS_HANDLER_THREADS:-1}
+OVS_REVALIDATOR_THREADS=${OVS_REVALIDATOR_THREADS:-1}
+
+# let ovn-controller know we are using an unix socket for SBDB communication instead of TLS/TCP
+ovs-vsctl set open . external-ids:ovn-remote="unix:/var/run/ovn/ovnsb_db.sock"
```
zshi-redhat, line 8: we probably want to leave this configuration to ovn-kubernetes.

mangelajo: Oh, I thought we had to do this, I was following your .md README, I can remove it now, and we can re-add it in your patch if it's necessary

packaging/systemd/configure-ovs-microshift.sh
mangelajo: I saw handling it on the logs, it's fine though even if it's not fully effective, will be when we switch to 2.17

mangelajo: I think you mean, it won't be applied with 2.17?

mangelajo: I wonder if we can leave it in place (if harmless) and have the ovs team re-admit that parameter so we don't need to touch the cpu affinity?

mangelajo:
@zshi-redhat @oglok I suggest we move on with this one, and do any follow-up changes, so we can unblock your patches and get it tested on CI, then we can finetune.

mangelajo:
rebasing and squashing

PR #788, opened 2022-07-21
remove the openshift-apiserver

deads2k:
removes the openshift-apiserver from microshift and fully uses the route CRD instead.


pkg/controllers/openshift-controller-manager.go
```diff
@@ -50,7 +50,7 @@ func NewOpenShiftControllerManager(cfg *config.MicroshiftConfig) *OCPControllerM
 
 func (s *OCPControllerManager) Name() string { return componentOCM }
 func (s *OCPControllerManager) Dependencies() []string {
-	return []string{"kube-apiserver", "openshift-apiserver"}
+	return []string{"kube-apiserver"}
```
fzdarsky, line 53: Would/should this depend on the openshift-crd-manager instead, to wait until the CRDs become available?

If so, you'd need to swap the order of adding these services in `run.go`, too.

assets/crd/0000_01_route.crd.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apiextensions.k8s.io/v1
```
fzdarsky, line 1: Where is this file copied from (for the rebase script), please?

oglok (changes requested):
We should use this enhanced CRD which has a lot more validations and default values:

https://github.com/openshift/api/pull/1228/files#diff-6f2a93a87b2c5d86ab777d70320bba54ba8395569f69dc4b4446ac9c3cb76480

fzdarsky (commented):
We may want to update the `go.mod` to source the versions "from kubernetes" where they are now sourced "from openshift-apiserver" and then rebase to the current release. That probably means reviewing the applied carry patches, so I'd be ok doing that in a separate PR post the first deliverable.

assets/crd/0000_01_route.crd.yaml
deads2k: now from https://github.com/openshift/api/pull/1228/files#diff-6f2a93a87b2c5d86ab777d70320bba54ba8395569f69dc4b4446ac9c3cb76480

Eventually from the openshift/api repo once the PR is green.

pkg/controllers/openshift-controller-manager.go
deads2k: Ordering is still fraught, but I've added this in another commit.

oglok:
In case the team decides to go for it, lgtm. (I'll be out on Monday)

assets/crd/0000_01_route.crd.yaml
deads2k: Now present in https://github.com/openshift/api/blob/master/route/v1/route.crd.yaml

ggiguash:
USHIFT-203

PR #789, opened 2022-07-21
provide more complete route schema.

deads2k:
like https://github.com/openshift/microshift/pull/788 with an auto-generated schema


oglok:
Why not changing the existing #788 PR?

@Miciah generated this CRD with validations and default values I think we should use:

https://github.com/openshift/api/pull/1228/files#diff-6f2a93a87b2c5d86ab777d70320bba54ba8395569f69dc4b4446ac9c3cb76480

deads2k:
> Why not changing the existing #788 PR?
> 
> @Miciah generated this CRD with validations and default values I think we should use:

Just making the delta clearer.  I've pushed into both now.

deads2k:
https://github.com/openshift/microshift/pull/788 is happy

/close

PR #791, opened 2022-07-21
Document Red Hat AMQ Broker 7.8 deployment and testing

ggiguash:
Document Red Hat AMQ Broker 7.8 deployment and testing
Closes [USHIFT-36](https://issues.redhat.com//browse/USHIFT-36)


docs/config/amq-broker.spec
```diff
@@ -0,0 +1,73 @@
+---
```
dhellmann, line 1: Is there a reason for this file to be `.spec` instead of `.yml`?

dhellmann (commented):
One question, otherwise looks good.

docs/config/amq-broker.spec
```diff
@@ -0,0 +1,73 @@
...
+    matchLabels:
+      app: amq-broker
+  template:
+    metadata:
+      labels:
+        app: amq-broker
+    spec:
+      containers:
+      - name: amq-broker
+        image: registry.redhat.io/amq7/amq-broker:7.8
```
mangelajo, line 26: Can the broker work without any volume/persistence?

docs/config/amq-broker.spec
ggiguash: I updated the file names to have .yaml extension

docs/config/amq-broker.spec
ggiguash: Currently, if the pod restarts, the queue-related data is lost. Do you think we need to demonstrate persistency in this doc?

mangelajo: Probably we should but we can do that step wise, may be start with this version and then add persistency?

dhellmann: Iterating makes sense to me.

ggiguash: I created a follow-up https://issues.redhat.com/browse/USHIFT-220 issue for implementing the persistence of queues. 
It was not straightforward to reconfigure AMQ Broker to use a volume directory for its journal file.

ggiguash: Can we please approve this PR if it's good in its state?

mangelajo: /lgtm

PR #793, opened 2022-07-22
Trigger br-ex configuration only once in MicroShift

mangelajo:
otherwise it breaks on reboot.


zshi-redhat:
lgtm

zshi-redhat:
/lgtm
it was verified manually

PR #795, opened 2022-07-22
Add the jq dependency

mangelajo:
microshift-ovs-init / configure-ovs.sh uses jq, and fails if it's not
installed. RHEL4Edge comes with jq by default but RHEL8 doesn't,
and we test on GCE with RHEL8.


dhellmann:
USHIFT-138

dhellmann:
USHIFT-186

PR #794, opened 2022-07-22
USHIFT-216: Fix kube-apiserver readiness check panic on kubeconfig error.

benluddy:
Errors returned by BuildConfigFromFlags were sent to the error
channel, but control flow continued despite the nil rest.Config
pointer.

PR #798, opened 2022-07-25
USHIFT-16: Use /readyz for kube-apiserver readiness checks instead of /healthz.

benluddy:
They're nearly equivalent, but there are behaviors that are controlled
by whether or not the apiserver has signaled readiness at least
once (e.g. warnings are logged about receiving requests before ready).

/hold

I'd originally included this in https://github.com/openshift/microshift/pull/794 and later separated it accounting reasons. This patch depends on the other.


mangelajo (commented):
/approve

PR #801, opened 2022-07-27
automatic rebasing of ovnk manifests

zshi-redhat:

### DIFF
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -31,7 +31,7 @@ PULL_SECRET_FILE="${HOME}/.pull-secret.json"
 
 EMBEDDED_COMPONENTS="openshift-controller-manager oauth-apiserver hyperkube etcd"
 EMBEDDED_COMPONENT_OPERATORS="cluster-kube-apiserver-operator cluster-kube-controller-manager-operator cluster-openshift-controller-manager-operator cluster-kube-scheduler-operator machine-config-operator"
-LOADED_COMPONENTS="cluster-dns-operator cluster-ingress-operator service-ca-operator"
+LOADED_COMPONENTS="cluster-dns-operator cluster-ingress-operator service-ca-operator cluster-network-operator"
 
 
 title() {
@@ -401,6 +401,12 @@ update_manifests() {
     sed -i 's|\${IMAGE}|{{ .ReleaseImage.service_ca_operator }}|' "${REPOROOT}"/assets/components/service-ca/deployment.yaml
     sed -i 's|REPLACE_TLS_SECRET|{{.TLSSecret}}|' "${REPOROOT}"/assets/components/service-ca/deployment.yaml
     sed -i 's|REPLACE_CA_CONFIG_MAP|{{.CAConfigMap}}|' "${REPOROOT}"/assets/components/service-ca/deployment.yaml
+
+    #-- ovn-kubernetes -----------------------------------
+    # 1) Adopt resource manifests
+    #    Replace all ovn-kubernetes manifests
+    rm -rf "${REPOROOT}"/assets/components/ovn/*
+    cp -r "${STAGING_DIR}"/cluster-network-operator/bindata/network/ovn-kubernetes/microshift/* "${REPOROOT}"/assets/components/ovn 2>/dev/null || true
 
     popd >/dev/null
 }


zshi-redhat:
depends on https://github.com/openshift/cluster-network-operator/pull/1517

fzdarsky (approved):

scripts/rebase.sh
```diff
@@ -402,6 +402,12 @@ update_manifests() {
     sed -i 's|REPLACE_TLS_SECRET|{{.TLSSecret}}|' "${REPOROOT}"/assets/components/service-ca/deployment.yaml
     sed -i 's|REPLACE_CA_CONFIG_MAP|{{.CAConfigMap}}|' "${REPOROOT}"/assets/components/service-ca/deployment.yaml
 
+    #-- ovn-kubernetes -----------------------------------
+    # 1) Adopt resource manifests
+    #    Replace all ovn-kubernetes manifests
+    rm -rf "${REPOROOT}"/assets/components/ovn/*
+    cp -r "${STAGING_DIR}"/cluster-network-operator/bindata/network/ovn-kubernetes/microshift/* "${REPOROOT}"/assets/components/ovn 2>/dev/null || true
```
mangelajo, line 409: So the cluster network operator will contain the manifests for microshift too?

zshi-redhat: Yes, I think this is what we want, please correct me if I'm wrong.
Having ovnk manifests in one place for easy maintainese, although need to have knowledge on how microshift consume ovnk.

dhellmann: Ideally we would be able to create a CI job to run when changes are proposed to either repository to ensure those manifests stay up to date as changes are made to cluster-network-operator.

I'm a little concerned with sending messages from the copy command to /dev/null and ignoring errors. If something goes wrong with the copy (permissions issue, files are renamed, etc.) we need to know and the rebase should fail, right?

zshi-redhat: > Ideally we would be able to create a CI job to run when changes are proposed to either repository to ensure those manifests stay up to date as changes are made to cluster-network-operator.

I have a question about creating the CI jobs (may not relevant to this PR):
My understanding is we will create a CI job that runs microshift tests in cluster-network-operator repo. How the CI job in microshift repo be triggered when changes are made to cluster-network-operator? Is it through the periodic rebase PR? or will there be an immediate CI run be triggered in microshift repo?

> 
> I'm a little concerned with sending messages from the copy command to /dev/null and ignoring errors. If something goes wrong with the copy (permissions issue, files are renamed, etc.) we need to know and the rebase should fail, right?

Agree, we don't want silent failure, this shall be notified in case any errors.
I noticed we have the /dev/null in other rebase copies, @fzdarsky  do you have other considerations by redirecting to `/dev/null`?

mangelajo: I guess having them in the cluster network operator would benefit things like SNO too

dhellmann: > > Ideally we would be able to create a CI job to run when changes are proposed to either repository to ensure those manifests stay up to date as changes are made to cluster-network-operator.
> 
> I have a question about creating the CI jobs (may not relevant to this PR):

Yes, let's not block this PR on this discussion.

>  My understanding is we will create a CI job that runs microshift tests in cluster-network-operator repo. How the CI job in microshift repo be triggered when changes are made to cluster-network-operator? Is it through the periodic rebase PR? or will there be an immediate CI run be triggered in microshift repo?

If we had one job that checked out the source for both repositories then built and tested MicroShift, we could use that whenever anything changes in either repo. If that's not easy to do, then a periodic job may be the next best thing. It would let us land breaking changes, but identify them quickly.

dhellmann: > I guess having them in the cluster network operator would benefit things like SNO too

Yes, that's a good point, too. I was thinking mostly of who should "own" the files and ensuring that they are kept up to date. Keeping them close together with other related code should make it more likely that they're updated as other similar changes are made for the networking code. We'll want to do something similar for most of the rest of the assets that we import, too.

zshi-redhat:
> /approve /hold (until the PR it depends on lands)

@mangelajo @fzdarsky @dhellmann @ggiguash 
The dependent PR is finally merged, I will go ahead and unhold this PR if no objections.

PR #802, opened 2022-07-27
Rebase arm64 images too

fzdarsky:
Makes the rebase script accept _two_ pull specs for release images, the first the one for the amd64 architecture (as before), the second the one for the arm64 architecture. Then uses the arm64 release image to update that architecture's container image digests.

[USHIFT-177](https://issues.redhat.com//browse/USHIFT-177)

### DIFF
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -61,7 +61,7 @@ download_release() {
     title "# Fetching release info for ${release_image_arm64} (arm64)"
     oc adm release info ${authentication} "${release_image_arm64}" -o json > release_arm64.json
 
-    title "# Extracing ${release_image_amd64} manifest content"
+    title "# Extracting ${release_image_amd64} manifest content"
     mkdir -p release-manifests
     pushd release-manifests >/dev/null
     content=$(oc adm release info ${authentication} --contents "${release_image_amd64}")
@@ -241,7 +241,7 @@ get_release_images() {
 
 # Updates the image digests in pkg/release/release*.go
 update_images() {
-    if [ ! -f "${STAGING_DIR}/release_amd64.json" ]; then
+    if [ ! -f "${STAGING_DIR}/release_amd64.json" ] || [ ! -f "${STAGING_DIR}/release_arm64.json" ]; then
         >&2 echo "No release found in ${STAGING_DIR}, you need to download one first."
         exit 1
     fi
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -42,7 +42,8 @@ title() {
 # Downloads a release's tools and manifest content into a staging directory,
 # then checks out the required components for the rebase at the release's commit.
 download_release() {
-    local release_image=$1
+    local release_image_amd64=$1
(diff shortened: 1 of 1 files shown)


scripts/rebase.sh
```diff
@@ -55,20 +56,20 @@ download_release() {
...
-    oc adm release extract ${authentication} --tools "${release_image}"
-    echo "Content of release.txt:"
-    cat release.txt
+    title "# Fetching release info for ${release_image_amd64} (amd64)"
+    oc adm release info ${authentication} "${release_image_amd64}" -o json > release_amd64.json
+    title "# Fetching release info for ${release_image_arm64} (arm64)"
+    oc adm release info ${authentication} "${release_image_arm64}" -o json > release_arm64.json
 
-    title "# Extracing ${release_image} manifest content"
+    title "# Extracing ${release_image_amd64} manifest content"
```
oglok, line 64: s/Extracing/Extracting

scripts/rebase.sh
```diff
@@ -240,21 +241,25 @@ get_release_images() {
 
 # Updates the image digests in pkg/release/release*.go
 update_images() {
-    if [ ! -f "${STAGING_DIR}/release.txt" ]; then
+    if [ ! -f "${STAGING_DIR}/release_amd64.json" ]; then
```
oglok, line 244: Don't you have to check also if release_arm64.json exists?

scripts/rebase.sh
```diff
@@ -55,20 +56,20 @@ download_release() {
...
+    oc adm release info ${authentication} "${release_image_amd64}" -o json > release_amd64.json
+    title "# Fetching release info for ${release_image_arm64} (arm64)"
+    oc adm release info ${authentication} "${release_image_arm64}" -o json > release_arm64.json
 
-    title "# Extracing ${release_image} manifest content"
+    title "# Extracing ${release_image_amd64} manifest content"
     mkdir -p release-manifests
     pushd release-manifests >/dev/null
-    content=$(oc adm release info ${authentication} --contents "${release_image}")
+    content=$(oc adm release info ${authentication} --contents "${release_image_amd64}")
```
oglok, line 67: Manifests are the same for any architecture, aren't they? Images will be multiarch IIUC

oglok (commented):
one typo and couple of questions. The rest looks good to me.

scripts/rebase.sh
fzdarsky: Good catch. thanks. Fixed.

scripts/rebase.sh
fzdarsky: They should always come together, but better to be safe. Fixed this.

scripts/rebase.sh
fzdarsky: Yes, they are. Or rather, they would be if both Intel and Arm release images were created together. As they are currently created independently, there could be a small delta, but that should be a rare event.

Multi-arch images are still experimental and not available for our version. Hopefully we'll be able to switch eventually.

PR #803, opened 2022-07-27
USHIFT-213: Enable OpenShift flavor of kube-apiserver.

benluddy:
Rather than maintain a parallel implementation of the kube-apiserver
cobra command's "Run", construct a KubeAPIServerConfig manifest file
and pass its path as a command argument. This preserves the
OpenShift's patched kube-apiserver behavior, including notable changes
to admission configuration: SecurityContextConstraint enforcement,
timeout, per-namespace runlevels, and custom resource validation and
defaulting for included OpenShift APIs.

### DIFF
--- a/assets/components/kube-apiserver/config-overrides.yaml
+++ b/assets/components/kube-apiserver/config-overrides.yaml
@@ -1,0 +1,20 @@
+apiVersion: kubecontrolplane.config.openshift.io/v1
+kind: KubeAPIServerConfig
+apiServerArguments:
+  # The following arguments are required to enable bound sa
+  # tokens. This is only supported post-bootstrap so these
+  # values must not appear in defaultconfig.yaml.
+  service-account-issuer:
+    - https://kubernetes.default.svc
+  api-audiences:
+    - https://kubernetes.default.svc
+  service-account-signing-key-file:
+    - /etc/kubernetes/static-pod-certs/secrets/bound-service-account-signing-key/service-account.key
+serviceAccountPublicKeyFiles:
+  # this being a directory means we cannot directly use the upstream flags.
+  # TODO make a configobserver that writes the individual values that we need.
+  - /etc/kubernetes/static-pod-resources/configmaps/sa-token-signing-certs
+  # The following path contains the public keys needed to verify bound sa
+  # tokens. This is only supported post-bootstrap.
+  - /etc/kubernetes/static-pod-resources/configmaps/bound-sa-token-signing-certs
+
--- a/assets/components/kube-apiserver/defaultconfig.yaml
+++ b/assets/components/kube-apiserver/defaultconfig.yaml
@@ -1,0 +1,179 @@
+apiVersion: kubecontrolplane.config.openshift.io/v1
+kind: KubeAPIServerConfig
+admission:
+  pluginConfig:
+    network.openshift.io/ExternalIPRanger:
+      configuration:
+        allowIngressIP: true
+        apiVersion: network.openshift.io/v1
+        externalIPNetworkCIDRs: null
+        kind: ExternalIPRangerAdmissionConfig
+      location: ""
+    PodSecurity:
+      configuration:
+        kind: PodSecurityConfiguration
+        apiVersion: pod-security.admission.config.k8s.io/v1alpha1
+        defaults:
+          enforce: "privileged"
+          enforce-version: "latest"
+          audit: "baseline"
+          audit-version: "latest"
+          warn: "baseline"
+          warn-version: "latest"
+apiServerArguments:
+  allow-privileged:
+    - "true"
+  anonymous-auth:
+    - "true"
+  authorization-mode:
+    - Scope
+    - SystemMasters
+    - RBAC
+    - Node
+  audit-log-format:
+    - json
+  audit-log-maxbackup:
+    - "10"
+  audit-log-maxsize:
+    - "100"
+  audit-log-path:
+    - /var/log/kube-apiserver/audit.log
+  audit-policy-file:
+    - /etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-audit-policies/policy.yaml
+  client-ca-file:
+    - /etc/kubernetes/static-pod-certs/configmaps/client-ca/ca-bundle.crt
+  disable-admission-plugins:
+    - PodSecurity
+  enable-admission-plugins:
+    - CertificateApproval
+    - CertificateSigning
+    - CertificateSubjectRestriction
+    - DefaultIngressClass
+    - DefaultStorageClass
+    - DefaultTolerationSeconds
+    - LimitRanger
+    - MutatingAdmissionWebhook
+    - NamespaceLifecycle
+    - NodeRestriction
+    - OwnerReferencesPermissionEnforcement
+    - PersistentVolumeClaimResize
+    - PersistentVolumeLabel
+    - PodNodeSelector
+    - PodTolerationRestriction
+    - Priority
+    - ResourceQuota
+    - RuntimeClass
+    - ServiceAccount
+    - StorageObjectInUseProtection
+    - TaintNodesByCondition
+    - ValidatingAdmissionWebhook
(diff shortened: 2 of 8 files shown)


oglok:
Something is conflicting with vendor/modules.txt

Should we run `go mod vendor` again?

benluddy:
https://github.com/openshift/microshift/pull/862 just merged -- rebased.

assets/components/kube-apiserver/defaultconfig.yaml
```diff
@@ -0,0 +1,179 @@
...
+    - ServiceAccount
+    - StorageObjectInUseProtection
+    - TaintNodesByCondition
+    - ValidatingAdmissionWebhook
+    - authorization.openshift.io/RestrictSubjectBindings
+    - authorization.openshift.io/ValidateRoleBindingRestriction
+    - config.openshift.io/DenyDeleteClusterConfiguration
+    - config.openshift.io/ValidateAPIServer
+    - config.openshift.io/ValidateAuthentication
+    - config.openshift.io/ValidateConsole
```
oglok, line 75: Should we enable this admission plugin if we're not deploying the console?

assets/components/kube-apiserver/defaultconfig.yaml
```diff
@@ -0,0 +1,179 @@
...
+    - config.openshift.io/ValidateFeatureGate
+    - config.openshift.io/ValidateImage
+    - config.openshift.io/ValidateOAuth
```
oglok, line 78: There's no OAuth endpoint in MicroShift's KAS: 

https://github.com/openshift/microshift/pull/846

Should we disable this as well?

assets/components/kube-apiserver/config-overrides.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kubecontrolplane.config.openshift.io/v1
```
oglok, line 1: All manifests under the `assets` directory will be applied once KAS notifies ready state. What are the implications of applying a new KAS configuration object afterwards?
In other control plane components, we create config files and store them in `/var/lib/microshift/` and pass them automatically. I'm not against this approach, I just want to understand the implications, and if we decide this is a better approach, we can move all configs to assets.

pkg/components/storage.go
```diff
@@ -30,15 +30,13 @@ func startCSIPLugin(kubeconfigPath string) error {
 			"assets/components/odf-lvm/topolvm-csi-resizer_rbac.authorization.k8s.io_v1_clusterrole.yaml",
 			"assets/components/odf-lvm/topolvm-node-scc_rbac.authorization.k8s.io_v1_clusterrole.yaml",
 			"assets/components/odf-lvm/topolvm-node_rbac.authorization.k8s.io_v1_clusterrole.yaml",
-			"assets/components/odf-lvm/topolvm-node-scc_rbac.authorization.k8s.io_v1_clusterrole.yaml",
```
oglok, line 33: Why are we removing the clusterrole and clusterrolebinding for topolvm? Is the new SCC enforcing already the persmissions?

pkg/controllers/kube-apiserver.go
```diff
@@ -16,130 +16,193 @@ limitations under the License.
...
+	s.kubeconfig = filepath.Join(cfg.DataDir, "resources", "kube-apiserver", "kubeconfig")
+	s.verbosity = cfg.LogVLevel
+
 	caCertFile := filepath.Join(cfg.DataDir, "certs", "ca-bundle", "ca-bundle.crt")
-	// certDir := filepath.Join(cfg.DataDir, "certs", s.Name())
-	// dataDir := filepath.Join(cfg.DataDir, s.Name())
 
 	if err := s.configureAuditPolicy(cfg); err != nil {
-		klog.Fatalf("Failed to configure kube-apiserver audit policy %v", err)
+		return fmt.Errorf("Failed to configure kube-apiserver audit policy: %w", err)
```
oglok, line 96: Why not using klog? It's our logging system for the entire project.

pkg/controllers/kube-apiserver.go
```diff
@@ -16,130 +16,193 @@ limitations under the License.
...
+			"kubelet-client-certificate":    {cfg.DataDir + "/resources/kube-apiserver/secrets/kubelet-client/tls.crt"},
+			"kubelet-client-key":            {cfg.DataDir + "/resources/kube-apiserver/secrets/kubelet-client/tls.key"},
+
+			"proxy-client-cert-file":           {cfg.DataDir + "/certs/kube-apiserver/secrets/aggregator-client/tls.crt"},
+			"proxy-client-key-file":            {cfg.DataDir + "/certs/kube-apiserver/secrets/aggregator-client/tls.key"},
+			"requestheader-client-ca-file":     {caCertFile},
+			"service-account-signing-key-file": {cfg.DataDir + "/resources/kube-apiserver/secrets/service-account-key/service-account.key"},
+			"tls-cert-file":                    {cfg.DataDir + "/certs/kube-apiserver/secrets/service-network-serving-certkey/tls.crt"},
+			"tls-private-key-file":             {cfg.DataDir + "/certs/kube-apiserver/secrets/service-network-serving-certkey/tls.key"},
+			"disable-admission-plugins": {
```
oglok, line 126: I see now that the KAS config manifests are just a template used to override values.

pkg/controllers/kube-apiserver.go
```diff
@@ -225,19 +292,36 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
 		close(ready)
 	}()
 
-	// Work around that that the NewAPIServerCommand hardcodes the stop channel to SIGTERM signals,
-	// so we cannot use the cobra RunE command directly.
-	completedOptions, err := kubeapiserver.Complete(s.serverOptions)
+	fd, err := os.CreateTemp("", "kube-apiserver-config-*.yaml")
```
oglok, line 295: What if instead of writing the config to a temp file, we store it under `/var/lib/microshift/` so users could modify it? The main problem (or advantage...) is that everytime MicroShift is restarted, the config file will be written as it's in the code base. We don't check if file is created, skip creation.

assets/components/kube-apiserver/defaultconfig.yaml
benluddy: This the unmodified static manifest from cluster-kube-apiserver-operator. The effective config passed to kube-apiserver has this disabled. (That should be true for the similar comments too).

assets/components/kube-apiserver/config-overrides.yaml
benluddy: This isn't a real type served by the API, it's only read from disk by OpenShift's patched kube-apiserver. Where are all manifests under assets being applied? I thought that individual asset paths had to explicitly listed for that to happen.

pkg/components/storage.go
benluddy: This path appeared in the list twice, probably due to an earlier mistake resolving a merge conflict, so it's just cleanup here.

pkg/controllers/kube-apiserver.go
benluddy: Fatalf is equivalent to log + `os.Exit(1)`. It's preferable to follow normal control flow here, return an error, and let the service manager gracefully stop all of the components before the process exits.

pkg/controllers/kube-apiserver.go
benluddy: We don't want users to modify (or even know about!) this file. In OpenShift, it's generated dynamically by the target config controller in cluster-kube-apiserver-operator (https://github.com/openshift/cluster-kube-apiserver-operator/blob/65fc794b8aaa36a881323b26618ab46b7ed7defd/pkg/operator/targetconfigcontroller/targetconfigcontroller.go) based on observed configuration.

oglok:
Looks good to me, I'll let one more reviewer to evaluate, and I'll tag it.

pacevedom (approved):

PR #804, opened 2022-07-27
more openshift/kubernetes

fzdarsky:
Now that we've dropped openshift-apiserver, stop vendoring modules as per openshift-apiserver's replace directives, but use o/k modules instead. This also allows us to drop all patches we're carrying still.

[USHIFT-224](https://issues.redhat.com//browse/USHIFT-224)


dhellmann:
Is this based on output of a changed rebase script, or did you do it by hand?

dhellmann:
> Is this based on output of a changed rebase script, or did you do it by hand?

Nevermind, I see this is from #805

go.mod
```diff
@@ -308,33 +308,33 @@ replace (
...
-	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.23.1 // from openshift-apiserver
-	k8s.io/component-base => k8s.io/component-base v0.23.1 // from openshift-apiserver
-	k8s.io/component-helpers => k8s.io/component-helpers v0.23.1 // from openshift-apiserver
-	k8s.io/controller-manager => k8s.io/controller-manager v0.23.1 // from openshift-apiserver
-	k8s.io/cri-api => k8s.io/cri-api v0.23.1 // from openshift-apiserver
-	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.23.1 // from openshift-apiserver
+	k8s.io/api => k8s.io/api v0.23.0 // from kubernetes
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.23.0 // from kubernetes
+	k8s.io/apimachinery => k8s.io/apimachinery v0.23.0 // from kubernetes
+	k8s.io/apiserver => github.com/openshift/kubernetes/staging/src/k8s.io/apiserver v0.0.0-20220510163048-3afdacbd0183 // staging kubernetes
```
benluddy, line 314: I was surprised to see that o/k has inconsistent replacements for the staging modules (e.g. https://github.com/openshift/kubernetes/blob/a6c02f3f66631afd36b8c99e02cf165ee6c4be21/go.mod#L489). Even if this is a bug in o/k (and it seems wrong to me... looking into it), Microshift should use the same replacements as o/k for these modules to avoid drift.

fzdarsky: I'd agree, except for that using the k/k version here doesn't compile and I'm puzzled how this can work in o/k.

For example, the build-time error:
```
# k8s.io/controller-manager/app
vendor/k8s.io/controller-manager/app/serve.go:51:42: too many arguments in call to "k8s.io/apiserver/pkg/server/filters".WithHTTPLogging
	have (http.Handler, nil)
	want (http.Handler)
```
comes from https://github.com/openshift/kubernetes/blob/3afdacbd0183254f82d302d4e8d14da3d311955f/staging/src/k8s.io/controller-manager/app/serve.go#L51 trying to call https://github.com/kubernetes/apiserver/blob/v0.23.0/pkg/server/filters/wrap.go#L63 instead of https://github.com/openshift/kubernetes/blob/3afdacbd0183254f82d302d4e8d14da3d311955f/staging/src/k8s.io/apiserver/pkg/server/filters/wrap.go#L63.

fzdarsky: I should note that k/k has the same inconsistency in replacing modules, but as staging modules and externally mirrored modules are the same it shouldn't matter in that case.

fzdarsky: And so the reason is that o/k (like k/k) ignores the content of go.mod w.r.t. k8s.io and instead has symbolic links in the vendor directory linking to staging. Therefore, we need to ignore those entries in MicroShift, too. Pushed an update that also addresses a few missing pieces.

PR #805, opened 2022-07-27
Add staging directive to rebase script

fzdarsky:
Extension of the rebase script to add the "staging" directive used by PR #804.

[USHIFT-224](https://issues.redhat.com//browse/USHIFT-224)

### DIFF
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -111,6 +111,18 @@ update_modulepath_version_from_release() {
     go mod tidy
 }
 
+# Updates the ReplaceDirective for an old ${modulepath} with the new modulepath
+# in the staging directory of openshift/kubernetes at the released version.
+update_modulepath_to_kubernetes_staging() {
+    local modulepath=$1
+
+    new_modulepath="github.com/openshift/kubernetes/staging/src/${modulepath}"
+    commit=$( cd "${STAGING_DIR}/kubernetes" && git rev-parse HEAD )
+    echo "go mod edit -replace ${modulepath}=${new_modulepath}@${commit}"
+    go mod edit -replace "${modulepath}=${new_modulepath}@${commit}"
+    go mod tidy
+}
+
 # Returns the line (including trailing comment) in the #{gomod_file} containing the ReplaceDirective for ${module_path}
 get_replace_directive() {
     local gomod_file=$1
@@ -183,6 +195,7 @@ valid_component_or_exit() {
 # and this is driven from keywords added as comments after each line of
 # ReplaceDirectives:
 #   // from ${component}     selects the replacement from the go.mod of ${component}
+#   // staging kubernetes    selects the replacement from the staging dir of openshift/kubernetes
 #   // release ${component}  uses the commit of ${component} as specified in the release image
 #   // override [${reason}]  keep existing replacement
 # Note directives without keyword comment are skipped with a warning.
@@ -202,6 +215,9 @@ update_go_mod() {
             component=${arguments%% *}
             valid_component_or_exit "${component}"
             update_modulepath_version_from_component "${modulepath}" "${component}"
+            ;;
+        staging)
+            update_modulepath_to_kubernetes_staging "${modulepath}"
             ;;
         release)
             component=${arguments%% *}
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -349,11 +349,20 @@ update_images() {
         # Compute the max length of image names incl. enclosing quotes
         w=$(awk "BEGIN {n=split(\"${images}\", images, \" \"); max=0; for (i=1;i<=n;i++) {if (length(images[i]) > max) {max=length(images[i])}}; print max+2; exit}")
         for i in ${images}; do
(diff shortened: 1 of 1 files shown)


benluddy:
If we pick up a new indirect dependency on a module from staging, will someone need to remember to add an explicit require directive and "staging" replace directive? It might be simplest to force a `require k8s.io/foo v0.0.0` for all direct and indirect dependencies from staging, then generate a `replace k8s.io/foo v0.0.0 => <replacement from openshift/kubernetes>` for each, instead of making the existing replace directives the source of truth.

scripts/rebase.sh
```diff
@@ -427,7 +443,7 @@ rebase_to() {
         title "## Updating vendor directory"
         go mod vendor
         title "## Patching vendor directory"
-        git apply scripts/rebase_patches/*
+        git apply scripts/rebase_patches/*.patch 2>/dev/null || true
```
dhellmann, line 446: Why hide the errors from this step?

dhellmann:
> If we pick up a new indirect dependency on a module from staging, will someone need to remember to add an explicit require directive and "staging" replace directive? It might be simplest to force a `require k8s.io/foo v0.0.0` for all direct and indirect dependencies from staging, then generate a `replace k8s.io/foo v0.0.0 => <replacement from openshift/kubernetes>` for each, instead of making the existing replace directives the source of truth.

The simplicity of forcing the replacements appeals to me. Are there any downsides?

benluddy:
> Are there any downsides?

I don't think so. It's effectively the same approach, except the source of truth becomes the list of direct and indirect requires instead of the existing replacements.

scripts/rebase.sh
mangelajo: Right, I'd keep those,  otherwise it can be hard to trace

fzdarsky: Soon having patches to apply will be the exception and I was bothered by CI highlighting the expected. But I agree and will instead apply conditionally on whether patches exist.

fzdarsky: done

dhellmann:

This PR doesn't change anything that runs in that job.

scripts/rebase.sh
```diff
@@ -92,6 +95,34 @@ download_release() {
...
+
+# Updates a replace directive using an embedded component's commit.
+# Caches component pseudoversions for faster processing.
+declare -A pseudoversions
+replace_using_component_commit() {
+    local modulepath=$1
+    local new_modulepath=$2
+    local component=$3
+
+    if [ "${pseudoversions[${component}]+foo}" ]; then
```
dhellmann, line 113: This test doesn't look right, there's no comparison. Should it have a `-z` or `-n` flag?

fzdarsky: I agree this shortcut isn't very clear. The empty string basically tests as true with single '['. Updated this to be more [idiomatic](https://mywiki.wooledge.org/BashFAQ/046).

fzdarsky:
Updated the commit to ensure Require- and ReplaceDirectives for all o/k staging repos to fix a few corner cases. Squashed with the previous attempt for easier readbility.

fzdarsky:
Also used the oppy to fix an issue where the rebase script overwrites the MicroShift-specific ovn-kubernetes container image with the one from the OpenShift release image. Once both variants get merged and made available in the release image, this workaround can be removed again.

scripts/rebase.sh
dhellmann: Oh, wow, I'd never seen that before in bash. I guess it's just like testing an empty string in python.

dhellmann:
I checked this out and tried with 4.10.26 and most everything worked except one of the steps close to the end

```
$ ./scripts/rebase.sh manifests
Extracting timestamp
Rebasing manifests
yq: -i/--in-place can only be used with -y/-Y
Script exited with error.
```

I don't know if that's caused by a change in this PR, a change in the release image, or a change in yq.

fzdarsky:
> I'm on RHEL
> 
> ```
> $ yq --version
> yq 2.12.2
> ```

I'm on RHEL8, too, but my `yq` version is vastly different (it may even be we're using different tools):
```
$ yq --version
yq (https://github.com/mikefarah/yq/) version 4.26.1
```
I've installed mine as in https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/microshift/rebase/openshift-microshift-rebase-commands.sh#L18-L27.

dhellmann:
> > I'm on RHEL
> > ```
> > $ yq --version
> > yq 2.12.2
> > ```
> 
> I'm on RHEL8, too, but my `yq` version is vastly different (it may even be we're using different tools):
> 
> ```
> $ yq --version
> yq (https://github.com/mikefarah/yq/) version 4.26.1
> ```
> 
> I've installed mine as in https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/microshift/rebase/openshift-microshift-rebase-commands.sh#L18-L27.

That explains it. I'm sure I just installed mine via yum. I'll replace it with this version and try again.

dhellmann:
Replacing the python version of yq I had with the one mentioned above made this work.

/lgtm

PR #811, opened 2022-07-29
Skip non-yaml files when generating bindata

zshi-redhat:

### DIFF
--- a/scripts/bindata.sh
+++ b/scripts/bindata.sh
@@ -18,10 +18,12 @@ if [ -z "$GOPATH" ]; then
 fi
 
 OUTPUT="pkg/assets/bindata.go"
+IGNORE_REGEXES=".+.tmpl$|.+.sh$"
 "${GOPATH}"/bin/go-bindata \
            $TIMESTAMP_ARGS \
            -nocompress \
            -prefix "pkg/assets" \
            -pkg assets \
+           -ignore "$IGNORE_REGEXES" \
            -o ${OUTPUT} "./assets/..."
 gofmt -s -w "${OUTPUT}"


mangelajo:
/approve
but waiting for @fzdarsky eyes

dhellmann:
What uses those asset files? Should we put them in another directory?

zshi-redhat:
> What uses those asset files? Should we put them in another directory?

These two files are used in rebase.sh script for updating dns daemonset yaml. I think they can be in a different directory, putting them in the same directory as openshift-dns manifest may be easier for maintaining. @fzdarsky  wdyt?

fzdarsky:
> What uses those asset files? Should we put them in another directory?

The rebase script. It copies the `update-node-resolver.sh` from the operator, then renders it into the `daemonset.yaml.tmpl`. That template is currently hardcoded in the operator's code, so we need to store it somewhere.

`/assets` still sounds reasonable, but if we cannot ignore those files when embedding bindata, maybe move to an excluded directory under assets.

dhellmann:
> > What uses those asset files? Should we put them in another directory?
> 
> The rebase script. It copies the `update-node-resolver.sh` from the operator, then renders it into the `daemonset.yaml.tmpl`. That template is currently hardcoded in the operator's code, so we need to store it somewhere.
> 
> `/assets` still sounds reasonable, but if we cannot ignore those files when embedding bindata, maybe move to an excluded directory under assets.

I don't have a problem with the approach in this PR, it just wasn't clear why those files were in the assets directory. 

This looks good to me, but I'll defer to @fzdarsky.

zshi-redhat:
> > What uses those asset files? Should we put them in another directory?
> 
> The rebase script. It copies the `update-node-resolver.sh` from the operator, then renders it into the `daemonset.yaml.tmpl`. That template is currently hardcoded in the operator's code, so we need to store it somewhere.
> 
> `/assets` still sounds reasonable, but if we cannot ignore those files when embedding bindata, maybe move to an excluded directory under assets.

Updated to ignore files with *.sh or *.tmpl suffix when generating bindata.go. (I didn't find a pattern to match negated `*.yaml`, thus using positive match).

PR #812, opened 2022-07-29
upgrade ovn database schema

zshi-redhat:
Closes [USHIFT-153](https://issues.redhat.com//browse/USHIFT-153)

### DIFF
--- a/assets/components/ovn/master/daemonset.yaml
+++ b/assets/components/ovn/master/daemonset.yaml
@@ -169,23 +169,23 @@ spec:
               - |
                 set -x
                 rm -f /var/run/ovn/ovnnb_db.pid
-                  echo "$(date -Iseconds) - nbdb - postStart - waiting for master to be selected"
+                echo "$(date -Iseconds) - nbdb - postStart - waiting for master to be selected"
 
-                  # Upgrade the db if required.
-                  DB_SCHEMA="/usr/share/ovn/ovn-nb.ovsschema"
-                  DB_SERVER="unix:/var/run/ovn/ovnnb_db.sock"
-                  schema_name=$(ovsdb-tool schema-name $DB_SCHEMA)
-                  db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
-                  target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
+                # Upgrade the db if required.
+                DB_SCHEMA="/usr/share/ovn/ovn-nb.ovsschema"
+                DB_SERVER="unix:/var/run/ovn/ovnnb_db.sock"
+                schema_name=$(ovsdb-tool schema-name $DB_SCHEMA)
+                db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
+                target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
 
-                  if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
-                    :
-                  elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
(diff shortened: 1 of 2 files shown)


assets/components/ovn/master/daemonset.yaml
```diff
@@ -169,23 +169,23 @@ spec:
...
-
-                  if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
-                    :
-                  elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
-                      echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), possibly an upgrade is partially complete?"
-                  else
-                      echo "Upgrading database $schema_name from schema version $db_version to $target_version"
-                      ovsdb-client -t 30 convert "$DB_SERVER" "$DB_SCHEMA"
-                  fi
+                echo "$(date -Iseconds) - nbdb - postStart - waiting for master to be selected"
```
dhellmann, line 172: A shell script this complex would be easier to review as a separate file where we would get syntax highlighting and not have the indentation introduced by the YAML. Changing that shouldn't hold up this PR, since it's obviously like that already, but maybe we can move it later on?

assets/components/ovn/master/daemonset.yaml
```diff
@@ -169,23 +169,23 @@ spec:
...
+                DB_SCHEMA="/usr/share/ovn/ovn-nb.ovsschema"
+                DB_SERVER="unix:/var/run/ovn/ovnnb_db.sock"
+                schema_name=$(ovsdb-tool schema-name $DB_SCHEMA)
+                db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
+                target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
+
+                if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
+                  :
+                elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
+                    echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), possibly an upgrade is partially complete?"
```
dhellmann, line 184: This seems like it's likely to happen in a MicroShift instance if a running system is upgraded, that upgrade fails, and then the ostree image is rolled back. In that case we would have older OVN code looking at a newer database. Will that work? Do we need to preserve the old database during an upgrade so we can use it if we have to rollback?

assets/components/ovn/master/daemonset.yaml
zshi-redhat: Yes, it is the same as how ovnk manifests are maintained in openshift (cluster-network-operator). I created the sub-task for later change: https://issues.redhat.com/browse/USHIFT-235

assets/components/ovn/master/daemonset.yaml
zshi-redhat: The ovsdb-client supports simple "upgrades" and "downgrades" on a database's schema as stated in man page. We probably just need to downgrade the schema when target version is lower than current. I will double check with ovn team and update the PR.

zshi-redhat: ovn database shema is automatically converted in standalone mode, which I will add support in ovnk manifests in a separate PR. In the standalone mode, ovn db schema will be tried to be converted (if version mismatch detected) every time when ovn containers start. A backup db file (named $DB_FILE.backup$version-$cksum) will be created before converting is executed.

slack thread discussion: https://coreos.slack.com/archives/C01G7T6SYSD/p1659322380242079

zshi-redhat: PR to switch ovn database mode: https://github.com/openshift/microshift/pull/821

zshi-redhat:
closing in favor of https://github.com/openshift/microshift/pull/821

PR #813, opened 2022-07-29
USHIFT-227: Cluster Policy Controller integration

oglok:

Closes [USHIFT-227](https://issues.redhat.com/browse/USHIFT-227)

This PR carry the following items:

- Enabling Cluster Policy Controller
- Disabling resource-quota and cluster-quota-reconciliation controllers.
- Creating of openshift-kube-controller-manager namespace (it's where the CreatedSCCRanges events happen)
- Applying csr-approver and namespace-security cluster roles and cluster role bindings. ([required](https://github.com/openshift/cluster-policy-controller/blob/release-4.10/pkg/cmd/controller/config.go) by the CPC)


oglok:
I'll test it more intensively on Monday

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,79 @@
...
+		return err
+	}
+	if err := rest.SetKubernetesDefaults(restConfig); err != nil {
+		return err
+	}
+	restConfig.NegotiatedSerializer = serializer.NewCodecFactory(runtime.NewScheme())
+
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+
+	if err := clusterpolicycontroller.WaitForHealthyAPIServer(kubeClient.Discovery().RESTClient()); err != nil {
```
dhellmann, line 68: I notice that [elsewhere](https://github.com/openshift/microshift/blob/main/pkg/controllers/kube-apiserver.go#L245) "health" and "ready" have different meanings and we want to wait for "ready". Is that the case here?

dhellmann (commented):
One question about the wait, but otherwise looks ok.

/assign @benluddy

pkg/controllers/cluster-policy-controller.go
benluddy: It's redundant in this case, because kube-apiserver is already a dependency of cluster-policy-controller in Microshift, so if we're running then kube-apiserver was observed ready.

oglok: Indeed, I didn't realize that the `clusterpolicycontroller.RunClusterPolicyController` already implements a waitForHealthyAPIServer, so I'll clean up this code.

pkg/controllers/infra-services-controller.go
```diff
@@ -41,10 +42,42 @@ func (s *InfrastructureServicesManager) Dependencies() []string {
...
+	}
 	// TO-DO add readiness check
 	if err := components.StartComponents(s.cfg); err != nil {
 		return err
 	}
 	klog.Infof("%s launched ocp componets", s.Name())
 	return ctx.Err()
 }
+
+func ApplyDefaultRBACs(cfg *config.MicroshiftConfig) error {
```
dhellmann, line 58: Does this function need to be public if it's only called from line 46 in this file?

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return err
+	}
+	ctrlConfig := &unstructuredv1.Unstructured{}
+	if err := runtime.DecodeInto(codec, encodedConfig, ctrlConfig); err != nil {
+		return err
+	}
+
+	const namespace = "openshift-kube-controller-manager"
+	builder := controllercmd.NewController(s.Name(), clusterpolicycontroller.RunClusterPolicyController).
```
dhellmann, line 70: Since the builder's validation isn't going to run until Run() is called, would it make sense to just instantiate it in Run() instead of instantiating and storing it?

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,97 @@
...
+		})
+	if err != nil {
+		return err
+	}
+	ctrlConfig := &unstructuredv1.Unstructured{}
+	if err := runtime.DecodeInto(codec, encodedConfig, ctrlConfig); err != nil {
+		return err
+	}
+
+	const namespace = "openshift-kube-controller-manager"
```
dhellmann, line 69: Is this duplicated in the asset file that creates the namespace?

pkg/controllers/cluster-policy-controller.go
```diff
@@ -14,82 +14,84 @@ package controllers
...
+			Controllers: []string{
+				"*",
+				"-openshift.io/resourcequota",
+				"-openshift.io/cluster-quota-reconciliation",
+			},
+		})
+	if err != nil {
+		return err
+	}
+	ctrlConfig := &unstructuredv1.Unstructured{}
```
ingvagabund, line 64: +1 for injecting the config directly without generating a config file

assets/crd/0000_03_securityinternal-openshift_02_rangeallocation.crd.yaml
```diff
@@ -0,0 +1,51 @@
+# 0000_03_securityinternal-openshift_02_rangeallocation.crd.yaml
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.openshift.io: https://github.com/openshift/api/pull/751
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+  name: rangeallocations.security.internal.openshift.io
```
ingvagabund, line 10: Would you more elaborate on why is this CRD needed?

assets/core/0000_10_csr_approver_clusterrole.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
ingvagabund: Would you please extend the commit message clarifying on where all the new assets comes from and why they are needed?

ingvagabund (changes requested):
There's a lot of changes forth and back. I have reviewed the changes commit by commit so my comments in one commit are in most cases obsolete in the following one. Would it be possible to clean the commits first? Making sure only the changes which are needed and agreed on are performed?

ingvagabund:
@oglok would you also please update the PR description with all the necessary steps that were carried? It will be easier to follow the changes and why their were needed.

pkg/controllers/infra-services-controller.go
oglok: this function can be private then. I'll address this.

pkg/controllers/cluster-policy-controller.go
oglok: the asset file will be applied by using bindata. This is just a constant to pass to the controller instantiation.

pkg/controllers/cluster-policy-controller.go
oglok: This is for consistency with the rest of the services. I agree we could instantiate it in the Run(), however, we should make a bit of refactoring after the first drop taking care of all the embedded components, signaling, etc. Do you think we can leave it for a subsequent PR?

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/openshift/api v0.0.0-20220414050251-a83e6f8f1d50
 	github.com/openshift/build-machinery-go v0.0.0-20220121085309-f94edc2d6874
 	github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
+	github.com/openshift/cluster-policy-controller v0.0.0-20220103093815-8e5b36511861
```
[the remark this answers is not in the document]
oglok: This drop of MicroShift is based on OpenShift 4.10.18 code base.

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright © 2021 Microshift Contributors
```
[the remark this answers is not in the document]
oglok: +1

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,79 @@
...
+	s.configure(cfg)
+	return s
+}
+
+func (s *ClusterPolicyController) Name() string           { return "cluster-policy-controller" }
+func (s *ClusterPolicyController) Dependencies() []string { return []string{"kube-apiserver"} }
+
+func (s *ClusterPolicyController) configure(cfg *config.MicroshiftConfig) {
+
+	s.kubeconfig = filepath.Join(cfg.DataDir, "resources", "kubeadmin", "kubeconfig")
```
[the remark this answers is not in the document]
oglok: I agree with this (or make a constant variable). Can you leave it for a following PR?

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,79 @@
...
+	// Use NewControllerCommandConfig to create a controller configuration struct
+	// with default values.
+	s.config = controllercmd.NewControllerCommandConfig(s.Name(), clusterpolicyversion.Get(), clusterpolicycontroller.RunClusterPolicyController)
```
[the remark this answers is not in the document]
oglok: Obsolete

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,79 @@
...
+
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+
+	if err := clusterpolicycontroller.WaitForHealthyAPIServer(kubeClient.Discovery().RESTClient()); err != nil {
+		klog.Fatal(err)
+		return err
+	}
+
+	go func() {
+		s.config.StartController(ctx)
```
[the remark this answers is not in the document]
oglok: Obsolete.

pkg/controllers/cluster-policy-controller.go
```diff
@@ -45,35 +43,53 @@ func (s *ClusterPolicyController) Dependencies() []string { return []string{"kub
...
-		return err
-	}
+	cmd := s.config.NewCommandWithContext(ctx)
+	s.flags.AddFlags(cmd)
+	cmd.Use = s.Name()
+	cmd.Short = "Start the cluster-policy-controller"
 
 	go func() {
-		s.config.StartController(ctx)
+		cli.Run(cmd)
```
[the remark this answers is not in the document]
oglok: Obsolete. We needed to configure the CPC and disable a couple of quota controllers. So the basic configuration provided by `StartController()` didn't work for us.

pkg/controllers/cluster-policy-controller.go
```diff
@@ -14,82 +14,84 @@ package controllers
...
+		// Without an explicit owner reference, the builder will try using POD_NAME or the
+		// first pod in the target namespace (and fail because we have no pod).
+		WithComponentOwnerReference(&corev1.ObjectReference{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "Namespace",
+			Name:       namespace,
+			Namespace:  namespace,
+		})
+
+	s.run = func(ctx context.Context) error {
```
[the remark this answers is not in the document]
oglok: /assign @benluddy

pkg/controllers/cluster-policy-controller.go
oglok: +1. @benluddy 's idea :-)

assets/crd/0000_03_securityinternal-openshift_02_rangeallocation.crd.yaml
oglok: /assign @benluddy

oglok:
> 

Sure, I'll update the PR description.

assets/core/0000_10_csr_approver_clusterrole.yaml
oglok: +1

assets/crd/0000_03_securityinternal-openshift_02_rangeallocation.crd.yaml
benluddy: The namespace SCC allocation controller needs it, context here: https://github.com/openshift/api/pull/751. There's no cluster-config-controller in Microshift to provide it (https://github.com/openshift/cluster-config-operator/blob/0840c6f2f7f25ad9aba1a36597760f36ff0ab097/Dockerfile.rhel7#L13).

pkg/controllers/cluster-policy-controller.go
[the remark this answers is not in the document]
benluddy: Microshift won't invoke Run on this component until kube-apiserver's readiness check has passed because it's a named dependency.

oglok:
> mmit are in most cases obsolete in the following one. Would it be possible to clean the commits first? Making sure only the

PR's commits are squashed and rebased.

pkg/controllers/cluster-policy-controller.go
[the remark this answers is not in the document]
benluddy: :+1: or inject it from above instead of relying on known file paths.

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,97 @@
...
+	s.kubeconfig = filepath.Join(cfg.DataDir, "resources", "kube-controller-manager", "kubeconfig")
+
+	scheme := runtime.NewScheme()
+	openshiftcontrolplanev1.AddToScheme(scheme)
+	codec := serializer.NewCodecFactory(scheme).LegacyCodec(openshiftcontrolplanev1.GroupVersion)
+
+	encodedConfig, err := runtime.Encode(codec,
+		&openshiftcontrolplanev1.OpenShiftControllerManagerConfig{
+			Controllers: []string{
+				"*",
```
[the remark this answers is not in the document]
benluddy: Are you sure? It looks like it should short-circuit here https://github.com/kubernetes/controller-manager/blob/a1e9c0affc6c83f4570ea350cfca8fd3d15b12c3/app/helper.go#L67 to disable specific controllers.

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,97 @@
...
+	return nil
+}
+
+func (s *ClusterPolicyController) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+	defer close(stopped)
+	if s.configErr != nil {
+		return fmt.Errorf("configuration failed: %w", s.configErr)
+	}
+
+	close(ready) // todo
```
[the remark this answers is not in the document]
benluddy: This channel could be wired to a cluster-policy-controller healthz endpoint, which is what OCP uses for a readiness probe, but I don't know how different that signal is from "`(*ControllerBuilder).Run(...)` has not returned yet".

pkg/controllers/infra-services-controller.go
```diff
@@ -41,10 +42,42 @@ func (s *InfrastructureServicesManager) Dependencies() []string {
...
 	// TO-DO add readiness check
 	if err := components.StartComponents(s.cfg); err != nil {
 		return err
 	}
 	klog.Infof("%s launched ocp componets", s.Name())
 	return ctx.Err()
 }
+
+func applyDefaultRBACs(cfg *config.MicroshiftConfig) error {
+	kubeconfigPath := cfg.DataDir + "/resources/kubeadmin/kubeconfig"
```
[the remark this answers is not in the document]
oglok: I'll update this.

assets/core/csr_approver_clusterrole.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
stlaz, line 1: shouldn't these assets be vendored (and possibly symlinked if you need them at this location)?

assets/core/namespace-openshift-infra.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  annotations:
+    openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
+  name: openshift-infra
+  labels:
+    # set value to avoid depending on kube admission that depends on openshift apis
+    openshift.io/run-level: "0"
```
stlaz, line 10: in case it's not vendored, why do we need run-levels here?

assets/core/csr_approver_clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
...
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
+  name: system:openshift:controller:cluster-csr-approver-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:openshift:controller:cluster-csr-approver-controller
+subjects:
+  - kind: ServiceAccount
+    name: cluster-csr-approver-controller
```
stlaz, line 13: This controller signs certificates for Prometheus scraper. Does the platform include running the openshift-monitoring stack?

pkg/controllers/cluster-policy-controller.go
```diff
@@ -0,0 +1,97 @@
...
+	scheme := runtime.NewScheme()
+	openshiftcontrolplanev1.AddToScheme(scheme)
+	codec := serializer.NewCodecFactory(scheme).LegacyCodec(openshiftcontrolplanev1.GroupVersion)
+
+	encodedConfig, err := runtime.Encode(codec,
+		&openshiftcontrolplanev1.OpenShiftControllerManagerConfig{
+			Controllers: []string{
+				"*",
+				"-openshift.io/resourcequota",
+				"-openshift.io/cluster-quota-reconciliation",
```
stlaz, line 58: unless you need certificates for the Prometheus metrics scraper, you'll want to disable `"csr-approver-controller"` too

oglok:
The `e2e-openshift-conformance-sig-storage` is failing. Apparently, there is something in the Cluster Policy Controller preventing the CSI registrar container to work. Investigating.

oglok:
There are two parameters in the node object that are not being written due to this PR (for whatever reason). 

The following annotation:
`      csi.volume.kubernetes.io/nodeid: '{"topolvm.cybozu.com":"maxwell.ogloknet.com"}'`

and the following label:

`      topology.topolvm.cybozu.com/node: maxwell.ogloknet.com`

pkg/controllers/cluster-policy-controller.go
benluddy: @deads2k suggested leaving that controller enabled to allow to optionally enabling Prometheus later.

assets/core/csr_approver_clusterrole.yaml
benluddy: This manifest comes from https://github.com/openshift/cluster-kube-controller-manager-operator/blob/e827c030ba3cd632668820cca73dfcbc986cc5a2/bindata/assets/kube-controller-manager/namespace-openshift-infra.yaml, and Microshift doesn't include operators, so it doesn't have a module dependency on kcm-operator (and isn't present in under vendor).

mangelajo: We have yaml manifests that are auto-imported from our rebase script (so if updates happen to the yamls, the yamls here will get updated). This would need to be one of those as far as I understand.

pkg/controllers/cluster-policy-controller.go
dhellmann: Where are we talking about running prometheus, and why?

pkg/controllers/cluster-policy-controller.go
dhellmann: Yes, consistency is a good reason to do it this way. If we decide to change, we can update all of them.

pkg/controllers/cluster-policy-controller.go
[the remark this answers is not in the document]
dhellmann: Let's save clean-ups like this for a follow-up.

assets/bindata_timestamp.txt
```diff
@@ -1 +1 @@
-1654679854
+1658914160
```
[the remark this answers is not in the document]
oglok: Do we need to update this file manually? I thought it was automatically done using the bindata.sh script.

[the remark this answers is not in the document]
dhellmann: No, sorry, my question wasn't clear. I see that this is changed. I expect that to only happen when the rebase script updates it as part of updating to a new release image, because the timestamp is supposed to come from the release file in the image. So, did this PR include a rebase? If so, maybe we can do that separately from enabling the new controller?

benluddy:
/retest-required

https://search.ci.openshift.org/?search=resource+Deployment%3Aopenshift-storage%2Ftopolvm-controller&maxAge=336h&context=-1&type=build-log&name=.*-microshift-.*&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

PR #815, opened 2022-08-01
Disable CGO when building MicroShift binary

zshi-redhat:
Disabling CGO would remove runtime dependency
of glibc and reduce the MicroShift binary size

Co-authored-by: Miguel Angel Ajo Pelayo <majopela@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,7 @@ GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi 
 GO_TEST_FLAGS=$(GO_BUILD_FLAGS)
 GO_TEST_PACKAGES=./cmd/... ./pkg/...
 
+# targets "all:" and "build:" defined in vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk
 # Disable CGO when building microshift binary
 all: export CGO_ENABLED=0
 
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,11 @@ GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi 
 GO_TEST_FLAGS=$(GO_BUILD_FLAGS)
 GO_TEST_PACKAGES=./cmd/... ./pkg/...
 
-# targets "all:" and "build:" defined in vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk
+# Disable CGO when building microshift binary
+all: export CGO_ENABLED=0
+
+build: export CGO_ENABLED=0
+
 microshift: build
 
 microshift-aio: build-containerized-all-in-one-amd64


dhellmann:
/approve

I'm interested in why including cgo is the default, and whether we were using it for anything. It seems like we're confident we don't need it now, if we ever did. 

I'll leave final approval to someone more familiar with the build.

fzdarsky:
It seems we're still using CGO in several vendored modules (according to `grep -R "import \"C\"" vendor | grep -vE "aix|bsd|dragonfly|plan9|solaris|windows"`). Is it used for anything else than mDNS?

Also, which functionality do we lose for mDNS?
- ability to announce routes to on-cluster services to clients?
- ability for MicroShift to discover external services announced via mDNS?
- ability for on-cluster services to discover external services announced via mDNS?

mangelajo:
> * ability for MicroShift to discover external services announced via mDNS?
> * ability for on-cluster services to discover external services announced via mDNS?

I believe those two:
    * ability for MicroShift to discover external services announced via mDNS? 
    * ability for on-cluster services to discover external services announced via mDNS?

Which, may be is a good reason to keep it (multiple microshifts on the same LAN)


> @dhellmann 
> /approve
> 
> I'm interested in why including cgo is the default, and whether we were using it for anything. It seems like we're confident we don't need it now, if we ever did.
> 
> I'll leave final approval to someone more familiar with the build.

I wonder if we keep cgo in openshift because it's just a single audit point (the audit and CVEs of the system libraries) instead of the audit + CVEs of go libraries + system libraries.

mangelajo:
Based on the discussion above I propose we:
/hold

zshi-redhat:
> It seems we're still using CGO in several vendored modules (according to `grep -R "import \"C\"" vendor | grep -vE "aix|bsd|dragonfly|plan9|solaris|windows"`). Is it used for anything else than mDNS?
> 
If any of these cgo functions are used, build shall fail with CGO_ENABLED=0, right?
I think the fact that it can build successfully with CGO disabled could mean no cgo functions are linked in the build stage.

Makefile
```diff
@@ -93,7 +93,11 @@ GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi
 GO_TEST_FLAGS=$(GO_BUILD_FLAGS)
 GO_TEST_PACKAGES=./cmd/... ./pkg/...
 
-# targets "all:" and "build:" defined in vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk
```
mangelajo, line 96: let's not remove the above line, it's an useful comment

Makefile
```diff
@@ -93,7 +93,11 @@ GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi
...
+# Disable CGO when building microshift binary
```
mangelajo, line 96:
```suggestion
# targets "all:" and "build:" defined in vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk
# Disable CGO when building microshift binary
```

mangelajo (changes requested):

Makefile
zshi-redhat: Done, accepted the suggested change.

mangelajo:
@ggiguash @oglok @copejon 

I believe this may be safe and it's worth trying.

ggiguash:
Searching for `import "C"` directive in the vendor directory produces the following results. Here's the analysis of these packages:
- [Container Advisor](https://github.com/google/cadvisor)
- [Go NVML Bindings](https://github.com/mindprince/gonvml)
- [Go Wrapper for GPGME library](https://github.com/mtrmac/gpgme)
- [runc API](https://github.com/opencontainers/runc/)
- [libseccomp interface](https://github.com/seccomp/libseccomp-golang)
- [Go unix API](https://pkg.go.dev/golang.org/x/sys/unix)
- [Go AST utilities](https://pkg.go.dev/golang.org/x/tools@v0.1.12/go/ast/astutil)

```
$ grep -nR "import \"C\"" vendor | grep -vE "aix|bsd|dragonfly|plan9|solaris|windows"
vendor/github.com/google/cadvisor/nvm/machine_libipmctl.go:22:import "C"
vendor/github.com/google/cadvisor/perf/collector_libpfm.go:26:import "C"
vendor/github.com/google/cadvisor/perf/types_libpfm.go:21:import "C"
vendor/github.com/google/cadvisor/perf/uncore_libpfm.go:25:import "C"
vendor/github.com/mindprince/gonvml/bindings.go:310:import "C"
vendor/github.com/mtrmac/gpgme/data.go:7:import "C"
vendor/github.com/mtrmac/gpgme/gpgme.go:9:import "C"
vendor/github.com/mtrmac/gpgme/unset_agent_info.go:6:import "C"
vendor/github.com/opencontainers/runc/libcontainer/seccomp/patchbpf/enosys_linux.go:64:import "C"
vendor/github.com/seccomp/libseccomp-golang/seccomp.go:26:import "C"
vendor/github.com/seccomp/libseccomp-golang/seccomp_internal.go:183:import "C"
vendor/golang.org/x/sys/unix/mkerrors.sh:442:   echo 'import "C"'
vendor/golang.org/x/tools/go/ast/astutil/imports.go:59:                 // Do not add to import "C", to avoid disrupting the
```

ggiguash:
The executable size reduction is 150MB vs 115MB when CGO is disabled.

MicroShift executable build with CGO support
```
$ ldd microshift 
        linux-vdso.so.1 (0x00007fff529f7000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fb7e7fcb000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fb7e7dab000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fb7e79e6000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fb7e81cf000)
$ ls -l microshift 
-rwxrwxr-x. 1 microshift microshift 151104128 Aug  7 08:42 microshift
```

MicroShift executable built without CGO support
```
$ ldd microshift 
        not a dynamic executable
$ ls -l microshift 
-rwxrwxr-x. 1 microshift microshift 119431168 Aug  8 01:33 microshift
```

ggiguash:
There is a gain of ~2GB in the process VIRT size when compiled without CGO support. This is due to the MicroShift executable being statically linked.
The size of the resident memory is slightly higher for CGO process, but this is inconclusive and should not be used as an argument for decision making.

CGO disabled
```top - 01:47:31 up  1:32,  0 users,  load average: 4.45, 5.46, 3.99
Tasks:   1 total,   0 running,   1 sleeping,   0 stopped,   0 zombie
%Cpu(s):  4.1 us,  3.8 sy,  0.0 ni, 89.2 id,  0.3 wa,  0.9 hi,  0.9 si,  0.8 st
MiB Mem :   3731.3 total,    250.3 free,   2217.9 used,   1263.1 buff/cache
MiB Swap:   3052.0 total,   2199.8 free,    852.2 used.   1235.4 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND           
  43317 root      20   0   11.1g 459192  85900 S  11.0  12.0   0:18.01 microshift        
```
CGO enabled
```op - 01:57:51 up  1:43,  0 users,  load average: 1.09, 4.40, 4.66
Tasks:   1 total,   0 running,   1 sleeping,   0 stopped,   0 zombie
%Cpu(s):  4.6 us,  2.7 sy,  0.0 ni, 91.4 id,  0.3 wa,  0.6 hi,  0.3 si,  0.1 st
MiB Mem :   3731.3 total,    130.2 free,   2181.7 used,   1419.4 buff/cache
MiB Swap:   3052.0 total,   1902.0 free,   1149.9 used.   1262.5 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND           
  65496 root      20   0   13.1g 462588 109944 S  13.6  12.1   0:38.40 microshift        
```

ggiguash:
Based on the above analysis, I tend to agree that we should try disabling CGO unless it becomes absolutely necessary to have it.
USHIFT-255
/lgtm

PR #816, opened 2022-08-01
Fix openvswitch issues when MicroShift service is disabled

zshi-redhat:
Closes [USHIFT-236](https://issues.redhat.com//browse/USHIFT-236)

### DIFF
--- a/hack/cleanup.sh
+++ b/hack/cleanup.sh
@@ -10,8 +10,8 @@ select yn in "Yes" "No"; do
 done
 
 # crictl redirect STDOUT.  When no objects (pod, image, container) are present, crictl dump the help menu instead.  This may be confusing to users.
-sudo bash -c '
-    echo "Stopping microshift"
+sudo bash -c "
+    echo Stopping MicroShift
     set +e
     systemctl stop --now microshift 2>/dev/null
     systemctl disable microshift 2>/dev/null
@@ -20,30 +20,36 @@ sudo bash -c '
     podman stop microshift 2>/dev/null
     podman stop microshift-aio 2>/dev/null
 
-    echo "Removing crio pods"
+    echo Removing non-OVN crio pods
+    
+    crictl pods | tail -n +2 | grep -vE openshift-ovn-kubernetes | awk '{print \$1}' | xargs crictl stopp
+    crictl pods | tail -n +2 | grep -vE openshift-ovn-kubernetes | awk '{print \$1}' | xargs crictl rmp
+    
+
+    echo Removing all crio pods
     until crictl rmp --all --force 1>/dev/null; do sleep 1; done
 
-    echo "Removing crio containers"
+    echo Removing crio containers
     crictl rm --all --force 1>/dev/null
 
-    echo "Removing crio images"
+    echo Removing crio images
     crictl rmi --all --prune 1>/dev/null
 
-    echo "Killing conmon, pause processes"
+    echo Killing conmon, pause processes
     pkill -9 conmon
     pkill -9 pause
     pkill -9 ovn-controller
     pkill -9 ovn-northd
     pkill -9 ovsdb-server
 
-    echo "Removing /var/lib/microshift"
+    echo Removing /var/lib/microshift
     crio wipe -f
     systemctl restart crio
-    echo "Reverting microshift-ovs-init.service configuration"
+    echo Reverting microshift-ovs-init.service configuration
     /usr/bin/configure-ovs.sh OpenShiftSDN
     rm -rf /var/lib/microshift
     rm -rf /var/lib/ovn
     rm -rf /var/run/ovn
 
-    echo "Cleanup succeeded"
-'
+    echo Cleanup succeeded
+"
--- a/hack/cleanup.sh
+++ b/hack/cleanup.sh
@@ -39,10 +39,11 @@ sudo bash -c '
     echo "Removing /var/lib/microshift"
     crio wipe -f
     systemctl restart crio
+    echo "Reverting microshift-ovs-init.service configuration"
+    /usr/bin/configure-ovs.sh OpenShiftSDN
     rm -rf /var/lib/microshift
     rm -rf /var/lib/ovn
     rm -rf /var/run/ovn
-    rm -rf /var/run/openvswitch
 
     echo "Cleanup succeeded"
 '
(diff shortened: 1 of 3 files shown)


ggiguash:
I think we should also enable openvswitch service in the blueprint:
- https://github.com/openshift/microshift/blob/879bbf1921b68fba5b2b18c4813bd4b485f84674/scripts/image-builder/config/blueprint_v0.0.1.toml#L47
-

zshi-redhat:
> I think we should also enable openvswitch service in the blueprint:
> 
> * https://github.com/openshift/microshift/blob/879bbf1921b68fba5b2b18c4813bd4b485f84674/scripts/image-builder/config/blueprint_v0.0.1.toml#L47

Enabled both openvswitch and NetworkManager in blueprint.

hack/cleanup.sh
```diff
@@ -39,10 +39,11 @@ sudo bash -c '
     echo "Removing /var/lib/microshift"
     crio wipe -f
     systemctl restart crio
+    echo "Reverting microshift-ovs-init.service configuration"
```
ggiguash, line 42: Do we need to still pkill the ovn processes above?

packaging/rpm/microshift.spec
```diff
@@ -217,6 +217,7 @@ fi
 sed -i -n -e '/^OPTIONS=/!p' -e '$aOPTIONS="--no-mlockall"' /etc/sysconfig/openvswitch
 %systemd_post microshift-ovs-init.service
 systemctl is-active --quiet NetworkManager && systemctl restart --quiet NetworkManager || true
+systemctl is-active --quiet openvswitch && systemctl restart --quiet openvswitch || true
```
ggiguash, line 220: Should we enable the services here instead of the blueprint?

ggiguash: We can add the following
`systemctl is-enabled --quiet openvswitch && systemctl enable --quiet openvswitch`

ggiguash (changes requested):

oglok:
Have we tested this and make sure it works? It does not in my local env.

hack/cleanup.sh
```diff
@@ -10,40 +10,47 @@ select yn in "Yes" "No"; do
...
-    systemctl restart crio
-    echo "Reverting microshift-ovs-init.service configuration"
-    /usr/bin/configure-ovs.sh OpenShiftSDN
-    rm -rf /var/lib/microshift
-    rm -rf /var/lib/ovn
-    rm -rf /var/run/ovn
-
-    echo "Cleanup succeeded"
-'
+echo "Stopping microshift"
```
ggiguash, line 13: Why did you remove "sudo"? I think we should keep it because it's mentioned without sudo in all the docs, etc.

hack/cleanup.sh
```diff
@@ -10,40 +10,47 @@ select yn in "Yes" "No"; do
...
+set +e
+systemctl stop --now microshift 2>/dev/null
+systemctl disable microshift 2>/dev/null
+systemctl stop --now microshift-containerized 2>/dev/null
+systemctl disable microshift-containerized 2>/dev/null
+podman stop microshift 2>/dev/null
+podman stop microshift-aio 2>/dev/null
+
+echo "Removing crio pods in order"
+NON_OVN_PODS=$(crictl pods | grep -vE ovn | grep -v CREATED | awk '{print $1}');
```
ggiguash, line 23: Instead of "grep -v CREATED" -> tail -n +2

hack/cleanup.sh
```diff
@@ -10,40 +10,47 @@ select yn in "Yes" "No"; do
...
+echo "Stopping microshift"
+set +e
+systemctl stop --now microshift 2>/dev/null
+systemctl disable microshift 2>/dev/null
+systemctl stop --now microshift-containerized 2>/dev/null
+systemctl disable microshift-containerized 2>/dev/null
+podman stop microshift 2>/dev/null
+podman stop microshift-aio 2>/dev/null
+
+echo "Removing crio pods in order"
```
ggiguash, line 22: What does the comment mean, sorry?

hack/cleanup.sh
ggiguash: I think this expression may be bit more selective. I'm a bit concerned with "ovn" pattern.

crictl pods | tail -n +2 | awk '{if ($7~/openshift-ovn-kubernetes/) next; print}' | awk '{print $1}'

ggiguash (changes requested):

hack/cleanup.sh
```diff
@@ -10,40 +10,47 @@ select yn in "Yes" "No"; do
...
+NON_OVN_PODS=$(crictl pods | grep -vE ovn | grep -v CREATED | awk '{print $1}');
+echo $NON_OVN_PODS
```
oglok, line 24: Remove this.

packaging/rpm/microshift.spec
zshi-redhat: Done

ggiguash (approved):

packaging/rpm/microshift.spec
```diff
@@ -217,6 +217,8 @@ fi
 sed -i -n -e '/^OPTIONS=/!p' -e '$aOPTIONS="--no-mlockall"' /etc/sysconfig/openvswitch
 %systemd_post microshift-ovs-init.service
 systemctl is-active --quiet NetworkManager && systemctl restart --quiet NetworkManager || true
+systemctl is-active --quiet openvswitch && systemctl start --quiet openvswitch || true
+systemctl is-enabled --quiet openvswitch && systemctl enable --quiet openvswitch || true
```
mangelajo, line 221:
```suggestion
systemctl enable --now --quiet openvswitch || true
```

packaging/systemd/configure-ovs.sh
```diff
@@ -4,8 +4,6 @@
 
 set -eux
 
-nmcli conn show br-ex >/dev/null && echo "br-ex already configured" && exit 0
```
mangelajo, line 7: Why are we disabling this now?

zshi-redhat: I think the rollback or reconfiguration issue was solved by restarting NM after installing NM-ovs.

zshi-redhat: In the cleanup script, it reverts the init ovs configuration by calling `configure.sh OpenShiftSDN` which doesn't work if we have the br-ex existense check.

mangelajo: Ahhh ok I missed that detail

mangelajo (approved):
/lgtm

mangelajo:
Oops I was able to merge it instead of the bot. Do we have a permission issue?

PR #818, opened 2022-08-01
Tolerate any taints in ovnkube-master daemonset

pacevedom:
Closes [USHIFT-237](https://issues.redhat.com/browse/USHIFT-237)

### DIFF
--- a/pkg/assets/bindata.go
+++ b/pkg/assets/bindata.go
@@ -171,7 +171,7 @@ func assetsComponentsOdfLvmCsiDriverYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "assets/components/odf-lvm/csi-driver.yaml", size: 247, mode: os.FileMode(420), modTime: time.Unix(1654679854, 0)}
+	info := bindataFileInfo{name: "assets/components/odf-lvm/csi-driver.yaml", size: 247, mode: os.FileMode(436), modTime: time.Unix(1654679854, 0)}
 	a := &asset{bytes: bytes, info: info}
 	return a, nil
 }
@@ -330,7 +330,7 @@ func assetsComponentsOdfLvmTopolvmController_deploymentYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "assets/components/odf-lvm/topolvm-controller_deployment.yaml", size: 4590, mode: os.FileMode(420), modTime: time.Unix(1654679854, 0)}
+	info := bindataFileInfo{name: "assets/components/odf-lvm/topolvm-controller_deployment.yaml", size: 4590, mode: os.FileMode(436), modTime: time.Unix(1654679854, 0)}
 	a := &asset{bytes: bytes, info: info}
 	return a, nil
 }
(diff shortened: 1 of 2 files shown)


zshi-redhat:
LGTM

pkg/assets/bindata.go
```diff
@@ -171,7 +171,7 @@ func assetsComponentsOdfLvmCsiDriverYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "assets/components/odf-lvm/csi-driver.yaml", size: 247, mode: os.FileMode(420), modTime: time.Unix(1654679854, 0)}
+	info := bindataFileInfo{name: "assets/components/odf-lvm/csi-driver.yaml", size: 247, mode: os.FileMode(436), modTime: time.Unix(1654679854, 0)}
```
ggiguash, line 174: Where all these "weird" 436 file permissions are coming from? Can we fix those to something more standard like 644, etc.?

ggiguash (changes requested):

pkg/assets/bindata.go
zshi-redhat: I think it is auto generated by go-bindata, I could reproduce the same changes locally.
It was using 436 for some of those generated manifests before the storage patch merging. for example: `assets/components/openshift-router/deployment.yaml` was changed in commit `remove hostpath provisioner manifests, bindata`

pacevedom: The binary offers some flags to override file permissions:
```
$ go-bindata --help
Usage: go-bindata [options] <input directories>
...
  -mode uint
    	Optional file mode override for all files.
```
However, the Makefile does not use them (https://github.com/openshift/microshift/blob/main/scripts/bindata.sh#L21), so it defaults to whatever the tool has configured.

dhellmann: > Where all these "weird" 436 file permissions are coming from? Can we fix those to something more standard like 644, etc.?

436 decimal is 644 octal.

ggiguash: Ah, that's a good point @dhellmann! No problem then.

mangelajo (commented):
The daemonset changes look good, I wonder why all the perm changes in bindata, and how could we fix that @fzdarsky  ?  ^

pkg/assets/bindata.go
dhellmann: It would be nice to know why those keep getting toggled back and forth between different permissions, though.

dhellmann (commented):
/approve

pkg/assets/bindata.go
zshi-redhat: The fileMode change may be related to system umask setting since git only tracks the execution bit (the following commit is from git repo). The file mode of resulting git tree could be changed based on system umask setting:

```
commit e44794706eeb57f2ee38ed1604821aa38b8ad9d2
Author: Linus Torvalds <torvalds@linux-foundation.org>
Date:   Sat Apr 16 22:26:31 2005 -0700

    Be much more liberal about the file mode bits.
    
    We only really care about the difference between a file being executable
    or not (by its owner). Everything else we leave for the user umask to
    decide.
```

When system umask is 0022, file mode is 644:
```
# umask 
0022
# ls -al assets/components/openshift-dns/dns/daemonset.yaml 
-rw-r--r--. 1 root root 3217 Aug  1 23:45 assets/components/openshift-dns/dns/daemonset.yaml
```
When umask is 0002, file mode is 664:
```
# umask 
0002
# ls -al assets/components/openshift-dns/node-resolver/daemonset.yaml
-rw-rw-r--. 1 zshi zshi 4823 May 27 14:14 assets/components/openshift-dns/node-resolver/daemonset.yaml
```

Maybe we shall agree on the umask setting while generating microshift bindata?

dhellmann: I filed #824 to try to address the permission issue

mangelajo:
it LGTM but needs a rebase

mangelajo:
I made a rebased version here: https://github.com/openshift/microshift/pull/843

PR #821, opened 2022-08-02
Switch to ovn standalone mode

zshi-redhat:
Get rid of cluster operations of ovn database by
switching from cluster to standalone mode.

The transaction processing steps are much simpler
in standalone mode compared with cluster mode.

Following is the sequence of actions for transaction execution in a standalone database: # Receive a transaction request.

    Execute the transaction on in-memory data.
    Write resulted changes on disk.
    Reply to a client.

And the same thing in a single-node cluster: # Receive a transaction request.

    Evaluate the transaction (same as execute, but changes are not yet committed).
    Propose resulted changes to a raft storage module.
    Update raft log with new changes.
    Write changes on disk.
    Mark proposed changes as applied.
    Read last applied change from raft storage.
    Re-evaluate changes and commit (basically, execute the transaction again).
    Reply to a client.

In standalone mode, ovn database schema is evaluated to
upgrade/downgrade every time when starting ovn database
using ovn-ctl script, which removes the need to add upgrade
logic in postStart hook of ovnk database containers.

Closes [USHIFT-153](https://issues.redhat.com//browse/USHIFT-153)

### DIFF
--- a/assets/components/ovn/master/daemonset.yaml
+++ b/assets/components/ovn/master/daemonset.yaml
@@ -1,6 +1,4 @@
 ---
-# The ovnkube control-plane components
-
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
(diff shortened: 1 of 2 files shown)


zshi-redhat:
Rebased to take the asset permission change.

mangelajo (commented):
/lgtm

PR #822, opened 2022-08-02
pkg/sysconfwatch: add non-linux variant to allow microshift binary to build

mfojtik:
The binary itself might not run as expected on Apple or other non-linux
platforms, but we should avoid adding linux specific code without
alternatives for other systems/platforms.

As a bonus, this allows building microshift binary on M1 MacBook, improving development workflow/experience for some contributors without emphasizing that microshift is supported running on non-linux platforms.

### DIFF
--- a/pkg/sysconfwatch/sysconfwatch.go
+++ b/pkg/sysconfwatch/sysconfwatch.go
@@ -1,135 +1,38 @@
-/*
-Copyright © 2022 Microshift Contributors
-
-Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


mangelajo:
Sounds ok to me, but what's the advantage of compiling in OSX if it's really unable to run?, at least you can check it compiles I guess..

mfojtik:
@mangelajo it runs, at least control-plane role, which is enough to debug and develop microshift itself for some people (or enough to write e2e tests against it, etc.).

PR #824, opened 2022-08-02
USHIFT-241: Asset permissions

dhellmann:
Address the permission issue noticed in #818

### DIFF
--- a/scripts/bindata.sh
+++ b/scripts/bindata.sh
@@ -17,6 +17,11 @@ if [ -z "$GOPATH" ]; then
     export GOPATH=$(go env GOPATH)
 fi
 
+# Ensure the assets have the expected permissions before we package
+# them, in case the user's umask changes them from the desired
+# settings.
+find ./assets -name '*.yaml' | xargs chmod 0644
+
 OUTPUT="pkg/assets/bindata.go"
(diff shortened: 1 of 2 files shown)


scripts/bindata.sh
```diff
@@ -17,6 +17,11 @@ if [ -z "$GOPATH" ]; then
     export GOPATH=$(go env GOPATH)
 fi
 
+# Ensure the assets have the expected permissions before we package
+# them, in case the user's umask changes them from the desired
+# settings.
+find ./assets -name '*.yaml' | xargs chmod 0644
```
ggiguash, line 23: Should we just update the umask instead of touching permissions on the file system?

dhellmann: The files are already checked out by the time this script runs, so I don't think just changing the umask will help, will it?

ggiguash: Ah, I thought umask is applied on permissions by bindata. If bindata "copies" permissions from the file system, not much we can do except resetting those.

ggiguash (approved):

scripts/bindata.sh
dhellmann: Yeah, the umask comes into account when the user checks the files out with git, but bindata takes what it reads from the filesystem.

PR #830, opened 2022-08-02
Ushift-230: CSI image templating

copejon:

Closes [USHIFT-230](https://issues.redhat.com//browse/USHIFT-230)


mangelajo:
@copejon the bindata seems to need a final update: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/830/pull-ci-openshift-microshift-main-verify/1554617697977765888#1:build-log.txt%3A24

assets/components/odf-lvm/topolvm-controller_deployment.yaml
```diff
@@ -26,7 +26,7 @@ spec:
       - command:
         - /topolvm-controller
         - --cert-dir=/certs
-        image: registry.redhat.io/odf4/odf-topolvm-rhel8@sha256:bd9fb330fc35f88fae65f1598b802923c8a9716eeec8432bdf05d16bd4eced64
+        image: {{ .ReleaseImage.odf_topolvm_rhel8 }}
```
fzdarsky, line 29: Why does the image name hardcode an OS version (`_rhel8`)? Is the intention to have different images for RHEL8 and RHEL9 eventually?

pkg/release/release.go
```diff
@@ -19,11 +19,17 @@ package release
...
-	"service_ca_operator": "quay.io/microshift/service-ca-operator:" + Base,
-	"ovn_kubernetes":      "quay.io/microshift/ovn-kubernetes-singlenode:" + Base,
+	"cli":                     "quay.io/microshift/cli:" + Base,
+	"coredns":                 "quay.io/microshift/coredns:" + Base,
+	"haproxy_router":          "quay.io/microshift/haproxy-router:" + Base,
+	"kube_rbac_proxy":         "quay.io/microshift/kube-rbac-proxy:" + Base,
+	"pause":                   "quay.io/microshift/pause:" + Base,
+	"service_ca_operator":     "quay.io/microshift/service-ca-operator:" + Base,
+	"ovn_kubernetes":          "quay.io/microshift/ovn-kubernetes-singlenode:" + Base,
+	"odf_topolvm_rhel8":       "registry.redhat.io/odf4/odf-topolvm-rhel8@sha256:bd9fb330fc35f88fae65f1598b802923c8a9716eeec8432bdf05d16bd4eced64",
```
fzdarsky, line 29: This file is the fall-back for "community-supported architectures" where RH doesn't have official images.
It looks like the digests are the same as for the `amd64` architecture. Therefore, this isn't ever going to be correct.

I suggest either adopting the same approach as for the images above (`"quay.io/microshift/..." + Base`) even if those don't exist for now. Eventually, me may want to consider removing the `release.go` fallback altogether.

assets/components/odf-lvm/topolvm-controller_deployment.yaml
copejon: I can't speak to why ODF embedded the OS version in the image name, but I'll drop it from field.

pkg/release/release.go
copejon: Changed the value to reflect the public access image repos, which don't yet exist.

dhellmann:
/hold

Both of the failing CI jobs have the messages that look related to TopoLVM:

```
case.go:256: resource DaemonSet:openshift-storage/topolvm-node: .status.numberAvailable: key is missing from map
```

and

```
case.go:256: resource Deployment:openshift-storage/topolvm-controller: .status.availableReplicas: key is missing from map
```

PR #832, opened 2022-08-04
Add required firewall configuration in post-networking RPM stage

ggiguash:
Test when firewalld is DOWN (sudo -i)
```
# systemctl is-active firewalld.service 
inactive
# firewall-offline-cmd --zone=trusted --list-sources

# systemctl is-active --quiet firewalld || firewall-offline-cmd -q --zone=trusted --add-source=10.42.0.0/16
# systemctl is-active --quiet firewalld && firewall-cmd         -q --zone=trusted --add-source=10.42.0.0/16
# firewall-offline-cmd --zone=trusted --list-sources
10.42.0.0/16
```
Test when firewalld is UP (sudo -i)
```
# systemctl is-active firewalld.service 
active
# firewall-cmd --zone=trusted --list-sources

# systemctl is-active --quiet firewalld || firewall-offline-cmd -q --zone=trusted --add-source=10.42.0.0/16
# systemctl is-active --quiet firewalld && firewall-cmd         -q --zone=trusted --add-source=10.42.0.0/16
# firewall-cmd --zone=trusted --list-sources
10.42.0.0/16
```

Closes [USHIFT-245](https://issues.redhat.com//browse/USHIFT-245)

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -218,6 +218,9 @@ sed -i -n -e '/^OPTIONS=/!p' -e '$aOPTIONS="--no-mlockall"' /etc/sysconfig/openv
 %systemd_post microshift-ovs-init.service
 systemctl is-active --quiet NetworkManager && systemctl restart --quiet NetworkManager || true
 systemctl enable --now --quiet openvswitch || true
+# configure the firewall rules for pods to intercommunicate
+systemctl is-active --quiet firewalld || firewall-offline-cmd -q --zone=trusted --add-source=10.42.0.0/16
+systemctl is-active --quiet firewalld && firewall-cmd         -q --zone=trusted --add-source=10.42.0.0/16
 
 %preun networking
 %systemd_preun microshift-ovs-init.service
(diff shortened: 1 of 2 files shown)


packaging/rpm/microshift.spec
```diff
@@ -218,6 +218,9 @@ sed -i -n -e '/^OPTIONS=/!p' -e '$aOPTIONS="--no-mlockall"' /etc/sysconfig/openv
 %systemd_post microshift-ovs-init.service
 systemctl is-active --quiet NetworkManager && systemctl restart --quiet NetworkManager || true
 systemctl enable --now --quiet openvswitch || true
+# configure the firewall rules for pods to intercommunicate
+systemctl is-active --quiet firewalld || firewall-offline-cmd -q --zone=trusted --add-source=10.42.0.0/16
```
zshi-redhat, line 222: Should we expect customized cluster IP range?

ggiguash: Can we extract this information from somewhere?

fzdarsky:
@ggiguash, so far we've deliberately avoided making firewall and similar security-impacting changes to the OS "under the hood" by the RPM installation scripts (or MicroShift FTM). The reason is we thought it a best-practice that sys admins should consciously and selectively open up their systems.

To @zshi-redhat's point: The problem is the `ClusterCIDR` is a runtime configuration parameters (https://github.com/openshift/microshift/blob/main/pkg/config/config.go#L97) and therefore there's no way to reliably add a rule at installation-time.

I understand the intention is to eliminate a common source of error; would it maybe make more sense to add some form of validation at MicroShift startup instead? Like "ERROR: You've configured a ClusterCIDR of 10.42.0.0/16, but didn't open the firewall ports."

PR #833, opened 2022-08-04
fix mdns unqulified name

zshi-redhat:
Recorded resolver name doesn't match mdns query name.

Assuming local hostname is "localhost.local", resolver
records domain name as "localhost.local", but a mdns
client would send query with name "localhost.local."
which results in no response from mdns server.


mangelajo (commented):
/lgtm

PR #840, opened 2022-08-05
Use the arm64 image for OVNk in release_arm64.go

mangelajo:
Enables ovnk in arm64


copejon:
test reports success, but tide doesn't seem to register that.  rerunning all

/retest

PR #847, opened 2022-08-09
PERFSCALE-1818: Initial commit of ansible-microshift automation

sjug:
- Pre-localhost setup and logging setup
- Add cadvisor to microshift node, run as service
- Fix systemd_service var and apply firewall rules sooner
- Manage prometheus config via template
- Need to restart cadvisor after CRI-O starts in order to pick up the CRI-O container metrics
- Split out EPEL repo install and add pbench agent tasks
- Add pbench to README & pbench_env to vars
- Set up pbench and kubectl
- Make pbench-user-benchmark async
- Updated inventory vars
- Move API checks to remote node and automate pbench data collection
- Write three stages of disk usage (pre-install, post-install, post-run) to file
- Modify timing tasks: reorder and small fixes
- New disk usage pattern + other small changes
- Add warm start measurement
- Rename install-localhost role to setup-localhost for clarity
- Remove microshift state, unused local cluster state tracking
- Move into ansible subdir
- Add something to README.md

This PR adds automation to measure the footprint and performance metrics of MicroShift via Ansible scripts.


dhellmann:
Let's link this to the Jira ticket

/retitle PERFSCALE-1818: Initial commit of ansible-microshift automation

ansible/README.md
```diff
@@ -0,0 +1,34 @@
+# ansible-microshift
+
+The purpose of this ansible automation is to help gain insight into the footprint and start times of MicroShift.
+At this stage we are capturing the first start as well as the second start (with cached images), as well as disk
```
ggiguash, line 4: Should we replace "as well" by a comma in the sentence?

ansible/README.md
```diff
@@ -0,0 +1,34 @@
...
+usage data from a number of points in time as well as performance tool data with pbench for each of the two starts.
+
+## Prereqs
+
+We are using [pbench](https://github.com/distributed-system-analysis/pbench) to capture a wide array of performance tool data.
+Pbench has Ansible playbooks available in Ansible Galaxy to install the pbench-agent on the MicroShift host.
+
+- Ansible
+  - `ansible-galaxy collection install pbench.agent`
```
ggiguash, line 13: Need instructions on installing ansible-galaxy as it's not available in standard repos.

ansible/inventory/inventory
```diff
@@ -0,0 +1,20 @@
+[microshift]
+microshift.example.com ansible_host= private_ip= # microshift node
+
+[microshift:vars]
+ansible_user=centos
```
ggiguash, line 5: Can we use `microshift` user here?

ansible/inventory/inventory
```diff
@@ -0,0 +1,20 @@
...
+ansible_become=yes
+pbench_key_url=
+pbench_config_url=
+
+[logging]
+logging.example.com ansible_host= private_ip= # logging node
+
+[logging:vars]
+ansible_user=centos
```
ggiguash, line 14: Can we use `microshift` user here?

ansible/roles/epel-repo-install/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+        state: present
+  when: ansible_distribution == "RedHat"
+
+- name: CentOS tasks
+  block:
+    - name: enable powertools repo (CentOS 8)
+      ansible.builtin.command: dnf config-manager --set-enabled powertools
+      when: ansible_distribution_version == "8"
+
+    # There's no CRI-O for RHEL/CentOS 9
```
ggiguash, line 19: We do not support RHEL 9 now, so can we remove this?

ansible/roles/install-logging/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
+---
+# install-logging tasks
+
+- name: CentOS tasks
+  block:
+    - name: enable powertools repo (CentOS 8)
+      ansible.builtin.command: dnf config-manager --set-enabled powertools
+      when: ansible_distribution_version == "8"
+
+    # There's no CRI-O for RHEL/CentOS 9
```
ggiguash, line 10: We do not support RHEL 9, so this is better removed, I think

ansible/roles/install-microshift/defaults/main.yml
```diff
@@ -0,0 +1,43 @@
+---
+# install-microshift default vars
+
+enable_repos:
```
ggiguash, line 4: Why do we need RHEL 9 repos for MicroShift? 
Should we just build RPMs here and install them?
See https://github.com/openshift/microshift/blob/main/docs/devenv_rhel8.md#build-microshift

ansible/roles/install-microshift/defaults/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  - http
+  - https
+  - mdns
+
+firewall_ports:
+  - 6443/tcp
+  - 8080/tcp
+  - 9100/tcp
+  - 9537/tcp
+  - 17001/tcp
```
ggiguash, line 28: We need firewall configuration for MicroShift as described in https://github.com/openshift/microshift/blob/main/docs/howto_firewall.md. Otherwise, not all pods are going to boot.

ansible/roles/install-microshift/tasks/main.yml
```diff
@@ -0,0 +1,237 @@
...
+
+- name: enable cri-o module
+  ansible.builtin.command: dnf module enable -y cri-o
+
+- name: install microshift package
+  ansible.builtin.dnf:
+    name:
+    - microshift
+    state: present
+    update_cache: true
```
ggiguash, line 134: I think we may want to either build RPMs locally, or download packages from brew.

- https://github.com/openshift/microshift/blob/main/docs/devenv_rhel8.md#build-microshift
- https://github.com/openshift/microshift/blob/main/docs/rpm_packages.md

ansible/roles/install-microshift/tasks/main.yml
```diff
@@ -0,0 +1,237 @@
...
+  until: remote_kubeconfig.stat.exists
+  retries: 30
+  delay: 10
+
+- name: wait for pods running
+  ansible.builtin.shell: kubectl get po -A --no-headers | grep Running | wc -l
+  retries: 60
+  delay: 5
+  register: result
+  until: result.stdout == "6"
```
ggiguash, line 193: Instead of counting Running pods, we may want to "grep -v Running" and wait for 0. This should be more robust when pods are added.

ansible/roles/install-microshift/templates/kubernetes.repo.j2
```diff
@@ -0,0 +1,7 @@
+[kubernetes]
+name=Kubernetes
+baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-$basearch
```
ggiguash, line 3: Should it be "el-8"?

ansible/inventory/inventory
dhellmann: Does the user depend on the image?

ansible/roles/install-microshift/tasks/main.yml
dhellmann: If microshift had a readiness check API endpoint, would we count those core pods as part of the answer it should return?

ansible/inventory/inventory
```diff
@@ -0,0 +1,20 @@
+[microshift]
+microshift.example.com ansible_host= private_ip= # microshift node
```
zshi-redhat, line 2: I got this error when filled in the ansible_host and private_ip:
```
fatal: [localhost]: FAILED! => {"msg": "[\"{{ hostvars['microshift'].ansible_host }}\", \"{{ hostvars['logging'].ansible_host }}\"]: \"hostvars['microshift']\" is undefined"}
```

```
microshift.example.com ansible_host=192.168.122.10 private_ip=192.168.122.10
```
How the host vars shall be populated?

ansible/roles/epel-repo-install/tasks/main.yml
zshi-redhat: This is skipped on rhel9:
```
TASK [install-logging : enable powertools repo (CentOS 9)] ******************************************************************************************************************
skipping: [logging.example.com] => {"changed": false, "skip_reason": "Conditional result was False"}
```

ansible/inventory/inventory
sjug: Sorry that was a bit hardcoded, my hosts did not have a hostname so it was looking for a host called `microshift`.
I've pushed a commit to dynamically build the host IPs based on the pattern specified in the inventory here.

ansible/README.md
sjug: Added Ansible installation doc step for Fedora.

ansible/inventory/inventory
sjug: Yeah it does, the AMIs for CentOS have user `centos`. It is a var after all so we can set it as necessary.
Once I move to image builder AMIs we can make the default user `microshift`.

ansible/inventory/inventory
sjug: Same as my previous comment for the `microshift:vars` for `ansible_user`.

ansible/roles/epel-repo-install/tasks/main.yml
sjug: It would be skipped on RHEL9 as it's a task for CentOS hosts.
I will change the comment so I don't mention RHEL.

ansible/roles/install-logging/tasks/main.yml
sjug: Would you like me to remove the "RHEL" text from the comment or all the CentOS 9 tasks?

ansible/roles/install-microshift/defaults/main.yml
sjug: I have created these scripts based upon installation of MicroShift from the COPR repo. 
Though even with building the RPMs, would the build not need prerequisite packages to be installed from the repos?

ansible/roles/install-microshift/templates/kubernetes.repo.j2
sjug: It should, but there's no `el8` repo.

ansible/README.md
sjug: Yeah that was a pretty bad sentence, let me know if it reads better now.

ansible/roles/epel-repo-install/tasks/main.yml
ggiguash: Yes, I think let's remove this. Otherwise, it's confusing.

ansible/roles/install-logging/tasks/main.yml
ggiguash: Yes, I think let's remove this. Otherwise, it's confusing.

ansible/roles/install-microshift/defaults/main.yml
ggiguash: COPR packages are outdated. We should probably check it it for the first pass and then discuss our next steps. We need support for RHEL8 as all the development and runtime is now on this OS.

ggiguash (approved):

ansible/roles/install-microshift/defaults/main.yml
sjug: Yes sir, updated firewall ports and CIDRs.

ansible/roles/install-microshift/defaults/main.yml
sjug: Agreed

ansible/roles/install-logging/tasks/main.yml
sjug: Removed

ansible/roles/epel-repo-install/tasks/main.yml
sjug: Removed

ansible/roles/install-microshift/tasks/main.yml
sjug: I've changed the check as you've suggested @ggiguash.
@dhellmann I'm not sure if you're asking me or Gregory, but I assume a readiness check would count the core pods?

Is there any chance of a race where some pods are not yet created but the first ones are in Running state?

ansible/roles/install-microshift/tasks/main.yml
sjug: That's an interesting problem as on one hand we're doing a build on super underpowered hardware (local on the microshift node). Alternatively brew is behind the VPN so not easy to pull to AWS or outside in general.

However we could either build the RPMs locally and then push them to the MicroShift node, or follow that same pattern to pull the RPMs from brew locally and then push them to the node.

ansible/roles/install-microshift/tasks/main.yml
sjug: The change broke the check, as the query returns "No resources found" for some time until some pods are created.
Any suggestions on what it should be if we don't like the pod running count check that I had before?

ggiguash: Since this is a shell command we're running, we should be able to add a "sleep 1" loop on "No resource found" condition and then proceeding to the next stage once we have at least one pod.

ansible/roles/install-microshift/tasks/main.yml
ggiguash: MicroShift binary by itself does not require much resources to be built, especially on development VMs.
Anyway, let's push this task to the second round of enhancements.

ggiguash:
Let's merge the first version of this change. I'm approving it and putting it on hold until @sjug feels convenient to release and merge it. Once merged, we should then discuss our next steps.
/hold 
/lgtm

ansible/roles/install-microshift/defaults/main.yml
sjug: I've added support for building and installing the MicroShift RPMs from source on RHEL8.

ansible/roles/install-microshift/tasks/main.yml
sjug: I've added the RPM building on the MicroShift node, takes around 9-10m.

ansible/roles/install-microshift/tasks/main.yml
sjug: Not really a fan of the current wait situation, but we can discuss this further for future changes.

sjug:
I've addressed pretty much all the review comments and we're in good shape for the most part.
Do you want me to squash and push one commit or do you squash merge?

PR #862, opened 2022-08-15
USHIFT-275: Use "apply" semantics for static resource appliers.

benluddy:
The existing library-go module dependency provides a resourceapply
package that supports nearly all of the types that are directly
applied in Microshift. Reusing this package provides consistency with
other OpenShift components and benefits over time from shared
maintenance of library-go.


benluddy:
One of the `kuttl` downloads apparently received an HTML page instead of the binary: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/862/pull-ci-openshift-microshift-main-e2e-openshift-conformance-sig-api-machinery/1559185638102667264#1:build-log.txt%3A207

/retest

dhellmann:
/approve

Nice.

oglok:
I've been digging a bit into the library-go module for applying static resources and I think it makes a lot of sense to use it and get aligned with the rest of the OpenShift. The code looks much cleaner now. Thanks @benluddy , this is a great PR.

/lgtm

PR #864, opened 2022-08-16
Fix early systemd readiness

pmtk:
Closes [USHIFT-242](https://issues.redhat.com/browse/USHIFT-242)

**Description**:
- Fixes premature systemd service readiness by clearing `NOTIFY_SOCKET` environment variable so apiserver, kubelet, or openshift-controller-manager won't send `READY=1`. The variable is restored when MicroShift is considered ready.
- Adds closure of `sysconfwatch`'s `ready` channel, so MicroShift's readiness is no longer blocked.
- Adds 2m service start timeout. If service does not start in that time, it'll be terminated and restarted.
- `systemctl start microshift` won't return until MicroShift is ready which now will happen much later (currently ~45s based on several tries).

### DIFF
--- a/pkg/sysconfwatch/sysconfwatch_linux.go
+++ b/pkg/sysconfwatch/sysconfwatch_linux.go
@@ -91,6 +91,9 @@ func (c *SysConfWatchController) Run(ctx context.Context, ready chan<- struct{},
 	var buf []byte = make([]byte, 8)
 	// Take a snapshot of the system and monototic clocks as a base reference
 	stimeRef, mtimeRef := getSysMonTimes()
+
+	klog.Infof("sysconfwatch-controller is ready")
+	close(ready)
 	for {
 		select {
 		case <-ticker.C:
--- a/packaging/systemd/microshift.service
+++ b/packaging/systemd/microshift.service
@@ -14,6 +14,7 @@ CPUAccounting=yes
 BlockIOAccounting=yes
 MemoryAccounting=yes
 LimitNOFILE=1048576
+TimeoutStartSec=2m
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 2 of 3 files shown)


pkg/sysconfwatch/sysconfwatch_linux.go
```diff
@@ -91,6 +91,11 @@ func (c *SysConfWatchController) Run(ctx context.Context, ready chan<- struct{},
 	var buf []byte = make([]byte, 8)
 	// Take a snapshot of the system and monototic clocks as a base reference
 	stimeRef, mtimeRef := getSysMonTimes()
+
+	go func() {
+		klog.Infof("sysconfwatch-controller is ready")
+		close(ready)
+	}()
```
benluddy, line 98: Can we do the same without a new goroutine?

```suggestion
	klog.Infof("sysconfwatch-controller is ready")
	close(ready)
```

benluddy (commented):
/lgtm

There's also https://pkg.go.dev/os#Unsetenv which might be closer to the intent. I'm OK with it either way. It would be surprising if go-systemd were changed such that it distinguished between empty and unset.

pmtk:
> There's also https://pkg.go.dev/os#Unsetenv which might be closer to the intent. I'm OK with it either way. It would be surprising if go-systemd were changed such that it distinguished between empty and unset.

Didn't notice that one. I like it better, thanks!

oglok:
Could you please rebase this PR? I'll have a look at it then.

dhellmann:
Nice! I'll leave this open for @oglok to approve since he said he wanted to take a look.

oglok:
/hold

Could you please squash the three commits into just one? It changes just 3 files, so it makes sense just one commit.

pmtk:
@oglok done :)

PR #866, opened 2022-08-16
update networking notes in design docs

dhellmann:
/assign @zshi-redhat @mangelajo

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -108,7 +108,6 @@ When deciding between different design options, we follow the following principl
 * MicroShift uses the Red Hat OpenShift Networking CNI driver, based on OVN-Kubernetes.
 * No Multus.
 * Single-node, so no API load balancer.
-* No service load balancer, by default.
 * Open issues / questions:
   * Ingress?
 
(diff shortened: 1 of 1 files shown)


docs/design.md
```diff
@@ -105,11 +105,11 @@ When deciding between different design options, we follow the following principl
 
 ### Networking
 * Host networking is configured by device management. MicroShift has to work with what it's been given by the host OS.
+* MicroShift uses the Red Hat OpenShift Networking CNI driver, based on OVN-Kubernetes.
 * No Multus.
+* Single-node, so no API load balancer.
+* No service load balancer, by default.
```
zshi-redhat, line 111: @dhellmann k8s service can be load balanced by ovn-network if there are multiple endpoints on single-node deployment. I can see your point that service won't be load balanced to multiple nodes. woud it make sense to say no service load balancer to multi-nodes or remove this line?

copejon: +1 Would be good to clarify for the reader.

dhellmann: If OVN supports the service type of load balancer we should remove this line.

dhellmann:
```suggestion
```

PR #868, opened 2022-08-17
USHIFT-197: Enable resource accounting in MicroShift systemd service

oglok:

Closes #[USHIFT-197
](https://issues.redhat.com/browse/USHIFT-197)

### DIFF
--- a/packaging/systemd/microshift.service
+++ b/packaging/systemd/microshift.service
@@ -9,6 +9,11 @@ ExecStart=microshift run
 Restart=always
 User=root
 Type=notify
+Delegate=yes
+CPUAccounting=yes
+BlockIOAccounting=yes
+MemoryAccounting=yes
+LimitNOFILE=1048576
 
 [Install]
 WantedBy=multi-user.target


oglok:
With this PR, local testing showed no error message:

```
I0718 15:57:18.377000 2180849 server.go:662] "Failed to get the kubelet's cgroup. Kubelet system container metrics may be missing." err="cpu and memory cgroup hierarchy not unified.  cpu: /user.slice, memory: \
/user.slice/user-1000.slice/session-1.scope" 
```

With these options in the systemd service file, we enable accounting as you can see in the status of the service itself:

```
systemctl status microshift
● microshift.service - MicroShift
   Loaded: loaded (/usr/lib/systemd/system/microshift.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2022-08-17 11:01:57 EDT; 8min ago
 Main PID: 1821 (microshift)
    Tasks: 25 (limit: 14213)
   Memory: 464.5M
      CPU: 51.887s
   CGroup: /system.slice/microshift.service
           └─1821 /usr/bin/microshift run
 ```

```
[root@localhost ~]# systemctl show microshift | grep -i accounting
CPUAccounting=yes
IOAccounting=no
BlockIOAccounting=yes
MemoryAccounting=yes
TasksAccounting=yes
IPAccounting=no
```

**Comment by oglok** [comment]
If required, I could add a new target to the Makefile so developers can run MicroShift from systemd in an easier way as @mrunalp suggested.

File: packaging/systemd/microshift.service [hunk 1]
```diff
@@ -9,6 +9,15 @@ ExecStart=microshift run
...
 Type=notify
+Delegate=yes
+CPUAccounting=yes
+BlockIOAccounting=yes
+MemoryAccounting=yes
+# Having non-zero Limit*s causes performance problems due to accounting overhead
+# in the kernel. We recommend using cgroups to do container-local accounting.
+LimitNOFILE=infinity
+LimitNPROC=infinity
+LimitCORE=infinity
```
ggiguash, line 20: Why are we resetting these 3 specific limits? How do we know we should not bump more to avoid the mentioned performance problems.

ggiguash: Also, can we add a link in the comment that describes the performance problem?

oglok: I'm not sure I follow @ggiguash , are you talking about the performance issues caused by a very high value in LimitNOFILE? I could reduce it to 1048576 as it seems to be the standard for other project.

oglok: https://github.com/coreos/fedora-coreos-tracker/issues/329

oglok: `LimitNOFILE = 1048576` seems to be the standard number for most of the projects that deal with containers from a systemd service.

ggiguash: Can we update the comment to something more concrete? I think it's important to note what part of the kernel is not performing well.
```
Having higher limit values cause file system performance problems due to accounting overhead in the kernel. 
We recommend using cgroup to do container-local accounting.
```

ggiguash (approved):

PR #870, opened 2022-08-18
API-1438: Split the global trust with kube-control-plane client certs

stlaz:
This PR attempts to split away the client certs for kube-scheduler and KCM from the global trust provided by the root CA.


dhellmann:
Linking to USHIFT-278

pkg/cryptomaterial/certpaths.go
```diff
@@ -0,0 +1,41 @@
+package cryptomaterial
```
oglok, line 1: What's the reasoning behind creating a new go package for this? Would it make sense to add it to the util package?

pkg/controllers/kube-apiserver.go
```diff
@@ -90,7 +92,9 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 	s.kubeconfig = filepath.Join(cfg.DataDir, "resources", "kube-apiserver", "kubeconfig")
 	s.verbosity = cfg.LogVLevel
 
-	caCertFile := filepath.Join(cfg.DataDir, "certs", "ca-bundle", "ca-bundle.crt")
+	certsDir := cryptomaterial.CertsDirectory(cfg.DataDir)
+	caCertFile := filepath.Join(certsDir, "ca-bundle", "ca-bundle.crt")
```
oglok, line 96: Why isn't there a helper function for the client CA bundle certficate path?

pkg/cmd/init.go
```diff
@@ -41,112 +55,205 @@ func loadCA(cfg *config.MicroshiftConfig) error {
...
 	// per https://kubernetes.io/docs/reference/access-authn-authz/node/#overview
 	if err := util.Kubeconfig(filepath.Join(cfg.DataDir, "/resources/kubelet/kubeconfig"),
+		clusterTrustBundlePEM,
 		"system:node:"+cfg.NodeName, []string{"system:nodes"}, cfg.Cluster.URL); err != nil {
 		return err
 	}
 	return nil
 }
+
+func generateClientCA(certsDir, signerDir, signerName string) (*crypto.CA, error) {
```
oglok, line 237: Shouldn't this be in cryptomaterial package or util?

oglok (commented):
Couple of questions. Otherwise, it looks good to me.

In order to understand this PR, it's good to have the following link as reference:

https://github.com/openshift/api/blob/master/tls/docs/kube-apiserver%20Client%20Certificates/README.md#kube-control-plane-signer

and this is just the first change to a better alignment with OpenShift's certificate management.

pkg/cryptomaterial/certpaths.go
stlaz: Having a separate package with a specific purpose seemed like a good practice.

I suspect it could exist as a package within the util packge, I don't have a strong opinion on its location.

pkg/controllers/kube-apiserver.go
stlaz: I didn't figure out what the `ca-bundle` directory stands for yet so I did not know how to name it.

pkg/cmd/init.go
stlaz: For now I treated it only as a shortcut function that is very specific to what's done in this file. Would it be useful elsewhere?

oglok: yeah, if this is not gonna be reuse, just keep it as a local function. I'm ok with it.

pkg/cryptomaterial/certpaths.go
oglok: If you're ok with it, let's put it under util where all the helper functions are located.

pkg/controllers/kube-apiserver.go
oglok: I just mean that `cryptomaterial.TotalClientCABundlePath` builds a path to the `client-ca.crt`. Why isn't there a helper function to build the path to `ca-bundle.crt`?

pkg/cryptomaterial/certpaths.go
ibihim: It is discouraged to have a util or commong pkg in Go.

stlaz: I think as long as we keep it structured, we might be ok. Currently it is just one package which does not really play well with the best practices, but I think `util/cryptomaterial` might be ok.

pkg/controllers/kube-apiserver.go
stlaz: I did not want to go and replace the reference everywhere, since we will be trying to get rid of this bundle :slightly_smiling_face:

But I can add a helper for that, at least we'll be able to find all the spots where it is used more easily later.

pkg/controllers/etcd.go
```diff
@@ -55,8 +56,10 @@ func (s *EtcdService) Name() string           { return "etcd" }
 func (s *EtcdService) Dependencies() []string { return []string{} }
 
 func (s *EtcdService) configure(cfg *config.MicroshiftConfig) {
-	caCertFile := filepath.Join(cfg.DataDir, "certs", "ca-bundle", "ca-bundle.crt")
-	certDir := filepath.Join(cfg.DataDir, "certs", s.Name())
+	certsDir := filepath.Join(cfg.DataDir)
+
+	caCertFile := cryptomaterial.UltimateTrustBundlePath(certsDir)
+	etcdCertDir := filepath.Join(certsDir, s.Name())
```
oglok, line 62: Tests are failing because etcd cannot start properly. It searches for 

`/var/lib/microshift/etcd/etcd-peer.crt`

when it should be:

`/var/lib/microshift/certs/etcd/etcd-peer.crt`

oglok (changes requested):

pkg/controllers/etcd.go
stlaz: Ooops, I forgot to replace the original `filepath.Join()`. That should now be fixed.

oglok:
Fatal!

`F0824 13:37:55.998994   19013 kubelet.go:150] Error in fetching depenedencies%!(EXTRA *fmt.wrapError=unable to load client CA file /var/lib/microshift/ca-bundle/ca-bundle.crt: open /var/lib/microshift/ca-bundle/ca-bundle.crt: no such file or directory)
goroutine 34528 [running]:`

oglok (approved):
/lgtm

PR #872, opened 2022-08-19
USHIFT-286: Run Service CA component as non root

oglok:

Closes [USHIFT-286](https://issues.redhat.com/browse/USHIFT-286)

### DIFF
--- a/assets/components/service-ca/deployment.yaml
+++ b/assets/components/service-ca/deployment.yaml
@@ -23,7 +23,6 @@ spec:
         app: service-ca
         service-ca: "true"
     spec:
-      securityContext: {}
       serviceAccount: service-ca
       serviceAccountName: service-ca
       containers:
@@ -33,8 +32,8 @@ spec:
         command: ["service-ca-operator", "controller"]
         ports:
         - containerPort: 8443
-        # securityContext:
-        #   runAsNonRoot: true
+        securityContext:
+          runAsNonRoot: true
         resources:
           requests:
             memory: 120Mi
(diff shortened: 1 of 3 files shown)


assets/components/service-ca/deployment.yaml
```diff
@@ -23,7 +23,10 @@ spec:
         app: service-ca
         service-ca: "true"
     spec:
-      securityContext: {}
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1001
```
dhellmann, line 28: Is there any reason to register or document this user ID to ensure that other workloads do not reuse it?

dhellmann (commented):
One question. If the answer is "no", feel free to remove the hold and merge this as it is.

/hold
/lgtm

assets/components/service-ca/deployment.yaml
dhellmann: OK, I don't know how UIDs like this are applied. If we have another pod with the same settings, would it be able to see or communicate with this one in a way we wouldn't want?

oglok: This isn't right. I've override the existing commit. The Cluster Policy Controller should generate the UID based on the security context policy, in this case, `runAsNonRoot`. 

After fixing the cluster role binding, now it works:

```
    securityContext:
      capabilities:
        drop:
        - KILL
        - MKNOD
        - SETGID
        - SETUID
      runAsNonRoot: true
      runAsUser: 1000070000
      ```

PR #876, opened 2022-08-22
Get node IP through ovn gateway interface when no default route

zshi-redhat:
MicroShift and some service components (e.g. kube-apiserver)
get nodeIP through the default route interface. When there is
no internet connection, default route won't exist which results
in failure starting services.

For example,

kubelet expects nodeIP to not be 127.0.0.1:

kubelet_node_status.go:622] "Failed to set some node status fields"
err="failed to validate nodeIP: nodeIP can't be loopback address"
node="localhost.localdomain"

kube-apiserver fails to find suitable network address:

service kube-apiserver exited with error: kube-apiserver configuration
error: Unable to find suitable network address.error='no default routes
found in "/proc/net/route" or "/proc/net/ipv6_route"'. Try to set the
AdvertiseAddress directly or provide a valid BindAddress to fix this.

Closes #[USHIFT-22](https://issues.redhat.com//browse/USHIFT-22)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -84,7 +84,7 @@ func NewMicroshiftConfig() *MicroshiftConfig {
 	}
 	nodeIP, err := util.GetHostIP()
 	if err != nil {
-		klog.Warningf("failed to get host IP: %v, using: %q", err, nodeIP)
+		klog.Fatalf("failed to get host IP: %v", err)
 	}
 
 	dataDir := findDataDir()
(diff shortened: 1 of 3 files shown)


zshi-redhat:
@benluddy  do you think there will be any issue if the advertise-address is always set?

zshi-redhat:
Note that this PR only fixes issues in the microshift main process.
There will be separate PRs to fix configure-ovs.sh and ovnk to support use case of no internet conneciton.

dhellmann:
> @benluddy do you think there will be any issue if the advertise-address is always set?

Maybe @tkashem can help while @benluddy is out.

mangelajo (commented):
/lgtm

PR #877, opened 2022-08-22
cleanup ovn kubernetes CNI binary and config file

zshi-redhat:
ovn-kubernetes CNI config file is written by ovnkube-node
process and will be left over if ovnkube-node doesn't
exit gracefully (which is the case when stopping MicroShift
service).

Crio waits for the existence of ovn-kubernetes CNI config
to mark the node as Ready. If ovn-kubernetes CNI config is
not cleaned up, node will become ready immediately on next
boot w/o waiting for ovn network become ready.

### DIFF
--- a/hack/cleanup.sh
+++ b/hack/cleanup.sh
@@ -40,6 +40,8 @@ sudo bash -c "
     echo Removing MicroShift and OVN configuration
     rm -rf /var/lib/{microshift,ovnk}
     rm -rf /var/run/ovn
+    rm -f /etc/cni/net.d/10-ovn-kubernetes.conf
+    rm -f /opt/cni/bin/ovn-k8s-cni-overlay
 
     echo Cleanup succeeded
 "


hack/cleanup.sh
```diff
@@ -40,6 +40,8 @@ sudo bash -c "
     echo Removing MicroShift and OVN configuration
     rm -rf /var/lib/{microshift,ovnk}
     rm -rf /var/run/ovn
+    rm -rf /etc/cni/net.d/10-ovn-kubernetes.conf
+    rm -rf /opt/cni/bin/ovn-k8s-cni-overlay
```
ggiguash, line 44: These are files, not directories, so can we use "rm -f" command?

ggiguash (changes requested):

ggiguash (approved):

dhellmann:
USHIFT-386

PR #878, opened 2022-08-22
USHIFT-293: Ansible GCP Compatibility Update

sjug:
- Check for presence & conditional installation of subscription-manager
  and oc/kubectl binary
- Small change to pod readiness count and command
- Small update to README for missing vars
- Formatting fixes

**Which issue(s) this PR addresses**: [USHIFT-293](https://issues.redhat.com/browse/USHIFT-293)

### DIFF
--- a/ansible/roles/install-microshift/defaults/main.yml
+++ b/ansible/roles/install-microshift/defaults/main.yml
@@ -37,8 +37,8 @@ firewall_trusted_cidr:
   - 10.42.0.0/16
   - 169.254.169.1/32
 
-root_bin_dir: /root/bin
-microshift_dir: /root/microshift
+root_bin_dir: /usr/bin
+microshift_dir: "{{ ansible_env.HOME }}/microshift"
 microshift_rpms: []
 
 systemd_services:
--- a/ansible/roles/microshift-start/tasks/main.yml
+++ b/ansible/roles/microshift-start/tasks/main.yml
@@ -55,12 +55,7 @@
   retries: 30
   delay: 1
 
-- name: wait for pods running
-  ansible.builtin.shell: kubectl get po -A --no-headers | grep Running | wc -l
-  retries: 60
-  delay: 1
-  register: result
-  until: result.stdout == "6"
+- include_tasks: roles/common/tasks/readinesscheck.yml
 
 - name: force update of current timestamp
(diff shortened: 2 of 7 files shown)


ansible/roles/install-microshift/defaults/main.yml
```diff
@@ -37,7 +37,7 @@ firewall_trusted_cidr:
   - 10.42.0.0/16
   - 169.254.169.1/32
 
-root_bin_dir: /root/bin
+root_bin_dir: /usr/bin
```
ggiguash, line 40: For microshift directory, can we use $HOME or ~/microshift - not a hardcoded value?

ansible/roles/microshift-start/tasks/main.yml
```diff
@@ -56,11 +56,11 @@
   delay: 1
 
 - name: wait for pods running
-  ansible.builtin.shell: kubectl get po -A --no-headers | grep Running | wc -l
+  ansible.builtin.shell: oc get po -A --no-headers | grep Running | wc -l
   retries: 60
   delay: 1
   register: result
-  until: result.stdout == "6"
+  until: result.stdout == "8"
```
ggiguash, line 63: This task seems to be duplicated in at least 2 places. Can we refactor into a common file and call it?

ggiguash (changes requested):

ansible/roles/install-microshift/defaults/main.yml
sjug: So this is for the `bin` dir. On AWS RHEL images $HOME/bin/ is in `$PATH` but on GCP images it's not.
That's why I changed it to something that's in the path for both.

ggiguash: I meant the microshift_dir variable. Sorry, comment is in the wrong place.

ansible/roles/microshift-start/tasks/main.yml
sjug: Agreed, pulled into common task.

ansible/roles/install-microshift/defaults/main.yml
sjug: Changed `microshift_dir` to be based on `$HOME`

PR #879, opened 2022-08-23
Use first available route interface

zshi-redhat:
Configure-ovs.sh gets ovn gateway interface using
default route, which may not exist in no internet
connection environment. This change falls back to
use first available route interface in the case
that no default route is found.


mangelajo (approved):
/lgtm

dhellmann:
USHIFT-385

PR #883, opened 2022-08-24
Add csi node user-facing configuration

copejon:

Closes [USHIFT-168](https://issues.redhat.com//browse/USHIFT-168)


ggiguash:
@copejon - could we reorganize the commits into at most 3?

pkg/components/components.go
```diff
@@ -13,7 +13,7 @@ func StartComponents(cfg *config.MicroshiftConfig) error {
 		return err
 	}
 
-	if err := startCSIPLugin(cfg.DataDir + "/resources/kubeadmin/kubeconfig"); err != nil {
+	if err := startCSIPlugin(cfg.DataDir+"/resources/kubeadmin/kubeconfig", cfg.CSIPluginConfig); err != nil {
```
ggiguash, line 16: Use filepath.join here?

oglok:
Just not to forget, please, squash the commits if possible to avoid having 25 commits in history.

pkg/config/storage/lvmd.go
```diff
@@ -0,0 +1,39 @@
...
+	"k8s.io/klog/v2"
+)
+
+const ForcedSocketDir = "/run/lvmd/lvmd.sock"
+
+// LvmdConfig holds the full structure of the lvmd configuration
+type LvmdConfig map[string]interface{}
+
+// UnmarshalYAML overrides the socket-dir when reading from yaml to LvmdConfig in order
+// to prevent users from breaking the CSIPlugin with missing or incorrect socket paths.
```
dhellmann, line 16: Another safer way to do this would be to add a "device classes" field to the MicroShift configuration file. That avoids exposing the `socket-name` option, so the user can't break it, but gives them control over the values they need to be able to change. It also lets us validate those settings, so we could fail to start if there is a list of device classes but none has the name "default", for example.

dhellmann (commented):
I'm a little worried about the depth at which we're exposing the config options here. I'm also not that keen on having our config contain embedded YAML (we can't do much about the lvmd.yaml spec, but we can avoid that pattern ourselves).

pkg/config/storage/lvmd.go
copejon: I'd done that initially in fact but there were a few things that bugged me with the approach:

- The lvmd process already validates the config, concisely logs validation errors, and crashes immediately.  So it's obvious that, when the lvmd pod is crashLooping, `oc logs` will produce less than 5 lines describing what's wrong.  If microshift were made to crash on lvmd validation errors, the cause is much less obvious with the amount of log spam the process generates.  

- copying the lvmd config API in microshift feels like bad practice, and it also implies reflecting how lvmd validates the config fields.  The lvmd version is pinned in microshift of course, but it would essentially create an invisible dependency on the lvmd code base.  Less horrible that making lvmd a direct dependency, admittedly.

- Users can view the lvmd configmap in the cluster, edit it, and test config changes live against the cluster.  It ends up being a much cleaner debugging approach.

pkg/components/render_test.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := renderLvmd(tt.args.b, tt.args.p)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("renderLvmd() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !bytes.Equal(got, tt.want) {
```
dhellmann, line 68: https://github.com/stretchr/testify has some nice features for showing where differences are in string comparisons like this. In another PR we should look at what adding that would do to the dependency list.

pkg/config/lvmd/lvmd.go
```diff
@@ -0,0 +1,104 @@
...
+// DeviceClass maps between device-classes and volume groups.
+type DeviceClass struct {
+	// Name for the device-class name
+	Name string `yaml:"name"`
+	// Volume group name for the device-class
+	VolumeGroup string `yaml:"volume-group"`
+	// Default is a flag to indicate whether the device-class is the default
+	Default bool `yaml:"default"`
+	// SpareGB is storage capacity in GiB to be spared
+	SpareGB uint `yaml:"spare-gb"`
```
dhellmann, line 33: Should this be a pointer so we can tell the difference between being willing to not leave any spare disk and not providing a value at all? Or does the value have to be > 0 to be valid?

pkg/config/config.go
```diff
@@ -73,8 +74,9 @@ type MicroshiftConfig struct {
 	ControlPlane ControlPlaneConfig `yaml:"controlPlane"`
 	Node         NodeConfig         `yaml:"node"`
 
-	Manifests []string    `yaml:"manifests"`
-	Debug     DebugConfig `yaml:"debug"`
+	Manifests       []string             `yaml:"manifests"`
+	Debug           DebugConfig          `yaml:"debug"`
+	CSIPluginConfig lvmd.CSIPluginConfig `yaml:"csiPluginConfig"`
```
dhellmann, line 79: Would it make sense to be more specific here and make the public name something like `odfLVM`? That gives us the ability to use a different name for a different plugin upstream, for example.

It's in the config file, so I think we can leave "config" out of the field name either way.

pkg/config/lvmd/lvmd.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func NewLVMDConfigForDeviceClasses(d []*DeviceClass) *lvmdConfig {
+	return &lvmdConfig{
+		SocketName:    defaultSockName,
+		DeviceClasses: d,
+	}
+}
+
+func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
```
dhellmann, line 66: Would it make sense to use the validation in the vendored code instead of reproducing it? Or does that drag in a bunch of other dependencies?

pkg/components/render_test.go
copejon: I like testify, but yeah, I didn't want to stack on more dependencies for this PR.

pkg/config/lvmd/lvmd.go
copejon: Lvmd will default a null `spare-gb` field to 10Gb, so users will have explicitly define the spare-gb field for any other value.

pkg/config/config.go
copejon: Good point, will fix.

pkg/config/lvmd/lvmd.go
dhellmann: How does that work? With the struct the way it is, I would expect the value of `SpareGB` to be `0` if the user does not provide anything. Then when we render the struct, I would expect to see the explicit `0` passed it to lvmd. Does lvmd convert the 0 to 10?

dhellmann (commented):
Sorry for nitpicking on naming. I do have one other question about the exposed data structure.

pkg/config/lvmd/lvmd.go
copejon: In lvmd, non-nil, non-negative values are valid inputs, and a nil value results in a 10gb default being used. IMO that seems overly cautious.  In prod, I would expect the system storage to remain static after the device is imaged, with the unallocated VG space being managed by lvmd.  Meaning the spared space would never be utilized. If that's a valid assumption, then defaulting to 0 seems reasonable.

pkg/components/render_test.go
```diff
@@ -0,0 +1,72 @@
...
+		c := template.Must(template.New("").Parse(
+			`# Source: topolvm/templates/lvmd/configmap.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: lvmd
+  namespace: openshift-storage
+data:
+  lvmd.yaml: {{ .lvmd }}
+`))
```
dhellmann, line 23: These tests look right, but I'd feel more comfortable with them if the expected values were spelled out by hand in the loop below, rather than using a duplicate template and rendering the same structures again in a different way.

dhellmann (commented):
This version looks good. I have one nit that can come in another PR if you want. I do think squashing that typo fix into the original commit would be good before we merge the PR.

pkg/components/render_test.go
copejon: Fixed! :)

dhellmann:
I'm happy with this version. I'd like to give @fzdarsky a chance to comment on the configuration layout and whether we want to put the storage-related settings in a subsection.

/lgtm
/hold

ggiguash:
I have two questions:
* Some files under controllers/, pkg/mdns and pkg/util contain tabulation changes in the license text. Do we want these in the current PR?
* Do we need to have user documentation updated to reflect this new functionality?

copejon:
> I have two questions:
> 
>     * Some files under controllers/, pkg/mdns and pkg/util contain tabulation changes in the license text. Do we want these in the current PR?

It seems that it won't pass the verify check otherwise.  Not sure why this is being caught now though.

>     * Do we need to have user documentation updated to reflect this new functionality?

I'll circle back and make sure the in-repo docs are accurate.

docs/default_csi_plugin.md
```diff
@@ -17,6 +19,25 @@ boots. [StorageCapacity tracking](https://kubernetes.io/docs/concepts/storage/st
...
+
+This configuration is specific to the LVMD runtime and informs how it interacts with the host. LVMD is a subcomponent of ODF-LVM, and in the case of MicroShift, deployed automatically as a DaemonSet.
+
+#### Specification
+
+Full documentation of the config spec can be found at [github.com/red-hat-storage/topolvm/blob/v4.11.0/docs/lvmd.md](https://github.com/red-hat-storage/topolvm/blob/v4.11.0/docs/lvmd.md).
+
+#### Path
+
+The user provided lvmd config should be written to the same directory as the Microshift config.  If a MicroShift config
```
pmtk, line 34:
```suggestion
The user provided lvmd config should be written to the same directory as the MicroShift config.  If a MicroShift config
```

docs/default_csi_plugin.md
```diff
@@ -17,6 +19,25 @@ boots. [StorageCapacity tracking](https://kubernetes.io/docs/concepts/storage/st
...
+doesn't exist, it will assume default lvmd values.  These paths will be checked for the config, depending on the user MicroShift
+is run as.
+
+1. User config dir: `~/.microshift/lvmd.yaml`
+2. Global config dir; `~/etc/microshift/lvmd.yaml`
```
pmtk, line 39:
```suggestion
2. Global config dir: `/etc/microshift/lvmd.yaml`
```

pkg/components/storage.go
```diff
@@ -58,6 +63,18 @@ func startCSIPLugin(cfg *config.MicroshiftConfig, kubeconfigPath string) error {
 			"assets/components/odf-lvm/topolvm-node-securitycontextconstraint.yaml",
 		}
 	)
+
+	// the lvmd file should be located in the same directory as the microshift config to minimize coupling with the
+	// csi plugin.
+	l, err := lvmd.NewLvmdConfigFromFileOrDefault(filepath.Join(filepath.Dir(cfg.ConfigFile), "lvmd.yaml"))
```
dhellmann, line 69: Renaming `l` to something like `lvmdSettings` here would make it more obvious what it holds.

pkg/config/lvmd/lvmd_test.go
```diff
@@ -0,0 +1,142 @@
...
+						SpareGB:     func() *uint64 { u := uint64(defaultSpareGB); return &u }(),
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			l := &Lvmd{
+				DeviceClasses: tt.fields.DeviceClasses,
```
dhellmann, line 61: Do we need to populate the `Lvmd` instance here in order to get valid defaults? It looks like `withDefaults` always returns the same data.

pkg/config/lvmd/types.go
```diff
@@ -0,0 +1,52 @@
+// DeviceClass type definition and validation is taken from:
+// https://github.com/red-hat-storage/topolvm/blob/release-4.10/lvmd/device_class_manager.go
```
pmtk, line 2: Nit: We could add an empty line between doc and `package`. Otherwise godoc will treat the comment like doc for the package

pkg/config/lvmd/test/config.yaml
```diff
@@ -0,0 +1,20 @@
+---
+lvmdConfigFile: ./lvmd.yaml
```
dhellmann, line 2: I think we want to remove this line, right?

scripts/tools.go
```diff
@@ -1,4 +1,7 @@
+//go:build ex
```
pmtk, line 1: so this makes sure that this package does not end up in the binary?

pkg/config/lvmd/test/lvmd.yaml
```diff
@@ -0,0 +1,10 @@
+socket-name: '/run/lvmd/lvmd.socket'
+device-classes:
+- name: "gold"
+  default: true
+  volume-group: "vg_1"
+  spare-gb: 5
+- name: "silver"
+  default: false
+  volume-group: "vg_2"
+  spare-gb: 5
```
pmtk, line 10: nit: missing endline

pkg/config/lvmd/test/valueMismatch.yaml
```diff
@@ -0,0 +1,10 @@
+socket-name: '/run/lvmd/lvmd.socket'
+device-classes:
+- name: "gold"
+  default: "true"
+  volume-group: "vg_1"
+  spare-gb: 5
+- name: "silver"
+  default: "false"
+  volume-group: "vg_2"
+  spare-gb: 5
```
pmtk, line 10: nit: missing endline

docs/default_csi_plugin.md
mangelajo, line 39:
```suggestion
2. Global config dir; `/etc/microshift/lvmd.yaml`
```

test/config.yaml
```diff
@@ -1,6 +1,7 @@
 --- 
 dataDir: /tmp/microshift/data
 auditLogDir: /tmp/microshift/logs
+lvmdConfigFile:
```
dhellmann, line 4: I think we want to remove this line, right?

mangelajo (changes requested):
a tiny doc nit

pkg/config/lvmd/lvmd_test.go
```diff
@@ -0,0 +1,142 @@
...
+						SpareGB:     iToP(5),
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "config value mismatch",
+			args: args{
+				"./test/valueMismatch.yaml",
```
dhellmann, line 122: What error is expected? Maybe you could add a comment to the file for now and then in a later PR update the test to check the error type or message?

scripts/tools.go
copejon: Not exactly.  These tool-import files are a bit of a hack to get go mod to fetch executables.  Unfortunately, `go test` can recognize this, but will still fail.  So slapping an arbitrary build constraint is an idiomatic way to enable `go test ./...`.  I meant to remove this one however, since it was just to make my IDE happy.

copejon, line 1:
```suggestion
```

test/config.yaml
copejon: Ah, yes.  It's ignored in the tests since the field doesn't exist on the go side. Will fix.

docs/default_csi_plugin.md
```diff
@@ -17,6 +19,25 @@ boots. [StorageCapacity tracking](https://kubernetes.io/docs/concepts/storage/st
...
+This configuration is specific to the LVMD runtime and informs how it interacts with the host. LVMD is a subcomponent of ODF-LVM, and in the case of MicroShift, deployed automatically as a DaemonSet.
+
+#### Specification
+
+Full documentation of the config spec can be found at [github.com/red-hat-storage/topolvm/blob/v4.11.0/docs/lvmd.md](https://github.com/red-hat-storage/topolvm/blob/v4.11.0/docs/lvmd.md).
+
+#### Path
+
+The user provided lvmd config should be written to the same directory as the MicroShift config.  If a MicroShift config
+doesn't exist, it will assume default lvmd values.  These paths will be checked for the config, depending on the user MicroShift
```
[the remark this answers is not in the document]
copejon:
```suggestion
doesn't exist, MicroShift will assume default lvmd values. These paths will be checked for the config, depending on the user MicroShift
```

pkg/config/lvmd/types.go
copejon:
```suggestion
// https://github.com/red-hat-storage/topolvm/blob/release-4.10/lvmd/device_class_manager.go

```

pkg/config/lvmd/lvmd_test.go
copejon: I've removed this test.  withDefaults returns a static type, which would be checked at compile time.  No point in testing for it as well.

pkg/config/lvmd/lvmd_test.go
copejon: I've gone ahead and removed the test.

dhellmann (commented):
/lgtm

PR #884, opened 2022-08-25
Reconcile ovnk in case node IP changes

zshi-redhat:
Closes #[NP-538](https://issues.redhat.com//browse/NP-538)

### DIFF
--- a/assets/components/ovn/master/daemonset.yaml
+++ b/assets/components/ovn/master/daemonset.yaml
@@ -215,10 +215,6 @@ spec:
           value: info
         - name: OVN_NORTHD_PROBE_INTERVAL
           value: "5000"
-        - name: K8S_NODE_IP
-          valueFrom:
-            fieldRef:
-              fieldPath: status.hostIP
         volumeMounts:
         - mountPath: /run/openvswitch/
           name: run-openvswitch
@@ -319,10 +315,6 @@ spec:
         env:
         - name: OVN_LOG_LEVEL
           value: info
-        - name: K8S_NODE_IP
-          valueFrom:
-            fieldRef:
-              fieldPath: status.hostIP
         volumeMounts:
         - mountPath: /run/openvswitch/
           name: run-openvswitch
@@ -349,6 +341,9 @@ spec:
             source "/env/_master"
             set +o allexport
           fi
+
+          # K8S_NODE_IP triggers reconcilation of this daemon when node IP changes
+          echo "$(date -Iseconds) - starting ovnkube-master, Node: ${K8S_NODE} IP: ${K8S_NODE_IP}"
 
           echo "I$(date "+%m%d %H:%M:%S.%N") - copy ovn-k8s-cni-overlay"
           cp -f /usr/libexec/cni/ovn-k8s-cni-overlay /cni-bin-dir/
@@ -431,6 +426,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: K8S_NODE_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.hostIP
         securityContext:
           privileged: true
         terminationMessagePolicy: FallbackToLogsOnError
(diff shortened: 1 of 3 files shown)


dhellmann (commented):
This is a much more extensive change than I expected to see.

Is it necessary to delete the entire Deployment and all of the configuration for the CNI plugin? Or can we scale it to 0 replicas and then let everything be re-rendered with the new settings when we start up again?

Do we need to add a `Delete()` method to the `readerApplier` interface, where we have to provide an implementation for everything, or can we make this a special case since it only applies to networking right now?

zshi-redhat:
> This is a much more extensive change than I expected to see.
> 
> Is it necessary to delete the entire Deployment and all of the configuration for the CNI plugin? Or can we scale it to 0 replicas and then let everything be re-rendered with the new settings when we start up again?

Yes, we only need to delete the Daemonset. It keeps all the rest configuration unchanged. Deployment is currently not used. According to test, the deletion may not be finished immediately after the execution, but the next boot or start will continue the deletion and restart the pods.
> 
> Do we need to add a `Delete()` method to the `readerApplier` interface, where we have to provide an implementation for everything, or can we make this a special case since it only applies to networking right now?

I updated the PR to remove `Delete()` method of the applier and add a special delete method to ovnk.

pkg/assets/apps.go
```diff
@@ -115,3 +117,23 @@ func ApplyDaemonSets(apps []string, render RenderFunc, params RenderParams, kube
 	ds.Client = appsClient(kubeconfigPath)
 	return applyApps(apps, ds, render, params)
 }
+
+func DeleteDaemonSets(apps []string, render RenderFunc, params RenderParams, kubeconfigPath string) error {
```
mangelajo, line 121: I think this will not work when:

1) microshift is stopped ( shutdown ) 
2) node boot with different IP address
3) microshift boots


I recommend instead during boot, we try to find incorrect daemonset configs (wrong IPs), and delete those before creating the new (or simply updating the old if that's possible).

mangelajo (changes requested):
See my comments, I believe this approach is brittle under some conditions.

pkg/assets/apps.go
zshi-redhat: @mangelajo  Currently we don't specify the node ip address in ovnk daemonset, it is auto-detected in ovnk golang code, so the idea here is to restart ovnk daemonset and let it discover the node ip change automatically.

Re the second step above, ovnk pods will restart and pick the new node ip.

zshi-redhat: The K8S_NODE env var in ovnk daemonset is from pod spec.nodeName, which will be updated by kubelet after microshift restart.

zshi-redhat: @dhellmann oh, good point! we actually can get rid of the logic in a smart way :-) 
I updated the daemonset to use K8S_NODE_IP as env var.

mangelajo: I liked this approach much more thanks @dhellmann & @zshi-redhat good work

mangelajo (approved):

zshi-redhat:
CI failed due to subscription error:
```
+ gcloud compute --project XXXXXXXXXXXXXXXXXXXXXX ssh --zone us-central1-c rhel8user@ci-op-ihc66yjk-68c08 --command 'sudo subscription-manager repos --enable rhocp-4.10-for-rhel-8-x86_64-rpms --enable fast-datapath-for-rhel-8-x86_64-rpms'
Error: 'rhocp-4.10-for-rhel-8-x86_64-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.
```

mangelajo:
let's retest in a while, it's probably some transient error in the CDN?  (I hope) %)

mangelajo:
``` # Loading microshift-container blueprint v0.0.1
ERROR: BlueprintsError: microshift-container: DNF error occured: RepoError: Error occurred when setting up repo: Failed to download metadata for repo '2': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
Execution time: 2 minutes ```

I'm seeing a similar issue locally when trying to build an iso

PR #885, opened 2022-08-25
Use OpenShift Infra NS from KCM operator

oglok:

The `openshift-infra` namespace is now gathered from the `cluster-kube-controller-manager-operator` assets.

### DIFF
--- a/assets/core/namespace-openshift-infra.yaml
+++ b/assets/core/namespace-openshift-infra.yaml
@@ -1,14 +1,6 @@
-kind: Namespace
 apiVersion: v1
+kind: Namespace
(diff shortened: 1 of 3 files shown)


dhellmann:
I see the asset being updated, but I don't see a change to the rebase script. What causes the manifest to come from the KCM operator?

oglok:
> I see the asset being updated, but I don't see a change to the rebase script. What causes the manifest to come from the KCM operator?

Good catch, I always forget about updating the rebase script. Done!

dhellmann:
/close

See #890 instead

PR #886, opened 2022-08-25
compact ovn databases periodically

zshi-redhat:
ovn database may grow to several Megabytes, compact
the database in a short cadence (default: 10s) other
than waiting until automatic compact in 24 hours.

### DIFF
--- a/assets/components/ovn/master/daemonset.yaml
+++ b/assets/components/ovn/master/daemonset.yaml
@@ -95,7 +95,7 @@ spec:
             memory: 10Mi
         terminationMessagePolicy: FallbackToLogsOnError
 
-      # nbdb: the northbound, or logical network object DB. In raft mode
+      # nbdb: the northbound, or logical network object DB. In standalone mode
       - name: nbdb
         image: {{ .ReleaseImage.ovn_kubernetes }}
         command:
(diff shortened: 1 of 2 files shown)


ggiguash:
@zshi-redhat, I'm no expert in this, but it seems to me that 10s cadence is too frequent for compaction operation. 
Should we make it less frequent, i.e. once in 10m or 30m?

zshi-redhat:
> @zshi-redhat, I'm no expert in this, but it seems to me that 10s cadence is too frequent for compaction operation. Should we make it less frequent, i.e. once in 10m or 30m?

@ggiguash That makes sense. I think the database size won't grow so fast that needs to be trimmed every 10s.
The period seconds was default (10s) in openshift manifests since it needs to check database leader for readiness which is not needed for microshift (standalone mode).
I updated the periodSeconds to 10m.

ggiguash (approved):

oglok:
How cpu intensive is this operation?

zshi-redhat:
> How cpu intensive is this operation?

@oglok good question!
If I run the compact cmd in a while loop, the cpu surges to 33.3% high
If I run the compact cmd in a while loop, with 1 second sleep, the max cpu observed is 0.7%
If I run the compact cmd in a while loop, with 5 seconds sleep, the max cpu observed is 0.3%
If I run the compact cmd in a while loop, with 10 seconds sleep, the max cpu observed is 0.3% stable

The above result is a summary of all ovsdb-server processes running on the host, so not specific to compact cmd (I didn't find a better way to measure the specific compact cmd).

The argument could be is it worth reducing database size by increase the cpu usage every 10 mins?

oglok:
Great analysis @zshi-redhat ! I was just wondering if the actual db compaction may create a CPU load spike that could create some inestability in the host. It doesn't seem to be the case with long periods, so 10 mins seems very reasonable.

dhellmann:
USHIFT-384

PR #888, opened 2022-08-26
USHIFT-136: Replace Microshift by MicroShift 

ggiguash:
Replace Microshift by MicroShift in two commits:
- With "Match Whole Word"
- ~~Without "Match Whole Word"~~


dhellmann:
/kind cleanup

I think we could fast-track a PR to update user-visible instances of the name and copy right strings. I don't know about the impact of the other changes and whether it's going to trigger rebases in ongoing work.

PR #889, opened 2022-08-26
Always do force cleanup

zshi-redhat:
Closes #[USHIFT-324](https://issues.redhat.com//browse/USHIFT-324)

### DIFF
--- a/hack/cleanup.sh
+++ b/hack/cleanup.sh
@@ -17,16 +17,8 @@ sudo bash -c "
     systemctl disable microshift 2>/dev/null
     pkill -9 microshift
 
-    echo Removing non-OVN crio pods
-    NOVN_PODS=\$(crictl pods | tail -n +2 | grep -vE openshift-ovn-kubernetes | awk '{print \$1}')
-    [ ! -z \"\${NOVN_PODS}\" ] && crictl stopp \${NOVN_PODS}
-    [ ! -z \"\${NOVN_PODS}\" ] && crictl rmp   \${NOVN_PODS}
-
-    echo Removing all crio pods
-    until crictl rmp --all --force 1>/dev/null; do sleep 1; done
-
     echo Removing crio container and image storage
-    crio wipe -f &>/dev/null
+    crio wipe -f &>/dev/null || true
     systemctl restart crio
 
     echo Killing conmon, pause and OVN processes


hack/cleanup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -eu
+set -u
```
ggiguash, line 2: I would not remove "fail on error" option. If you'd like a given command to always succeed, please add "|| true" to it.

For example:
crio wipe -f || true

ggiguash (changes requested):

hack/cleanup.sh
zshi-redhat: done

oglok:
I don't understand, if we only use `crio wipe` we don't have control over the order of which pods will be removed first. If it happens to be the OVN ones, we'll hit the issue we had previously about `no ovn network found`. 

Am I missing something?

zshi-redhat:
> I don't understand, if we only use `crio wipe` we don't have control over the order of which pods will be removed first. If it happens to be the OVN ones, we'll hit the issue we had previously about `no ovn network found`.
> 
> Am I missing something?

`crio wipe -f` force wipes all container images regardless of ovn pods status, together with `systemctl restart crio`, all running containers will be stopped and removed.  so this commit takes a force approach here to avoid cases that ovnk pods may not always be available.
 
For example: cleanup.sh script is executed after a reboot, not executed immediately followed by stopping microshift service, in such case, ovnk pods are not available to respond CNI request (from `crictl rmp/stopp`) which results in infinite loop.

oglok:
> > I don't understand, if we only use `crio wipe` we don't have control over the order of which pods will be removed first. If it happens to be the OVN ones, we'll hit the issue we had previously about `no ovn network found`.
> > Am I missing something?
> 
> `crio wipe -f` force wipes all container images regardless of ovn pods status, together with `systemctl restart crio`, all running containers will be stopped and removed. so this commit takes a force approach here to avoid cases that ovnk pods may not always be available.
> 
> For example: cleanup.sh script is executed after a reboot, not executed immediately followed by stopping microshift service, in such case, ovnk pods are not available to respond CNI request (from `crictl rmp/stopp`) which results in infinite loop.

I see, to be honest I'm used to do cleanups right after I stop MicroShift's service and not after reboots. If this has been tested, I'm ok with merging it.

dhellmann:
> I see, to be honest I'm used to do cleanups right after I stop MicroShift's service and not after reboots. If this has been tested, I'm ok with merging it.

I'm used to doing that, too. However, if the system isn't in exactly the right state when the current implementation is used then it doesn't work. If you run `cleanup.sh` twice in a row, for example, the second attempt hangs trying to kill the extra containers.

PR #893, opened 2022-08-27
skip extra make step in dev environment automation

dhellmann:
Building the rpm runs make to build the binary so there is no need to
run it separately in the dev environment configure script.

/kind cleanup
/assign @ggiguash


ggiguash:
USHIFT-315
/lgtm

PR #895, opened 2022-08-28
update create-vm.sh to support taking inputs via environment variables

dhellmann:
This change makes it possible to create a wrapper script that sets
variables instead of passing arguments on the command line, which
makes it easier to remember what each value means.

/type cleanup
/assign @ggiguash


ggiguash:
USHIFT-315
/lgtm

PR #896, opened 2022-08-29
Make nodeport range configurable

zshi-redhat:
Setting service nodeport range to KubeAPIServerConfig.ServicesNodePortRange
doesn't take effect, needs to use service-node-port-range in
KubeAPIServerConfig.APIServerArguments

Closes #[USHIFT-334](https://issues.redhat.com//browse/USHIFT-334)

### DIFF
--- a/assets/components/openshift-router/service-cloud.yaml
+++ b/assets/components/openshift-router/service-cloud.yaml
@@ -15,8 +15,6 @@ spec:
     - name: http
       port: 80
       targetPort: 80
-      nodePort: 30001
     - name: https
       port: 443
       targetPort: 443
-      nodePort: 30002
--- a/pkg/controllers/kube-apiserver.go
+++ b/pkg/controllers/kube-apiserver.go
@@ -127,6 +127,7 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 			"proxy-client-key-file":            {filepath.Join(kasSecretsDir, "aggregator-client", "tls.key")},
 			"requestheader-client-ca-file":     {clientCABundlePath},
 			"service-account-signing-key-file": {cfg.DataDir + "/resources/kube-apiserver/secrets/service-account-key/service-account.key"},
+			"service-node-port-range":          {cfg.Cluster.ServiceNodePortRange},
 			"tls-cert-file":                    {filepath.Join(servingCertsDir, "tls.crt")},
 			"tls-private-key-file":             {filepath.Join(servingCertsDir, "tls.key")},
 			"disable-admission-plugins": {
--- a/pkg/assets/bindata.go
+++ b/pkg/assets/bindata.go
@@ -2695,11 +2695,9 @@ spec:
     - name: http
       port: 80
       targetPort: 80
-      nodePort: 30001
     - name: https
       port: 443
       targetPort: 443
-      nodePort: 30002
 `)
 
 func assetsComponentsOpenshiftRouterServiceCloudYamlBytes() ([]byte, error) {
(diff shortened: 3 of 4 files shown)


assets/components/openshift-router/service-cloud.yaml
```diff
@@ -15,8 +15,6 @@ spec:
     - name: http
       port: 80
       targetPort: 80
-      nodePort: 30001
```
dhellmann, line 18: Do we need the rebase script to automate this change?

pkg/controllers/kube-apiserver.go
```diff
@@ -125,6 +125,7 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 			"proxy-client-key-file":            {cfg.DataDir + "/certs/kube-apiserver/secrets/aggregator-client/tls.key"},
 			"requestheader-client-ca-file":     {clientCABundlePath},
 			"service-account-signing-key-file": {cfg.DataDir + "/resources/kube-apiserver/secrets/service-account-key/service-account.key"},
+			"service-node-port-range":          {cfg.Cluster.ServiceNodePortRange},
```
dhellmann, line 128: How is this different from the setting on line 171?

assets/components/openshift-router/service-cloud.yaml
zshi-redhat: Yes, the node ports are added in the rebase script.
Updated the PR to remove the nodeport customization.

@fzdarsky  I noticed you added the node port customization in rebase script, is there any speical consideration of using nodeport 30001 and 30002?

pkg/controllers/kube-apiserver.go
zshi-redhat: I think line 171 (ServicesNodePortRange) was an internal struct that holds the value of service-node-port-range back in 4.1 code: https://github.com/openshift/origin/blob/release-4.1/pkg/cmd/openshift-kube-apiserver/openshiftkubeapiserver/flags.go#L141

In latest openshift, it only exists in the vendor folder for setting the default port range value: https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/openshift-kube-apiserver/configdefault/kubecontrolplane_default.go#L67, not sure if it is actually referenced in any other code.

[service-node-port-range](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go#L203) is the supported option in upstream kube-apiserver.

@deads2k Could you help confirm that ServicesNodePortRange is not used for specifying customized port range (I noticed from git history that you used to update code related to the service-node-port-range)?

assets/components/openshift-router/service-cloud.yaml
mangelajo: What is happening when we remove nodePort here

zshi-redhat: @mangelajo It will allocate a random node port from the specified range.
This change may have side effect if we rely on hardcoded node port in microshift. I searched the repo and didn't find other reference to the fixed node port, but I'd like to double check.

benluddy:
I also don't see anything reading the value of ServicesNodePortRange inside KubeAPIServerConfig. On OCP, this is configured in APIServerArguments by one of the (Network) config observers inside cluster-kube-apiserver-operator. I'm sure we'll discuss how to better unify OCP and MicroShift configuration, but that's independent of ServicesNodePortRange vs. APIServerArguments.

/lgtm

zshi-redhat:
fixed merge conflict.

PR #898, opened 2022-08-29
AUTH-296: Separate kubelet client trust

stlaz:
This PR separates the CAs that are trusted by the kubelet for client cert authentication.

The `cmd/init.go` is starting to be a little messy, we will probably want to refactor it eventually.

---

This is currenly untested, I would be surprised if everything worked because this finally cuts the trust of a kubelet to the ultimate bundle for client cert authentication, but let's see what happens,

### DIFF
--- a/pkg/util/cryptomaterial/certpaths.go
+++ b/pkg/util/cryptomaterial/certpaths.go
@@ -13,8 +13,9 @@ const (
 	ClientCertFileName = "client.crt"
 	ClientKeyFileName  = "client.key"
 
-	ClientCAValidityDays   = 60
-	ClientCertValidityDays = 30
+	ClientCAValidityDays                  = 60
+	ClientCertValidityDays                = 30
+	AdminKubeconfigClientCertValidityDays = 365 * 10
 )
 
 func CertsDirectory(dataPath string) string { return filepath.Join(dataPath, "certs") }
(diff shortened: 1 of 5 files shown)


dhellmann:
also related to USHIFT-278

stlaz:
/test e2e-openshift-conformance-sig-network

The sig-network test failed with:
`{  failed to acquire lease for "gcp-quota-slice": resources not found}`

stlaz:
/test all
I do not know how this repo CI works but perhaps all needs to be run in order for the tests to succeed

stlaz:
Looks like a subscription issue. I should point out that all the tests except `sig-network` passed before, and this test did not pass because of gpc-quotas.

pkg/util/cryptomaterial/certpaths.go
```diff
@@ -11,8 +13,9 @@ const (
 	ClientCertFileName = "client.crt"
 	ClientKeyFileName  = "client.key"
 
-	ClientCAValidityDays   = 60
-	ClientCertValidityDays = 30
+	ClientCAValidityDays                  = 60
+	ClientCertValidityDays                = 30
+	AdminKubeconfigClientCertValidityDays = 365 * 10
```
pacevedom, line 18: Is this documented/required anywhere?

pacevedom (commented):
One minor comment.

I am having the same issues with subscription manager in my local env too.

pkg/util/cryptomaterial/certpaths.go
stlaz: the documentation is to be found here: https://github.com/openshift/api/blob/master/tls/docs/kube-apiserver%20Client%20Certificates/README.md#admin-kubeconfig-signer

I noticed that the client CAs are actually a bit more diverse in their validity configurations, I will be fixing that up in a follow up along with a refactoring of how this works.

PR #900, opened 2022-08-29
Update RPM post install script to return 0 correctly

### LINKED ISSUE #899: USHIFT-361: RPM post install script has non-zero return
#### What happened:

When installing the microshift RPMs the post install script results in a non-zero return code.

RPM post install script should have a 0 return code on normal system installs.

1. dnf install <microshift rpms files"

#### Environment:
- MicroShift Version: 4.10.0-0.microshift-2022-08-08-151458
  Base OCP Version: 4.10.18
- RHEL 8
- 4.18.0-372.19.1.el8_6.x86_64

#### Relevant Logs 

Here is the original error messsage:

```
2022-08-29T15:20:29+0000 INFO warning: %post(microshift-4.10.0-2022_08_08_151458.el8.x86_64) scriptlet failed, exit status 3

2022-08-29T15:20:29+0000 ERROR Error in POSTIN scriptlet in rpm package microshift
```

The error is because of this section of the script
```
%systemd_post microshift.service

# only for install, not on upgrades
if [ $1 -eq 1 ]; then
        # if crio was already started, restart it so it will catch /etc/crio/crio.conf.d/microshift.conf
fi
```

The issue is the use of the "AND" operator. If crio is not running, the first command returns "3" so the overall script return is 3. The script shouldn't return a non-zero code for normal operations. I'm submitting a PR that fixes this issue and returns 0 in either case.

dmc5179:
Closes #899

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -192,7 +192,10 @@ install -m644 packaging/selinux/microshift.pp.bz2 %{buildroot}%{_datadir}/selinu
 # only for install, not on upgrades
 if [ $1 -eq 1 ]; then
 	# if crio was already started, restart it so it will catch /etc/crio/crio.conf.d/microshift.conf
-	systemctl is-active --quiet crio && systemctl restart --quiet crio
+	if systemctl is-active --quiet crio
+	then
+		systemctl restart --quiet crio
+	fi
 fi
 
 %post selinux


packaging/rpm/microshift.spec
```diff
@@ -192,7 +192,10 @@ install -m644 packaging/selinux/microshift.pp.bz2 %{buildroot}%{_datadir}/selinu
 # only for install, not on upgrades
 if [ $1 -eq 1 ]; then
 	# if crio was already started, restart it so it will catch /etc/crio/crio.conf.d/microshift.conf
-	systemctl is-active --quiet crio && systemctl restart --quiet crio
+	if systemctl is-active --quiet crio
+	then
+		systemctl restart --quiet crio
+	fi
```
ggiguash, line 198:
```suggestion
        systemctl is-active --quiet crio && systemctl restart --quiet crio || true
```

ggiguash (changes requested):

packaging/rpm/microshift.spec
ggiguash: There are few other similar statements in the file, so for the consistency sake...

ggiguash:
The fix implemented in #911 
/close

PR #905, opened 2022-08-30
align dev install with prod and simplify

fzdarsky:
Changes the dev env documentation + automation script to align with the way we'd expect users to install MicroShift:
* microshift*.rpms (should) pull in package deps, so don't install them manually.
* MicroShift's service units (should) take care of starting service units they depend on, so do not start them manually.
* Install oc and kubectl clients from the repos rather than side-loading them.


dhellmann:
Good points

/lgtm

PR #909, opened 2022-08-31
Remove KAS kubeconfig

stlaz:
The kube-apiserver kubeconfig is likely unused. This PR removes this file with privileged credentials.

The kubeconfig appears to have only been used to generate a REST client for KAS "/readyz" check. Let's see if we can easily replace that by just supplying an anonymous client instead for this check.


dhellmann:
The "likely unused" part of the description, along with a lack of confidence in our CI coverage, combined make me concerned about merging this patch without understanding how to test it. Is there something I can do to poke at a running instance to see if the change is working properly?

stlaz:
What Ben says - I found it odd that the kube-apiserver would need a kubeconfig but I did not want to rule anything out.

Turns out it's only ever used for the /readyz checking, and so this PR is safe to merge

pkg/controllers/kube-apiserver.go
```diff
@@ -262,14 +260,15 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
-			restConfig, err := clientcmd.BuildConfigFromFlags("", s.kubeconfig)
+			restConfig, err := clientcmd.BuildConfigFromFlags(s.masterURL, "")
 			if err != nil {
 				return false, err
 			}
 			if err := rest.SetKubernetesDefaults(restConfig); err != nil {
 				return false, err
 			}
 			restConfig.NegotiatedSerializer = serializer.NewCodecFactory(runtime.NewScheme())
+			restConfig.CAFile = s.servingCertPath
```
pacevedom, line 271: Should we use the CA from https://github.com/openshift/microshift/blob/main/pkg/controllers/kube-apiserver.go#L96?

stlaz: Since this is just a startup test and we are specifically interested in checking this very server, I believe a simple trust such as we've got here is adequate

pacevedom (approved):
/lgtm

PR #913, opened 2022-08-31
USHIFT-364: add command to show default configuration

dhellmann:
Add a command to show the default configuration, modified to make it
suitable for global use on the host.

Related to [USHIFT-19](https://issues.redhat.com//browse/USHIFT-19)

### DIFF
--- a/pkg/cmd/showConfig.go
+++ b/pkg/cmd/showConfig.go
@@ -1,0 +1,59 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/yaml"
+
+	"github.com/openshift/microshift/pkg/config"
+)
+
+type showConfigOptions struct {
+	Mode string
+	genericclioptions.IOStreams
+}
+
+func NewShowConfigCommand(ioStreams genericclioptions.IOStreams) *cobra.Command {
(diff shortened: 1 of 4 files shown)


fzdarsky:
I wonder whether users may also want/need the ability to show the _effective_ configuration, i.e. the configuration MicroShift uses after merging default, inputs from config.yaml, env vars, and auto-detected values? In which case we may want to think whether the selection btw. default and effective would be a flag to a `config` or `show-config` command?

dhellmann:
/assign @fzdarsky 

Updated to `show-config`

pkg/cmd/showConfig.go
```diff
@@ -0,0 +1,59 @@
...
+	cmd := &cobra.Command{
+		Use:   "show-config",
+		Short: "Print MicroShift's configuration",
+		Run: func(cmd *cobra.Command, args []string) {
+
+			switch opts.Mode {
+			case "default":
+				// Override a few settings to always use the global locations
+				cfg.DataDir = config.DefaultGlobalDataDir
+				cfg.Manifests = cfg.Manifests[:2]
```
mangelajo, line 35:
```suggestion
				cfg.Manifests = cfg.Manifests[:len(cfg.Manifests)-1]
```
Just in case we add other paths in the future, although I don't see that likely..

mangelajo (changes requested):
Small suggestion

dhellmann:
/retitle USHIFT-364: add command to show default configuration

dhellmann:
@mangelajo I've updated the PR as you suggested.

PR #914, opened 2022-08-31
USHIFT-19: add example config file to rpm

### LINKED ISSUE #663: USHIFT-19: Ship default config file in rpm

Commonly- Linux packaged software includes an example configuration file with the defaults, or optionals
in comments shipped to the etc directory, making customization straight forward during development.

#### Why is this needed:

More common workflow
Easier for developers.

For example I needed several minutes of crawling through documents, while simply looking for the etc
file is the first thing I attempted.

The etc config could contain explanations to the fields, becoming a self-document file.

dhellmann:
Add /etc/microshift/microshift.yaml with the current defaults,
all commented out.

Closes #663


packaging/rpm/microshift.spec
```diff
@@ -166,6 +166,9 @@ install -p -m644  packaging/network-manager-conf/microshift-nm.conf %{buildroot}
 install -d -m755 %{buildroot}/%{_unitdir}
 install -p -m644 packaging/systemd/microshift.service %{buildroot}%{_unitdir}/microshift.service
 
+install -d -m755 %{buildroot}/%{_sysconfdir}/microshift
+install -p -m644 packaging/microshift/microshift.yaml %{buildroot}%{_sysconfdir}/microshift/microshift.yaml
```
dhellmann, line 170: I wasn't sure whether to install the file to the real location or to an "example.yaml" location. I decided to go for the real location, but I'm open to changing that.

PR #916, opened 2022-08-31
USHIFT-336: Inject embedded component names into klog headers.

benluddy:
Adds servicemanager service names as a new field to klog headers, like this:

```
Aug 31 21:43:28 microshift.localdomain microshift[598329]: I0831 21:43:28.008100  598329 kube-apiserver kube-apiserver.go:295] "kube-apiserver" is ready
Aug 31 21:43:28 microshift.localdomain microshift[598329]: I0831 21:43:28.008132  598329 kube-scheduler manager.go:114] Starting kube-scheduler
Aug 31 21:43:28 microshift.localdomain microshift[598329]: I0831 21:43:28.008151  598329 kube-controller-manager manager.go:114] Starting kube-controller-manager
Aug 31 21:43:28 microshift.localdomain microshift[598329]: I0831 21:43:28.008156  598329 openshift-crd-manager manager.go:114] Starting openshift-crd-manager
Aug 31 21:43:28 microshift.localdomain microshift[598329]: I0831 21:43:28.008787  598329 openshift-crd-manager crd.go:121] Applying openshift CRD assets/crd/0000_03_securityinternal-openshift_02_rangeallocation.crd.yaml
Aug 31 21:43:28 microshift.localdomain microshift[598329]: I0831 21:43:28.136916  598329 kube-scheduler serving.go:348] Generated self-signed cert in-memory
Aug 31 21:43:28 microshift.localdomain microshift[598329]: I0831 21:43:28.199799  598329 kube-controller-manager serving.go:348] Generated self-signed cert in-memory
```

This patch takes advantage of existing runtime support for goroutine labels (https://github.com/golang/proposal/blob/master/design/17280-profile-labels.md), which are propagated to spawned goroutines from the spawning goroutine. I feel bad about it.

### DIFF
--- a/scripts/rebase_patches/0020-klog-component-names.patch
+++ b/scripts/rebase_patches/0020-klog-component-names.patch
@@ -1,0 +1,63 @@
+From 362ddacd8704c1e124d4831b2a301aeec0d34641 Mon Sep 17 00:00:00 2001
+From: Ben Luddy <bluddy@redhat.com>
+Date: Wed, 31 Aug 2022 17:36:35 -0400
+Subject: [PATCH] Hack to tag klog headers with embedded component names.
+
+---
+ vendor/k8s.io/klog/v2/goroutine_labels.go | 30 ++++++++++++++++++++++++++++++
+ vendor/k8s.io/klog/v2/klog.go             |  2 ++
+ 2 files changed, 32 insertions(+)
+ create mode 100644 goroutine_labels.go
+
+diff --git a/vendor/k8s.io/klog/v2/goroutine_labels.go b/vendor/k8s.io/klog/v2/goroutine_labels.go
+new file mode 100644
+index 0000000..b9a7356
+--- /dev/null
++++ b/vendor/k8s.io/klog/v2/goroutine_labels.go
+@@ -0,0 +1,30 @@
++package klog
++
++import (
++	"context"
++	"runtime/pprof"
++	"unsafe"
++)
++
++//go:linkname runtime_getProfLabel runtime/pprof.runtime_getProfLabel
++func runtime_getProfLabel() unsafe.Pointer
++
++func getMicroshiftLoggerComponent() string {
++	labels := (*map[string]string)(runtime_getProfLabel())
++	if labels == nil {
++		return "???"
++	}
++
++	c, ok := (*labels)["microshift_logger_component"]
++	if !ok {
++		return "???"
++	}
++
++	return c
++}
++
++func WithMicroshitLoggerComponent(c string, f func()) {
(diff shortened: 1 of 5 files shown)


scripts/rebase_patches/0020-klog-component-names.patch
```diff
@@ -0,0 +1,57 @@
+From 56e7ca38133a9f6c5edc3778e27308be9f8d397c Mon Sep 17 00:00:00 2001
+From: Ben Luddy <bluddy@redhat.com>
+Date: Wed, 31 Aug 2022 17:36:35 -0400
+Subject: [PATCH] Upsetting hack to tag klog with embedded component names.
```
dhellmann, line 4: Would it be less upsetting to write a logr plugin that did this instead of using a patch to klog? I'm fine either way, but if it eases your conscience...

scripts/rebase_patches/0020-klog-component-names.patch
```diff
@@ -0,0 +1,57 @@
...
++	"unsafe"
++)
++
++//go:linkname runtime_getProfLabel runtime/pprof.runtime_getProfLabel
++func runtime_getProfLabel() unsafe.Pointer
++
++func getMicroshiftLoggerComponent() string {
++	labels := (*map[string]string)(runtime_getProfLabel())
++	if labels != nil {
++		return (*labels)["microshift_logger_component"]
```
dhellmann, line 32: Do any of the components themselves start separate goroutines that wouldn't inherit the label? Should this access check if the key is present and return a "???" if it is not?

dhellmann (commented):
Nice. I miss thread-local storage.

scripts/rebase_patches/0020-klog-component-names.patch
benluddy: I don't mind carrying the patch (given its size) as much the dependency on getProfLabel. I hope it will be a relatively stable interface for us given the age of the label feature, though.

scripts/rebase_patches/0020-klog-component-names.patch
benluddy: Since these labels are passed from spawner to spawnee, that should be OK. I think your suggestion is still useful for cases where MicroShift might be using Kubernetes packages directly / outside the scope of an embedded component.

benluddy: Done.

mangelajo:
Awesome @benluddy , I had an idea which was much more complex than this, I didn't know you could label goroutines. My idea involved looking back at the caller stack (slower..., and more error prone)

I love this patch :)

dhellmann:
/retitle USHIFT-336: Inject embedded component names into klog headers.

PR #917, opened 2022-09-01
Wait for default route to become available after boot

zshi-redhat:
MicroShift service runs right after system boot when
dhcp address may not be available on node interface.
This results in configure-ovs.sh script getting non
default route interface on lo device which cannot be
added to ovn gateway interface br-ex.

Closes #[USHIFT-340](https://issues.redhat.com//browse/USHIFT-340)


mangelajo (commented):
/lgtm

PR #918, opened 2022-09-01
USHIFT-256 OpenShift tests static analysis

pmtk:
Closes [USHIFT-256](https://issues.redhat.com/browse/USHIFT-256)

### DIFF
--- a/hack/tests-static-analysis/README
+++ b/hack/tests-static-analysis/README
@@ -1,0 +1,3 @@
+# Static analysis of OpenShift tests to get API usage
+
+See comments in main.go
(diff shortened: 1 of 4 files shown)


dhellmann:
Would it make more sense to put this code in the repo with the tests?

ggiguash:
> Would it make more sense to put this code in the repo with the tests?

Since this is the first implementation and we do not know yet how it's going to be used in CI, I asked to have this code in MicroShift repo. Once we complete the task and understand where it's plugged in CI procedures, we will move the code.

PR #920, opened 2022-09-01
USHIFT-333: update validation for tls in route crd

dhellmann:
Update to copy of file from
openshift/api@0b39f81154fa42abe26933181141ed4b4e3ad247

Note to reviewers: There were a lot of whitespace changes in the YAML file being updated, so I split this PR into 2 commits. The first updates the "base" by using a recent commit in openshift/api that was older than the validation change we care about. The second includes only that validation update.

### DIFF
--- a/assets/crd/0000_01_route.crd.yaml
+++ b/assets/crd/0000_01_route.crd.yaml
@@ -69,6 +69,10 @@ spec:
               - properties:
                   path:
                     maxLength: 0
+              - properties:
+                  tls:
+                    enum:
+                    - null
               - not:
                   properties:
                     tls:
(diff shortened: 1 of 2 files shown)


dhellmann:
I need to make sure we have the right process in place to keep this file up to date.

/hold

dhellmann:
We're updating by hand for now, and have USHIFT-363 to remember to incorporate this file into the rebase automation when we are pulling content from HEAD again.

/hold cancel

benluddy:
/lgtm

Confirmed that Routes generated by the ingress-to-route controller are passing validation with this schema update.

mangelajo (commented):
/lgtm

PR #921, opened 2022-09-02
Use lorax outside of container during iso creation

mangelajo:
This saves a couple of minutes of building in my environment since lorax
is installed just once on the system the first time.


dhellmann:
/override ci/prow/e2e-openshift-conformance-sig-api-machinery

Changes to developer tools won't affect CI jobs.

PR #929, opened 2022-09-06
expand configuration section of design guidelines

dhellmann:
Capture some guidelines that we have been implicitly using in the
discussion of how to expose CSI configuration options.

/assign @copejon @fzdarsky
/type documentation

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -91,6 +91,17 @@ When deciding between different design options, we follow the following principl
 * MicroShift uses as few configuration options as possible. Where it provides configuration options, they are intuitive and have sensible defaults respectively are auto-configured.
 * MicroShift is preferably configured through config files, but allows overriding of parameters via environment variables (for use in containers, systemd) and command line flags (for ad-hoc use).
 * MicroShift can use both user-local and system-wide configuration.
+* Components built into the `microshift` binary are configured by modifying `config.yaml` and restarting the service, instead of through Kubernetes APIs as in standalone OpenShift. This ensures that:
+  * MicroShift can act as an "application running on RHEL", which avoids requiring Kubernetes expertise to deploy it.
+  * Configuration changes can be packaged as part of an ostree update with the OS.
+  * The configurable options can be limited and abstracted for the MicroShift use case.
+  * The complexity of applying configuration changes is minimized by not attempting to reconfigure a running component on the fly.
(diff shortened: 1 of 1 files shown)


dhellmann:
/assign @zshi-redhat @mangelajo @oglok 

FYI

docs/design.md
```diff
@@ -91,6 +91,13 @@ When deciding between different design options, we follow the following principl
 * MicroShift uses as few configuration options as possible. Where it provides configuration options, they are intuitive and have sensible defaults respectively are auto-configured.
 * MicroShift is preferably configured through config files, but allows overriding of parameters via environment variables (for use in containers, systemd) and command line flags (for ad-hoc use).
 * MicroShift can use both user-local and system-wide configuration.
+* Components built into the `microshift` binary are configured by modifying `microshift.conf` and restarting the service, instead of through Kubernetes APIs. This ensures that
```
fzdarsky, line 94: `microshift.conf` --> `/etc/microshift/config.yaml` ?

docs/design.md
```diff
@@ -91,6 +91,13 @@ When deciding between different design options, we follow the following principl
...
+  * MicroShift can act as an "application running on RHEL", which avoids requiring Kubernetes expertise to deploy it.
+  * Configuration changes can be packaged as part of an ostree update with the OS.
+  * The configurable options can be limited and abstracted for the MicroShift use case.
+  * The complexity of applying configuration changes is minimized by not attempting to reconfigure a running component on the fly.
+* Components bundled with MicroShift but not built into the `microshift` binary (CSI and CNI drivers, for example) may be configured with configuration files native to the bundled application (`lvmd` or OVN-K). This avoids importing the configuration API of those bundled components into MicroShift and will make replacing those components in some configurations more straightforward.
```
fzdarsky, line 99: Would we expect hosted components other than CSI and CNI implementations to require this?

I'd argue CSI and CNI implementations (like CRI) are special because users _may_ want to swap them out (therefore we'd want their config to be reasonably opaque to MicroShift). Also, they have a "bottom-half" in the shape of a daemon running on and configuring the host OS - which is what I think we're mainly configuring here.

docs/design.md
dhellmann: Yes, I keep making that mistake.

docs/design.md
dhellmann: I've tried to expand this section a bit. Let me know what you think.

fzdarsky: That captures it nicely, thank you!

fzdarsky (approved):

PR #930, opened 2022-09-06
USHIFT-458: replace openshift-controller-manager with route-controller-manager

atiratree:
Removes  openshift-controller-manager dependency in favor of route-controller-manager

TODO:

- [x] requires kube update to at least 1.24


atiratree:
hack commit is in place to make this buildable https://github.com/openshift/microshift/pull/930/commits/9227fff69840d1adefcf313bdb7c5a156a7c4e04 
/hold

pkg/cmd/run.go
```diff
@@ -104,7 +104,7 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 		util.Must(m.AddService(controllers.NewKubeScheduler(cfg)))
 		util.Must(m.AddService(controllers.NewKubeControllerManager(cfg)))
 		util.Must(m.AddService(controllers.NewOpenShiftCRDManager(cfg)))
-		util.Must(m.AddService(controllers.NewOpenShiftControllerManager(cfg)))
+		util.Must(m.AddService(controllers.NewRouteControllerManager(cfg)))
```
atiratree, line 107: are the commands running with admin privileges? Is RBAC going to be introduced in some way for all these components?

pkg/controllers/openshift-route-controller-manager.go
```diff
@@ -80,34 +80,20 @@ kubeClientConfig:
...
 servingInfo:
   bindAddress: "0.0.0.0:8445"
-  certFile: ` + cfg.DataDir + `/resources/openshift-controller-manager/secrets/tls.crt
-  keyFile:  ` + cfg.DataDir + `/resources/openshift-controller-manager/secrets/tls.key
+  certFile: ` + cfg.DataDir + `/resources/route-controller-manager/secrets/tls.crt
+  keyFile:  ` + cfg.DataDir + `/resources/route-controller-manager/secrets/tls.key
   clientCA: ` + cryptomaterial.UltimateTrustBundlePath(cryptomaterial.CertsDirectory(cfg.DataDir)) + `
 controllers:
 - "openshift.io/ingress-ip"
 - "openshift.io/ingress-to-route"
```
atiratree, line 88: are these configs going to be always hardcoded? Or can a user acccess these via the API?

dhellmann: We don't expect to expose these configuration options to the user.

pkg/cmd/run.go
dhellmann: I think that's part of the work @stlaz is doing, isn't it? See #898 for example.

oglok:
@atiratree I'd need a bit more context here, due to my lack of knowledge. Part of the work in the control plane side was to split all the controllers from OCM to individual ones that can be selected in MicroShift. 

This PR replaces the entire OCM for just the route-controller-manager.

- Does this mean that the rest of the functionality within OCM is completely useless for MicroShift? Or will there be more PRs to enable others?

- Does the route-controller-manager handles validation and dynamic defaulting of values that we cannot achieve using the route CRD currently?

Thanks for your work here!

atiratree:
> @atiratree I'd need a bit more context here, due to my lack of knowledge. Part of the work in the control plane side was to split all the controllers from OCM to individual ones that can be selected in MicroShift.
> 
> This PR replaces the entire OCM for just the route-controller-manager.
> 
> Does this mean that the rest of the functionality within OCM is completely useless for MicroShift? Or will there be more PRs to enable others?

Yes, as far as I know, only the route controllers are required from OCM.

> 
>Does the route-controller-manager handles validation and dynamic defaulting of values that we cannot achieve using the route CRD currently?

the API part will be part of the patched kube-apiserver and this should be tracked in jira https://issues.redhat.com/browse/API-1419

route-controller-manager only handles the controller part, you can see the details in the README here https://github.com/openshift/route-controller-manager

pkg/controllers/openshift-route-controller-manager.go
```diff
@@ -24,48 +24,48 @@ import (
...
 
-	var configFilePath = cfg.DataDir + "/resources/openshift-controller-manager/config/config.yaml"
+	var configFilePath = cfg.DataDir + "/resources/route-controller-manager/config/config.yaml"
 	s.kubeconfig = filepath.Join(cfg.DataDir, "resources", "kubeadmin", "kubeconfig")
 	s.ConfigFilePath = configFilePath
 	s.Output = os.Stdout
 }
 
-func (s *OCPControllerManager) writeConfig(cfg *config.MicroshiftConfig) error {
+func (s *OCPRouteControllerManager) writeConfig(cfg *config.MicroshiftConfig) error {
```
oglok, line 68: Another comment that came to mind. The idea is to standardize the way we configure the embeded components to avoid writing config files in disk. 

Could you please update this PR to use structs instead like the CPC?

https://github.com/openshift/microshift/blob/main/pkg/controllers/cluster-policy-controller.go#L53

atiratree: yes, planning to add it, although it might not be that pretty as CPC since we are not using the controller builder (yet)

atiratree:
I have added direct passing of the config

pkg/controllers/openshift-route-controller-manager.go
```diff
@@ -56,44 +58,37 @@ func (s *OCPRouteControllerManager) Dependencies() []string {
...
-- "openshift.io/ingress-ip"
-- "openshift.io/ingress-to-route"
-`)
-
-	path := filepath.Join(cfg.DataDir, "resources", "route-controller-manager", "config", "config.yaml")
-	os.MkdirAll(filepath.Dir(path), os.FileMode(0700))
-	return ioutil.WriteFile(path, data, 0644)
+	c := &openshiftcontrolplanev1.OpenShiftControllerManagerConfig{
+		KubeClientConfig: configv1.KubeClientConfig{
+			KubeConfig: s.kubeconfig,
```
oglok, line 69: Could you please add the ContentType param of ConnectionOverrides like this:

https://github.com/openshift/microshift/pull/988/commits/c630434974e3f61bb5ab5938e292ba0ed68417f3

We would want to avoid issues in case protobuf is the default format for communication.

atiratree: done, and also set `BindNetwork` to  `tcp4`

atiratree:
done!

PR #932, opened 2022-09-07
add mtu config option

zshi-redhat:
Closes #[USHIFT-369](https://issues.redhat.com//browse/USHIFT-369)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -43,6 +43,7 @@ type ClusterConfig struct {
 	ServiceNodePortRange string `yaml:"serviceNodePortRange"`
 	DNS                  string `yaml:"dns"`
 	Domain               string `yaml:"domain"`
+	MTU                  string `yaml:"mtu"`
 }
 
 type ControlPlaneConfig struct {
@@ -104,6 +105,7 @@ func NewMicroshiftConfig() *MicroshiftConfig {
 			ServiceNodePortRange: "30000-32767",
 			DNS:                  "10.43.0.10",
 			Domain:               "cluster.local",
+			MTU:                  "1400",
 		},
 		ControlPlane: ControlPlaneConfig{},
 		Node:         NodeConfig{},
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -52,6 +52,7 @@ func TestCommandLineConfig(t *testing.T) {
 					ServiceNodePortRange: "1024-32767",
 					DNS:                  "cluster.dns",
 					Domain:               "cluster.local",
+					MTU:                  "1200",
 				},
 				Manifests: []string{defaultManifestDirLib, defaultManifestDirEtc, "/tmp/microshift/data/manifests"},
 			},
@@ -75,6 +76,7 @@ func TestCommandLineConfig(t *testing.T) {
 		flags.StringVar(&config.Cluster.ServiceNodePortRange, "service-node-port-range", "", "")
 		flags.StringVar(&config.Cluster.DNS, "cluster-dns", "", "")
 		flags.StringVar(&config.Cluster.Domain, "cluster-domain", "", "")
+		flags.StringVar(&config.Cluster.MTU, "mtu", "1200", "")
 
 		// parse the flags
 		flags.Parse([]string{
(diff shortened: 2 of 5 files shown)


fzdarsky:
/hold

This PR's approach to passing in configuration is not consistent with the one we converged on for the CSI (result of the discussion captured in #929) and would thus need more thorough discussion.

@dhellmann ^

mangelajo:
> /hold
> 
> This PR's approach to passing in configuration is not consistent with the one we converged on for the CSI (result of the discussion captured in #929) and would thus need more thorough discussion.
> 
> @dhellmann ^

In this case it's different, this won't configure MTU on br-ex, this will configure internal MTU for the pods,

 we cannot use a config from the system as ovnk does not have that. We could discover the MTU on the system and apply that - overhead. Although, in the case of the issue the customer is facing they need this as a solution when the next hops still have smaller MTU than br-ex or eth0, and the PMTU discovery is not working for some reason (VPN does not emit the needfrag ICMPs correctly or other reasons)

given those arguments and the intent of the patch I believe the PR is good

fzdarsky:
> In this case it's different, this won't configure MTU on br-ex, this will configure internal MTU for the pods,

So it's not what the OpenShift docs call "cluster network MTU" and that gets configured by the NO on the host (via NM), right?

Is this specific to OVNk or something that's common in CNI drivers? I'm asking because the proposal is to extend MicroShift's configuration API by adding the parameter `NetworkPlugin.OVNKubernetesConfig.MTU`.

zshi-redhat:
> > In this case it's different, this won't configure MTU on br-ex, this will configure internal MTU for the pods,
> 
> So it's not what the OpenShift docs call "cluster network MTU" and that gets configured by the NO on the host (via NM), right?

The MTU won't be configured to host physical interfaces (eth0), but to `br-int` and pod interfaces.

> 
> Is this specific to OVNk or something that's common in CNI drivers? I'm asking because the proposal is to extend MicroShift's configuration API by adding the parameter `NetworkPlugin.OVNKubernetesConfig.MTU`.

Specific to OVNK which uses overlay (geneve tunnel), other CNIs may or may not use tunnel.
This MTU setting tells OVNK which value should be used when creating CNI-network pods.

mangelajo:
Talking to @fzdarsky here, we believe that a possible approach is to move the MTU parameter up with the ClusterCIDR and ServiceCIDR parameters, as almost any driver will setup/accept MTU on the pods and cluster network.

This way we dont leak up OVNk in the config, and we can still keep in sync the ClusterCIDR and ServiveCIDR parameters in ovnkube.conf (which would be hard if we moved that file out)

pkg/components/networking.go
```diff
@@ -63,6 +63,7 @@ func startOVNKubernetes(cfg *config.MicroshiftConfig, kubeconfigPath string) err
 		return err
 	}
 	params := assets.RenderParams{
+		"MTU":            cfg.Cluster.NetworkPlugin.OVNKubernetesConfig.MTU,
```
mangelajo, line 66:
```suggestion
		"MTU":            cfg.Cluster.ClusterMTU
```

or just MTU?

pkg/config/config.go
```diff
@@ -35,14 +35,23 @@ var (
 	validRoles   = []string{"controlplane", "node"}
 )
 
+type OVNKubernetesConfig struct {
```
mangelajo, line 38: Let's not make a separate tree of structs, and use the clusterconfig level as the ClusterCIDR and ServiceCIDR parameters; those must go at the same level

pkg/config/config_test.go
```diff
@@ -90,6 +96,7 @@ func TestCommandLineConfig(t *testing.T) {
 			"--service-node-port-range=" + tt.config.Cluster.ServiceNodePortRange,
 			"--cluster-dns=" + tt.config.Cluster.DNS,
 			"--cluster-domain=" + tt.config.Cluster.Domain,
+			"--cluster-networkplugin-ovnkubernetesconfig-mtu=" + tt.config.Cluster.NetworkPlugin.OVNKubernetesConfig.MTU,
```
mangelajo, line 99:
```suggestion
			"--cluster-mtu=" + tt.config.Cluster.MTU,
```

zshi-redhat:
> Talking to @fzdarsky here, we believe that a possible approach is to move the MTU parameter up with the ClusterCIDR and ServiceCIDR parameters, as almost any driver will setup/accept MTU on the pods and cluster network.
> 
@mangelajo @fzdarsky should I unhold ?

PR #933, opened 2022-09-07
USHIFT-251: Sync more core and CRD manifests

fzdarsky:
Updates rebase script to copy in more of the assets/core and assets/crd manifests from the respective OpenShift components.

Closes [USHIFT-251](https://issues.redhat.com//browse/USHIFT-251)


scripts/rebase.sh
```diff
@@ -389,12 +389,21 @@ update_manifests() {
 
     #-- OpenShift control plane ---------------------------
     # 1) Adopt resource manifests
+    #    Selectively copy in only those core manifests that MicroShift is already using
+    cp "${STAGING_DIR}/cluster-openshift-controller-manager-operator/bindata/v3.11.0/openshift-controller-manager/ns.yaml" "${REPOROOT}"/assets/core/0000_50_cluster-openshift-controller-manager_00_namespace.yaml
+    cp "${STAGING_DIR}/cluster-kube-controller-manager-operator/bindata/assets/kube-controller-manager/csr_approver_clusterrole.yaml" "${REPOROOT}"/assets/core
```
pacevedom, line 394: What do you think about using a for? Would make this easier to read, maybe?

fzdarsky: Yeah, I agree it looks like lots of repetition. Was thinking about a loop over a map from source to dest, but if you look at how little these have in common and if we want to comment a lot, I wasn't confident this would achieve a lot.

Maybe worth a try and a potential follow up PR?

PR #934, opened 2022-09-07
Use yaml key in the example config file

zshi-redhat:


fzdarsky:
I suspect the example got generated from the new `microshift show-config` command. That command marshals the MicroShiftConfig struct (https://github.com/openshift/microshift/blob/main/pkg/cmd/showConfig.go#L47-L50) and therefore generates capitalised names.

We may want to fix up the show-config command to output the names in the annotations instead. I believe using the `https://github.com/ghodss/yaml` package that we're already vendoring would do the trick.

WDYT @dhellmann?

fzdarsky:
> I suspect the example got generated from the new `microshift show-config` command. That command marshals the MicroShiftConfig struct (https://github.com/openshift/microshift/blob/main/pkg/cmd/showConfig.go#L47-L50) and therefore generates capitalised names.

Correction: It generates capitalised names because `pkg/config/config.go` uses `yaml:` annotations expected by `gopkg.in/yaml.v2` while `pkg/cmd/showConfig.go` uses `sigs.k8s.io/yaml` which expects `json:` annotations instead.

My suggestion would be to merge this PR so the example is correct, then follow up with a separate PR to switch `pkg/config/config.go` to K8s' implementation, too.

zshi-redhat:
[Cluster](https://github.com/openshift/microshift/blob/main/packaging/microshift/config.yaml#L5) yaml key was not commented out. Once changed to lowercase `cluster`, it starts to pass empty config to microshift, which may be the reason of CI failure.

Updated the PR to comment out `cluster` in config.yaml and added `mtu` example.

dhellmann:
> > I suspect the example got generated from the new `microshift show-config` command. That command marshals the MicroShiftConfig struct (https://github.com/openshift/microshift/blob/main/pkg/cmd/showConfig.go#L47-L50) and therefore generates capitalised names.
> 
> Correction: It generates capitalised names because `pkg/config/config.go` uses `yaml:` annotations expected by `gopkg.in/yaml.v2` while `pkg/cmd/showConfig.go` uses `sigs.k8s.io/yaml` which expects `json:` annotations instead.
> 
> My suggestion would be to merge this PR so the example is correct, then follow up with a separate PR to switch `pkg/config/config.go` to K8s' implementation, too.

See #943

zshi-redhat:
Closing in favor of #943 .

PR #936, opened 2022-09-07
USHIFT-373: add priorityClassName to topoLVM

pacevedom:
Add priorityClassName to TopoLVM pods for better scheduling/preemption capabilities.

Currently, MicroShift does not have a default priority for pods, which means they resolve to 0. In order to have pods properly scheduled for the cluster to be in healthy status, and also be able to support features like [graceful node shutdown](https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown), we must have all critical pods with the correct priority.
[Pod preemption reference](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/)

### DIFF
--- a/assets/components/odf-lvm/topolvm-controller_deployment.yaml
+++ b/assets/components/odf-lvm/topolvm-controller_deployment.yaml
@@ -130,6 +130,7 @@ spec:
         volumeMounts:
         - mountPath: /certs
           name: certs
+      priorityClassName: system-cluster-critical
       restartPolicy: Always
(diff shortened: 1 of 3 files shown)


oglok:
Hey @pacevedom ,

Could you please add some description to the PR? Why is priorityClassName needed?

pacevedom:
> Hey @pacevedom ,
> 
> Could you please add some description to the PR? Why is priorityClassName needed?

Done

PR #939, opened 2022-09-08
AUTH-296: certificate chains init refactoring

stlaz:
This PR adds a refactoring of the current certificate initialization code.

The new code allows describing certificate chains by Go structures, feeding them into a builder object, and then creating them all at once once the developer is happy with what they have.

The new structure currently only allows to retrieve leaf certificate certs/key PEMs in a path-like manner, but might be extensible to perform queries on *x509.Certificate objects in the future, so as to allow for checking expirations and could be thus fed to a controller that is capable of handling rotations over the structure.

/cc @ibihim

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift/microshift
 
-go 1.18
+go 1.17
 
 require (
 	github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e // openshift-controller-manager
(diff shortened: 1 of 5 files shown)


oglok:
I've had a look at this PR. I'd need a bit more of context from OpenShift cert management, but in general it looks good to me. I understand it generalizes the build of signers and client certificates. It works in my dev environment.

@ibihim is referenced in the PR, are we expecting him to review it too?

PR #941, opened 2022-09-08
update ovnk image from release payload

zshi-redhat:
rebase ovnk image to ovn-kubernetes-microshift in release payload

Closes #[NP-596](https://issues.redhat.com//browse/NP-596)


pkg/release/release.go
```diff
@@ -19,17 +19,17 @@ package release
...
-	"ose_csi_node_registrar":  "quay.io/microshift/ose-csi-node-driver-registrar" + Base,
-	"ose_csi_livenessprobe":   "quay.io/microshift/ose-csi-livenessprobe" + Base,
+	"cli":                       "quay.io/microshift/cli:" + Base,
+	"coredns":                   "quay.io/microshift/coredns:" + Base,
+	"haproxy_router":            "quay.io/microshift/haproxy-router:" + Base,
+	"kube_rbac_proxy":           "quay.io/microshift/kube-rbac-proxy:" + Base,
+	"pause":                     "quay.io/microshift/pause:" + Base,
+	"service_ca_operator":       "quay.io/microshift/service-ca-operator:" + Base,
+	"openssl":                   "quay.io/microshift/openssl" + Base,
+	"ovn_kubernetes-microshift": "quay.io/microshift/ovn-kubernetes-microshift:" + Base,
```
fzdarsky, line 29: You'll probably want to use underscores only (i.e. `ovn_kubernetes_microshift`). We're using this same name in the Go templating of the resource manifests and IIRC, the templating language stumbles over hyphens.

BTW: You'd need to update your manifests in this same PR.

pkg/release/release_amd64.go
```diff
@@ -21,18 +21,18 @@ package release
...
+		"cli":                       "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ba44dead03ea74107f90d58525106fb52d27a120b73c6cc8e2be31d37043ca1c",
+		"coredns":                   "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2b423e88cdd37f307aff51cbb0f53fc45deff9618f5b4f12bfb78bea7aff51a2",
+		"haproxy_router":            "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5f9ee3afa744e790dbb61d08f44e30370c9a5ff041054bf99dc1afe58792cd7b",
+		"kube_rbac_proxy":           "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3df935837634adb5df8080ac7263c7fb4c4f9d8fd45b36e32ca4fb802bdeaecc",
+		"pause":                     "k8s.gcr.io/pause:3.6",
+		"service_ca_operator":       "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:8b0e194c1cb6babeb2d8326091deb2bba63c430938abe056eb89398c78733eb6",
+		"openssl":                   "registry.access.redhat.com/ubi8/openssl@sha256:8b41865d30b7947de68a9c1747616bce4efab4f60f68f8b7016cd84d7708af6b",
+		"ovn_kubernetes-microshift": "quay.io/microshift/ovn-kubernetes-singlenode@sha256:e97d6035754fad1660b522b8afa4dea2502d5189c8490832e762ae2afb4cf142",
```
fzdarsky, line 31: Same as above.

pkg/release/release_arm64.go
```diff
@@ -21,13 +21,13 @@ package release
...
+		"cli":                       "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:65889fde898f3025462f1344bd93e4a1d8ae5a764224b61b4f0b7e46cc1108e4",
+		"coredns":                   "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2619cbcbc5ea990bd8349848d410cc4040aea0265821b9d1e46b364f0519481b",
+		"haproxy_router":            "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:972f3fadc1b7eab2eaa02eff4128149eabe9577943dcb3e56f44dad7fc14059e",
+		"kube_rbac_proxy":           "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2d8319a33aa781918a3173e2061635d8534cc5e4a39171775830ff7e2ce646ac",
+		"pause":                     "k8s.gcr.io/pause:3.6",
+		"service_ca_operator":       "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:97a7d56ac6d6a38f7977b0759bb1b97fe3895e6020364e60a8a2926f3fd0084e",
+		"openssl":                   "registry.access.redhat.com/ubi8/openssl@sha256:3f781a07e59d164eba065dba7d8e7661ab2494b21199c379b65b0ff514a1b8d0",
+		"ovn_kubernetes-microshift": "quay.io/microshift/ovn-kubernetes-singlenode@sha256:012e743363b5f15f442c238099d35a0c70343fd1d4dc15b0a57a7340a338ffdb",
```
fzdarsky, line 31: Same as above.

fzdarsky (changes requested):

pkg/release/release.go
zshi-redhat: Fixed the image name.
@fzdarsky Do you mean I need to update the image name in ovnk manifest?  The image name in ovnk manifests are rendered from config parameters. Other parts of the ovnk manifests are up-to-date I think.

pkg/release/release_amd64.go
zshi-redhat: Fixed the image name.

pkg/release/release_arm64.go
zshi-redhat: Fixed.

pkg/release/release.go
fzdarsky: > Fixed the image name. @fzdarsky Do you mean I need to update the image name in ovnk manifest? The image name in ovnk manifests are rendered from config parameters. Other parts of the ovnk manifests are up-to-date I think.

Thank you.

Yes, you need to update the manifests (https://github.com/openshift/microshift/blob/main/assets/components/ovn/master/daemonset.yaml#L47 et al.), because we're now adding the whole `release.Image` map to the rendering parameters by default (see https://github.com/openshift/microshift/blob/main/pkg/components/render.go#L16), so we no longer have to add them on a case-by-case basis. Result is the names need to match, though.

zshi-redhat: I see, thanks!
I forgot to update the image var in manifest.
updated the image var to ovn_kubernetes_microshift

fzdarsky:
Thank you for the update.

/lgtm

PR #946, opened 2022-09-09
Fix config params and tests

fzdarsky:
This PR fixes tests in pkg/config/config_tests.go that were either flaky in CI or not correctly testing the function under test.
Also, adds command line flags for parameters for which we already support configuration via config.file.

Closes [USHIFT-383](https://issues.redhat.com/browse/USHIFT-383)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -213,23 +213,51 @@ func (c *MicroshiftConfig) updateManifestList() {
 }
 
 func (c *MicroshiftConfig) ReadFromCmdLine(flags *pflag.FlagSet) error {
-	if dataDir, err := flags.GetString("data-dir"); err == nil && flags.Changed("data-dir") {
-		c.DataDir = dataDir
+	if s, err := flags.GetString("data-dir"); err == nil && flags.Changed("data-dir") {
+		c.DataDir = s
 		// if the defaults are present, rebuild based on the new data-dir
 		c.updateManifestList()
 	}
-	if auditLogDir, err := flags.GetString("audit-log-dir"); err == nil && flags.Changed("audit-log-dir") {
-		c.AuditLogDir = auditLogDir
+	if s, err := flags.GetString("audit-log-dir"); err == nil && flags.Changed("audit-log-dir") {
+		c.AuditLogDir = s
 	}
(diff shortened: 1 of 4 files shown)


pkg/config/config_test.go
```diff
@@ -55,53 +56,66 @@ func TestCommandLineConfig(t *testing.T) {
...
-		flags.StringVar(&config.NodeName, "node-name", "", "")
-		flags.StringVar(&config.NodeIP, "node-ip", "", "")
-		flags.StringVar(&config.Cluster.URL, "cluster-url", "", "")
-		flags.StringVar(&config.Cluster.ClusterCIDR, "cluster-cidr", "", "")
-		flags.StringVar(&config.Cluster.ServiceCIDR, "service-cidr", "", "")
-		flags.StringVar(&config.Cluster.ServiceNodePortRange, "service-node-port-range", "", "")
-		flags.StringVar(&config.Cluster.DNS, "cluster-dns", "", "")
-		flags.StringVar(&config.Cluster.Domain, "cluster-domain", "", "")
-		flags.StringVar(&config.Cluster.MTU, "mtu", "1200", "")
+		// bind the config file flag to the struct
```
dhellmann, line 71: Shouldn't this test just call addRunFlags to do this?

pkg/config/config.go
```diff
@@ -211,23 +213,51 @@ func (c *MicroshiftConfig) updateManifestList() {
...
-		c.Roles = roles
+	if ss, err := flags.GetStringSlice("roles"); err == nil && flags.Changed("roles") {
+		c.Roles = ss
 	}
-	if pprofFlag, err := flags.GetBool("debug.pprof"); err == nil && flags.Changed("debug.pprof") {
-		c.Debug.Pprof = pprofFlag
+	if s, err := flags.GetString("node-name"); err == nil && flags.Changed("node-name") {
+		c.NodeName = s
 	}
+	if s, err := flags.GetString("node-ip"); err == nil && flags.Changed("node-ip") {
```
dhellmann, line 233: I wouldn't expect to need to do all of this work to copy options one at a time if the flags are bound to the config struct by addRunFlags. Maybe I don't understand how that works?

dhellmann:
This unblocks some other PRs, so let's take it and keep iterating.

/lgtm

PR #952, opened 2022-09-14
USHIFT-399: update list of enabled APIs

dhellmann:
When we removed the openshift-apiserver dependency, most of the APIs
from the authorization.openshift.io group were removed. Update the
documentation to reflect that change.


benluddy (approved):
/lgtm

PR #954, opened 2022-09-14
USHIFT-261: clean up path handling for kubeconfig files

dhellmann:
Provide a consistent way to calculate the path to a set of managed
kubeconfig files so we can avoid errors with path management
throughout the code.

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -1,33 +1,33 @@
 package components
 
 import (
-	"path/filepath"
-
 	"github.com/openshift/microshift/pkg/config"
 	"k8s.io/klog/v2"
 )
 
 func StartComponents(cfg *config.MicroshiftConfig) error {
-	if err := startServiceCAController(cfg, filepath.Join(cfg.DataDir, "/resources/kubeadmin/kubeconfig")); err != nil {
+	kubeAdminConfig := cfg.KubeConfigPath(config.KubeAdmin)
+
+	if err := startServiceCAController(cfg, kubeAdminConfig); err != nil {
 		klog.Warningf("Failed to start service-ca controller: %v", err)
 		return err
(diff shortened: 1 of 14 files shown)


pkg/config/config.go
```diff
@@ -78,6 +78,21 @@ type MicroshiftConfig struct {
...
+
+const (
+	KubeAdmin             KubeConfigID = "kubeadmin"
+	KubeControllerManager KubeConfigID = "kube-controller-manager"
+	KubeScheduler         KubeConfigID = "kube-scheduler"
+	Kubelet               KubeConfigID = "kubelet"
+)
+
+// KubeConfig returns the path to the specified kubeconfig file.
+func (cfg *MicroshiftConfig) KubeConfig(id KubeConfigID) string {
```
stlaz, line 92: Let's just call it `KubeConfigPath()`

pkg/components/components.go
```diff
@@ -1,33 +1,33 @@
...
 import (
-	"path/filepath"
-
 	"github.com/openshift/microshift/pkg/config"
 	"k8s.io/klog/v2"
 )
 
 func StartComponents(cfg *config.MicroshiftConfig) error {
-	if err := startServiceCAController(cfg, filepath.Join(cfg.DataDir, "/resources/kubeadmin/kubeconfig")); err != nil {
+	kubeAdminConfig := cfg.KubeConfig(config.KubeAdmin)
```
stlaz, line 9: :+1:

stlaz (commented):
looks good, I think I would be just a little bit more explicit with the name of the function that returns the path

pkg/config/config.go
dhellmann: Good call, done

PR #955, opened 2022-09-15
USHIFT-401: Carried out patch to disable APIrequestcount controller

oglok:

The APIrequestcount controller cannot be disabled via configuration param in the OpenShift KAS. This carried out patch disables it and avoids recurrent calls to this non-existing API.

### DIFF
--- a/scripts/rebase_patches/0002-disable-APIrequestcount-controller.patch
+++ b/scripts/rebase_patches/0002-disable-APIrequestcount-controller.patch
@@ -1,0 +1,47 @@
+diff --git a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
+index fb66eece..046e20bc 100644
+--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
++++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
+@@ -8,7 +8,6 @@ import (
+ 	"github.com/openshift/apiserver-library-go/pkg/admission/imagepolicy/imagereferencemutators"
+ 	"github.com/openshift/apiserver-library-go/pkg/admission/quota/clusterresourcequota"
+ 	"github.com/openshift/apiserver-library-go/pkg/securitycontextconstraints/sccadmission"
+-	apiclientv1 "github.com/openshift/client-go/apiserver/clientset/versioned/typed/apiserver/v1"
+ 	configclient "github.com/openshift/client-go/config/clientset/versioned"
+ 	configv1informer "github.com/openshift/client-go/config/informers/externalversions"
+ 	quotaclient "github.com/openshift/client-go/quota/clientset/versioned"
+@@ -33,7 +32,6 @@ import (
+ 	"k8s.io/kubernetes/openshift-kube-apiserver/admission/autoscaling/managementcpusoverride"
+ 	"k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/nodeenv"
+ 	"k8s.io/kubernetes/openshift-kube-apiserver/enablement"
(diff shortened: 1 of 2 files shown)


dhellmann:
/assign @benluddy 

This looks OK to me, but I'm not an expert in this area.

Do we have a ticket on the KAS team's backlog to make this more configurable in the future?

oglok:
> /assign @benluddy
> 
> This looks OK to me, but I'm not an expert in this area.
> 
> Do we have a ticket on the KAS team's backlog to make this more configurable in the future?

Not yet, we were discussing if making this configurable would makes sense if only MicroShift would be the only platform disabling it.

I've been experimenting with disabling also informers for `clusterresourcequota` and `user` API endpoints, so we get rid of all the annoying logs like:

```
E0915 19:14:15.874904  109954 kube-apiserver reflector.go:138] github.com/openshift/client-go/user/informers/externalversions/factory.go:101: Failed to watch *v1.Group: failed to list *v1.Group: the server could not find the requested resource (get groups.user.openshift.io)
```

I could create a carried patch to disable all of them at once, and get a discussion going about making them configurable.

wdyt @benluddy ?

dhellmann:
> I could create a carried patch to disable all of them at once, and get a discussion going about making them configurable.
> 
> wdyt @benluddy ?

I'd be OK with incremental changes here, too. Do whatever is easier.

oglok:
@benluddy are we ok with merging this?

scripts/rebase_patches/0002-disable-APIrequestcount-controller.patch
```diff
@@ -0,0 +1,47 @@
...
+-	deprecatedAPIClient, err := apiclientv1.NewForConfig(makeJSONRESTConfig(genericConfig.LoopbackClientConfig))
+-	if err != nil {
+-		return err
+-	}
+-	deprecatedAPIRequestController := deprecatedapirequest.NewController(deprecatedAPIClient.APIRequestCounts(), nodeFor())
+-	genericConfig.AddPostStartHook("openshift.io-deprecated-api-requests-filter", func(context genericapiserver.PostStartHookContext) error {
+-		go deprecatedAPIRequestController.Start(context.StopCh)
+-		return nil
+-	})
+-	genericConfig.BuildHandlerChainFunc, err = BuildHandlerChain(
```
benluddy, line 35: Will MicroShift ever need the patches for the OAuth metadata endpoint, impersonation backwards-compat, and console redirect that are installed here?

/cc @stlaz

scripts/rebase_patches/0002-disable-APIrequestcount-controller.patch
```diff
@@ -0,0 +1,47 @@
...
+@@ -80,26 +78,6 @@ func OpenShiftKubeAPIServerConfigPatch(genericConfig *genericapiserver.Config, k
+ 	)
+ 	// END ADMISSION
+ 
+-	// HANDLER CHAIN (with oauth server and web console)
+-	deprecatedAPIClient, err := apiclientv1.NewForConfig(makeJSONRESTConfig(genericConfig.LoopbackClientConfig))
+-	if err != nil {
+-		return err
+-	}
+-	deprecatedAPIRequestController := deprecatedapirequest.NewController(deprecatedAPIClient.APIRequestCounts(), nodeFor())
```
stlaz, line 30: should we remove this controller?

scripts/rebase_patches/0002-disable-APIrequestcount-controller.patch
stlaz: OAuth metadata endpoint would only be useful if we allowed deploying the OpenShift authentication, which I don't makes sense for our use-case, so I don't think it's going to be used.

The console redirect would likely come handy if we had console, which we don't, right?

As for the user impersonation backwards compat, IIRC it only makes sense with OpenShift authentication, so it should be safe to remove it.

scripts/rebase_patches/0002-disable-APIrequestcount-controller.patch
oglok: This is the goal of the patch: to remove the APIrequestController because MicroShift does not provide the APIrequestcount API endpoint.

PR #956, opened 2022-09-16
USHIFT-536: add user-facing config for ovn-kubernetes CNI

zshi-redhat:
Closes [USHIFT-536](https://issues.redhat.com//browse/USHIFT-536)

### DIFF
--- a/packaging/microshift/config.yaml
+++ b/packaging/microshift/config.yaml
@@ -19,9 +19,6 @@ cluster:
   # URL of the API server for the cluster
   #url: https://127.0.0.1:6443
 
-  # MTU for CNI
-  #mtu: "1400"
-
 # Log verbosity (0-5)
 #logVLevel: 0
 
--- a/packaging/systemd/configure-ovs.sh
+++ b/packaging/systemd/configure-ovs.sh
@@ -626,6 +626,15 @@ fi
 # print initial state
 print_state
 if [ "$1" == "OVNKubernetes" ]; then
+  # Skip configuring NICs onto OVS bridge "br-ex" when disableOVSInit is true
+  MICROSHIFT_OVN_CONFIG_FILE_PATH="/etc/microshift/ovn.yaml"
+  if [ -f "$MICROSHIFT_OVN_CONFIG_FILE_PATH" ]; then
+    disableOVSInit=$(cat "$MICROSHIFT_OVN_CONFIG_FILE_PATH" | awk "/disableOVSInit:/ && ! /#.*disableOVSInit:/ {print \$2}")
+    if [ "$disableOVSInit" == "true" ]; then
+      echo "disableOVSInit is true, skipped configure-ovs.sh "
+      exit 0
+    fi
+  fi
   # Configures NICs onto OVS bridge "br-ex"
   # Configuration is either auto-detected or provided through a config file written already in Network Manager
   # key files under /etc/NetworkManager/system-connections/
--- a/pkg/components/networking.go
+++ b/pkg/components/networking.go
@@ -1,14 +1,17 @@
 package components
 
 import (
+	"fmt"
 	"path/filepath"
 
 	"github.com/openshift/microshift/pkg/assets"
 	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/config/ovn"
+	"github.com/openshift/microshift/pkg/util"
 	"k8s.io/klog/v2"
 )
 
-func startOVNKubernetes(cfg *config.MicroshiftConfig, kubeconfigPath string) error {
+func startCNIPlugin(cfg *config.MicroshiftConfig, kubeconfigPath string) error {
 	var (
 		ns = []string{
 			"components/ovn/namespace.yaml",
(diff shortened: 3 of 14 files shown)


zshi-redhat:
cc @fzdarsky @mangelajo

dgrisonnet:
Putting a hold for the API team to take some time to consider how to handle this new configuration parameter going forward. We are currently working on rewriting the existing configuration to be more in line with OCP so adding a dependency on the existing configuration of MicroShift will be problematic for us.

/hold

packaging/microshift/config.yaml
```diff
@@ -41,3 +41,9 @@ cluster:
 
 # The name of the node (defaults to hostname)
 #nodeName: ""
+
+# OVNKubernetes network plugin config
+#ovnkubernetesConfig:
```
dhellmann, line 46: We found a way to have all of the configuration for ODF-LVM in a file outside of MicroShift's main config file. That will, eventually, allow us to provide a different CSI implementation that can also have its own separate file.

Some of the future use cases for MicroShift, especially upstream, make it desirable to be able to replace the CNI implementation as well, and following the same pattern with a similar clean separation for the configuration will make that more straightforward. 

Let's talk about the approach we want to take this week.

zshi-redhat: @dhellmann ack+, I updated the PR to follow the CSI approach and have a separate config file for ovn-kubernetes.

zshi-redhat:
> Putting a hold for the API team to take some time to consider how to handle this new configuration parameter going forward. We are currently working on rewriting the existing configuration to be more in line with OCP so adding a dependency on the existing configuration of MicroShift will be problematic for us.
> 
> /hold

@dgrisonnet The PR is updated to use a separate config file for ovn-kubernetes.

packaging/microshift/config.yaml
dhellmann: That looks good.

I'm not entirely comfortable with parsing YAML from bash, but what you have looks like it will work and we can consider whether we need something more "robust" in the future if we run into problems.

zshi-redhat: I thought about using `yq` in the bash to parse yaml file, but it's not available in the existing repos. Adding the installation dependency sounds a bit heavy for this change in configure-ovs.sh since it is currently the only one I'm aware of that requires parsing yaml in bash in microshift project.
Another idea might be to use json config file and `jq` to parse it, `jq` is installed by default. The downside is that it is inconsistent with csi and microshift config files. I could write a small golang program to parse the yaml if we will have similar needs in microshift scripts. Or any other ideas?

dhellmann: Like I said, I think what you have is fine. We can consider other options as an improvement later if we have issues or establish a pattern based on other config.

dgrisonnet:
Unholding since the initial concerns were cleared in this thread: https://github.com/openshift/microshift/pull/1030#discussion_r1006118685

/unhold

packaging/systemd/configure-ovs.sh
```diff
@@ -626,6 +626,23 @@ fi
 # print initial state
 print_state
 if [ "$1" == "OVNKubernetes" ]; then
+  # Skip configuring NICs onto OVS bridge "br-ex" when disableOVSInit is true
+  MICROSHIFT_CONFIG_DIRS=(~/.microshift /etc/microshift)
+  MICROSHIFT_CONFIG_FILE_NAME="config.yaml"
+  MICROSHIFT_OVN_CONFIG_FILE_NAME="ovn.yaml"
+  for dir in "${MICROSHIFT_CONFIG_DIRS[@]}"; do
+    if [ -f "$dir/$MICROSHIFT_CONFIG_FILE_NAME" ] && [ -f "$dir/$MICROSHIFT_OVN_CONFIG_FILE_NAME" ]; then
+      disableOVSInit=$(cat "$dir/$MICROSHIFT_OVN_CONFIG_FILE_NAME" | awk "/disableOVSInit:/ && ! /#.*disableOVSInit:/ {print \$2}")
```
dhellmann, line 635: Can we use yq to parse the YAML? How big is it and is it OK as a dependency?

If not, would it make sense to use a file format that is friendly for parsing with bash?

packaging/systemd/configure-ovs.sh
```diff
@@ -626,6 +626,23 @@ fi
 # print initial state
 print_state
 if [ "$1" == "OVNKubernetes" ]; then
+  # Skip configuring NICs onto OVS bridge "br-ex" when disableOVSInit is true
+  MICROSHIFT_CONFIG_DIRS=(~/.microshift /etc/microshift)
```
dhellmann, line 630: Is it likely that a systemd unit would find something in ~/.microshift? Should we just look in /etc?

pkg/components/networking.go
```diff
@@ -63,6 +77,7 @@ func startOVNKubernetes(cfg *config.MicroshiftConfig, kubeconfigPath string) err
 		return err
 	}
 	extraParams := assets.RenderParams{
+		"MTU":            ovnConfig.MTU,
```
dhellmann, line 80: What do you think about passing ovnConfig here so the template can access any future settings without us having to extend this parameter list?

dhellmann (commented):
Let's work on this, but wait until we branch for 4.13 to merge it.

packaging/systemd/configure-ovs.sh
zshi-redhat: Good catch!
agree, systemd may not find the config file in user root. I'll update to use /etc/ directly.

pkg/components/networking.go
zshi-redhat: Good idea, will update to pass ovnConfig.

packaging/systemd/configure-ovs.sh
mangelajo: I believe @zshi-redhat solution is OK, we avoid installing yq which based on his comments doesn't seem to be available in RHEL, and we keep yaml which is a more familiar config format in the k8s world, also more human friendly than json.

mangelajo:
Waiting for last comments to be handled, looks good otherwise.

pkg/components/networking.go
zshi-redhat: done.

packaging/systemd/configure-ovs.sh
zshi-redhat: done.

mangelajo (commented):
/lgtm

PR #960, opened 2022-09-19
Add support for process-exporter

sjug:
- Fix Prometheus services and templates
- Don't check GPG for process-exporter
- Manage role permissions

Provides system process resource monitoring for prometheus.


ansible/inventory/inventory
```diff
@@ -1,11 +1,8 @@
 [microshift]
-microshift.example.com ansible_host= # microshift node
+microshift-dev.localhost ansible_host= private_ip=
```
ggiguash, line 2: .localhost name always resolves to 127.0.0.1. Let's just have "microshift-dev" default.

ansible/roles/install-logging/tasks/main.yml
```diff
@@ -1,40 +1,28 @@
...
-      ansible.builtin.command: dnf config-manager --set-enabled crb
-      when: ansible_distribution_version == "9"
-
-    - name: install EPEL repo
+    - name: upgrade all packages
+      ansible.builtin.dnf:
+        name: "*"
+        state: latest
+      notify: reboot machine
+
```
ggiguash, line 11: Why do we need to upgrade all packages when installing logging?

ansible/roles/install-logging/tasks/main.yml
```diff
@@ -1,40 +1,28 @@
...
+  when: (ansible_distribution == "CentOS") or (ansible_distribution == "RedHat") or (ansible_distribution == "Fedora")
 
 - name: install prometheus
-  ansible.builtin.dnf:
-    name:
-      - golang-github-prometheus
+  community.general.pacman:
+    name: prometheus
     state: present
+  when: (ansible_distribution == "Archlinux")
```
ggiguash, line 25: ArchLinux? What is it for?

ggiguash (changes requested):

PR #963, opened 2022-09-20
USHIFT-415: Remove scripts and files for unsupported procedures

ggiguash:
Including minor documentation updates.

Tested the following procedures:
* make
* make rpm
* make srpm
* make iso
* packaging/rpm/make-microshift-images-rpm.sh (includes paack.py calls)

Closes [USHIFT-415](https://issues.redhat.com//browse/USHIFT-415)


fzdarsky:
This PR got pretty large and changes many things at once, making it hard to review and merge, but also hard to restore things from.

The way I see it we have 5 categories of changes that may warrant separate PRs (to merge faster and to not make it an "all-or-nothing decision):
- Fixes of paths and docs that are important but unrelated to "dropping unsupported procedures". I'd suggest factoring those out into a separate PR.
- Deleting of things that are probably non-controversial (`.sites`, `.image-references.`, `hack/install.sh`, `hack/all-in-one`), because no longer working/used at all.
- Container build of MicroShift
- MicroShift AIO image
- Container builds of hosted components for architectures not yet productised (e.g. RISC-V)

The last two are something I'm relatively sure the community will want to bring back, so they're less obvious to me.

ggiguash:
@fzdarsky, I can organize these in separate commits, maybe? This should allow us for easier restore if necessary

Makefile
```diff
@@ -195,76 +191,6 @@ image-build-iso: rpm
 iso: image-build-configure image-build-iso
 .PHONY: iso
 
-###############################
```
fzdarsky, line 198: When removing the containerized builds, you'll also want to remove their systemd units (`packaging/systemd/microshift-aio.service` and `packaging/systemd/microshift-containerized.service`) and drop them from the RPM package.

PR #965, opened 2022-09-20
Rebase to 4.11.5

fzdarsky:
Rebases MicroShift to the 4.11 stable series as stepping-stone to 4.12.

Closes USHIFT-417.


fzdarsky:
@copejon there's a problem with the TopoLVM Controller pod not coming up. I assume it's still a version for 4.10 not adapted to the new SCCs v2 and PSA. Could you look into that, please?

PR #967, opened 2022-09-20
Switch to ovnk local gateway mode

zshi-redhat:
ovnk local gateway mode allows pod egress traffic to use system routes, also NodePorts will work over multiple interfaces.


mangelajo:
Oh I need to update bindata on the patch

PR #968, opened 2022-09-21
USHIFT-427: Disable clusterresourcequota and user informers

oglok:
These two APIs are not available in MicroShift, so informers are useless here.


stlaz:
The removal of RBR controller LGTM.

@benluddy please chime in on the cluster quotas

oglok:
Closing this PR because a fix directly in OCP would be preferred instead of a carried patch like this.

PR #969, opened 2022-09-22
Update dns comments in config.yaml

zshi-redhat:


oglok (commented):
/lgtm

packaging/microshift/config.yaml
```diff
@@ -7,10 +7,10 @@ cluster:
   # IP range for use by the cluster
   #clusterCIDR: 10.42.0.0/16
 
-  # DNS server IP
+  # DNS server IP is the k8s service IP address where pods query for name resolution
```
mangelajo, line 10:
```suggestion
  # DNS server IP is the k8s service IP address space where pods query for name resolution
```
?

oglok: What do you mean by space? It's not a CIDR, but a single IP address...

fzdarsky: Ack, it's not an address space, but it'd be good to`s/where/which/`.

In general, wondering whether we need this at all, as CoreDNS is running always on the 10th IP of the subnet? But that's not for this PR.

zshi-redhat: updated to use which.

There was some confusion on how these dns options shall be used, thus the update.

PR #970, opened 2022-09-22
AUTH-296: Certificate-related fixes

stlaz:
This PR introduces a couple of smaller fixes to the current code related to certificates.

Namely:
- actually use the serving CA to generate serving cert for the OCM
- don't generate new sub-CA certificates for the CSR signer with every restart
- fix the CA validity days to match OCP

/cc @oglok @ibihim

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -172,6 +172,17 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 			"service-ca",
 			cryptomaterial.ServiceCADir(certsDir),
 			cryptomaterial.ServiceCAValidityDays,
+		).WithServingCertificates(
+			&cryptomaterial.ServingCertificateSigningRequestInfo{
+				CertificateSigningRequestInfo: cryptomaterial.CertificateSigningRequestInfo{
+					Name:         "openshift-controller-manager-serving",
+					ValidityDays: cryptomaterial.ServiceCAServingCertValidityDays,
+				},
+				Hostnames: []string{
+					"controller-manager.openshift-controller-manager.svc",
+					"controller-manager.openshift-controller-manager.svc.cluster.local",
+				},
+			},
 		),
 	).WithCABundle(
 		cryptomaterial.TotalClientCABundlePath(certsDir),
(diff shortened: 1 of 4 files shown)


pkg/cmd/init.go
```diff
@@ -172,6 +172,17 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 			"service-ca",
 			cryptomaterial.ServiceCADir(certsDir),
 			cryptomaterial.ServiceCAValidityDays,
+		).WithServingCertificates(
+			&cryptomaterial.ServingCertificateSigningRequestInfo{
+				CertificateSigningRequestInfo: cryptomaterial.CertificateSigningRequestInfo{
+					Name:         "openshift-controller-manager-serving",
+					ValidityDays: 730,
```
oglok, line 179: Why not adding a constant as the rest of validity days ones? or just using the `ServiceCAServingCertValidityDays`?

pkg/cmd/init.go
```diff
@@ -138,7 +138,7 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 		cryptomaterial.NewCertificateSigner(
 			"kubelet-signer",
 			cryptomaterial.KubeletCSRSignerSignerCertDir(certsDir),
-			cryptomaterial.ClientCAValidityDays,
+			cryptomaterial.KubeControllerManagerCSRSignerSignerCAValidityDays,
```
oglok, line 141: Why do we need the kube-controller-manager-csr-signer-signer if this is used by the kube-controller-manager-operator which MicroShift does not use or ship?

oglok (commented):
Couple of Q's, but the rest is good!

pkg/cmd/init.go
stlaz: I am in fact only using this to stick to the OpenShift schema.

From what I can see, this certificate signs two others:
- the csr-signer CA certificate
- the kubelet client certificates during bootstrap

I think that we may in fact be able to live without this certificate in Microshift, for now I would stick with the current OCP schema until I figure out whether there is any use for this `signer -> intermediate` scheme here.

pkg/cmd/init.go
stlaz: Turns out such a constant already exists, only I forgot to use it :+1:

PR #973, opened 2022-09-22
Add Standa to approvers

pacevedom:
/cc @oglok 
/cc @stlaz


OWNERS
```diff
@@ -10,6 +10,7 @@ approvers:
 - zshi-redhat
 - pmtk
 - pacevedom
+- stlaz
```
oglok, line 13: Shouldn't we add him also as reviewer?

dhellmann: I think approvers automatically get reviewer rights, too, but it's better to be safe.

stlaz:
/lgtm
Thanks!

PR #974, opened 2022-09-23
USHIFT-179: PR creator

pmtk:
Closes  [USHIFT-179](https://issues.redhat.com/browse/USHIFT-179)

Adds python script for PR creation using `pygithub` library to interact with GitHub API (more info in script's docstring)

Needs to be merged before https://github.com/openshift/microshift/pull/974

### DIFF
--- a/scripts/create_pr.py
+++ b/scripts/create_pr.py
@@ -17,6 +17,7 @@ App requires following permissions:
 import os
 import sys
 import subprocess
+from git import Repo # GitPython
 from github import GithubIntegration, Github # pygithub
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


scripts/create_pr.py
```diff
@@ -0,0 +1,63 @@
...
+key_path = try_get_env(KEY_ENV)
+org = try_get_env(ORG_ENV)
+repo = try_get_env(REPO_ENV)
+
+integration = GithubIntegration(app_id, Path(key_path).read_text())
+app_installation = integration.get_installation(org, repo)
+if app_installation == None:
+    sys.exit(f"Failed to get app_installation for {org}/{repo}. Response: {app_installation.raw_data}")
+
+installation_access_token = integration.get_access_token(app_installation.id).token
```
copejon, line 47: Would it be worth nesting the integration calls into a try/catch block or do they fail cleanly enough to be debugable?

copejon: Ditto for package calls further down.

pmtk: When working with the library, the errors were quite nice (with the exception of app_installation), but we can wrap them just in case if there's a need

PR #975, opened 2022-09-23
AUTH-296: work out etcd certificates

stlaz:
I was not yet able to learn what's important for etcd serving/peer certs so I'm going with minimal config where client certificates only follow the kube CN and O schemes, and serving certificates get their hostnames in DNS/IPs SANs.

The PR is rebased on top of https://github.com/openshift/microshift/pull/970 and can only merge once this one merges.

It's very likely that we may not see any failures in etcd because no peer communication is likely to happen with just one etcd instance. I'll need to test the cert setup with at least 2 etcds or ask someone with the necessary knowledge about how the peer certs should look like.

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -67,19 +67,6 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 		return nil, nil, err
 	}
 
-	// based on https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml#L19
-	if err := util.GenCerts("etcd-server", filepath.Join(certsDir, "/etcd"),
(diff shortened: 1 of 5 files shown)


stlaz:
This PR now follows the certificate configuration from https://github.com/stlaz/etcd-certs.

oglok (commented):
What's the purpose of the `.image_references` empty file? What do we gain in terms of security by making this repo unimportable?

stlaz:
> What's the purpose of the .image_references empty file? What do we gain in terms of security by making this repo unimportable?
I do not know its purpose. The repository is currently unimportable because the filename is malformed, the commit is fixing it.

stlaz:
somehow a lonely `ultimateTrustBundle` variable appeared in the code, removed now, all tests should hopefully pass again :slightly_smiling_face:

stlaz:
@oglok I think this might be good for re-tagging :slightly_smiling_face:

PR #976, opened 2022-09-23
ETCD-319: adding transient systemd unit for etcd

dusk125:
This is the first step to get etcd running un-bundled from MicroShift; this change makes etcd run in a transient systemd unit whose lifetime is bound to the MicroShift execution. Also, as a temporary measure, I added a subcommand to MicroShift that just runs the bundled etcd; this is so we can test etcd's execution with the systemd unit and this will be removed and replaced by an invocation of the etcd binary that will be laid down by an rpm.

The end goal, as described in this [Microshift Enhancement](https://github.com/openshift/enhancements/pull/1242) is to remove the current embedded etcd service and replace it with a service that launches etcd in a transient systemd unit.

### DIFF
--- a/etcd/cmd/microshift-etcd/run.go
+++ b/etcd/cmd/microshift-etcd/run.go
@@ -97,6 +97,7 @@ func (s *EtcdService) Run() error {
 	}
 	<-e.Server.ReadyNotify()
 
+	// Wait to be stopped.
 	sigTerm := make(chan os.Signal, 1)
 	signal.Notify(sigTerm, os.Interrupt, syscall.SIGTERM)
 	sig := <-sigTerm
--- a/pkg/controllers/etcd.go
+++ b/pkg/controllers/etcd.go
@@ -46,15 +46,39 @@ func (s *EtcdService) Name() string           { return "etcd" }
 func (s *EtcdService) Dependencies() []string { return []string{} }
 
 func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+	defer close(stopped)
+
+	// Check to see if we should run as a systemd run or directly as a binary.
+	runningAsSvc := os.Getenv("INVOCATION_ID") != ""
+
+	// Get the path to the etcd binary based on the MicroShift binary location.
 	microshiftExecPath, err := os.Executable()
 	if err != nil {
 		return fmt.Errorf("%v failed to get exec path: %v", s.Name(), err)
 	}
 	etcdPath := filepath.Join(filepath.Dir(microshiftExecPath), "microshift-etcd")
+	// Not running the etcd binary directly, the proper etcd arguments are handled
+	//		in etcd/cmd/microshift-etcd/run.go.
 	args := []string{"run"}
 
+	// If we're launching MicroShift as a service, we need to do the same
+	//		with etcd, so wrap it in a transient systemd-unit that's tied
+	//		to the MicroShift service lifetime.
+	var exe string
+	if runningAsSvc {
+		args = append([]string{
+			"--uid=root",
+			"--scope",
+			"--property", "BindsTo=microshift.service",
+			"--unit", "microshift-etcd",
+			etcdPath,
+		}, args...)
+		exe = "systemd-run"
+	} else {
+		exe = etcdPath
+	}
 	// Not using context as canceling ctx sends SIGKILL to process
-	cmd := exec.Command(etcdPath, args...)
+	cmd := exec.Command(exe, args...)
 
 	wd, err := os.Getwd()
 	if err != nil {
@@ -63,7 +87,7 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
 	cmd.Dir = wd
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	if err := cmd.Start(); err != nil {
+	if err = cmd.Start(); err != nil {
 		return fmt.Errorf("%s failed to start: %v", s.Name(), err)
 	}
 
@@ -87,6 +111,7 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
 	klog.Info("etcd is ready!")
 	close(ready)
 
+	// Wait for MicroShift to be done
 	<-ctx.Done()
 	return ctx.Err()
 }
--- a/hack/cleanup.sh
+++ b/hack/cleanup.sh
@@ -18,6 +18,9 @@ sudo bash -c "
     systemctl disable microshift 2>/dev/null
     pkill -9 microshift
 
+    systemctl stop --now microshift-etcd 2>/dev/null
+    systemctl reset-failed microshift-etcd 2>/dev/null
+
     echo Removing crio container and image storage
     crio wipe -f &>/dev/null || true
     systemctl restart crio


dusk125:
@haircommander Would you mind taking a look at this; mainly that this will be able to do what I've outlined above please?

haircommander:
added a review comment on the proposal, this doesn't seem like the right direction to me (though I certainly have missed many conversations)

dusk125:
The most recent update adds @deads2k suggestion to have the transient systemd unit launch the currently bundled etcd so that we can start testing the systemd launching mechanism while we work to get the rpm with etcd laid down. Once we have etcd installing from an rpm, I'll update this to completely remove bundled etcd and edit the systemd-run command to launch the rpm etcd.

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (s *ExtEtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+	defer close(stopped)
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("%s failed to get workdir: %v", s.Name(), err)
+	}
+
```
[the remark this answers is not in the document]
dusk125: Maybe, I'm currently working on getting the etcd systemd unit tied to the MicroShift unit such that if the latter is killed, etcd comes down as well.

If we'd be more comfortable, for sanity sake, issuing the stop command for microshift-etcd just in case, I'm fine to add it.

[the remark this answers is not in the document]
dusk125: With the most recent update, the etcd unit will go away when MS goes away (even if MS dies or is killed)

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,87 @@
...
+	// fake readiness probe
+	go func() {
+		<-time.After(time.Second * 15)
+		klog.Info("etcd is ready!")
+		close(ready)
+	}()
+
+	<-ctx.Done()
+
+	_ = cmd.Process.Kill()
```
deads2k, line 79: with BindsTo is this needed?

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,87 @@
...
+
+	to, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	cmd = exec.CommandContext(to, "systemctl", "stop", "microshift-etcd.scope")
```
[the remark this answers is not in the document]
dusk125: Just tested, looks like no!

pkg/controllers/systemd-etcd.go
dusk125: Just tested and no it's not!

deads2k:
This PR as it stands
1. starts the transient systemd unit
2. ties the systemd unit to the lifecycle of microshift.service so termination of microshift graceful and non-graceful (kill -9) also terminates etcd
3. does not bring in a new RPM dependency (yet).  that is desired in the near future, but this gives the proper systemd lifecycle and process separation to reduce the risk on introducing the new RPM
4. passes the existing configuration arguments to the separate etcd process

/lgtm

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,154 @@
...
+	// 	// TODO alray - we might not need peers since we're single node
+	// 	fmt.Sprintf("--peer-cert-file=%v", s.PeerCertFile),
+	// 	fmt.Sprintf("--peer-key-file=%v", s.PeerKeyFile),
+	// 	fmt.Sprintf("--peer-trusted-ca-file=%v", s.PeerTrustedCAFile),
+	// 	"--peer-client-cert-auth=true",
+	// 	fmt.Sprintf("--advertise-client-urls=%v", s.AdvertiseClientUrls),
+	// 	fmt.Sprintf("--listen-client-urls=%v", s.ListenClientUrls),
+	// 	"--metrics=extensive",
+	// 	fmt.Sprintf("--listen-metrics-urls=%v", s.ListenMetricsUrls),
+	// }
```
stlaz, line 127: why is this still here?

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,154 @@
...
+
+	<-ctx.Done()
+
+	// TODO alray - remove if not needed because of BindsTo
+	// _ = cmd.Process.Kill()
+
+	// to, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	// defer cancel()
+	// cmd = exec.CommandContext(to, "systemctl", "stop", "microshift-etcd.scope")
+	// _ = cmd.Run()
```
stlaz, line 151: why is this still here?

stlaz (commented):
remove TODO comments?

stlaz:
@dhellmann is the hold still valid?

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,154 @@
...
+	caCertFile := cryptomaterial.UltimateTrustBundlePath(certsDir)
+	etcdCertDir := filepath.Join(certsDir, s.Name())
+
+	s.CertFile = filepath.Join(etcdCertDir, "etcd-serving.crt")
+	s.KeyFile = filepath.Join(etcdCertDir, "etcd-serving.key")
+	s.TrustedCAFile = caCertFile
+
+	s.PeerCertFile = filepath.Join(etcdCertDir, "etcd-peer.crt")
+	s.PeerKeyFile = filepath.Join(etcdCertDir, "etcd-peer.key")
+	s.PeerTrustedCAFile = caCertFile
```
stlaz, line 56: You'll want to fix these with the functions from https://github.com/openshift/microshift/blob/main/pkg/util/cryptomaterial/certpaths.go#L118-L126

See https://github.com/openshift/microshift/blob/main/pkg/controllers/etcd.go#L82-L88

stlaz:
/lgtm cancel
/test all
the cert paths need fixing

dhellmann:
I think at this point this change is a good candidate for early 4.13. Let's catch our breath, get the enhancement in order, and line up the other work in a small number of incremental PRs to land after we branch 4.12.

pkg/controllers/systemd-etcd.go
dusk125: Removed the unneeded information, but kept the todo since the args there are what we should be using once we have the etcd binary.

dusk125:
The current state of this PR:
- The transient systemd unit will launch embedded etcd through a temporary `microshift` subcommand.
- This should automatically detect if MicroShift is running in a systemd unit or not and change if etcd is launched in a systemd unit or not.
- The external etcd launching code is there and locked behind an environment variable feature flag; if a developer sets the environment variable `EXTERNAL_ETCD`  to any non-empty value, my changes will attempt to launch an etcd binary that's on disk in the same location as the MicroShift binary. This flag will be removed once the [submodule PR](https://github.com/openshift/microshift/pull/1028) has been accepted.

pmtk:
@dusk125 just fyi: the go submodule PR is not complete - it's more in a WIP+POC state. However it might be easier for others to just review complete solution (it's not that big). Next sprint I can take a look again at it, do all the things and the we could set my PR to merge into yours and review it, WDYT?

dhellmann:
> I'm open to just going ahead and including the submodule work here as well and forgoing the feature flag, but I'm also fine to continue the tiered approach to these changes. Thoughts @deads2k @pmtk @copejon ?

How do we test the change in CI if it is behind a feature flag?

dusk125:
@pmtk That's kinda what I was thinking, and that sounds good to me if the others are ok reviewing one large PR.

@dhellmann Currently, the stuff behind the feature flag doesn't work because we don't have an etcd binary laid down, and when we do, we can remove the feature flag and it will run in CI.

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,153 @@
...
+	execPath, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("%v failed to get exec path: %v", s.Name(), err)
+	}
+
+	var etcdPath string
+	var args []string
+	// This is a temporary measure to run embedded etcd with the transient systemd unit.
+	// TODO remove this once we have the etcd binary laid down.
+	if os.Getenv("EXTERNAL_ETCD") == "" {
```
pmtk, line 84: What I did in my poc was using `INVOCATION_ID` env var - if MicroShift is running as systemd service - it'll be set

pkg/controllers/systemd-etcd.go
```diff
@@ -0,0 +1,153 @@
...
+			"--client-cert-auth=true",
+			// TODO - we might not need peers since we're single node
+			"--peer-cert-file=" + s.PeerCertFile,
+			"--peer-key-file=" + s.PeerKeyFile,
+			"--peer-trusted-ca-file=" + s.PeerTrustedCAFile,
+			"--peer-client-cert-auth=true",
+			"--advertise-client-urls=" + s.AdvertiseClientUrls,
+			"--listen-client-urls=" + s.ListenClientUrls,
+			"--metrics=extensive",
+			"--listen-metrics-urls=" + s.ListenMetricsUrls,
```
pmtk, line 111: So this looks like a foundation for situation when we use etcd binary directly (with no microshift-etcd, shims, etc.). Do we actually need that right now? I think this might be confusing when reading code

dusk125: We don't need it right now, which is why I put it under a flag so we could start manually testing with the binary.

pmtk:
I refreshed my https://github.com/openshift/microshift/pull/1028 - it only includes etcd as a separate module. I think next step would be merge our changes and remove old etcd in MicroShift

dusk125:
@dhellmann do we still need the hold, or is this ready to move forward?

dhellmann:
> @dhellmann do we still need the hold, or is this ready to move forward?

I want to l and the enhancement before we start landing code changes. That way the code reviewers know the design is settled and they can review with it in mind.

dusk125:
The hold here needs to stay until this PR is updated again; we decided to get Patryk's submodule work (quoted below) in first which will simplify this PR as we won't have to have the feature flag disabling the systemd launch (which was added to wait for the submodule work).

> I refreshed my #1028 - it only includes etcd as a separate module. I think next step would be merge our changes and remove old etcd in MicroShift

dusk125:
#1028 has been merged and this PR has been updated with its changes. This is ready for review @dhellmann @pmtk

pkg/controllers/etcd.go
```diff
@@ -36,25 +37,119 @@ var (
...
+	// Get the path to the etcd binary based on the MicroShift binary location.
 	microshiftExecPath, err := os.Executable()
 	if err != nil {
 		return fmt.Errorf("%v failed to get exec path: %v", s.Name(), err)
 	}
 	etcdPath := filepath.Join(filepath.Dir(microshiftExecPath), "microshift-etcd")
-	args := []string{"run"}
 
+	// All of the etcd parameters
+	args := []string{
```
pmtk, line 114: Take a look at `microshift/etcd/cmd/microshift-etcd/run.go` - we're not running `etcd` binary directly so we don't need all these arguments

pkg/cmd/run.go
```diff
@@ -93,7 +93,7 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 	}
 
 	m := servicemanager.NewServiceManager()
-	util.Must(m.AddService(controllers.NewEtcd(cfg)))
+	util.Must(m.AddService(controllers.NewExtEtcd(cfg)))
```
pmtk, line 96: I don't see `NewExtEtcd()` in this PR

pkg/controllers/etcd.go
dusk125: Ahh missed that, thanks.

pkg/controllers/etcd.go
```diff
@@ -46,15 +46,39 @@ func (s *EtcdService) Name() string           { return "etcd" }
...
+	var exe string
+	if runningAsSvc {
+		args = append([]string{
+			"--uid=root",
+			"--scope",
+			"--property", "BindsTo=microshift.service",
+			"--unit", "microshift-etcd",
+			etcdPath,
+		}, args...)
+		exe = "systemd-run"
```
pmtk, line 76: Is this `systemd-run` execution a blocking one?
Or does it simply starts the unit and exits? - in this case, cmd.Process used in `defer func()` after `cmd.Start()` can cause some misleading logs to appear.

If that's non-blocking, then maybe for transient unit we want `cmd.Run()` instead of `cmd.Start()`?
We'd add `TimeoutStartSec` to unit and `github.com/coreos/go-systemd/daemon.daemon.SdNotify(false, daemon.SdNotifyReady)` (like we have [here](https://github.com/openshift/microshift/blob/main/pkg/cmd/run.go#L142)) to the `microshift-etcd`'s `run.go`.
In such case `cmd.Run()` would run to completion which would be "starting the unit and informing whether it was started successfully".

I know `microshift-etcd` will be tied to `microshift`, but maybe we could use that as a fail safe, rather than default path.
WDYT about shutting down `microshift-etcd` unit gracefully (i.e. issuing stop and waiting for exit) upon microshift's shutdown? I'm not sure what SIG will systemd send to `microshift-etcd`.
Because with etcd logs in different unit we could miss things when looking at `journalctl -u microshift`.

dusk125: systemd-run is non-blocking; it creates/starts the unit and exits. 
> In such case cmd.Run() would run to completion which would be "starting the unit and informing whether it was started successfully".
Yeah this makes sense, I'll make this change.

I'm fine adding a bit more to not rely on systemd to shut us down when shutting down ushift.

pkg/controllers/etcd.go
```diff
@@ -63,7 +89,7 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
 	cmd.Dir = wd
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	if err := cmd.Start(); err != nil {
+	if err := cmd.Run(); err != nil {
```
pmtk, line 92: What I meant with `Start` vs `Run` is use `Start()` for executing `microshift-etcd` directly - developer (non systemd) mode and use `Run()` for transient systemd unit

pkg/controllers/etcd.go
```diff
@@ -87,6 +113,17 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
 	klog.Info("etcd is ready!")
 	close(ready)
 
+	// Once we're ready and running as a service, tell systemd that we're ready.
+	if runningAsSvc {
+		if supported, err := daemon.SdNotify(false, daemon.SdNotifyReady); err != nil {
```
pmtk, line 118: I think this needs to be in the `microshift-etcd` binary - `microshift` already has this and we don't want to publish this just after etcd is up, but after every component is up and running.

Let me rephrase my idea:
- `microshift` executes `systemd-run ... microshift-etcd run` with blocking `Run()`
- `microshift-etcd` can:
  - exit with error code (which systemd-run should propagate)
  - timeout - didn't sent READY ntfy in N seconds (would 30 be a sane default?), systemd-run kills it and propagates error
  - send READY, continue running in background, systemd-run exits with success (0)
- `Run()` function returns, `microshift` inspects error code, if success, then process with testing etcd endpoint, if failure then exit

All of this is based on a assumption that systemd supports and honors `SdNotifyReady` for transient units. Because we're not monitoring unit's health, it could be that we fire and forget the systemd-run but are left without etcd, right?

dusk125: Thanks for rephrasing, that makes much more sense!

pkg/controllers/etcd.go
pmtk: I'm sorry @dusk125 
I just checked systemd-run docs and it states:
> If a command is run as transient scope unit, (...).
> Execution in this case is synchronous, and will return only when the command finishes. This mode is enabled via the --scope switch (see below).

Coming back to my questions:
> Is this systemd-run execution a blocking one?
> (...)
> If that's non-blocking, then maybe for transient unit we want cmd.Run() instead of cmd.Start()?

So it looks like we don't have the luxury of "reusing systemd unit's timeout" which I meant there.
We just need to treat `systemd-run` just like we'd treat `microshift-etcd`, right?
This means that there's no use of "Run()" because it'll block indefinitely.
Also all what I wrote about `daemon.SdNotify()` is unnecessary (because there's no such mechanism for transient units)

pmtk: So it seems that we can revert changes related to that and just keep wrapping microshift-etcd with systemd... Sorry for that, I should've check that earlier...

dusk125: Will do, thanks!

pmtk:
This looks good to me, I'd like others to take a look
@dhellmann @pacevedom @copejon

etcd/cmd/microshift-etcd/run.go
```diff
@@ -95,15 +97,29 @@ func (s *EtcdService) Run() error {
 	if err != nil {
 		return fmt.Errorf("microshift-etcd failed to start: %v", err)
 	}
-	<-e.Server.ReadyNotify()
+	defer func() {
+		e.Server.Stop()
+		<-e.Server.StopNotify()
+	}()
+
+	to := 30 * time.Second
```
dhellmann, line 105: How was this timeout selected? What happens if it takes longer for etcd to start up? (Say there's a database format change or something and the content has to be migrated.)

This check appears to be happening inside the etcd process, is that right? Why here instead of in the caller?

pkg/controllers/etcd.go
```diff
@@ -46,15 +46,39 @@ func (s *EtcdService) Name() string           { return "etcd" }
 func (s *EtcdService) Dependencies() []string { return []string{} }
 
 func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+	defer close(stopped)
+
+	// Check to see if we should run as a systemd run or directly as a binary.
+	runningAsSvc := os.Getenv("INVOCATION_ID") != ""
```
dhellmann, line 52: The main thing we wanted was to always use the microshift-etcd binary in the same directory as the microshift binary, to support the developer use case. Does it matter if that developer binary is run in a systemd unit? Would always using system simplify anything below?

dhellmann (commented):
We've been trying to improve the git log messages with new changes. You could consider adding some of the detail from the PR message to the commit message.

I had a couple of questions inline, but nothing I think should block this.

/hold cancel
/approve

I'll leave final approval for @pmtk.

pkg/controllers/etcd.go
pmtk: That's a good idea. Only thing is that etcd logs would be separate, but I don't think we need them that much in day to day development

etcd/cmd/microshift-etcd/run.go
pmtk: I think we could even drop it, we have a health-check in microshift that checks etcd health (via etcd client)

dhellmann: OK, maybe we want to land this PR and then remove the check separately?

pkg/controllers/etcd.go
dhellmann: Maybe this is another area to simplify after the initial version is merged?

dusk125: IIRC, there was some pushback in a previous meeting about not executing etcd as a systemd unit in development, I don't remember the reason given.

etcd/cmd/microshift-etcd/run.go
dusk125: I just dropped it in a new update

pkg/controllers/etcd.go
dhellmann: I may have been worried that it wouldn't work, or wouldn't stop, or something. I don't have a preference, as long as it works. We can land it this way and experiment with removing it when we have some time.

etcd/cmd/microshift-etcd/run.go
dhellmann: Thanks!

PR #978, opened 2022-09-27
[WIP] Move CI build spec to break the cyclic dependency

copejon:
Move the build image file to this repo.  This addresses a chicken and egg cycle that happens when code is submitted that requires build changes.

PR openshift/release#32625 depends on this change.


packaging/images/openshift-ci/rpm.Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM registry.redhat.io/ubi8/ubi
+
+ENV GOPATH=/go
+
+WORKDIR $GOPATH/src/github.com/openshift/microshift
+
+RUN dnf update -y &&  \
+    dnf install --setopt=tsflags=nodocs -y \
+    git \
+    golang-1.17.12-1.module+el8.6.0+16014+a372c00b && \
```
dhellmann, line 10: Is there some way to base this image on a golang builder image so we don't have to keep updating the version here for every release?

PR #979, opened 2022-09-27
rebase.sh: Use shallow clone

vfreex:
This will speed up the rebase process and save a lot of bandwidth.

### DIFF
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -76,6 +76,7 @@ download_release() {
     echo "${commits}" | jq -r '.references.spec.tags[] | "\(.name) \(.annotations."io.openshift.build.source-location") \(.annotations."io.openshift.build.commit.id")"' > source-commits
 
     git config --global advice.detachedHead false
+    git config --global init.defaultBranch main
     while IFS="" read -r line || [ -n "$line" ]
     do
         component=$(echo "${line}" | cut -d ' ' -f 1)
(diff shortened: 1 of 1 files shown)


scripts/rebase.sh
```diff
@@ -83,8 +83,10 @@ download_release() {
         commit=$(echo "${line}" | cut -d ' ' -f 3)
         if [[ "${EMBEDDED_COMPONENTS}" == *"${component}"* ]] || [[ "${LOADED_COMPONENTS}" == *"${component}"* ]] || [[ "${EMBEDDED_COMPONENT_OPERATORS}" == *"${component}"* ]]; then
             title "## Cloning ${repo} at commit ${commit}..."
-            git clone "${repo}"
```
mangelajo, line 86: why not simply `git clone --depth=1` "${repo}" ?

vfreex: `git clone --depth=1 "${repo}"` will have a single commit that the `master` or `main` branch tip points to. It might not be the commit we are interested in here.

fzdarsky: /approve

That LGTM and drastically speeds up the cloning, thank you!

Would it make sense to add a `git config --global init.defaultBranch main`, so we don't pollute the logs with the hints about missing default branch (which at least I am getting with my git version 2.31)?

fzdarsky: Hope you don't mind I added this to the PR, so we can merge it.

mangelajo: Right!, sorry @vfreex

PR #981, opened 2022-09-29
AUTH-296: generate kubelet certs properly

stlaz:
This PR fixed cert generation for kubelet client cert to use the CSR signer instead of the root CA and uses this same signer to generate the serving cert for it.

/assign @oglok

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -153,6 +153,14 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 					// userinfo per https://kubernetes.io/docs/reference/access-authn-authz/node/#overview
 					UserInfo: &user.DefaultInfo{Name: "system:node:" + cfg.NodeName, Groups: []string{"system:nodes"}},
 				},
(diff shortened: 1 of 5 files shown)


pkg/cmd/init.go
```diff
@@ -139,17 +139,29 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 			"kubelet-signer",
 			cryptomaterial.KubeletCSRSignerSignerCertDir(certsDir),
 			cryptomaterial.KubeControllerManagerCSRSignerSignerCAValidityDays,
-		).WithClientCertificates(
```
oglok, line 142: So instead of creating the client certificates for kubelet using the kubelet signer, you're creating a new signer using a sub CA (kube-csr-signer) that provides client AND serving certificates for kubelet. Is this correct?

oglok: Is this aligned with OpenShift's cert management? https://github.com/openshift/api/tree/master/tls/docs

stlaz:
the CI is still broken it seems :(

PR #986, opened 2022-10-02
Fix wait-for-pods command in performance test scripts 

ggiguash:
Fix wait-for-pods command in performance test scripts not to depend on the number of pods
Closes [USHIFT-294](https://issues.redhat.com//browse/USHIFT-294)

### DIFF
--- a/ansible/roles/common/tasks/readinesscheck.yml
+++ b/ansible/roles/common/tasks/readinesscheck.yml
@@ -2,9 +2,8 @@
 # Wait for pods to be running
 
 - name: wait for pods running
-  ansible.builtin.shell: oc get po -A --no-headers | grep Running | wc -l
+  ansible.builtin.shell: oc get pods -A -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}' | grep -c False
   retries: 60
   delay: 5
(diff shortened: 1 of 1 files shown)


ansible/roles/common/tasks/readinesscheck.yml
```diff
@@ -2,9 +2,8 @@
 # Wait for pods to be running
 
 - name: wait for pods running
-  ansible.builtin.shell: oc get po -A --no-headers | grep Running | wc -l
+  ansible.builtin.shell: oc get pods -A -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}' | grep -c False
```
pacevedom, line 5: Just as a suggestion, did you try `oc wait --for=condition=Ready pods --all --all-namespaces`? You would get rid of the ansible loop.

pacevedom (approved):

ansible/roles/common/tasks/readinesscheck.yml
ggiguash: I tried this. Unfortunately, it seems to return true before all containers of a given pod are in Ready state.

PR #988, opened 2022-10-03
USHIFT-209: Refactor OpenShift Controller Manager configuration

oglok:

Closes [USHIFT-209](https://issues.redhat.com//browse/USHIFT-209)

### DIFF
--- a/pkg/controllers/openshift-controller-manager.go
+++ b/pkg/controllers/openshift-controller-manager.go
@@ -106,8 +106,6 @@ func (s *OCPControllerManager) configure(cfg *config.MicroshiftConfig) {
 func (s *OCPControllerManager) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
 	defer close(stopped)
 
-	errorChannel := make(chan error, 1)
-
 	// run readiness check
 	go func() {
 		healthcheckStatus := util.RetryTCPConnection("127.0.0.1", "8445")
@@ -129,16 +127,10 @@ func (s *OCPControllerManager) Run(ctx context.Context, ready chan<- struct{}, s
 		return err
 	}
 
-	go func(ctx context.Context) error {
+	if err := openshift_controller_manager.RunOpenShiftControllerManager(&s.config, clientConfig); err != nil {
+		return err
+	}
 
-		errorChannel <- openshift_controller_manager.RunOpenShiftControllerManager(&s.config, clientConfig)
-
-		select {
-		case <-ctx.Done():
-			klog.Infof("OpenShift Controller Manager received stop signal. exiting.")
-			return nil
-		}
-	}(ctx)
-
-	return <-errorChannel
+	<-ctx.Done()
+	return nil
 }
--- a/pkg/controllers/openshift-controller-manager.go
+++ b/pkg/controllers/openshift-controller-manager.go
@@ -17,10 +17,8 @@ package controllers
 
 import (
 	"context"
-	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/library-go/pkg/config/helpers"
 	"k8s.io/klog/v2"
@@ -32,6 +30,9 @@ import (
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
+
+	"github.com/openshift/library-go/pkg/config/configdefaults"
+	"github.com/openshift/library-go/pkg/config/leaderelection"
 )
 
 type OCPControllerManager struct {
@@ -80,12 +81,6 @@ func (s *OCPControllerManager) configure(cfg *config.MicroshiftConfig) {
 			"-openshift.io/templateinstancefinalizer",
 			"-openshift.io/unidling",
 		},
-		LeaderElection: configv1.LeaderElection{
-			Disable:       true,
-			LeaseDuration: metav1.Duration{Duration: 270 * time.Second},
-			RenewDeadline: metav1.Duration{Duration: 240 * time.Second},
-			RetryPeriod:   metav1.Duration{Duration: 60 * time.Second},
-		},
 		ServingInfo: &configv1.HTTPServingInfo{
 			ServingInfo: configv1.ServingInfo{
 				BindAddress: "0.0.0.0:8445",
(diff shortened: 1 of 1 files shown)


pkg/controllers/openshift-controller-manager.go
```diff
@@ -129,8 +122,10 @@ func (s *OCPControllerManager) Run(ctx context.Context, ready chan<- struct{}, s
...
 	}
 
-	options := openshift_controller_manager.OpenShiftControllerManager{Output: os.Stdout}
-	options.ConfigFilePath = s.ConfigFilePath
+	clientConfig, err := helpers.GetKubeClientConfig(configv1.KubeClientConfig{KubeConfig: s.kubeconfig})
+	if err != nil {
+		return err
+	}
 
-	return options.StartControllerManager(ctx.Done())
```
benluddy, line 135: How will the MicroShift service manager tell OCM to stop?

benluddy: With `StartControllerManager`, OCM should stop after it sees the context's done channel has closed. Under the proposed patch, `RunOpenShiftControllerManager` will ignore shutdown from the service manager and will keep running until the process dies.

fzdarsky:
What are our goals with this refactoring? I understand it's to avoid having config files on disk. However, there is another goal we may want to keep in mind, which is the ability to programmatically compare parameters used in OCP with that of MicroShift during rebasing. And I'm wondering whether moving away from config.yamls is making this harder?

pkg/controllers/openshift-controller-manager.go
oglok: > With `StartControllerManager`, OCM should stop after it sees the context's done channel has closed. Under the proposed patch, `RunOpenShiftControllerManager` will ignore shutdown from the service manager and will keep running until the process dies.

Do you think we can pass the openshiftcontrolplanev1.OpenShiftControllerManagerConfig struct with all the configs to StartControllerManager ?

oglok:
> What are our goals with this refactoring? I understand it's to avoid having config files on disk. However, there is another goal we may want to keep in mind, which is the ability to programmatically compare parameters used in OCP with that of MicroShift during rebasing. And I'm wondering whether moving away from config.yamls is making this harder?

I wasn't aware of this goal regarding automated rebases. I was aware of exposing configurations, but using structs and being opinionated on which params we allow users to modify sounded good to me anyways. Is the rebase script already comparing params that way?

fzdarsky:
> > What are our goals with this refactoring? I understand it's to avoid having config files on disk. However, there is another goal we may want to keep in mind, which is the ability to programmatically compare parameters used in OCP with that of MicroShift during rebasing. And I'm wondering whether moving away from config.yamls is making this harder?
> 
> I wasn't aware of this goal regarding automated rebases. I was aware of exposing configurations, but using structs and being opinionated on which params we allow users to modify sounded good to me anyways. Is the rebase script already comparing params that way?

Yeah, I agree those are good goals, I just wanted to raise this additional consideration, because - to my knowledge - we haven't discussed more broadly how we'd like this to work yet. No, the rebase script only does this comparison for hosted components, not embedded components, _because_ I wouldn't know how to programmatically do that with the current structure.

pkg/controllers/openshift-controller-manager.go
oglok: I've created a go routing to pass the context as if the `StartControllerManager` function does.

pkg/controllers/openshift-controller-manager.go
```diff
@@ -129,8 +124,21 @@ func (s *OCPControllerManager) Run(ctx context.Context, ready chan<- struct{}, s
...
+	go func(ctx context.Context) error {
+
+		errorChannel <- openshift_controller_manager.RunOpenShiftControllerManager(&s.config, clientConfig)
+
+		select {
+		case <-ctx.Done():
+			klog.Infof("OpenShift Controller Manager received stop signal. exiting.")
+			return nil
+		}
+	}(ctx)
```
benluddy, line 141: Since RunOpenShiftControllerManager doesn't block, I don't think that calling it from this goroutine is providing any benefit over something like:

```suggestion
	if err := openshift_controller_manager.RunOpenShiftControllerManager(&s.config, clientConfig); err != nil {
		return err
	}
	<-ctx.Done()
	return nil
```

pkg/controllers/openshift-controller-manager.go
benluddy: My assumption that StartControllerManager wires the provided channel to shutdown was wrong, so I don't think it necessary needs to be addressed here.

pkg/controllers/openshift-controller-manager.go
oglok: Suggestion accepted

atiratree:
I am doing something similar and it should be superseded by https://github.com/openshift/microshift/pull/930 as we only want to use route-controller-manager from now on

pkg/controllers/openshift-controller-manager.go
```diff
@@ -56,58 +57,50 @@ func (s *OCPControllerManager) Dependencies() []string {
...
+			"-openshift.io/image-trigger",
+			"-openshift.io/origin-namespace",
+			"-openshift.io/serviceaccount",
+			"-openshift.io/serviceaccount-pull-secrets",
+			"-openshift.io/templateinstance",
+			"-openshift.io/templateinstancefinalizer",
+			"-openshift.io/unidling",
+		},
+		LeaderElection: configv1.LeaderElection{
+			Disable:       true,
```
atiratree, line 84: this is not going to work now since we would need to use a builder in the controller to utilize this, but it will be possible once we refactor route-controller-manager to BuilderController

pkg/controllers/openshift-controller-manager.go
atiratree: it should be possible to gracefully shutdown once we get this in https://github.com/openshift/route-controller-manager/pull/6

oglok:
@benluddy If you agree, let's merge this untill we finally shift to the route controller, but at least we can deliver this in the next drop.

pkg/controllers/openshift-controller-manager.go
```diff
@@ -56,58 +58,47 @@ func (s *OCPControllerManager) Dependencies() []string {
...
-	// If a non-zero length list is specified, only controllers enabled in the list are started.  Unlisted controllers
-	// are therefore disabled.  Enable controllers by appending their name to `controllers:`. Disable a controller by
-	// prepending "-" to the name, e.g. `controllers: ["-openshift.io/build"]
-	// Disabled OCM controllers are included in the list for documentary purposes.
-	data := []byte(`apiVersion: openshiftcontrolplane.config.openshift.io/v1
-kind: OpenShiftControllerManagerConfig
-kubeClientConfig:
-  kubeConfig: ` + s.kubeconfig + `
-  connectionOverrides:
-    contentType: "application/json"
```
benluddy, line 83: Where is this in the new revision?

oglok: I'll add it!

oglok:
OCM is completely useless at its current state for MicroShift. A configuration refactor would not make any difference.

PR #990, opened 2022-10-03
USHIFT-445: Add prometheus exporter for RHEL4EDGE

sjug:
This PR adds a `-prometheus` flag to the image-builder build script. At the moment that is specifically the process-exporter which provides pidstat like data for all host processes. 

Small bugfixes for `build.sh` also included as well as firewall customization to allow inbound connections to the exporter.

### DIFF
--- a/scripts/image-builder/build.sh
+++ b/scripts/image-builder/build.sh
@@ -13,8 +13,9 @@ OCP_PULL_SECRET_FILE=
 MICROSHIFT_RPM_SOURCE=${ROOTDIR}/_output/rpmbuild/
 AUTHORIZED_KEYS_FILE=
 AUTHORIZED_KEYS=
+PROMETHEUS=false
 STARTTIME=$(date +%s)
-BUILDDIR=${ROOTDIR}/_output/image-builder/
+BUILDDIR=${ROOTDIR}/_output/image-builder
 
 trap ${SCRIPTDIR}/cleanup.sh INT
 
@@ -170,6 +171,10 @@ while [ $# -gt 0 ] ; do
         [ -z "${AUTHORIZED_KEYS_FILE}" ] && usage "Authorized keys file not specified"
         shift
         ;;
+    -prometheus)
+        PROMETHEUS=true
+        shift
+        ;;
     *)
         usage
         ;;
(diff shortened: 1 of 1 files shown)


scripts/image-builder/build.sh
```diff
@@ -246,8 +251,22 @@ if [ ! -z ${CUSTOM_RPM_FILES} ] ; then
     createrepo custom-rpms >/dev/null
 fi
 
+# Install prometheus process exporter
+rm -rf prometheus-rpms 2>/dev/null || true
+if ${PROMETHEUS} ; then
```
ggiguash, line 256: Let's us a title ".." message here so that user sees RPMs downloading, etc.

scripts/image-builder/config/blueprint_v0.0.1.toml
```diff
@@ -43,5 +43,8 @@ version = "*"
 
 [customizations]
 
+[customizations.firewall]
```
ggiguash, line 46: This change cannot be unconditional. Please, append the firewall settings to the copied blueprint only if prometheus option is selected

scripts/image-builder/config/prometheus-rpms.toml.template
```diff
@@ -0,0 +1,7 @@
+id = "prometheus-rpms"
```
ggiguash, line 1: This file is probably not required since you're reusing the custom RPM script option, right?

scripts/image-builder/build.sh
```diff
@@ -246,8 +251,22 @@ if [ ! -z ${CUSTOM_RPM_FILES} ] ; then
...
+  owner=ncabatoff
+  repo=process-exporter
+  version=$(curl -s https://api.github.com/repos/${owner}/${repo}/releases/latest | jq -r '.name')
+  url=https://github.com/${owner}/${repo}/releases/download/${version}/
+  file=${repo}_${version#v}_linux_amd64.rpm
+
+  wget -q -P prometheus-rpms ${url}${file}
+  CUSTOM_RPM_FILES+="$(pwd)/prometheus-rpms/${file},"
```
ggiguash, line 264: This is a bit inconsistent. We should either create a new repo and leave custom RPMs alone, or we should download to custom-rpms directory and not create a new repo.

ggiguash (changes requested):

scripts/image-builder/build.sh
sjug: Custom RPMs has an existing mechanism to modify the blueprint and I'm reusing that rather than duplicating it.

ggiguash: OK, if we want to reuse custom RPMs, let's move the code of changing CUSTOM_RPM_FILES setting before the line 244. This way, we should not need prometheus-rpms.toml template at all - all the prometheus RPMs will be copied to custom-rpms folder and included automatically.

ggiguash (changes requested):

ggiguash:
@sjug please squash the commits when you make the last change.

PR #992, opened 2022-10-04
add etcd shim to microshift until we package proper etcd separately

deads2k:
this will allow `microshift etcd` to behave like a standalone etcd process so we can start merging PRs that run a transient systemd unit before packaging is complete.

/assign @dusk125


deads2k:
included in #976

PR #994, opened 2022-10-05
USHIFT-442: Refactor logging exporters installation and update docs

ggiguash:
Closes [USHIFT-442](https://issues.redhat.com//browse/USHIFT-442)

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,26 +1,23 @@
-# ansible-microshift
+# Ansible MicroShift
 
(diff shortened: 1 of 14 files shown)


sjug:
I don't think we should remove pbench, we used it specifically for the CI execution case for which it is well suited.
Afterwards I added support for the devenv and rhel4edge images, and used Prometheus as it doesn't have external dependencies for these development environments.

There should be two different playbooks that exists for different targets but I don't think that removing pbench is productive.

sjug (approved):
/lgtm

PR #924, opened 2022-09-02
[WIP] Enable kube-apiserver option send-retry-after-while-not-ready-once

tkashem:
Relates to: [API-1437](https://issues.redhat.com//browse/API-1437): Clients (including internal clients) must not use an unready Kubernetes apiserver

This is expected to fail since https://github.com/openshift/kubernetes/pull/1346 has not merged yet, and we need to wait microshift  rebase 1.25 is done.

PR #995, opened 2022-10-05
API-1437: Enable kube-apiserver option send-retry-after-while-not-ready-once

benluddy:
Rebase of https://github.com/openshift/microshift/pull/924.

### DIFF
--- a/pkg/controllers/kube-apiserver.go
+++ b/pkg/controllers/kube-apiserver.go
@@ -150,7 +150,8 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 				"quota.openshift.io/ClusterResourceQuota",
 				"quota.openshift.io/ValidateClusterResourceQuota",
 			},
-			"enable-admission-plugins": {},
+			"enable-admission-plugins":              {},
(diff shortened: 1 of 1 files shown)


pkg/controllers/kube-apiserver.go
```diff
@@ -150,7 +150,8 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 				"quota.openshift.io/ClusterResourceQuota",
 				"quota.openshift.io/ValidateClusterResourceQuota",
 			},
-			"enable-admission-plugins": {},
+			"enable-admission-plugins":              {},
+			"send-retry-after-while-not-ready-once": {"true"},
```
oglok, line 154: Has this option been merged into OpenShift's KAS?

Upstream seems to be targeting 1.26 IIUC: https://github.com/kubernetes/kubernetes/commit/6e3923d0a4f4720d2d9f628eb9c073d2d3ee291a

benluddy: It's a carried patch in openshift/kubernetes. We don't expect it to ever exist upstream, because the usual recommendation for this problem is to run an external load balancer.

PR #996, opened 2022-10-05
USHIFT-313: Refactor microshift-start standalone role

sjug:
- Make disk space task common
- Change readiness check error handling
- Install grafana on logging host
- Update README with new vars

This closes [USHIFT-313](https://issues.redhat.com//browse/USHIFT-313), streamlines rhel4edge perf workflow.


ansible/roles/microshift-start/tasks/main.yml
```diff
@@ -46,16 +79,40 @@
...
+  retries: 10
 
-- include_tasks: roles/common/tasks/readinesscheck.yml
+- name: wait for pods running
+  ansible.builtin.shell: oc get pods -A -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}' | grep -c False
+  retries: 120
+  delay: 5
+  register: result
+  until: result.stdout == "0"
+  failed_when: result.stderr != ''
```
ggiguash, line 115: Let's move these wait for any and running back to the separate task to keep the file shorter.

ansible/roles/microshift-start/tasks/main.yml
```diff
@@ -46,16 +79,40 @@
...
+
+- name: create copy of kubeconfig
+  become: yes
+  ansible.builtin.copy:
+    src: "{{ remote_kubeconfig_path }}"
+    dest: "/home/{{ ansible_user }}/.kube/config"
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+    remote_src: yes
+    force: yes
```
ggiguash, line 101: kubeconfig init could also be moved to a separate task in commons, together with ~/.kube directory creation

ansible/roles/microshift-start/tasks/main.yml
```diff
@@ -1,21 +1,48 @@
...
   block:
   - name: source pbench-agent & register-tool-set
     ansible.builtin.shell: source /etc/profile.d/pbench-agent.sh && pbench-register-tool-set
   
+  - name: set new pidstat interval
+    ansible.builtin.shell: source /etc/profile.d/pbench-agent.sh && pbench-register-tool --name=pidstat -- --interval={{ sample_interval }}
+
   - name: start pbench-user-benchmark recording
-    ansible.builtin.shell: source /etc/profile.d/pbench-agent.sh && pbench-user-benchmark --config=microshift2 -- sleep {{ pbench_record_duration }}
+    ansible.builtin.shell: source /etc/profile.d/pbench-agent.sh && pbench-user-benchmark --config=microshift -- sleep {{ pbench_record_duration }}
```
ggiguash, line 45: Don't we need to add condition here for pbench presence? We may want to collect stats using Prometheus while pbench is not installed.

ggiguash: Again, we may also wait to refactor this start/stop code to a separate task in commons to keep this file managable

ansible/roles/microshift-start/tasks/main.yml
```diff
@@ -69,22 +126,29 @@
 - name: Add boot info to local file
   lineinfile:
     path: boot.txt
-    line: "Second boot: {{ runtime }}"
+    line: "First boot: {{ runtime }}"
     create: yes
   delegate_to: localhost
 
 - name: finish pbench capture
+  become: yes
```
ggiguash, line 134: This code needs to be conditional so that we can run Prometheus stats for the start procedure.
It would be nice to have it moved to common tasks as well.

ansible/vars/all.yml
```diff
@@ -2,9 +2,11 @@ manage_subscription: false
 rhel_username:
 rhel_password:
 rhel_pool_id:
+manage_repos: false
 
-prometheus_logging: true
 setup_microshift_host: false
+prometheus_logging: true
+install_microshift: false
```
ggiguash, line 9: This is true by default for all the cases except RHEL4Edge, right? Should we default to true and update the README.md with one more case?

ggiguash (changes requested):
I have a few more comments for refactoring the code and making sure we allow for both pbench and Prometheus collection.

ansible/roles/microshift-start/tasks/main.yml
sjug: Doesn't that make it harder to read jumping around files? It was in common before because it was used in multiple locations, but that's no longer the case.

ansible/vars/all.yml
sjug: Correct, the defaults I've set are for the RHEL4Edge case. I'll add an examples section to the README.

ansible/roles/install-logging/defaults/main.yml
```diff
@@ -1,2 +1,10 @@
 ---
 # install-logging default vars
+
+logging_packages:
+ - golang-github-prometheus
+ - grafana
```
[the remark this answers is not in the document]
sjug: I'm just installing the package and starting the service, the bulk grafana work will come as part of USHIFT-448.
I can remove anything to do with grafana from this PR if you'd like.

ansible/README.md
```diff
@@ -94,9 +96,32 @@ manage_subscription: false
...
 
+setup_microshift_host: false
 prometheus_logging: true
+install_microshift: true
+build_etcd_binary: false
+```
+
+#### RHEL4Edge Host Example Variables
+
+If the user has deployed a RHEL4Edge image built with MicrosShift, the scripts can be run with a few more variables disabled as the host is fully configured out of the box.
```
ggiguash, line 109: MicroShift is not spelled correctly.

ansible/roles/microshift-start/tasks/main.yml
ggiguash: I think this is an implementation detail, which is not related to the current task. I should be easier to find/fix such procedures if they are kept in common/tasks, I think.

ansible/roles/microshift-start/tasks/main.yml
ggiguash: I must be missing where this condition is. Can you point me to the line?

ansible/roles/microshift-start/tasks/main.yml
sjug: Line 35 to Line 55 is a logical block (L37), the entire block of tasks to "start bench-user-benchmark recording" is conditional on `install_pbench` being true (L53).

ansible/roles/microshift-start/tasks/main.yml
sjug: This one is on line 146

PR #997, opened 2022-10-06
USHIFT-449: Fix auto-apply for manifests to run in a non-default namespace

ggiguash:
Closes [USHIFT-449](https://issues.redhat.com//browse/USHIFT-449)

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -75,7 +75,7 @@ The list of manifest locations can be customized via configuration using the abo
 
 ## Manifest Example
 
-The example demonstrates automatic deployment of an `nginx` container using `kustomize` manifests in the `/etc/microshift/manifests` directory.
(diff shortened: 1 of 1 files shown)


docs/howto_config.md
```diff
@@ -75,51 +75,57 @@ The list of manifest locations can be customized via configuration using the abo
...
       containers:
-      - name: nginx
-        image: NGINX_IMAGE
-        ports:
-        - containerPort: 8080
+      - name: busybox
+        image: BUSYBOX_IMAGE
+        command:
+          - sleep
+          - "3600"
```
mangelajo, line 110: I'm checking if something like this would work with this image
```suggestion
    - name: hello-world-http
      image: busybox
      command: ['sh','-c', 'echo "hello world" > index.html && /bin/httpd -p 9000 -f']
      ports:                 # A port this container is exposing
      - containerPort: 9000
        protocol: TCP
```

PR #1011, opened 2022-10-10
[WIP] User-facing Configuration API Discussion Kickoff

dinhxuanvu:
User-facing Configuration API Discussion Kickoff

Closes #<Issue Number>

### DIFF
--- a/pkg/config/api/config.go
+++ b/pkg/config/api/config.go
@@ -1,0 +1,46 @@
+package api
+
+type Config struct {
+	Components Components `json:components`
+}
+
+type Components struct {
+	Network Network `json:network`
+	DNS     DNS     `json:dns`
+}
+
+type Network struct {
+	// IP address pool to use for pod IPs.
+	// This field is immutable after installation.
+	ClusterNetwork []ClusterNetworkEntry `json:"clusterNetwork"`
+
+	// IP address pool for services.
+	// Currently, we only support a single entry here.
+	// This field is immutable after installation.
+	ServiceNetwork []string `json:"serviceNetwork"`
+
+	// The port range allowed for Services of type NodePort.
+	// If not specified, the default of 30000-32767 will be used.
+	// Such Services without a NodePort specified will have one
+	// automatically allocated from this range.
+	// This parameter can be updated after the cluster is
+	// installed.
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])-([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`
+	ServiceNodePortRange string `json:"serviceNodePortRange,omitempty"`
+}
+
+type ClusterNetworkEntry struct {
+	// The complete block for pod IPs.
+	CIDR string `json:"cidr"`
+}
+
+type DNS struct {
+	// baseDomain is the base domain of the cluster. All managed DNS records will
+	// be sub-domains of this base.
+	//
+	// For example, given the base domain `openshift.example.com`, an API server
+	// DNS record may be created for `cluster-api.openshift.example.com`.
+	//
+	// Once set, this field cannot be changed.
+	BaseDomain string `json:"baseDomain"`
+}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -37,37 +37,37 @@ var (
 )
 
 type ClusterConfig struct {
-	URL string `json:"url"`
+	// URL string `json:"url"`
 
-	ClusterCIDR          string `json:"clusterCIDR"`
-	ServiceCIDR          string `json:"serviceCIDR"`
-	ServiceNodePortRange string `json:"serviceNodePortRange"`
-	DNS                  string `json:"dns"`
-	Domain               string `json:"domain"`
-	MTU                  string `json:"mtu"`
+	// ClusterCIDR          string `json:"clusterCIDR"`
+	// ServiceCIDR          string `json:"serviceCIDR"`
+	// ServiceNodePortRange string `json:"serviceNodePortRange"`
+	// DNS string `json:"dns"`
+	// Domain               string `json:"domain"`
+	// MTU string `json:"mtu"`
 }
 
 type DebugConfig struct {
-	Pprof bool `json:"pprof"`
+	// Pprof bool `json:"pprof"`
 }
 
 type MicroshiftConfig struct {
-	ConfigFile string `json:"configFile"`
-	DataDir    string `json:"dataDir"`
-
-	AuditLogDir string `json:"auditLogDir"`
-	LogVLevel   int    `json:"logVLevel"`
-
-	Roles []string `json:"roles"`
-
-	NodeName string `json:"nodeName"`
-	NodeIP   string `json:"nodeIP"`
-
-	Cluster ClusterConfig `json:"cluster"`
-
-	Manifests []string `json:"manifests"`
-
-	Debug DebugConfig `json:"debug"`
+	// ConfigFile string `json:"configFile"`
+	// DataDir    string `json:"dataDir"`
+	//
+	// AuditLogDir string `json:"auditLogDir"`
+	// LogVLevel   int    `json:"logVLevel"`
+	//
+	// Roles []string `json:"roles"`
+	//
+	// NodeName string `json:"nodeName"`
+	// NodeIP   string `json:"nodeIP"`
+	//
+	// Cluster ClusterConfig `json:"cluster"`
+	//
+	// Manifests []string `json:"manifests"`
+	//
+	// Debug DebugConfig `json:"debug"`
 }
 
 // KubeConfigID identifies the different kubeconfigs managed in the DataDir


pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
-
-	NodeName string `json:"nodeName"`
-	NodeIP   string `json:"nodeIP"`
-
-	Cluster ClusterConfig `json:"cluster"`
-
-	Manifests []string `json:"manifests"`
-
-	Debug DebugConfig `json:"debug"`
+	// ConfigFile string `json:"configFile"`
```
fzdarsky, line 55: We probably no longer have a need to keep this user-configurable.

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
+	// DataDir    string `json:"dataDir"`
```
fzdarsky, line 56: We probably no longer have a need to keep this user-configurable.

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
+	//
+	// AuditLogDir string `json:"auditLogDir"`
+	// LogVLevel   int    `json:"logVLevel"`
```
fzdarsky, line 59: How else would we configure the verbosity level?

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
+	//
+	// Roles []string `json:"roles"`
```
fzdarsky, line 61: We'll need this once we split control plane and node roles again, but until then this doesn't have to be user-configurable.

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
-
-	Cluster ClusterConfig `json:"cluster"`
-
-	Manifests []string `json:"manifests"`
-
-	Debug DebugConfig `json:"debug"`
+	// ConfigFile string `json:"configFile"`
+	// DataDir    string `json:"dataDir"`
+	//
+	// AuditLogDir string `json:"auditLogDir"`
```
fzdarsky, line 58: Is the proposal to log these to stdout instead?

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
+	// LogVLevel   int    `json:"logVLevel"`
+	//
+	// Roles []string `json:"roles"`
+	//
+	// NodeName string `json:"nodeName"`
```
fzdarsky, line 63: We'll eventually need this override as host names cannot be guaranteed to be unique or stable.

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
+	// NodeIP   string `json:"nodeIP"`
+	//
+	// Cluster ClusterConfig `json:"cluster"`
+	//
+	// Manifests []string `json:"manifests"`
```
fzdarsky, line 68: This is where the bootstrap / workload manifests live. Probably doesn't have to be user-configurable, but we'll need well-known locations (in /etc, /var) for different use cases.

fzdarsky (commented):
It's a good idea to weed out config params that are no longer needed or can better be auto-configured and to rename for better UX. Just wondering what motivated the proposed structure / naming? In how far is this aligning with OCP?

pkg/config/config.go
benluddy: It could be selected arbitrarily (even randomly) and persisted to address uniqueness and stability. Is uniqueness an important consideration for MicroShift? My understanding is that there would never be more than one node, so even a fixed node name would be fine. What happens if a user reconfigures the node name after first start?

pkg/config/api/config.go
```diff
@@ -0,0 +1,46 @@
...
+	ServiceNetwork []string `json:"serviceNetwork"`
+
+	// The port range allowed for Services of type NodePort.
+	// If not specified, the default of 30000-32767 will be used.
+	// Such Services without a NodePort specified will have one
+	// automatically allocated from this range.
+	// This parameter can be updated after the cluster is
+	// installed.
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])-([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`
+	ServiceNodePortRange string `json:"serviceNodePortRange,omitempty"`
```
atiratree, line 29: are we planning to add/support `ExternalIPNetworkCIDRs` for LoadBalancer Services? 

https://access.redhat.com/documentation/en-us/openshift_container_platform/3.3/html/cluster_administration/admin-guide-unique-external-ips-ingress-traffic

atiratree: in OpenShift 4 configured via `networks.config/cluster` and `.spec.externalIP.AutoAssignCIDRs` field

https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html-single/networking/index#nw-service-externalip-create_configuring-ingress-cluster-traffic-service-external-ip

pkg/config/api/config.go
```diff
@@ -0,0 +1,46 @@
+package api
+
+type Config struct {
+	Components Components `json:components`
+}
+
+type Components struct {
```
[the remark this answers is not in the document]
dinhxuanvu: We certainly don't have to use the `Components` term as it is something we randomly picked. We can change it to `Inputs` to differentiate it from other components.

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
+	// ServiceCIDR          string `json:"serviceCIDR"`
+	// ServiceNodePortRange string `json:"serviceNodePortRange"`
+	// DNS string `json:"dns"`
+	// Domain               string `json:"domain"`
+	// MTU string `json:"mtu"`
 }
 
 type DebugConfig struct {
-	Pprof bool `json:"pprof"`
+	// Pprof bool `json:"pprof"`
```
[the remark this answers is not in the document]
dinhxuanvu: We would like to enable it by default.

pkg/config/config.go
dinhxuanvu: Would it be acceptable to use a default value that will be sufficient for most use cases such as level 3 or 4?

pkg/config/api/config.go
```diff
@@ -0,0 +1,46 @@
...
+
+type Network struct {
+	// IP address pool to use for pod IPs.
+	// This field is immutable after installation.
+	ClusterNetwork []ClusterNetworkEntry `json:"clusterNetwork"`
+
+	// IP address pool for services.
+	// Currently, we only support a single entry here.
+	// This field is immutable after installation.
+	ServiceNetwork []string `json:"serviceNetwork"`
```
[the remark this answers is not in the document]
dinhxuanvu: This is modelled after the [Network](https://pkg.go.dev/github.com/openshift/api/config/v1#NetworkSpec) type in OpenShift config v1.

pkg/config/config.go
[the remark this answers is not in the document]
benluddy: Enabled by default would be consistent with upstream/OCP: https://github.com/kubernetes/kubernetes/blob/4516c7972d306918b8848488255d492f94d15aac/staging/src/k8s.io/apiserver/pkg/server/config.go#L348.

pkg/config/api/config.go
dinhxuanvu: We model the Network part of the config using [Network](https://pkg.go.dev/github.com/openshift/api/config/v1#NetworkSpec) type in OS config v1. That type spec does have `ExternalIP` config which we can use in the future if there is a need to support this feature in the future.
For now, we just rework the existing fields in the config into a new config that is aligned with OS config v1 API.

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
-	Domain               string `json:"domain"`
-	MTU                  string `json:"mtu"`
+	// URL string `json:"url"`
+
+	// ClusterCIDR          string `json:"clusterCIDR"`
+	// ServiceCIDR          string `json:"serviceCIDR"`
+	// ServiceNodePortRange string `json:"serviceNodePortRange"`
+	// DNS string `json:"dns"`
+	// Domain               string `json:"domain"`
+	// MTU string `json:"mtu"`
```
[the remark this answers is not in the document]
fzdarsky: @zshi-redhat @mangelajo does MTU have to be user-configurable or can we auto-detect it somehow?

pkg/config/config.go
```diff
@@ -37,37 +37,37 @@ var (
...
 type ClusterConfig struct {
-	URL string `json:"url"`
-
-	ClusterCIDR          string `json:"clusterCIDR"`
-	ServiceCIDR          string `json:"serviceCIDR"`
-	ServiceNodePortRange string `json:"serviceNodePortRange"`
-	DNS                  string `json:"dns"`
-	Domain               string `json:"domain"`
-	MTU                  string `json:"mtu"`
+	// URL string `json:"url"`
```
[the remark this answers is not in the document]
dinhxuanvu: @benluddy Any comments on this?

pkg/config/config.go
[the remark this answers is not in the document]
zshi-redhat: @fzdarsky Until the path mtu is verified to work in ovnk, we need the mtu to be configurable.

pkg/config/config.go
fzdarsky: We're starting with single node, because that's where 80% of the use cases are, but expect to need 3-node (HA) eventually. I guess we can cross that bridge when we get there, though.

Right now, the use cases I believe we need to address the following scenarios:
a) User has resource manifests selecting on node (name) on OCP that shall run unmodified on MicroShift, too. Could be handled by defaulting node name == host name (current behaviour).
b) Host changes name during runtime, either as result of DHCP lease renewal or the device owner / device agent deliberately changing name from the image defaults for better host identification after MicroShift booted for the first time.

pkg/config/config.go
fzdarsky: To start with, I think so, yes.

pkg/config/api/config.go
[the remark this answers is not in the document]
dinhxuanvu: I think the intention is "this is a group of user-configurable parameters".

pkg/config/config.go
dhellmann: Uniqueness will be important for fleet management, but do we have a separate cluster ID for that? Can we have MicroShift set the node name to a hardware device ID of some sort instead of asking users to configure it?

pkg/config/api/config.go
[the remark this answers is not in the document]
deads2k: If you change something like the servicenetwork, then you will end up with a cluster that has services persisted in it that are not in a valid network range.  Protection like this is part of the reason to unify the types.

pkg/config/config.go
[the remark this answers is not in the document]
deads2k: @benluddy @dgrisonnet MTU is stored on NetworkStatus.  You sure you don't want the entire config objects as opposed to just spec?

pkg/config/config.go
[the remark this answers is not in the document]
deads2k: The overhead is neglible and the endpoint is protected by the standard kube-apiserver authn/authz

pkg/config/config.go
deads2k: Keeping an on-disk layout consistent with OCP reduces our documentation, training, and tooling load.  It's not configurable on OCP, let's use the same locations.

log rotation is handled in-process by the kube-apiserver.

When the config.openshift.io types are brought in, there's an option for audit logging levels.  Disabling audit logging is accompanied with a warning that, "if something fails, it's quite likely we won't be able to find it without audit logs"

pkg/config/config.go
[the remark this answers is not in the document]
mangelajo: Sometimes PMTU won't work because something on a remote network is broken and out of the customer's control. In this case, they need to be able to limit MTU locally to workaround issues.

pkg/config/config.go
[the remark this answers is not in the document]
dgrisonnet: > but is instead used to specify the IP and port this instance's API server binds to.

I don't see why we would want that to be configurable, the address the API server binds to should always be the same.

The url of the API server will also be present in the kubeconfig so other instances can use the kubeconfig to connect to the server.

pkg/config/config.go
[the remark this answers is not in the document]
dgrisonnet: Using the NetworkStatus that is supposed to be read-only in OCP to specify a configuration in Microshift is not ideal and kind of confusing, I'd rather diverge from the existing config spec and do conversion later on than bringing the status.

pkg/config/config.go
dgrisonnet: We need to have a top-level log-level parameter for convenience purposes. In OCP we usually have those defined at the operator level, but in the case of MicroShift since we don't have any operator, I think such an option should be handled at the MicroShift level directly.

> Would it be acceptable to use a default value that will be sufficient for most use cases such as level 3 or 4?

I don't think so unless you are debugging a verbosity level of 0 is enough for the users. Also in OCP the default is 0, so I think we should stick to that.

With regards to the actual name of this option as well as the values it can take, I think we should reuse the existing mechanism we have for OCP operators to keep it consistent with the existing products: https://github.com/openshift/api/blob/622889ac07cf3dd8787679af0d3176bf292b3185/operator/v1/types.go#L55-L62

pkg/config/api/config.go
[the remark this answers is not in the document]
dhellmann: We have 2 reasons we can't say these fields are immutable:
* We can't keep people from editing the configuration file. 
* MicroShift hosts are portable and may transit between networks, so we're going to eventually need to protect against the case where the host lands on a network that conflicts with these internal networks.

That use case isn't a priority right now, but we can't ignore it.

pkg/config/api/config.go
dhellmann: @zshi-redhat or @mangelajo do either of you have an opinion about @atiratree's question about ExternalIPNetworkCIDR?

pkg/config/config.go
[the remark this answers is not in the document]
dhellmann: We can't assume that network addresses stay the same. MicroShift hosts can be autonomous devices that move between networks.

Hosts may have multiple NICs and multiple IPs.

Given those constraints, how do we ensure that the correct API service IP is used?

pkg/config/config.go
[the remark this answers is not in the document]
dhellmann: I agree with @dgrisonnet. MTU is an example of a way that MicroShift needs to be configurable that (right now) OCP does not.

pkg/config/config.go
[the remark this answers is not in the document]
dhellmann: What's the benefit of having profiling on by default? Can we quantify the overhead?

pkg/config/config.go
dhellmann: The log level is going to need to be configurable in MicroShift.

pkg/config/config.go
dhellmann: If we agree we can use a hardware ID, we'll want to leave the field in place for now and file a ticket to do the work to change it later.

dhellmann (commented):
I did not originally realize that this proposal was tied to moving the schema definition out of openshift/microshift and into openshift/api. We cannot do that, for several reasons including the fact that we intend to take this project back upstream and we don't want a community project bound to reviews in the OpenShift repo. So, let's focus on organizing the schema to align with OCP where it makes sense, and then bring the implementation of those changes into this repository.

deads2k:
> I did not originally realize that this proposal was tied to moving the schema definition out of openshift/microshift and into openshift/api. We cannot do that, for several reasons including the fact that we intend to take this project back upstream and we don't want a community project bound to reviews in the OpenShift repo.

I do not agree.  openshift/api is where the openshift API contracts that ship on an OCP cadence are managed and reviewed for consistency and maintainability over time.  Simply wishing the project wasn't subject to an external API review for maintainability isn't a compelling reason to avoid it.  Even in this review we've seen cases where the current configuration for microshift doesn't actually reflect the reality of configuration changes that can be made over time.  This should be discussed with staff-eng, I'll add it to the agenda.  API reviews are not a barrier to getting things done.  They provide consistency and longevity for project direction over time.

deads2k:
> We have 2 reasons we can't say these fields are immutable:
We can't keep people from editing the configuration file.
MicroShift hosts are portable and may transit between networks, so we're going to eventually need to protect against the case where the host lands on a network that conflicts with these internal networks.
That use case isn't a priority right now, but we can't ignore it.

I don't think you understand. That's not a choice you get to make.  Changing the service network after it has been established causes existing services to exist in the wrong network range.  This in turn creates a situation the cluster doesn't function when you start it again.  You can pretend that its possible to change it, but it's simply not a thing you can do and maintain a functional cluster.

pkg/config/config.go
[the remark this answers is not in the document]
dgrisonnet: Then we need dns if the ips are dynamic and we can't trust them. Adding a route to reach out to the API server doesn't sound too bad if we need to support scenarios where hosts have multiple NICs and IPs and they still need an immutable way to reach out to the API server.

Also, can you confirm that these constraints are coming from the initial customers of microshift? We are trying to come up with a minimal version of the config to avoid having options configurable that we don't want to support in the future but can't remove because some of the initial customers may already depend on them.

pkg/config/config.go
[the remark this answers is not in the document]
derekwaynecarr: I agree with @deads2k that its best to leave this on.  Simplifies support interactions and overhead is minimal.

pkg/config/api/config.go
zshi-redhat: @dhellmann sorry I missed this question.

Yes, ovn-kubernetes supports k8s service of ExternalIP type and it can be used today with prerequisites that the network infrastructure (outside MicroShift cluster) needs to route traffic for the external IP address to MicroShift cluster. However, we probably don't need `ExternalIPNetworkCIDRs` or `AutoAssignCIDRs` since these two API fields are exposed by OpenShift operators which are not supposed to run in regular MicroShift deployment.
So we don't have external IP pool management in MicroShift due to lack of operators support, but user can attach the selected external IP manually when creating/patching the k8s service.

dinhxuanvu:
Hey folks, thank you all for valuable inputs and feedbacks. This PR has served its purpose to kick off the conversation regarding MicroShift config changes that we wish to pursue. It is time to close it out to focus on spinoff PRs.

- For new config spec discussion, please see this PR: #1030 
- For changes to existing config, please see these PRs (and more to come): #1026 #1051

pkg/config/api/config.go
atiratree: in that case I have created a PR that disables the controller so it doesn't have to run https://github.com/openshift/microshift/pull/1058

PR #1017, opened 2022-10-13
Revert "microshift: compact ovn databases periodically"

mangelajo:
This reverts commit a81d1d39b4093cbd142db166c7f5949ff71afd00.

Closes-Issue: https://issues.redhat.com/browse/USHIFT-451

### DIFF
--- a/assets/components/ovn/master/daemonset.yaml
+++ b/assets/components/ovn/master/daemonset.yaml
@@ -95,7 +95,7 @@ spec:
             memory: 10Mi
         terminationMessagePolicy: FallbackToLogsOnError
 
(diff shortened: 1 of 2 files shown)


mangelajo:
We need to find a different way to do compaction, the readiness probe is not the right place.

zshi-redhat:
> We need to find a different way to do compaction, the readiness probe is not the right place.

I agree, thanks for the fix @mangelajo !
/lgtm

mangelajo:
> > We need to find a different way to do compaction, the readiness probe is not the right place.
> 
> I agree, thanks for the fix @mangelajo ! /lgtm

Thanks for having an eye on it zshi! :D

PR #1019, opened 2022-10-14
USHIFT-247: Add images command

fzdarsky:
Adds a "microshift images" command that prints the embedded image references. By default, prints them space-separated (e.g. for use in "podman pull"). The output format can be changed to json, yaml, or toml. The latter format can be used in ImageBuilder blueprints.

Closes [USHIFT-247](https://issues.redhat.com//browse/USHIFT-247)

### DIFF
--- a/pkg/cmd/images.go
+++ b/pkg/cmd/images.go
@@ -1,0 +1,61 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/openshift/microshift/pkg/release"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+type imagesOptions struct {
+	Output string
+}
+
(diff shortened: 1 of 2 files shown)


pkg/cmd/images.go
```diff
@@ -0,0 +1,67 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "images",
+		Short: "Print container images used by this MicroShift version",
+		Run: func(cmd *cobra.Command, args []string) {
+			var marshalled []byte
+			var err error
+
+			switch opts.Output {
+			case "":
```
dhellmann, line 31: It would be good to have an explicit "list" mode, but we could add that in a follow-up.

pkg/cmd/images.go
```diff
@@ -0,0 +1,67 @@
...
+				var buf strings.Builder
+				for _, v := range release.Image {
+					if buf.Len() > 0 {
+						fmt.Fprintf(&buf, " ")
+					}
+					fmt.Fprintf(&buf, v)
+				}
+				fmt.Fprintf(ioStreams.Out, "%s", buf.String())
```
dhellmann, line 39: Is there any reason not to print directly to the stream in this case?

fzdarsky: Good point. FIxed.

PR #1020, opened 2022-10-14
USHIFT-484: Fix template data ref

copejon:

Fixes [USHIFT-484](https://issues.redhat.com//browse/USHIFT-484)

### DIFF
--- a/assets/components/odf-lvm/topolvm-node_daemonset.yaml
+++ b/assets/components/odf-lvm/topolvm-node_daemonset.yaml
@@ -12,7 +12,7 @@ spec:
   template:
     metadata:
       annotations:
-        odf-lvm.microshift.io/lvmd_config_sha256sum: "{{ Sha256sum .Config }}"
+        odf-lvm.microshift.io/lvmd_config_sha256sum: "{{ Sha256sum .lvmd }}"
       labels:
         app: topolvm-node
       name: lvmcluster-sample
--- a/pkg/assets/bindata.go
+++ b/pkg/assets/bindata.go
@@ -1297,7 +1297,7 @@ spec:
   template:
     metadata:
       annotations:
-        odf-lvm.microshift.io/lvmd_config_sha256sum: "{{ Sha256sum .Config }}"
+        odf-lvm.microshift.io/lvmd_config_sha256sum: "{{ Sha256sum .lvmd }}"
       labels:
         app: topolvm-node
       name: lvmcluster-sample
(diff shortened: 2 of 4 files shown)


assets/components/odf-lvm/topolvm-node_daemonset.yaml
```diff
@@ -12,7 +12,7 @@ spec:
   template:
     metadata:
       annotations:
-        odf-lvm.microshift.io/lvmd_config_sha256sum: "{{ Sha256sum .Config }}"
+        odf-lvm.microshift.io/lvmd_config_sha256sum: "{{ if .lvmd }}{{ Sha256sum .lvmd }}{{ end }}" # do not generate sha's for 0 length strings
```
dhellmann, line 15: Is the comment always going to end up in the output? Is there a template language syntax for a comment that would strip it when the template is rendered?

Would this be clearer if we put the logic in the template data passed in to the render call, instead of the template? For example, an `.lvmd` field and a separate `.lvmdChecksum` field?

fzdarsky: In https://github.com/openshift/microshift/blob/main/pkg/components/render.go#L41, could you please try adding `template.New("").Funcs(templateFuncs).Option("missingkey=error")`? I suspect you're missing the key and then the correct way to write something conditionally is to use `{{if ne (index .YourStruct "your_field") ""}}`. `{{if}}` cannot deal with `nil`, but `index` can and will return an empty string if `nil`.

dhellmann:
/override ci/prow/e2e-openshift-conformance-sig-api-machinery

assets/components/odf-lvm/topolvm-node_daemonset.yaml
copejon: added, thanks!

dhellmann:
Overriding broken CI jobs

/override ci/prow/e2e-openshift-conformance-sig-cli
/override ci/prow/e2e-openshift-conformance-sig-node
/override ci/prow/e2e-openshift-conformance-sig-arch
/override ci/prow/e2e-openshift-conformance-sig-network
/override ci/prow/e2e-openshift-conformance-sig-instrumentation
/override ci/prow/e2e-openshift-conformance-sig-apps
/override ci/prow/e2e-openshift-conformance-sig-auth
/override ci/prow/e2e-openshift-conformance-sig-api-machinery
/override ci/prow/e2e-openshift-conformance-sig-storage
/override ci/prow/e2e-openshift-conformance-sig-scheduling

PR #1023, opened 2022-10-17
rebase-4.12.0-0.nightly-2022-10-15-094115

pmtk:
etcd is not (and won't be - check diff for rebase.sh) updated


go.mod
```diff
@@ -292,42 +288,42 @@ replace (
 	go.etcd.io/etcd/raft/v3 => github.com/openshift/etcd/raft/v3 v3.5.1-0.20220707134052-31b6b2d9b4d7 // from etcd
```
fzdarsky, line 288: You may want to make it explicit that we're pinning the etcd version here and why, by replacing the `// from etcd` with `// override <reason>`. Then the rebase script should leave this alone.

pmtk: Is that the case for other etcds in the replace? Can you check if change I made in `rebase.sh` is even legit? :D

fzdarsky: I don't think your change to `rebase.sh` is necessary if you used the override here in the go.mod.

pmtk:
obsoleted by https://github.com/openshift/microshift/pull/1032

PR #1025, opened 2022-10-17
USHIFT-363: Use route CRD from cluster-ingress-operator repo in rebase script

oglok:

Closes [USHIFT-363](https://issues.redhat.com//browse/USHIFT-363)


assets/crd/route.crd.yaml
```diff
@@ -69,10 +69,6 @@ spec:
               - properties:
                   path:
                     maxLength: 0
-              - properties:
```
dhellmann, line 72: Weren't these lines necessary to fix a bug in the past?

fzdarsky:
The `openshift/api` version differs between components and therefore potentially also the route CRD, e.g. currently

    cluster-ingress-operator/go.mod:	github.com/openshift/api v0.0.0-20220801162917-f4ac46b878fe
    kubernetes/go.mod:	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68

Are we certain the one from o/k gets applied and not the one from CIO?

assets/crd/route.crd.yaml
oglok: I see, you're talking about this one: https://github.com/openshift/microshift/commit/7506328bde4effc43620c80772b22548305165a8

I'll fix it.

oglok: > The `openshift/api` version differs between components and therefore potentially also the route CRD, e.g. currently
> 
> ```
> cluster-ingress-operator/go.mod:	github.com/openshift/api v0.0.0-20220801162917-f4ac46b878fe
> kubernetes/go.mod:	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68
> ```
> 
> Are we certain the one from o/k gets applied and not the one from CIO?

The canonical location of this CRD is https://github.com/openshift/api/blob/master/route/v1/route.crd.yaml and according to the Network Edge team, this is the only one maintained. They don't control the files vendored by operators (except CIO). So my suggestion would be to clone the api repo as part of the rebase procedure. Does it make sense ?

dhellmann: Yeah, we should use the version of the file they consider canonical.

oglok:
> The `openshift/api` version differs between components and therefore potentially also the route CRD, e.g. currently
> 
> ```
> cluster-ingress-operator/go.mod:	github.com/openshift/api v0.0.0-20220801162917-f4ac46b878fe
> kubernetes/go.mod:	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68
> ```
> 
> Are we certain the one from o/k gets applied and not the one from CIO?

It'd be better to use the original location of the route CRD, however, due to the approach of the rebase script (pulling from container images from the release payload) we can use the vendored file from the cluster-ingress-operator.

pkg/assets/bindata.go
```diff
@@ -0,0 +1,6928 @@
+// Code generated for package assets by go-bindata DO NOT EDIT. (@generated)
```
pmtk, line 1: Please remove this file

pmtk, line 1: Please remove this file

pmtk (changes requested):

pkg/assets/bindata.go
oglok: Done. I did the rebase and I'm not sure why this remained as part of the PR. Sorry!

PR #1026, opened 2022-10-17
USHIFT-500: Remove ConfigFile, DataDir and Manifests fields off MicroShift config

dinhxuanvu:
The three fields ConfigFile, DataDir and Manifests don't need to be user-configurable. Those values should be set to be default values at all time. As a result, those fields can be removed off microshift config to avoid supporting these user-facing configuration in the future.

Closes #<Issue Number>

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -27,6 +27,8 @@ import (
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 )
 
+var microshiftDataDir = config.GetDataDir()
+
 func initAll(cfg *config.MicroshiftConfig) error {
 	// create CA and keys
 	clusterTrustBundlePEM, certChains, err := initCerts(cfg)
@@ -42,7 +44,7 @@ func initAll(cfg *config.MicroshiftConfig) error {
 }
 
 func loadCA(cfg *config.MicroshiftConfig) error {
-	return util.LoadRootCA(filepath.Join(cfg.DataDir, "/certs/ca-bundle"), "ca-bundle.crt", "ca-bundle.key")
+	return util.LoadRootCA(filepath.Join(microshiftDataDir, "/certs/ca-bundle"), "ca-bundle.crt", "ca-bundle.key")
 }
 
 func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.CertificateChains, error) {
@@ -56,7 +58,7 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 		return nil, nil, err
 	}
 
-	certsDir := cryptomaterial.CertsDirectory(cfg.DataDir)
+	certsDir := cryptomaterial.CertsDirectory(microshiftDataDir)
 	// store root CA for all
 	//TODO generate ca bundles for each component
 	clusterTrustBundlePEM, _, err := util.StoreRootCA("https://kubernetes.svc", filepath.Join(certsDir, "/ca-bundle"),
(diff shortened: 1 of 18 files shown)


dinhxuanvu:
@dgrisonnet @benluddy PTAL

pkg/cmd/init.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 )
 
+var DataDir = config.FindDataDir()
```
benluddy, line 30: I think this should be written in a way that guarantees everything in the process uses the same directory. By exporting FindDataDir, which can return different values from call to call, there's the risk of future issues with multiple data directories.

dinhxuanvu: I can construct a struct with directory-related configuration that will be generated at the beginning and pass it along as a parameter to those functions.

fzdarsky: If we want to move to a single, fixed location for the data dir (`/var/lib/microshift`), then it means this simply reduces to an accessor function to a global-scope constant, or? What I don't understand is why we want to do this if we wire through the instance config (`cfg`) anyway and often need that info in the same functions?

pkg/config/config.go
```diff
@@ -181,46 +173,29 @@ func StringInList(s string, list []string) bool {
...
 
-func (c *MicroshiftConfig) ReadFromConfigFile() error {
-	if len(c.ConfigFile) == 0 {
-		return nil
-	}
-
-	contents, err := os.ReadFile(c.ConfigFile)
+// Note: add a configDir parameter here because of unit test requiring custom
+// local directory
+func (c *MicroshiftConfig) ReadFromConfigFile(configDir string) error {
```
dhellmann, line 178: Is the argument the full path to the file or just the directory name? It looks like it's used as a filename on the next line, but it's named like it's a directory.

ggiguash:
/retitle USHIFT-500: Remove ConfigFile, DataDir and Manifests fields off MicroShift config

benluddy:
/approve

/assign @fzdarsky

This looks in line with the discussion on Vu's original PR. Are there any outstanding concerns about this particular change?

dinhxuanvu:
@dhellmann please take a look when you have the time. Thanks

pkg/cmd/run.go
```diff
@@ -67,8 +63,7 @@ func NewRunMicroshiftCommand() *cobra.Command {
 }
 
 func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
-
-	if err := cfg.ReadAndValidate(flags); err != nil {
+	if err := cfg.ReadAndValidate("", flags); err != nil {
```
dhellmann, line 66: I think instead of having ReadAndValidatate() be responsible for setting a default, this higher level function should do it. We can fix that in a follow-up PR, though.

dhellmann (commented):
I think we can point to a few cleanups, but let's get this in now and iterate on those separately.

/lgtm

dhellmann:
I realized after approving that the sample file (https://github.com/openshift/microshift/blob/main/packaging/microshift/config.yaml) also mentions some of these options. That should be a quick follow-up PR, too.

benluddy:
@dhellmann @dinhxuanvu I took the liberty of opening https://issues.redhat.com/browse/USHIFT-528 to keep track of the requested follow-up changes. Please add anything I missed.

pkg/config/config.go
```diff
@@ -179,46 +188,29 @@ func StringInList(s string, list []string) bool {
 	return false
 }
 
-func (c *MicroshiftConfig) ReadFromConfigFile() error {
-	if len(c.ConfigFile) == 0 {
-		return nil
-	}
```
benluddy, line 185: ^

PR #1028, opened 2022-10-18
USHIFT-487: microshift-etcd binary

pmtk:
- Added `github.com/openshift/microshift/etcd` module based on original EtcdService that is a standalone binary `microshift-etcd` running etcd server configured with MicroShift config files
- `make` and `make all` build both `microshift` and `microshift-etcd`
- `make etcd` produces only `microshift-etcd`
- `microshift-etcd` and `microshift` are packaged in the MicroShift RPM and deployed in the same directory
- `microshift-etcd` has 2 commands: `run` and `version`


ggiguash:
This is a 4.13 change, so putting this on hold until the 4.12 branch is created.
/hold

pmtk:
Tomorrow I want to verify couple thing, but it should be pretty close to ready for review
FYI it might be easier to review via https://github.dev/openshift/microshift/pull/1028
/cc @dusk125 @dhellmann @pacevedom

etcd/cmd/microshift-etcd/main.go
```diff
@@ -0,0 +1,29 @@
...
+	"k8s.io/component-base/logs"
+)
+
+func main() {
+	logs.InitLogs()
+	defer logs.FlushLogs()
+
+	cmd := &cobra.Command{
+		Use:   "microshift-etcd",
+		Short: "MicroShift, a minimal OpenShift",
```
dhellmann, line 18: nit: We probably want a different description string?

etcd/cmd/microshift-etcd/run.go
```diff
@@ -0,0 +1,117 @@
...
+	etcdSignerCertPath := cryptomaterial.CACertPath(cryptomaterial.EtcdSignerDir(certsDir))
+	dataDir := filepath.Join(microshiftDataDir, s.Name())
+
+	// based on https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml#L19
+	s.etcdCfg = etcd.NewConfig()
+	s.etcdCfg.ClusterState = "new"
+	//s.etcdCfg.ForceNewCluster = true //TODO
+	s.etcdCfg.Logger = "zap"
+	s.etcdCfg.Dir = dataDir
+	s.etcdCfg.APUrls = setURL([]string{cfg.NodeIP}, ":2380")
```
dhellmann, line 71: Does this need to be the public node IP? Could it be localhost, for example, like the other endpoints?

etcd/cmd/microshift-etcd/run.go
```diff
@@ -0,0 +1,117 @@
...
+
+	e.Server.Stop()
+	<-e.Server.StopNotify()
+	return nil
+}
+
+func setURL(hostnames []string, port string) []url.URL {
+	urls := make([]url.URL, len(hostnames))
+	for i, name := range hostnames {
+		u, err := url.Parse("https://" + name + port)
```
dhellmann, line 110: Assembling a URL like this doesn't work well if the "name" portion is an IPv6 address. Will we always get a name here?

Do we need a colon between the name and the port?

dhellmann (commented):
I have a couple of nits that we can deal with separately. This looks good to me.

etcd/cmd/microshift-etcd/run.go
pmtk: Looks like `127.0.0.1` works

etcd/cmd/microshift-etcd/run.go
pmtk: Used [net.JoinHostPort](https://pkg.go.dev/net#JoinHostPort) which is IPv6 aware

etcd/cmd/microshift-etcd/main.go
pmtk: it's not user facing app so I removed it altogether

dhellmann:
I'm happy with this version. I'll leave the final sign-off for @dusk125

dusk125:
lgtm!

dusk125:
Wasn't sure if I have reviewer, but I'll go ahead and try
/lgtm

dhellmann:
This also closes ETCD-332

PR #1030, opened 2022-10-19
USHIFT-607: Introduce a new config format for microshift config

dinhxuanvu:
Separate component-specific config and microshift-specific config into 2 sections within top-level config. The components config is modelled after OpenShift config v1 API while microshift config contains fields that are specific to microshift usage.

Closes #<Issue Number>

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -10,48 +10,53 @@ The MicroShift configuration file must be located at `~/.microshift/config.yaml`
 The format of the `config.yaml` configuration file is as follows.
 
 ```yaml
-cluster:
-  clusterCIDR: ""
-  serviceCIDR: ""
-  serviceNodePortRange: ""
-  domain: ""
-  url: ""
-nodeIP: ""
-nodeName: ""
 subjectAltNames:
   - ""
-logVLevel: ""
+nodeName: ""
+nodeIP: ""
+url: ""
+clusterDomain: ""
+network:
+  clusterNetwork:
+    - cidr: ""
+  serviceNetwork:
+    - ""
+  serviceNodePortRange: ""
+debugging:
+  logLevel: ""
 ```
 
 The configuration settings alongside with the supported command line arguments and environment variables are presented below.
 
-| Field Name          | CLI Argument              | Environment Variable                    | Description |
-|---------------------|---------------------------|-----------------------------------------|-------------|
-| clusterCIDR         | --cluster-cidr            | MICROSHIFT_CLUSTER_CLUSTERCIDR          | A block of IP addresses from which Pod IP addresses are allocated
-| serviceCIDR         | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
-| serviceNodePortRange| --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort (WARNING: see the dedicated section)
-| domain              | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
-| url                 | --url                     | MICROSHIFT_CLUSTER_URL                  | URL of the API server for the cluster.
-| nodeIP              | --node-ip                 | MICROSHIFT_NODEIP                       | The IP address of the node, defaults to IP of the default route
-| nodeName            | --node-name               | MICROSHIFT_NODENAME                     | The name of the node, defaults to hostname
+| Field Name            | CLI Argument              | Environment Variable                    | Description |
+|-----------------------|---------------------------|-----------------------------------------|-------------|
+| cidr (clusterNetwork) | --cluster-cidr            | MICROSHIFT_CLUSTER_CLUSTERCIDR          | A block of IP addresses from which Pod IP addresses are allocated
+| serviceNetwork        | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
+| serviceNodePortRange  | --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort
+| clusterDomain         | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
+| url                   | --url                     | MICROSHIFT_CLUSTER_URL                  | URL of the API server for the cluster.
+| nodeIP                | --node-ip                 | MICROSHIFT_NODEIP                       | The IP address of the node, defaults to IP of the default route
+| nodeName              | --node-name               | MICROSHIFT_NODENAME                     | The name of the node, defaults to hostname
+| logLevel              | --v                       | MICROSHIFT_LOGVLEVEL                    | Log verbosity (Normal, Debug, Trace, TraceAll)
 | subjectAltNames     | --subject-alt-names       | MICROSHIFT_SUBJECTALTNAMES              | Subject Alternative Names for apiserver certificates
-| logVLevel           | --v                       | MICROSHIFT_LOGVLEVEL                    | Log verbosity (0-5)
 
 ## Default Settings
 
 In case `config.yaml` is not provided, the following default settings will be used.
 
 ```yaml
-cluster:
-  clusterCIDR: 10.42.0.0/16
-  serviceCIDR: 10.43.0.0/16
-  serviceNodePortRange: 30000-32767
-  domain: cluster.local
-  url: https://127.0.0.1:6443
-nodeIP: ""
 nodeName: ""
-subjectAltNames: nil
-logVLevel: 0
+nodeIP: ""
+url: https://127.0.0.1:6443
+clusterDomain: cluster.local
+network:
+  clusterNetwork:
+    - cidr: 10.42.0.0/16
+  serviceNetwork:
+    - 10.43.0.0/16
+  serviceNodePortRange: 30000-32767
+debugging:
+  logLevel: "Normal"
 ```
 
 ## Service NodePort range
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -70,6 +70,48 @@ type MicroshiftConfig struct {
 	Ingress IngressConfig `json:"-"`
 }
 
+// Top level config
+type Config struct {
+	NodeName        string    `json:"nodeName"`
+	NodeIP          string    `json:"nodeIP"`
+	URL             string    `json:"url"`
+	ClusterDomain   string    `json:"clusterDomain"`
+	Network         Network   `json:"network"`
+	Debugging       Debugging `json:"debugging"`
+	SubjectAltNames []string  `json:"subjectAltNames"`
+}
+
+type Network struct {
+	// IP address pool to use for pod IPs.
+	// This field is immutable after installation.
+	ClusterNetwork []ClusterNetworkEntry `json:"clusterNetwork,omitempty"`
+
+	// IP address pool for services.
+	// Currently, we only support a single entry here.
+	// This field is immutable after installation.
+	ServiceNetwork []string `json:"serviceNetwork,omitempty"`
+
+	// The port range allowed for Services of type NodePort.
+	// If not specified, the default of 30000-32767 will be used.
+	// Such Services without a NodePort specified will have one
+	// automatically allocated from this range.
+	// This parameter can be updated after the cluster is
+	// installed.
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])-([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`
+	ServiceNodePortRange string `json:"serviceNodePortRange,omitempty"`
+}
+
+type ClusterNetworkEntry struct {
+	// The complete block for pod IPs.
+	CIDR string `json:"cidr,omitempty"`
+}
+
+type Debugging struct {
+	// Valid values are: "Normal", "Debug", "Trace", "TraceAll".
+	// Defaults to "Normal".
+	LogLevel string `json:"logLevel"`
+}
+
 func GetConfigFile() string {
 	return configFile
 }
@@ -219,9 +261,36 @@ func (c *MicroshiftConfig) ReadFromConfigFile(configFile string) error {
 	if err != nil {
 		return fmt.Errorf("reading config file %q: %v", configFile, err)
 	}
+	var config Config
+	if err := yaml.Unmarshal(contents, &config); err != nil {
+		return fmt.Errorf("decoding config file %s: %v", configFile, err)
+	}
 
-	if err := yaml.Unmarshal(contents, c); err != nil {
-		return fmt.Errorf("decoding config file %q: %v", configFile, err)
+	// Wire new Config type to existing MicroshiftConfig
+	c.LogVLevel = config.GetVerbosity()
+	if config.NodeName != "" {
+		c.NodeName = config.NodeName
+	}
+	if config.NodeIP != "" {
+		c.NodeIP = config.NodeIP
+	}
+	if config.URL != "" {
+		c.Cluster.URL = config.URL
+	}
+	if len(config.Network.ClusterNetwork) != 0 {
+		c.Cluster.ClusterCIDR = config.Network.ClusterNetwork[0].CIDR
+	}
+	if len(config.Network.ServiceNetwork) != 0 {
+		c.Cluster.ServiceCIDR = config.Network.ServiceNetwork[0]
+	}
+	if config.Network.ServiceNodePortRange != "" {
+		c.Cluster.ServiceNodePortRange = config.Network.ServiceNodePortRange
+	}
+	if config.ClusterDomain != "" {
+		c.Cluster.Domain = config.ClusterDomain
+	}
+	if len(config.SubjectAltNames) > 0 {
+		c.SubjectAltNames = config.SubjectAltNames
 	}
 
 	return nil
@@ -382,4 +451,22 @@ func HideUnsupportedFlags(flags *pflag.FlagSet) {
 	})
 
 	flags.MarkHidden("version")
+}
+
+// GetVerbosity returns the numerical value for LogLevel which is an enum
+func (c *Config) GetVerbosity() int {
+	var verbosity int
+	switch c.Debugging.LogLevel {
+	case "Normal":
+		verbosity = 2
+	case "Debug":
+		verbosity = 4
+	case "Trace":
+		verbosity = 6
+	case "TraceAll":
+		verbosity = 8
+	default:
+		verbosity = 2
+	}
+	return verbosity
 }
--- a/test/config.yaml
+++ b/test/config.yaml
@@ -1,12 +1,15 @@
 ---
-logVLevel: 4
 subjectAltNames:
   - '1.2.3.40'
 nodeName: node1
 nodeIP: '1.2.3.4'
-cluster:
-  url: https://127.0.0.1:6443
-  clusterCIDR: '10.20.30.40/16'
-  serviceCIDR: '40.30.20.10/16'
-  domain: cluster.local
+url: https://127.0.0.1:6443
+clusterDomain: 'cluster.local'
+network:
+  clusterNetwork:
+    - cidr: '10.20.30.40/16'
+  serviceNetwork:
+    - '40.30.20.10/16'
   serviceNodePortRange: 30000-32767
+debugging:
+  logLevel: 'Debug'


pkg/config/config.go
```diff
@@ -37,38 +37,79 @@ var (
...
-	ClusterCIDR          string `json:"clusterCIDR"`
-	ServiceCIDR          string `json:"serviceCIDR"`
-	ServiceNodePortRange string `json:"serviceNodePortRange"`
-	DNS                  string `json:"dns"`
-	Domain               string `json:"domain"`
-	MTU                  string `json:"mtu"`
+// Top level config
+type Config struct {
+	Components Components       `json:components`
+	Microshift MicroshiftConfig `json:microshift`
```
dhellmann, line 43: It feels odd to imply that only a subset of the configuration file for MicroShift is for configuring MicroShift.

Maybe "logging" is a component? Then we could have something like

```
logging:
  level: 3
```

Can the MTU value go into the networking block?

And then can we move the "components" to each be top level entries? That would retain the nice organization of the per-component structs without introducing an extra level of nesting.

dinhxuanvu:
@benluddy @dgrisonnet @fzdarsky Please take a look.
Just a note here that this PR only contains the config API spec/definition and there is not wiring yet. I will do the wiring after we agree upon the spec.

pkg/config/config.go
fzdarsky: @dinhxuanvu based on our discussion yesterday, I believe we said that MicroShift-specific params should not be gathered into a single region, but be "overlaid" onto the OpenShift config region it belongs to conceptually if that exists (e.g. "MTU" into networking, "bindAddress" into apiServer) and into the top level (or a new grouping) if not (e.g. log-level).

pkg/config/config.go
```diff
@@ -37,38 +37,81 @@ var (
...
+	// If not specified, the default of 30000-32767 will be used.
+	// Such Services without a NodePort specified will have one
+	// automatically allocated from this range.
+	// This parameter can be updated after the cluster is
+	// installed.
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])-([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`
+	ServiceNodePortRange string `json:"serviceNodePortRange,omitempty"`
+
+	// Note: additional field to config v1 Network type
+	MTU string `json:"mtu"`
```
deads2k, line 73: let's model it like

```go
type Network struct{
  ClusterNetwork []ClusterNetworkEntry`json:"clusterNetwork"`
  ServiceNetwork []string`json:"serviceNetwork"`

  MicroshiftNetwork `json:",inline"`
}

type MicroshiftNetwork struct{
  MTU string `json:"mtu"`
}
```

doing this will keep the region segregated and make the additional fields clearly opt-in at the code level while keeping the serialization looking like

```yaml
network:
  serviceNetwork: valueOne
  mtu: valueHere
```

pkg/config/config.go
deads2k: > into the top level (or a new grouping) if not (e.g. log-level).

This is what I expected for logLevel.  I'm flexible for the containing type change, though `logging` seems to suffer from "what else would ever go into that region".  `microshift` doesn't offend my sensibilities, but it looks like a component to me.  `debugging` would be more general `logging`.

pkg/config/config.go
```diff
@@ -37,38 +37,81 @@ var (
...
+// Top level config
+type Config struct {
+	Components Components       `json:components`
+	Microshift MicroshiftConfig `json:microshift`
 }
 
-type DebugConfig struct {
-	Pprof bool `json:"pprof"`
+// Components config
+type Components struct {
```
fzdarsky, line 47: Did we agree on a more suitable name than "Components" (if what we end up with contains this in the serialisation at all)? For example, the "Core" that @deads2k suggested?

pkg/config/config.go
zshi-redhat: Agree that the `mtu` field shall be moved out from the network main spec to align with openshift [network.config](https://github.com/openshift/api/blob/master/config/v1/types_network.go#L34) spec.
This `mtu` value is correlated to CNI implementation. For ovn-kubernetes, it is the tunnel interface mtu, instead of the uplink mtu. It would be good to move it to the CNI specific struct, such as OVNKubernetesConfig, just like how it is defined in the [network operator API](https://github.com/openshift/api/blob/master/operator/v1/types_network.go#L421). This puts mtu in the context of ovn-kubernetes CNI and should help to reduce the confusion.

Perhaps the `MicroshiftNetwork` can be renamed to `OVNKubernetesNetwork` for our current CNI choice. Other CNI network struct (if there will be) can be simply appended to the `Network` struct.

dhellmann: @zshi-redhat makes a good point. If the MTU value is only used by CNI, we may want that in a different config file entirely. We have all of the TopoLVM config in a separate file, for example, and that's going to make it much easier to decouple that work so the community can swap in different CSI implementations. We will want to be able to do the same for CNI.

pkg/config/config.go
```diff
@@ -37,38 +37,81 @@ var (
...
+	BaseDomain string `json:"baseDomain"`
+}
 
-	NodeName string `json:"nodeName"`
-	NodeIP   string `json:"nodeIP"`
+type APIServer struct {
+	// servingCert is the TLS cert info for serving secure traffic. If not specified, operator managed certificates
+	// will be used for serving secure traffic.
+	ServingCerts APIServerServingCerts `json:"servingCerts"`
+}
```
dgrisonnet, line 96: I don't think this is required for the first version of the config

pkg/config/config.go
dgrisonnet: Thank you all for stepping up to discuss, it took a direction we didn't expect, but it sounds better than the original proposal.

My question is should we really decouple that from the microshift configuration? Having everything in one place would make it easier for the users IMO. We could have dedicated entries in the config for `CNIPlugins` and `CSIPlugins`, with a similar layout as what we have today in the operator configs: https://github.com/openshift/api/blob/4226c2167e40c62b09e2db083d1c0051fafba7da/operator/v1/types_network.go#L196-L212

pkg/config/config.go
dinhxuanvu: `Core` sounds reasonable to me.

pkg/config/config.go
```diff
@@ -37,38 +37,81 @@ var (
...
+	// names is a optional list of explicit DNS names (leading wildcards allowed) that should use this certificate to
+	// serve secure traffic. If no names are provided, the implicit names will be extracted from the certificates.
+	// Exact names trump over wildcard names. Explicit names defined here trump over extracted implicit names.
+	Names []string `json:"names,omitempty"`
+}
 
-	Debug DebugConfig `json:"debug"`
+// Microshift specific config
+type MicroshiftConfig struct {
+	LogLevel int `json:"logLevel"`
```
dinhxuanvu, line 114: The current `LogLevel` is accepting an `int`. However, @dgrisonnet has a suggestion to model this LogLevel after [Operator v1 API](https://github.com/openshift/api/blob/622889ac07cf3dd8787679af0d3176bf292b3185/operator/v1/types.go#L55-L62) and then we will convert it the back to `int` verbosity. That way it will align better with OCP and users don't need why those numbers mean instead of using a plain English terminology such as "Normal", "Debug", "Trace", "TraceAll". What do you think? @dhellmann @deads2k @fzdarsky @benluddy

pkg/config/config.go
dinhxuanvu: The `debugging` struct sounds reasonable IMO. So I will construct it as:

```
type Debugging struct {
	LogLevel int `json:"logLevel"`
}
```

pkg/config/config.go
dinhxuanvu: David's suggestion sounds good to me. Will make the changes for the struct changes.

pkg/config/config.go
fzdarsky: In general I like plain English, but then it becomes unclear how these map to the Kubernetes [logging level conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md), or?

dinhxuanvu: Oh sorry. Forgot to include the convention code that is currently used in OCP: https://github.com/openshift/cluster-kube-apiserver-operator/blob/0378867a15c8f34b8286e1c92243a0a6bbdda383/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L492-L504

pkg/config/config.go
```diff
@@ -37,38 +37,81 @@ var (
...
-	DataDir    string `json:"dataDir"`
+type Network struct {
+	// IP address pool to use for pod IPs.
+	// This field is immutable after installation.
+	ClusterNetwork []ClusterNetworkEntry `json:"clusterNetwork"`
+
+	// IP address pool for services.
+	// Currently, we only support a single entry here.
+	// This field is immutable after installation.
+	ServiceNetwork []string `json:"serviceNetwork"`
```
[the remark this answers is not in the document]
dinhxuanvu: @zshi-redhat Please take a look

pkg/config/config.go
dinhxuanvu: @zshi-redhat Please take a look

pkg/config/config.go
```diff
@@ -37,38 +37,81 @@ var (
...
+	APIServer APIServer `json:apiServer`
 }
 
-type MicroshiftConfig struct {
-	ConfigFile string `json:"configFile"`
-	DataDir    string `json:"dataDir"`
+type Network struct {
+	// IP address pool to use for pod IPs.
+	// This field is immutable after installation.
+	ClusterNetwork []ClusterNetworkEntry `json:"clusterNetwork"`
```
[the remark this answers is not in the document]
dhellmann: I'm not a networking expert. The main concern we've had with hard-coding subnet ranges is that hosts that might be mobile, and therefore might connect to different networks at different times. Does the subnet you propose allow for that?

Do we have to worry about IPv6?

pkg/config/config.go
dhellmann: Using an enum with conversion makes sense to me.

pkg/config/config.go
```diff
@@ -37,38 +37,63 @@ var (
...
-
-	ClusterCIDR          string `json:"clusterCIDR"`
-	ServiceCIDR          string `json:"serviceCIDR"`
-	ServiceNodePortRange string `json:"serviceNodePortRange"`
-	DNS                  string `json:"dns"`
-	Domain               string `json:"domain"`
-	MTU                  string `json:"mtu"`
+// Top level config
+type Config struct {
+	Core Core `json:core`
```
dhellmann, line 42: We're now down to 1 member here. Why not move the members of Core here and eliminate a level of nesting?

pkg/config/config.go
```diff
@@ -37,38 +37,63 @@ var (
...
+	// The port range allowed for Services of type NodePort.
+	// If not specified, the default of 30000-32767 will be used.
+	// Such Services without a NodePort specified will have one
+	// automatically allocated from this range.
+	// This parameter can be updated after the cluster is
+	// installed.
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])-([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`
+	ServiceNodePortRange string `json:"serviceNodePortRange,omitempty"`
+
+	MicroshiftNetwork MicroshiftNetwork `json:",inline"`
```
dhellmann, line 71: In the other PR we talked about moving MTU to a separate config file. I think that would be more consistent with our plans to allow different CNI plugins to be swapped in by the community. @zshi-redhat are you OK with that?

dhellmann (commented):
We've had a few conversations about these changes in different places, so it would be useful to capture  the reasoning in an enhancement. We could continue to use this PR to hammer out the final structure, and then document it retroactively.

pkg/config/config.go
```diff
@@ -37,38 +37,63 @@ var (
...
-	URL string `json:"url"`
-
-	ClusterCIDR          string `json:"clusterCIDR"`
-	ServiceCIDR          string `json:"serviceCIDR"`
-	ServiceNodePortRange string `json:"serviceNodePortRange"`
-	DNS                  string `json:"dns"`
-	Domain               string `json:"domain"`
-	MTU                  string `json:"mtu"`
+// Top level config
+type Config struct {
```
dhellmann, line 41: Should we add a metadata section so we can be explicit about versioning, too? What's our standard practice with other files like this?

pkg/config/config.go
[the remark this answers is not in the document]
zshi-redhat: @dhellmann made a good point.
A corner case: cluster admin knows the clusterNetwork range might overlap with mobile subnet, so would like to change the clusterNetwork subnet to another range.
Other than that, I don't see major issues of fixed clusterNetwork subnet.

pkg/config/config.go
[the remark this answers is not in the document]
zshi-redhat: I agree with @dgrisonnet that addressing space shall not be a problem for single node deployment. We may have the same corner case that the fixed serviceNetwork conflicts with node subnet as mentioned in the clusterNetwork comment.

pkg/config/config.go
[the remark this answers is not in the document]
zshi-redhat: IPv6 is another good point, since we allow dual-stack and single-stack configuration, user may not always want fixed dual-stack.

pkg/config/config.go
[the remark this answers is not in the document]
zshi-redhat: Same consideration for dual-stack & single-stack. serviceNetwork may contain one entry for single-stack or multiple entries for dual-stack.

pkg/config/config.go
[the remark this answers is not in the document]
zshi-redhat: An example of dual-stack config with openshift:
```
networking:
  clusterNetworks:
  - cidr: 10.128.0.0/14
    hostPrefix: 23
  - cidr: fd01::/48
    hostPrefix: 64
  machineNetwork:
  - cidr: 10.73.116.0/23
  - cidr: 2620:52:0:4974::/64
  networkType: OVNKubernetes
  serviceNetwork:
  - 172.30.0.0/16
  - fd02::/112
 ```

File: pkg/config/config.go (hunk 9 above)
Comment (dinhxuanvu) [reply pkg/config/config.go:71]: Based on the suggestion here, this [PR 956](https://github.com/openshift/microshift/pull/956/) should be able to contain the MTU information.

File: pkg/config/config.go (hunk 10 above)
Comment (benluddy) [reply pkg/config/config.go:41]: Can we put off that decision until we find ourselves needing to make a breaking change?

File: pkg/config/config.go (hunk 8 above)
Comment (benluddy) [reply pkg/config/config.go:42]: I'll offer a preference, but not a substantial argument: Leaving as-is retains the option of backwards-compatibly adding a new top-level field that is unlike a core configuration domain. For example, if we were to introduce fields that describe the configuration itself (i.e. an embedded TypeMeta), at the config root, they would be interspersed with fields that configure the cluster. I like the flexibility, and imagine users accustomed to managing Kubernetes manifests won't be bothered by nesting that's comparable to the "spec" convention.

File: pkg/config/config.go [hunk 11]
```diff
@@ -37,37 +37,64 @@ var (
...
+	Core Core `json:core`
 }
 
-type DebugConfig struct {
-	Pprof bool `json:"pprof"`
+// Core config
+type Core struct {
+	Network   Network   `json:network`
+	DNS       DNS       `json:dns`
+	Debugging Debugging `json:debugging`
```
benluddy, line 49: At the risk of entering bikeshed territory... is this a cross-cutting concern? On OCP, flags for things like log verbosity live within the operator/v1 group because they relate to the operation of the cluster components and not their functional behavior. When we discussed configuration a couple weeks ago, I stuck log level outside of core in the doc example because I thought there was consensus.

pkg/config/config.go
dhellmann: We could. Would we get any benefit from declaring this a beta or even alpha version?

pkg/config/config.go
dhellmann: Good point, I always forget that the version field is on its own, not in the metadata block.

pkg/config/config.go
dhellmann: I think that adding support for the separate configuration file in another PR makes sense. Whether that's #956 or something else is up to @zshi-redhat

pkg/config/config.go
dhellmann: I'd be happy with a log level option at the top level if that's what the team agrees on.

pkg/config/config.go
zshi-redhat: @dinhxuanvu @dhellmann I will update #965 to include mtu in ovnk CNI config file.

zshi-redhat: I updated the CNI PR (#956) to include the mtu config change.

dinhxuanvu:
Hey folks, I want to provide some updates on the PR:

- Several PRs have been merged to remove several fields (or they are no longer user-configurable) from the existing config as they were deemed unnecessary or can be auto-configured.
- The remaining configurable fields are being mapped into structs that are based on existing OpenShift configure v1 (under `Core` struct)
- There are two fields, `NodeIP` and `NodeName` that we are trying to find a solution to have them auto-configured but haven't found a consensus yet. As a result, they will stay in the new config schema under `NodeConfig`.
- The new `Config` API is a user-facing API that is for config file AKA the json schema structure that config file must follow in order to be decoded properly. Internally, we keep the original `MicroshiftConfig` struct which is no longer used for config file. Data in new `Config` struct is mapped/wired into `MicroShift` and passing along to other components. This is done to avoid code change and keep this PR small and easy to review. In the future, we can do a complete rewiring and getting rid of `MicroshiftConfig` altogether if there is a desire to do so.

Would love to hear more feedbacks on the PR regarding the new config structure and hopefully we can move this PR forward.
Thanks.

fzdarsky:
Once we've agreed on the config format, we'll need to add updates to the default config file (`packaging/microshift/config.yaml`) and documentation (`docs/howto_config.md`) to this PR, too.

test/config.yaml
```diff
@@ -1,10 +1,17 @@
...
-  serviceNodePortRange: 30000-32767
+node:
+  nodeName: node1
+  nodeIP: '1.2.3.4'
+core:
+  apiServer:
+    bindHost: https://1.2.3.4:6443
+  network:
+    clusterNetwork:
+      - cidr: '10.20.30.40/16'
```
fzdarsky, line 10: Having a `cidr` field here but not in `serviceNetwork` is one obvious example to me of where we sacrifice UX / intuitiveness of the config API for the sake of alignment.

fzdarsky:
Thanks a lot for all the work incorporating feedback, @dinhxuanvu, this already looks quite decent now. 

> * The remaining configurable fields are being mapped into structs that are based on existing OpenShift configure v1 (under `Core` struct)

This `Core` struct is the last bit that I find a bit off. I understand it's to separate OpenShift user config from non-OpenShift user config, but with our agreement on the need to potentially add MicroShift-specific extensions to those structs, I no longer see a technical reason to separate the OpenShift config bits out. From a UX perspective, users wouldn't care about the difference and the selection of what is in `Core` and what not would seem arbitrary to them.

Shall we remove the `Core` struct altogether and move its sub-structs to `Config`? (Would also address the comment to move `Debug` to top-level.)

Alternatively, if we want more structure, I believe the original structure that separated between config specific to the _instance_ of MicroShift (e.g. debug) from that which applies to the _cluster_ as a whole (network, control plane components) was more meaningful.

test/config.yaml
dinhxuanvu: I agreed and it is unfortunate that there is a bit of a difference on how `serviceNetwork` and `clusterNetwork` are constructed on config API side. At this point, it is essentially a choice to make. Do we want to strictly follow the config v1 struct or can we make a slight modification here? @benluddy @deads2k

dinhxuanvu:
> Thanks a lot for all the work incorporating feedback, @dinhxuanvu, this already looks quite decent now.
> 
> > * The remaining configurable fields are being mapped into structs that are based on existing OpenShift configure v1 (under `Core` struct)
> 
> This `Core` struct is the last bit that I find a bit off. I understand it's to separate OpenShift user config from non-OpenShift user config, but with our agreement on the need to potentially add MicroShift-specific extensions to those structs, I no longer see a technical reason to separate the OpenShift config bits out. From a UX perspective, users wouldn't care about the difference and the selection of what is in `Core` and what not would seem arbitrary to them.
> 
> Shall we remove the `Core` struct altogether and move its sub-structs to `Config`? (Would also address the comment to move `Debug` to top-level.)

I think I will be happy to flat this structure out a bit by removing the `Core` layer given we don't have to be too rigid on separating things out when there is already a layer of config v1 struct that should maintain reasonable separation between different components.

> 
> Alternatively, if we want more structure, I believe the original structure that separated between config specific to the _instance_ of MicroShift (e.g. debug, node) from that which applies to the _cluster_ as a whole and therefore needs to match across all nodes (network, control plane components) was more meaningful.

From all of conversations so far and particularly your point earlier about "users wouldn't care about the difference", I feel ultimately a config should be simple and easy to construct. Trying to separate things out too much for the shake of how engineers feel about keep things compartmentalized is rather counterintuitive. As I said earlier, I now favor keeping things a bit flat if possible.
cc @benluddy

test/config.yaml
fzdarsky: We have agreed to align where possible, so let's just keep this as is.

dinhxuanvu: Sounds fair.

dinhxuanvu:
Pushed changes to flat out the config struct as @fzdarsky requested.

pkg/config/config.go
```diff
@@ -189,9 +245,33 @@ func (c *MicroshiftConfig) ReadFromConfigFile(configFile string) error {
...
 		return fmt.Errorf("reading config file %q: %v", configFile, err)
 	}
+	var config Config
+	if err := yaml.Unmarshal(contents, config); err != nil {
+		return fmt.Errorf("decoding config file %s: %v", configFile, err)
+	}
 
-	if err := yaml.Unmarshal(contents, c); err != nil {
-		return fmt.Errorf("decoding config file %q: %v", configFile, err)
+	// Wire new Config type to existing MicroshiftConfig
```
dhellmann, line 253: Is there a plan to use the new Config type directly in the future?

pkg/config/config.go
```diff
@@ -65,6 +65,62 @@ type MicroshiftConfig struct {
...
+	NodeName  string    `json:nodeName`
+	NodeIP    string    `json:nodeIP`
+	Network   Network   `json:network`
+	DNS       DNS       `json:dns`
+	Debugging Debugging `json:debugging`
+	ApiServer ApiServer `json:apiServer`
+}
+
+type ApiServer struct {
+	BindHost string `json:bindHost`
```
dhellmann, line 79: The example for BindHost shows this as a URL, but the variable implies to me that I would set an IP or hostname.

dhellmann (commented):
I'm happy with this version. I do have a couple of questions about clarity, but I don't think they're worth holding this up. I'll leave the PR open since several other reviewers had feedback on earlier versions.

pkg/config/config.go
dinhxuanvu: Yes. The idea is to have a subsequent PR to remove MicroShift completely and only use Config moving forward. For now, the MicroShift remains to minimize the code change.

pkg/config/config.go
dinhxuanvu: This variable name is proposed in the design doc that we discussed in the past. I'm happy to change it to something else more clear if we can agree on a new name. @benluddy

dgrisonnet: Looking into it again, it seems that we are only using this option to set the cluster URL in the various kubeconfig files generated by MicroShift on intialization: https://github.com/openshift/microshift/blob/main/pkg/cmd/init.go#L364-L438

As such it feels a bit weird to have this option under apiserver when it is used globally to configure many components. Wouldn't it make more sense to keep this one as a top level option and rename it to something like `ServerURL` or `ClusterURL`?

cc @benluddy

docs/howto_config.md
```diff
@@ -10,44 +10,54 @@ The MicroShift configuration file must be located at `~/.microshift/config.yaml`
...
+| serviceNetwork        | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
+| serviceNodePortRange  | --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort
+| baseDomain            | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
+| bindHost              | --url                     | MICROSHIFT_CLUSTER_URL                  | URL of the API server for the cluster.
+| nodeIP                | --node-ip                 | MICROSHIFT_NODEIP                       | The IP address of the node, defaults to IP of the default route
+| nodeName              | --node-name               | MICROSHIFT_NODENAME                     | The name of the node, defaults to hostname
+| logLevel              | --v                       | MICROSHIFT_LOGVLEVEL                    | Log verbosity (Normal, Debug, Trace, TraceAll)
```
dgrisonnet, line 40: These changes should be reflected in https://github.com/openshift/microshift/blob/main/pkg/cmd/run.go#L30-L40

test/config.yaml
```diff
@@ -1,10 +1,15 @@
...
+network:
+  clusterNetwork:
+    - cidr: '10.20.30.40/16'
+  serviceNetwork:
+    - '40.30.20.10/16'
   serviceNodePortRange: 30000-32767
+dns:
+  baseDomain: 'cluster.local'
+debugging:
+  logLevel: 'Normal'
```
dgrisonnet, line 15: For the sake of keeping the same value as before and to be consistent with the other tests in https://github.com/openshift/microshift/blob/main/pkg/config/config_test.go, it would be better if we set the logLevel to Debug here.

pkg/config/config.go
dgrisonnet: I synced with Ben as I forgot some context.

`BindHost` is the replacement for the original `URL` option which is used to set the server URL in the various kubeconfig files. However, we also noticed that this option was used to configure the port of the bind address of kube-apiserver: 
- https://github.com/openshift/microshift/blob/main/pkg/controllers/kube-apiserver.go#L108
- https://github.com/openshift/microshift/blob/main/pkg/config/config.go#L151-L170

The idea today is to change a bit the UX by having the kube-apiserver bind address explicitly configurable and generate the cluster URL for the kubeconfig files from it.

This wasn't the original way this PR was handling this option so we would have to update it. Also it might be worth renaming the option to `BindAddress` which is a bit clearer and follow what we have in OCP (https://github.com/openshift/api/blob/94ac871f4a586d03b0ae11344f3604016065a458/config/v1/types.go#L48).

Let me know if that sounds right to you @dhellmann.

dinhxuanvu:
@dgrisonnet Pushed a change to do the rebase and addressed one comment. The rest is at your mercy ;). There is an unfortunate addition of a field `subjectAtlNames` in the config which was merged recently. I do not take any actions on that addition yet and keep it as it is. FYI. Thanks.

pkg/config/config.go
dgrisonnet: Discussed that briefly with @dhellmann offline, we will need an enhancement to change the existing behavior, so we will do that change in another PR.

mangelajo:
/approve

The new format seems reasonable.

docs/howto_config.md
```diff
@@ -10,48 +10,55 @@ The MicroShift configuration file must be located at `~/.microshift/config.yaml`
...
+nodeIP: ""
+url: ""
+network:
+  clusterNetwork:
+    - cidr: ""
+  serviceNetwork:
+    - ""
+  serviceNodePortRange: ""
+dns:
+  baseDomain: ""
```
mangelajo, line 25:
```suggestion
  baseDomain: ""
  clusterDomain: ""
```

baseDomain is used for public DNS/router records, while clusterDomain is for the service network domains, what typically is cluster.local

docs/howto_config.md
```diff
@@ -10,48 +10,55 @@ The MicroShift configuration file must be located at `~/.microshift/config.yaml`
...
+nodeIP: ""
+url: https://127.0.0.1:6443
+network:
+  clusterNetwork:
+    - cidr: 10.42.0.0/16
+  serviceNetwork:
+    - 10.43.0.0/16
+  serviceNodePortRange: 30000-32767
+dns:
+  baseDomain: cluster.local
```
mangelajo, line 59:
```suggestion
  baseDomain: cluster.local
```

This does not align with the explanation in https://github.com/openshift/microshift/pull/1030/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R113

Probably I see now that it's just that description that needs to be changed?

mangelajo:
I see some inconsistency with baseDomain/domain/clusterDomain

I would suggest we change that parameter to clusterDomain, and we update the description. So this would fully align with the name used in the openshift DNS operator: https://github.com/openshift/cluster-dns-operator/blob/master/pkg/operator/controller/controller_dns_configmap.go#L80

pkg/config/config.go
```diff
@@ -70,6 +70,59 @@ type MicroshiftConfig struct {
...
+
+type DNS struct {
+	// baseDomain is the base domain of the cluster. All managed DNS records will
+	// be sub-domains of this base.
+	//
+	// For example, given the base domain `openshift.example.com`, an API server
+	// DNS record may be created for `cluster-api.openshift.example.com`.
+	//
+	// Once set, this field cannot be changed.
+	BaseDomain string `json:"baseDomain"`
```
mangelajo, line 117:
```suggestion
	// baseDomain is the base domain of the cluster. All managed DNS records will
	// be sub-domains of this base.
	//
	// For example, given the base domain `cluster.local`, an service would be
	// discoverable inside the service network as  `service.namespace.cluster.local`.
	//
	// Once set, this field cannot be changed.
	ClusterDomain string `json:"clusterDomain"`
```

pkg/config/config.go
```diff
@@ -219,9 +272,36 @@ func (c *MicroshiftConfig) ReadFromConfigFile(configFile string) error {
...
+	}
+	if len(config.Network.ServiceNetwork) != 0 {
+		c.Cluster.ServiceCIDR = config.Network.ServiceNetwork[0]
+	}
+	if config.Network.ServiceNodePortRange != "" {
+		c.Cluster.ServiceNodePortRange = config.Network.ServiceNodePortRange
+	}
+	if config.DNS.BaseDomain != "" {
+		c.Cluster.Domain = config.DNS.BaseDomain
+	}
```
mangelajo, line 302:
```suggestion
	if config.DNS.ClusterDomain != "" {
		c.Cluster.Domain = config.DNS.ClusterDomain
	}
```

This change is quite indicative of what I was explaining above.

test/config.yaml
```diff
@@ -1,12 +1,16 @@
...
-  domain: cluster.local
+url: https://127.0.0.1:6443
+network:
+  clusterNetwork:
+    - cidr: '10.20.30.40/16'
+  serviceNetwork:
+    - '40.30.20.10/16'
   serviceNodePortRange: 30000-32767
+dns:
+  baseDomain: 'cluster.local'
```
mangelajo, line 14:
```suggestion
  clusterDomain: 'cluster.local'
```

mangelajo:
For me it would work to merge and iterate.

docs/howto_config.md
```diff
@@ -10,48 +10,55 @@ The MicroShift configuration file must be located at `~/.microshift/config.yaml`
...
 nodeName: ""
-subjectAltNames: nil
-logVLevel: 0
+nodeIP: ""
+url: https://127.0.0.1:6443
+network:
+  clusterNetwork:
+    - cidr: 10.42.0.0/16
+  serviceNetwork:
+    - 10.43.0.0/16
```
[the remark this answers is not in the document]
fzdarsky: @mangelajo that's because it's defined this way in OpenShift APIs (see https://github.com/openshift/api/blob/master/config/v1/types_network.go#L89-L98) and we're aligning with that.

docs/howto_config.md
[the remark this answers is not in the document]
dinhxuanvu: So we model this DNS struct based on the DNS config on openshift config v1 (https://pkg.go.dev/github.com/openshift/api/config/v1#DNSSpec). It only contains the `baseDomain` field. So if this `baseDomain` field doesn't match what the current `domain` field in the original microshift config does, then it is best to get rid of this `DNS` struct altogether and keep the `ClusterDomain` as a field at the top level just like` NodeIP`.
cc @dgrisonnet

dgrisonnet:
@mangelajo I renamed the option to `clusterDomain` and move it to the top-level following @dinhxuanvu's suggestion in https://github.com/openshift/microshift/pull/1030#discussion_r1040360267.

docs/howto_config.md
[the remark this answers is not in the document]
mangelajo: Thanks for looking at this @dinhxuanvu

docs/howto_config.md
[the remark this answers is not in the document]
mangelajo: oh, right. We are just inheriting it.

mangelajo:
The test/config.yaml needs an update, I guess it's simply not used yet. We can follow up on that as we explained

/approve

dgrisonnet:
I updated test/config.yaml.

PR #1032, opened 2022-10-19
rebase-4.12.0-0.nightly-2022-10-18-192348 with pinned etcd

pmtk:
etcd is pinned so it's opentelemetry dependency doesn't conflict with k8s'


pmtk:
obsoleted by https://github.com/openshift/microshift/pull/1055

PR #1035, opened 2022-10-20
ec4 images

pmtk:

PR #1036, opened 2022-10-20
ec4 images

dhellmann:
obsoletes #1035


mangelajo:
it seems to work (so far...) with those images

mangelajo:
We don't need it in the end.

PR #1038, opened 2022-10-21
Some minor fixes for configure scripts and add arm64 image mapping

praveenkumar:


scripts/image-builder/configure.sh
```diff
@@ -1,13 +1,19 @@
...
-sudo dnf install -y git osbuild-composer composer-cli \
-    cockpit-composer bash-completion podman genisoimage \
-    createrepo syslinux yum-utils selinux-policy-devel jq wget lorax rpm-build
+pkg_list="git osbuild-composer composer-cli \
+	cockpit-composer bash-completion podman genisoimage \
+	createrepo yum-utils selinux-policy-devel jq wget lorax rpm-build"
+
+if [ $(uname -p) != "aarch64" ]; then
+    pkg_list="${pkg_list} syslinux"
+fi
```
ggiguash, line 10: We should be able to simply remove syslinux package from the list on all platforms. It'll be pulled in by lorax when necessary.

scripts/image-builder/configure.sh
```diff
@@ -1,13 +1,19 @@
...
+
+sudo dnf install -y ${pkg_list}
 sudo systemctl enable osbuild-composer.socket --now
 sudo systemctl enable cockpit.socket --now
-sudo firewall-cmd -q --add-service=cockpit
-sudo firewall-cmd -q --add-service=cockpit --permanent
+sudo firewall-cmd --add-service=cockpit
+sudo firewall-cmd --add-service=cockpit --permanent
```
ggiguash, line 16: Is there no -q option on aarch64? Why remove it?

ggiguash: Also, let's remove the duplicate command on line 15.

ggiguash (changes requested):

scripts/image-builder/configure.sh
praveenkumar: @ggiguash looks like you already removed it.

PR #1040, opened 2022-10-21
Fix go version substitution in configure-vm.sh

pmtk:


ggiguash:
/retitle USHIFT-494: Fix go version substitution in configure-vm.sh

PR #1041, opened 2022-10-21
USHIFT-384: ovsdb compaction

pmtk:


assets/components/ovn/master/daemonset.yaml
```diff
@@ -120,32 +120,26 @@ spec:
           trap quit TERM INT
 
           bracketify() { case "$1" in *:*) echo "[$1]" ;; *) echo "$1" ;; esac }
+          compact() {
+            while true; do
+              /usr/bin/ovn-appctl -t /var/run/ovn/ovn${1}_db.ctl --timeout=5 ovsdb-server/compact 2>/dev/null || true
+              sleep 600
```
mangelajo, line 126: I was going to suggest inverting sleep and ovn-appctl , to give the process some breathing room at startup, but that would come at the cost of harder testing, now that is triggered right away, and if there is an error we will find it If we exchange the sleep and appctl errors could be hidden.

assets/components/ovn/master/daemonset.yaml
```diff
@@ -233,7 +227,7 @@ spec:
         - /bin/bash
         - -c
         - |
-          set -xm
+          set -xem
```
mangelajo, line 230: +1

assets/components/ovn/master/daemonset.yaml
```diff
@@ -120,32 +120,26 @@ spec:
...
-
-          if [[ ! -e ${ovn_db_file} ]]; then
-            initialize="true"
-          fi
+          exec /usr/share/ovn/scripts/ovn-ctl \
+            ${OVN_ARGS} \
+            --ovn-nb-log="-vconsole:${OVN_LOG_LEVEL} -vfile:off -vPATTERN:console:%D{%Y-%m-%dT%H:%M:%S.###Z}|%05N|%c%T|%p|%m" \
+            run_nb_ovsdb &
 
-          if [[ "${initialize}" == "true" ]]; then
```
mangelajo, line 136: I see this is a cleanup, since both paths of the if do the same.

mangelajo (commented):
looks good to me, have we tested the pod stop? I wonder if the while loop will exit on the signal to stop the process, probably yes.

assets/components/ovn/master/daemonset.yaml
zshi-redhat: +1

assets/components/ovn/master/daemonset.yaml
zshi-redhat: yes.

assets/components/ovn/master/daemonset.yaml
```diff
@@ -120,32 +120,26 @@ spec:
...
-                wait $!
-          else
-            exec /usr/share/ovn/scripts/ovn-ctl ${OVN_ARGS} \
-              --ovn-nb-log="-vconsole:${OVN_LOG_LEVEL} -vfile:off -vPATTERN:console:%D{%Y-%m-%dT%H:%M:%S.###Z}|%05N|%c%T|%p|%m" \
-              run_nb_ovsdb &
-
-              wait $!
-          fi
+          db_pid=$!
+          compact $db &
```
zshi-redhat, line 141: I guess there is no need to wait for the compact process, wdyt?

zshi-redhat (commented):
Once this PR is merged, the same change needs to be updated in CNO repo as well.

assets/components/ovn/master/daemonset.yaml
pmtk: Maybe a middle ground: `sleep 15` before while loop? Or other adequate value

assets/components/ovn/master/daemonset.yaml
pmtk: It's not waiting, just running it in the background. So to answer Miguel's question:
> I wonder if the while loop will exit on the signal to stop the process, probably yes.

I'd say that it'll exit just fine because we're not waiting on it, just running it in the background and when db exits (wait $db), this background subprocess will go down as well

pmtk:
/hold

conformance tests are not executed when manifest was changed but bindata no longer exists

assets/components/ovn/master/daemonset.yaml
mangelajo: makes sense

PR #1045, opened 2022-10-23
USHIFT-492: Document getting-started procedure from microshift-testing copr repo

ggiguash:
Closes [USHIFT-492](https://issues.redhat.com//browse/USHIFT-492)

### DIFF
--- a/docs/config/microshift-starter.ks
+++ b/docs/config/microshift-starter.ks
@@ -58,12 +58,11 @@ rm -f selinux-policy*.rpm
 dnf copr enable -y @redhat-et/microshift-testing
 dnf install -y microshift
 
-# Enable MicroShift service to be started after reboot
-systemctl enable microshift.service
+# MicroShift service should be enabled later after setting up CRI-O with the pull secret
 
 # Configure firewalld
-sudo firewall-cmd --permanent --zone=trusted --add-source=10.42.0.0/16 
-sudo firewall-cmd --permanent --zone=trusted --add-source=169.254.169.1
+firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16
+firewall-offline-cmd --zone=trusted --add-source=169.254.169.1
 
 # Install the oc and kubectl utilities
 curl -LO -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz
--- a/docs/config/microshift-starter.ks
+++ b/docs/config/microshift-starter.ks
@@ -1,0 +1,73 @@
+lang en_US.UTF-8
+keyboard us
+timezone UTC
+text
+reboot
+
+# Configure network to use DHCP and activate on boot
+network --bootproto=dhcp --device=link --activate --onboot=on --hostname=microshift-starter.local --noipv6
+
+# Partition disk with a 1GB boot XFS partition and a 8GB LVM volume containing system root
+# The remainder of the volume will be used by the CSI driver for storing data
+zerombr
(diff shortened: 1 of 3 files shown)


docs/config/microshift-starter.ks
```diff
@@ -0,0 +1,73 @@
...
+# Allow the default user to run sudo commands without password
+echo -e 'redhat\tALL=(ALL)\tNOPASSWD: ALL' > /etc/sudoers.d/redhat
+
+# Update selinux-policy packages from CentOS 8 Stream
+CENTOS8BASE=http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages
+curl -LO -s $CENTOS8BASE/selinux-policy-3.14.3-96.el8.noarch.rpm
+curl -LO -s $CENTOS8BASE/selinux-policy-devel-3.14.3-96.el8.noarch.rpm
+curl -LO -s $CENTOS8BASE/selinux-policy-targeted-3.14.3-96.el8.noarch.rpm
+dnf localinstall -y selinux-policy*.rpm
+rm -f selinux-policy*.rpm
```
ggiguash, line 55: This is necessary because MicroShift was built against the patched version of RHEL 8.6. 
The RHEL 8.6 base DVD does not work for MicroShift unless security-policy packages are updated.

The alternative is to require users to register their RHEL 8.6 systems, which complicates the flow.

docs/config/microshift-starter.ks
```diff
@@ -0,0 +1,72 @@
+lang en_US.UTF-8
+keyboard us
+timezone UTC
+text
+reboot
+
+# Configure network to use DHCP and activate on boot
+network --bootproto=dhcp --device=link --activate --onboot=on --hostname=microshift-starter.local --noipv6
```
mangelajo, line 8: @ggiguash why do we disable IPv6, do we have any issue related to that?

docs/config/microshift-starter.ks
```diff
@@ -0,0 +1,72 @@
...
+dnf install -y microshift
+
+# MicroShift service should be enabled later after setting up CRI-O with the pull secret
+
+# Configure firewalld
+firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16
+firewall-offline-cmd --zone=trusted --add-source=169.254.169.1
+
+# Install the oc and kubectl utilities
+curl -LO -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz
```
mangelajo, line 68: We should use https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp-dev-preview/4.12.0-ec.4/openshift-client-linux-4.12.0-ec.4.tar.gz instead, which handles the small differences in the pod security API.

docs/config/microshift-starter.ks
ggiguash: @mangelajo - MicroShift is not using/supporting it at the moment. I've been doing it in the devenv. Just to be on the safe side.

docs/config/microshift-starter.ks
ggiguash: Right, but that would require more testing. I can make this change and retest when the latest version of MicroShift is rebuilt at copr.

docs/config/microshift-starter.ks
mangelajo: We haven't tested the IPv6 features, but having ipv6 on the host would not hurt, anyway, better to be on the safe side I guess 👍

PR #1051, opened 2022-10-25
API-1449: Enable kube-apiserver audit logging.

benluddy:
Audit logs from kube-apiserver are used in virtually every support engagement with the API team, so they're useful to enable out-of-the-box. I've also removed the audit log path config field, instead using the same path used by OpenShift (under /var/log).

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -37,7 +37,6 @@ func addRunFlags(cmd *cobra.Command, cfg *config.MicroshiftConfig) {
 	cmd.MarkFlagFilename("config", "yaml", "yml")
 	// All other flags will be read after reading both config file and env vars.
 	flags.String("data-dir", cfg.DataDir, "The directory for storing runtime data.")
-	flags.String("audit-log-dir", cfg.AuditLogDir, "The directory for storing audit logs.")
 	flags.StringSlice("roles", cfg.Roles, "The roles of this MicroShift instance.")
 	flags.String("node-name", cfg.NodeName, "The hostname of the node.")
 	flags.String("node-ip", cfg.NodeIP, "The IP address of the node.")
@@ -93,7 +92,6 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 	}
 
 	os.MkdirAll(cfg.DataDir, 0700)
-	os.MkdirAll(cfg.AuditLogDir, 0700)
 
 	// TODO: change to only initialize what is strictly necessary for the selected role(s)
 	if _, err := os.Stat(filepath.Join(cfg.DataDir, "certs")); errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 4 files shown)


benluddy:
/assign @fzdarsky

Frank, any objection to this change? KAS rotates and truncates its own audit logs, so we should not expect to see problems with disk space exhaustion.

pkg/cmd/run.go
```diff
@@ -37,7 +37,6 @@ func addRunFlags(cmd *cobra.Command, cfg *config.MicroshiftConfig) {
 	cmd.MarkFlagFilename("config", "yaml", "yml")
 	// All other flags will be read after reading both config file and env vars.
 	flags.String("data-dir", cfg.DataDir, "The directory for storing runtime data.")
-	flags.String("audit-log-dir", cfg.AuditLogDir, "The directory for storing audit logs.")
```
dhellmann, line 40: There's a sample configuration file in https://github.com/openshift/microshift/blob/main/packaging/microshift/config.yaml that has this config value, too. We can clean that up in a separate PR.

dhellmann (commented):
I see how this removes the configuration option. I don't see how it ensures that the audit logs are written out. Is that a default behavior that doesn't need to be enabled? Is it because the output directory is now being created? Or am I missing something in reading the changes?

pkg/cmd/run.go
benluddy: Good catch. Opened https://github.com/openshift/microshift/pull/1056.

benluddy:
@dhellmann Specifying an audit log path in itself enables audit logging (https://github.com/benluddy/microshift/blob/f233b7f9b38eec3c97ad8a2686340b929de9d82e/vendor/k8s.io/apiserver/pkg/server/options/audit.go#L501-L503). I've also manually verified that, with this change, audit logs are written to /var/log/kube-apiserver/audit.log.

pkg/controllers/kube-apiserver.go
```diff
@@ -113,7 +113,6 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 	overrides := &kubecontrolplanev1.KubeAPIServerConfig{
 		APIServerArguments: map[string]kubecontrolplanev1.Arguments{
 			"advertise-address": {cfg.NodeIP},
-			"audit-log-path":    {cfg.AuditLogDir},
```
dhellmann, line 116: It looks like this removes the path setting, but in https://github.com/openshift/microshift/pull/1051#issuecomment-1292300737 you say specifying the path is what turns on audit logging. 

Are we relying on some defaulting behavior elsewhere?

I believe you that this works, I just don't understand how. Sorry to be obtuse.

benluddy: > Are we relying on some defaulting behavior elsewhere?

MicroShift configures its kube-apiserver by merging overrides on top of the base OCP defaults, so this PR is simply removing the override. The value comes from https://github.com/benluddy/microshift/blob/f233b7f9b38eec3c97ad8a2686340b929de9d82e/assets/components/kube-apiserver/defaultconfig.yaml#L45.

dhellmann: Aha! Thank you.

PR #1053, opened 2022-10-26
USHIFT-506: Remove pkg/assets/bindata.go

pmtk:


stlaz (changes requested):
you'll want to fix unit tests which still seem to contain references to the assets.

`pkg/components/render_test.go`, for example

pkg/components/render_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"text/template"
 
+	embedded "github.com/openshift/microshift/assets"
```
stlaz, line 12: I would rather see:
1. remove the bindata.go
2. rename the package in `microshift/assets/embed.go` back to `assets` and change this wherever needed
3. profit :slightly_smiling_face:

stlaz (changes requested):

pkg/components/render_test.go
pmtk: Problem is we already have `pkg/assets` package which imports `assets/`

stlaz: I see... I think I'll try to fix this afterwards, lets' stick with what you've got at this moment, then :slightly_smiling_face:

PR #1058, opened 2022-10-27
disable ingress-ip controller as we are not allowing configuration of IngressIPNetworkCIDR

atiratree:
external IPs on services should be assigned manually and not automatically in microshift

followup to https://github.com/openshift/microshift/pull/1011#discussion_r1004232794 conversation


atiratree:
@dhellmann @zshi-redhat

dhellmann:
So we don't need this controller because we are going to require users to set IPs on services themselves? I assume that only applies in the case where they want an IP other than the default one for the cluster?

atiratree:
> So we don't need this controller because we are going to require users to set IPs on services themselves? I assume that only applies in the case where they want an IP other than the default one for the cluster?

Yes, applies to LoadBalancer services that need an external ingress IP set

PR #1060, opened 2022-10-28
USHIFT-529: fix(storage): correct lvmd file path

dinhxuanvu:
The lvmd file is expected to live in the same path with config file. Previous PR mistakenly changed it to manifests data directory. It's now changed it back to the correct config file directory.

Closes #<Issue Number>


dinhxuanvu:
@benluddy @dhellmann Please take a look. Thanks
@pmtk FYI

benluddy:
/lgtm
/approve

There should be some kind of functional test that is ensuring config files are being used as expected, could you file an issue for that please?

PR #1062, opened 2022-10-28
USHIFT-532: Remove pprof flag.

benluddy:
The pprof endpoints are already exposed by default by kube-apiserver.


benluddy:
This change is based on the discussion in https://github.com/openshift/microshift/pull/1011/files#r991634061.

PR #1063, opened 2022-10-28
Remove NodeIP configurable option from microshift config

dinhxuanvu:
The NodeIP is always the host IP.

Closes #<Issue Number>

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -27,7 +27,10 @@ import (
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 )
 
-var microshiftDataDir = config.GetDataDir()
+var (
+	microshiftDataDir = config.GetDataDir()
+	nodeIP            = config.GetNodeIP()
+)
 
 func initAll(cfg *config.MicroshiftConfig) error {
 	// create CA and keys
@@ -148,7 +151,7 @@ func initCerts(cfg *config.MicroshiftConfig) ([]byte, *cryptomaterial.Certificat
 						Name:         "kubelet-server",
 						ValidityDays: cryptomaterial.ServingCertValidityDays,
 					},
-					Hostnames: []string{cfg.NodeName, cfg.NodeIP},
+					Hostnames: []string{cfg.NodeName, nodeIP},
 				},
 			),
 		),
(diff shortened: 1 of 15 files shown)


dinhxuanvu:
@benluddy PTAL

pkg/config/config.go
```diff
@@ -78,6 +78,18 @@ func GetManifestsDir() []string {
 	return manifestsDir
 }
 
+func GetNodeIP() string {
+	return nodeIP
+}
+
+func findNodeIP() string {
+	hostIP, err := util.GetHostIP()
```
dhellmann, line 86: We expect hosts to have multiple IPs. How does the user influence how this function picks the IP to use?

zshi-redhat: It seems the nodeIP option is not repected in main branch, even it is user configurable. We probably want to fix it. This influence approach shall also work for microshift-ovs-init service script, which sets up the nodeIP interface for ovn-kubernetes CNI plugin.

pkg/cmd/init.go
```diff
@@ -27,7 +27,10 @@ import (
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 )
 
-var microshiftDataDir = config.GetDataDir()
+var (
+	microshiftDataDir = config.GetDataDir()
+	nodeIP            = config.GetNodeIP()
```
dgrisonnet, line 32: instead of defining that globally every time, could we perhaps create a structure that could be passed on to the various init functions alongside the configuration?

pkg/config/config.go
dgrisonnet: In which scenario would a user want to pick a specific IP? The IP to which core components of microshift bind to shouldn't really matter to the end-user since their applications will most likely only communicate via the different services.

mangelajo:
I believe NodeIP is important to let the services know which specific IP we will be listening to if it's not the one on default route. Or if we want it to listen to 0.0.0.0 ?

Also bear in mind that IPv6 values should be accepted here.

pkg/config/config.go
mangelajo: I remember that some popular distros also supported specifying the interface name or something like IP(eth0) to make it work with DHCP too

dgrisonnet: If I am not mistaken, the nodeIP is not even configurable in OCP today. The closest thing I could find is this effort: https://github.com/openshift/enhancements/blob/ef85659d01738b9f89958d5f0da31cff05bb1182/enhancements/network/ip-interface-selection.md but this seem way too complex for MicroShift to need something like that right now.

> I remember that some popular distros also supported specifying the interface name or something like IP(eth0) to make it work with DHCP too

I am not sure why that would be beneficial to do for control plane components.

Another concern I have is what would happen if microshift was first started with a nodeIP, but then a user decided to change it on the fly. Everything that used to depend on the previous nodeIP will not be able to reach the new nodeIP if we don't have a way to propagate the update. I don't think that's supported in k8s in general since one would most likely have to delete the node to update its IP.

dhellmann: Unlike OCP hosts, we have to deal with the fact that a host running MicroShift may change its IP or move between networks. So we either need a way to pick a node IP that will never change, or we need to support it being updated. Could we always use a localhost IP or the hardware's IPv6 local address, for example? What effect would that have? What needs to use the node IP?

dinhxuanvu:
It looks like the opinion here is that we will need to keep this NodeIP option configurable given the possibility of multiple IP options and/or changes. That sounds fair. I would recommend folks to comment on #1030 PR on where to keep this option in the new config. Thanks.

PR #1064, opened 2022-10-28
USHIFT-535: Remove dns configurable option from MicroShift config

dinhxuanvu:
Cluster DNS is set to the 10th IP of Service CIDR

Closes #<Issue Number>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/openshift/microshift
 go 1.18
 
 require (
+	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e // openshift-controller-manager
 	github.com/kelseyhightower/envconfig v1.4.0 // microshift
 	github.com/miekg/dns v1.1.35 // microshift
@@ -19,7 +20,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	go.etcd.io/etcd/server/v3 v3.5.4
 	golang.org/x/sys v0.0.0-20220728004956-3c1f35247d10
-	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.25.2
 	k8s.io/apiextensions-apiserver v0.25.0
 	k8s.io/apimachinery v0.25.2
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -35,7 +35,6 @@ func addRunFlags(cmd *cobra.Command, cfg *config.MicroshiftConfig) {
 	flags.String("cluster-cidr", cfg.Cluster.ClusterCIDR, "The IP range in CIDR notation for pods in the cluster.")
 	flags.String("service-cidr", cfg.Cluster.ServiceCIDR, "The IP range in CIDR notation for services in the cluster.")
 	flags.String("service-node-port-range", cfg.Cluster.ServiceNodePortRange, "The port range to reserve for services with NodePort visibility. This must not overlap with the ephemeral port range on nodes.")
-	flags.String("cluster-dns", cfg.Cluster.DNS, "Comma-separated list of DNS server IP address. This value is used for containers DNS server in case of Pods with \"dnsPolicy=ClusterFirst\".")
 	flags.String("cluster-domain", cfg.Cluster.Domain, "Domain for this cluster.")
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -3,11 +3,13 @@ package config
 import (
 	"errors"
 	"fmt"
+	"net"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
 
+	"github.com/apparentlymart/go-cidr/cidr"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/pflag"
@@ -43,7 +45,7 @@ type ClusterConfig struct {
 	ClusterCIDR          string `json:"clusterCIDR"`
 	ServiceCIDR          string `json:"serviceCIDR"`
 	ServiceNodePortRange string `json:"serviceNodePortRange"`
-	DNS                  string `json:"dns"`
+	DNS                  string `json:"-"`
 	Domain               string `json:"domain"`
 }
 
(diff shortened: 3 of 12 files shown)


dinhxuanvu:
@benluddy PTAL

pkg/config/config.go
```diff
@@ -45,7 +47,6 @@ type ClusterConfig struct {
 	ClusterCIDR          string `json:"clusterCIDR"`
 	ServiceCIDR          string `json:"serviceCIDR"`
 	ServiceNodePortRange string `json:"serviceNodePortRange"`
-	DNS                  string `json:"dns"`
```
benluddy, line 48: Would you be OK with leaving this field here and changing its tag to `json:"-"`? Then, assigning a value to it could happen as part of ReadAndValidate, and the callers don't need to call a new method or handle possible errors.

go.mod
```diff
@@ -241,6 +241,8 @@ require (
 	github.com/stretchr/objx v0.2.0 // indirect
 )
 
+require github.com/apparentlymart/go-cidr v1.1.0
```
benluddy, line 244: I was going to object to the new dependency, but I see that this is identical to what cluster-dns-operator is doing (https://github.com/openshift/cluster-dns-operator/blob/b9c9dfe826b51c4b4c93dc0619178f67842ab385/pkg/operator/controller/controller.go#L560-L568).

Any idea why there are so many separate `require` stanzas here?

benluddy:
There are some mentions of this config field in `packaging/microshift/config.yaml` and the config docs, too.

go.mod
dhellmann: I think the require structure has to do with the rebase script. @fzdarsky am I remembering that correctly?

pkg/config/config.go
dinhxuanvu: I'm fine with that.

zshi-redhat:
Do we know why dns option was added before?

dinhxuanvu:
> Do we know why dns user-configurable option was added before?

I do not know. This was added before my time. Maybe @fzdarsky knows.

ggiguash:
/retitle USHIFT-535: Remove dns configurable option from MicroShift config

Created a JIRA issue for this work.

mangelajo:
> Do we know why dns user-configurable option was added before?

I would suspect that we just didn't know the .10 was a hard default?

go.mod
fzdarsky: That's an annoying behaviour by `go mod` since 1.17. It should only have two blocks, one for direct and one for indirect, but it randomly creates new blocks. Haven't figured out the system behind that. Anyway, the rebase script is just yet another victim of this behaviour.

fzdarsky:
> > Do we know why dns user-configurable option was added before?
> 
> I would suspect that we just didn't know the .10 was a hard default?

It's configurable on other K8s distros, so we probably just thought it should be configurable here, too.

pkg/cmd/run.go
```diff
@@ -60,6 +59,14 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
...
 	}
 
+	if cfg.Cluster.ServiceCIDR != "" {
+		clusterDNS, err := config.GetClusterDNS(cfg.Cluster.ServiceCIDR)
+		if err != nil {
+			klog.Fatalf("failed to get DNS IP: %v", err)
+		}
+		cfg.Cluster.DNS = clusterDNS
+	}
+
```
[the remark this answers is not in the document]
benluddy: Since this PR removes "DNS" from the configuration API, I disagree that we want errors generated by the computation of this IP to be treated as validation errors, or for that computation to happen as part of config reading/validation. Looking at the implementation, it seems that currently the only way this would return an error is if the Service CIDR is invalid or has fewer than 10 IPs. Those might be candidates for validation?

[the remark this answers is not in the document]
benluddy: I'd also expect an error from the DNS Service IP computation if there is no Service CIDR configured instead of ignoring it and pushing the failure later.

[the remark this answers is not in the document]
fzdarsky: I don't think we're disagreeing here. My point is that following a `ReadAndValidate()` you either have a valid configuration (whether the parameter is directly configurable by the user like Service CIDR or auto-configured based on user configurable input like Cluster DNS) or you don't. In the latter case, you need to fail fatally.

So, yes, you need to validate that Service CIDR is non-empty and in valid format plus that it's big enough to contain the Cluster IP. Whether you test that in a way that the following `net.ParseCIDR(serviceCIDR)` and `cidr.Host(service, 10)` can never fail or you let them fail and report that failure is probably a matter of taste.

pkg/components/controllers.go
```diff
@@ -201,6 +201,7 @@ func startDNSController(cfg *config.MicroshiftConfig, kubeconfigPath string) err
 		klog.Warningf("Failed to apply", "namespace", ns, "err", err)
 		return err
 	}
+
```
mangelajo, line 204:
```suggestion
```

pkg/node/kubelet.go
```diff
@@ -90,7 +90,6 @@ func (s *KubeletServer) configure(cfg *config.MicroshiftConfig) {
 func (s *KubeletServer) writeConfig(cfg *config.MicroshiftConfig) error {
 	certsDir := cryptomaterial.CertsDirectory(microshiftDataDir)
 	servingCertDir := cryptomaterial.KubeletServingCertDir(certsDir)
-
 	data := []byte(`
```
mangelajo, line 93:
```suggestion

	data := []byte(`
```

mangelajo:
some nits

pkg/config/config.go
```diff
@@ -248,9 +246,30 @@ func (c *MicroshiftConfig) ReadAndValidate(configFile string, flags *pflag.FlagS
 		return err
 	}
 
+	// validate serviceCIDR
+	if c.Cluster.ServiceCIDR != "" {
```
benluddy, line 250: Why the exception for `""`? Isn't that invalid?

pkg/config/config.go
```diff
@@ -248,9 +246,28 @@ func (c *MicroshiftConfig) ReadAndValidate(configFile string, flags *pflag.FlagS
 		return err
 	}
 
+	// validate serviceCIDR
+	_, err := GetClusterDNS(c.Cluster.ServiceCIDR)
```
dhellmann, line 250: Why not go ahead and store the value in the struct here instead of making the caller compute it again and handle the error separately? GetClusterDNS could be a private function.

dhellmann (commented):
One nit, but if other reviewers are happy we can merge this and continue to iterate.

dinhxuanvu:
@dhellmann Addressed your feedback. Looking for a lgtm and approve. Thanks.

fzdarsky (approved):

PR #1065, opened 2022-10-28
USHIFT-533: Remove role (controlplane/node) config.

benluddy:
Following earlier discussion (https://github.com/openshift/microshift/pull/1011#discussion_r991640607), let's wait to define the configuration surface until split role support arrives.


fzdarsky:
/hold cancel

Looking at this again, after removing the config option the remaining wiring probably isn't awfully relevant.

PR #1068, opened 2022-10-30
USHIFT-514: Fix rpm file name creation based on latest release tag

ggiguash:
In addition to minor make_rpm.sh script optimizations, the following names are now used for RPM files.

1. If a tag does not exist, a timestamp and "untagged" is concatenated to the name, i.e. microshift-4.12.0-2022_10_30_113709_untagged.el8.x86_64.rpm
2. If a tag exists, it is used in the name with the current commit SHA, i.e. microshift-4.10.0-2022_09_08_132255_261_gc4c9645.el8.x86_64.rpm
```
$ git describe --tags --abbrev=7 --match 4.10*
4.10.0-0.microshift-2022-09-08-132255-262-g486b833
```

Closes USHIFT-514


fzdarsky:
IIRC when we discussed on Slack the need to fix RPM versioning, we concluded we should move away from git tags completely, but we hadn’t concluded on the exact replacement scheme, e.g. what to use to replace the buildtime which isn’t very useful.

ggiguash:
> IIRC when we discussed on Slack the need to fix RPM versioning, we concluded we should move away from git tags completely, but we hadn’t concluded on the exact replacement scheme, e.g. what to use to replace the buildtime which isn’t very useful.
Can we use this implementation as a temporary fix until there is a decision. The current code eliminates the duplication and makes the tagging consistent

mangelajo (commented):
/lgtm

Even though we are not in agreement about the git tags, there are several other improvements to the scripts, and it will be a good approach until we have decided on the non-git-tag-based strategy.

PR #1070, opened 2022-10-31
Use formatting infof when logging variable

zshi-redhat:


pkg/config/lvmd/lvmd.go
```diff
@@ -59,15 +59,15 @@ func newLvmdConfigFromFile(p string) (*Lvmd, error) {
 func NewLvmdConfigFromFileOrDefault(path string) (*Lvmd, error) {
 	if _, err := os.Stat(path); err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			klog.Info("lvmd file not found, assuming default values")
+			klog.Infof("lvmd file not found, assuming default values")
```
ggiguash, line 62: Do we need Infof here as well? There is not formatting.

zshi-redhat: We don't strictly need infof here, I tool this opportunity to change it as well just because most of the klog uses infof.

PR #1074, opened 2022-11-01
WIP: add microshift inspect command

mfojtik:
This is a prototype of `microshift inspect` command which copies the `oc adm inspect` command inside microshift binary.
The first commit is just pure copy and wiring, the second commit allows to run `inspect` command without additional parameters with defaults set to microshift (admin kubeconfig, resources, etc).

The output of this command is a directory that matches the `oc must-gather` 1-1,  including events and directory structure. That allows for future analytic tools to be built and shared between microshift and OCP.

As for the next steps/GA, we need to figure out how to share the `oc adm inspect` between `oc` and `microshift`, and possibly extract that command into library-go or to shared space.
In addition, the inspect command need to collect microshift binary logs (!) and audit logs. This can be added as an optional step  into "inspect" command (a post-run-hook or something).

### DIFF
--- a/cmd/microshift/main.go
+++ b/cmd/microshift/main.go
@@ -10,6 +10,7 @@ import (
 
 	cmds "github.com/openshift/microshift/pkg/cmd"
 	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/inspect"
 )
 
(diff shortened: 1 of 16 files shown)


mangelajo:
very cool

PR #1075, opened 2022-11-01
USHIFT-539: fix(config): remove user ability to configure ingress config

dinhxuanvu:
The ingress config doesn't need to be user-configurable. The ingress config will be generated by default.

Closes #<Issue Number>


benluddy:
/cc @stlaz

Synced out of band, and this field is here for plumbing purposes, not to be part of the user-facing config API.

PR #1082, opened 2022-11-03
[WIP] Consolidate NodeIP and URL usage

dinhxuanvu:
The current host name portion of URL field is unused and they can be used
to configure node IP information which should be used to set `advertise-address`,
`etcd` config, `kubelet` config and others which should use the same IP.

For the default setting value, this host name will be localhost IP. User can
configure the hostname which will serve as node IP using config file, env var and cli.

Closes #<Issue Number>

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -17,7 +17,6 @@ cluster:
   dns: ""
   domain: ""
   url: ""
-nodeIP: ""
 nodeName: ""
 logVLevel: ""
 ```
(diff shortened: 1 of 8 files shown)


dhellmann:
I feel like we've tried a few times to remove or change this setting, and the motivation still isn't quite clear to me. So, I want to make sure I understand what this proposal is actually doing.

> The current host name portion of URL field is unused and they can be used to configure node IP information which should be used to set `advertise-address`, `etcd` config, `kubelet` config and others which should use the same IP.
> 
> For the default setting value, this host name will be localhost IP. User can configure the hostname which will serve as node IP using config file, env var and cli.

I think that what you're saying here is that nothing is currently depending on the hostname portion of the existing URL field, and so we could use that to set the NodeIP instead of having a separate field for the NodeIP. Do I understand that correctly?

Are we sure that's true? How is that URL configuration value used? Does that URL end up in any certificates, for example? Is it OK for that URL to use an IP, and not a name?

Is there ever a reason the URL might need to be different from the NodeIP? For example, although we emphasize that often MicroShift deployments will be firewalled, that will not always be the case and sometimes clients will need to access them remotely. In that case I would expect the URL to be able to include a name. What happens if the user puts a name in the field instead of an IP? Will the other uses of the NodeIP mentioned be able to cope?

dhellmann:
> Are we sure that's true? How is that URL configuration value used? Does that URL end up in any certificates, for example? Is it OK for that URL to use an IP, and not a name?

#1086 adds the IP to certificates.

dinhxuanvu:
This PR is obsolete at this point given all of changes that are already made into the config. I'm closing this PR and if we need to revisit this topic, will open a new PR. Thanks.

PR #1084, opened 2022-11-04
USHIFT-430: use HostNetwork publishing strategy for router

zshi-redhat:


scripts/auto-rebase/rebase.sh
```diff
@@ -547,6 +547,8 @@ update_manifests() {
     #    Add hostPorts for routes and metrics (needed as long as there is no load balancer)
     yq -i '.spec.template.spec.containers[0].ports[0].hostPort = 80' "${REPOROOT}"/assets/components/openshift-router/deployment.yaml
     yq -i '.spec.template.spec.containers[0].ports[1].hostPort = 443' "${REPOROOT}"/assets/components/openshift-router/deployment.yaml
+    #    Set hostNetwork to true, as we're using HostNetwork publishing strategy.
+    yq -i '.spec.template.spec.hostNetwork = true' "${REPOROOT}"/assets/components/openshift-router/deployment.yaml
```
fzdarsky, line 551: @zshi-redhat as you've pointed out in another channel, it's the cluster-ingress-operator that sets this programmatically. Would it make sense for better documentation just to move these lines up into section 2) (it's a change from rendering by the operator, not a MicroShift-specific change that we add)?

mangelajo: If we do this, we would not need the hostPorts in L548 and 549 but we must make the router listen on 80/443 directly, wouldn't that also make it a privileged container?

But we should be aware that we would also expose the router metrics over the host network. I'm not sure how good of an idea is, even though it aligns better with the ingress operator.  

@fzdarsky WDYT?

mangelajo:
/hold

Let's agree first if we really want this

scripts/auto-rebase/rebase.sh
zshi-redhat: Default hostPorts (80 & 443) are explicitly set in router deployment if omitted in openshift, including the metrics port: https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/controller.go#L452-L461

These ports are made configurable in openshift in case they are already in use. The router metrics is used for health check by external load balancer: https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L646-L659. So it has to use one of the host ports, maybe question is do we want that metrics port to be configurable?

mangelajo: it would be cool if we could setup metrics on localhost only on the router.

zshi-redhat: @mangelajo  Metrics port is used for health checking according to the api doc, if set to localhost only, it may not work for external load balancer, right?

mangelajo:
Closing this one in favor of the LoadBalancer approach.

PR #1054, opened 2022-10-26
USHIFT-507: remove router cluster-service

zshi-redhat:
hostPort exposes router service to host ports 80 & 443,
removing cluster-service which exposes router service
to random node port.

PR #1085, opened 2022-11-04
USHIFT-507: Remove service-cloud from rebase.sh

pmtk:


scripts/auto-rebase/rebase.sh
```diff
@@ -494,6 +494,8 @@ update_manifests() {
     cp "${STAGING_DIR}"/cluster-openshift-controller-manager-operator/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrolebinding.yaml "${REPOROOT}"/assets/components/openshift-router || true
     #    Restore the openshift-router's service-ca ConfigMap
     git restore "${REPOROOT}"/assets/components/openshift-router/configmap.yaml
+    # Remove service-cloud.yaml (#1054 / USHIFT-507)
```
ggiguash, line 497: It's a nit, but can you replace the references to other PRs/JIRA by a summary text if necessary?

ggiguash (changes requested):

scripts/auto-rebase/rebase.sh
pmtk: I don't quite get it. I don't see any other refs to PR/JIRA in the file (this is my invention :D I'm thinking it'll be hard to find when that file was deleted), so guess this only applies to this?
Are you suggesting the comment should be more of like: `#1054: USHIFT-507: remove router cluster-service`? Or that I should try to find tickets for other exceptions?

ggiguash: I'm not sure we should embed references to PR numbers or JIRA issues in the comments. This information can be obtained by following up on code changes, JIRA references, etc. 

In this case, I would:
- /retitle USHIFT-507: Remove service-cloud from rebase.sh
- Change the comment to "Remove router cluster-service because ...."

pmtk: got it

PR #1087, opened 2022-11-04
USHIFT-569: Add efi partition to Image Builder kickstart template

praveenkumar:
As of now generated iso fails to run on the EFI enabled system with following error.
```
Starting automated install.Saving storage configuration...
.Failed to save boot loader configuration
Failed to find a suitable stage1 device: EFI System Partition cannot be of type xfs.; EFI System Partition must be mounted on one of /boot/efi.; EFI System Partition cannot be of type lvmpv.
```

### DIFF
--- a/docs/rhel4edge_iso.md
+++ b/docs/rhel4edge_iso.md
@@ -92,6 +92,7 @@ The artifact of the build is the `_output/image-builder/microshift-installer-${V
 The `kickstart.ks` file is configured to partition the main disk using `Logical Volume Manager` (LVM). Such parititioning is required for the data volume to be utilized by the MicroShift CSI driver and it allows for flexible file system customization if the disk space runs out.
 
 By default, the following partition layout is created and formatted with the `XFS` file system:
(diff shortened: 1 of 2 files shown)


dhellmann (commented):
Is this something we want everywhere, or should we have a couple of templates for people to choose from?

ggiguash:
This section needs to be updated with the change.
https://github.com/openshift/microshift/blob/main/docs/rhel4edge_iso.md#disk-partitioning

ggiguash:
/retitle: USHIFT-569: Add efi partition to Image Builder kickstart template

ggiguash:
> Is this something we want everywhere, or should we have a couple of templates for people to choose from?

I think we would want to have one template for simplicity if it's technically possible and does not break anything on non-EFI systems. Otherwise, this needs to be conditional and set from the build.sh script.

ggiguash:
/retitle USHIFT-569: Add efi partition to Image Builder kickstart template

mangelajo (commented):
/approve

PR #1088, opened 2022-11-04
Fix pod/MicroShift systemd services shutdown order

mangelajo:
Control shutdown order by declaring the MicroShift service to start Before the kubepods.slice transient systemd unit; this makes system shutdown delay MicroShift shut down until all the pod containers are down. This is important because some services need to talk to the MicroShift API during shutdown (i.e. releasing leader election locks or cleaning up other resources) MicroShift restart or manual stop will not stop the kubepods.

Related-Issue: https://issues.redhat.com/browse/USHIFT-542

### DIFF
--- a/packaging/systemd/microshift.service
+++ b/packaging/systemd/microshift.service
@@ -3,6 +3,13 @@ Description=MicroShift
 Wants=network-online.target crio.service openvswitch.service microshift-ovs-init.service
 After=network-online.target crio.service openvswitch.service microshift-ovs-init.service
 
+# Control shutdown order by declaring this service to start Before the kubepods.slice
+# transient systemd unit; this makes system shutdown delay MicroShift shutdown until
+# all the pod containers are down. This is important because some services need to talk
+# to the MicroShift API during shutdown (i.e. releasing leader election locks or cleaning
+# up other resources) MicroShift restart or manual stop will not stop the kubepods.
+Before=kubepods.slice
+
 [Service]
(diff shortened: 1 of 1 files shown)


fzdarsky:
Two questions on this one, please:
1) Does this imply we can no longer restart the microshift.service without disrupting workloads?
2) Does this imply pods won't start until microshift.service signals it completed start-up (whereby some service may depends on pods)?

mangelajo:
> Two questions on this one, please:
> 
> 1. Does this imply we can no longer restart the microshift.service without disrupting workloads?

No, I tried to explain in both the service file, and the commit, MicroShift restart or stop does not affect workloads.

The only thing that changes is system shutdown, instead of shuting down kubepods.slice and MicroShift in parallel it will happen sequentially.

Also, it would affect startup if kubepods.slice was not a transient systemd initiated by the kubelet :)

> 2. Does this imply pods won't start until microshift.service signals it completed start-up (whereby some service may depend on pods)?

2. Theoretically yes. But pods shouldn't really be attempted to be started until MicroShift/kubelet is ready. In this way you also prevent any other more critical services from being split out of MicroShift :)

What I have seen after this change is much faster reboot times. (things don't stay hanging waiting for the API -never there anymore- to respond)

benluddy:
This looks reasonable to me from what I understand from the systemd docs. Setting a hold to allow time for other reviewers.

/hold
/lgtm

PR #1089, opened 2022-11-04
[Bug] Docs - add caveat to storage capacity tracking

copejon:

- Add documentation on storage capacity tracking to warn users of the potential for over-provisioning LVs.

Relates to [USHIFT-507](https://issues.redhat.com//browse/USHIFT-507)

### DIFF
--- a/docs/default_csi_plugin.md
+++ b/docs/default_csi_plugin.md
@@ -19,6 +19,13 @@ boots. [StorageCapacity tracking](https://kubernetes.io/docs/concepts/storage/st
 Pods with an LVMS PVC are not scheduled if the requested storage is greater than the volume group's remaining free
 storage.
 
+> **IMPORTANT!** StorageCapacity tracking poses a race condition in which the node's storage capacity is not up-to-date
+> before reconciling a PVC.  This can lead to a situtation where total provisioned LVs exceed the capacity of the node, 
+> calculated as VolumeGroupTotalCapacity - SpareGigabyte.  When this happens, the over-scheduled LVs will be reclaimed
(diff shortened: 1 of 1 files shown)


docs/default_csi_plugin.md
```diff
@@ -19,6 +19,13 @@ boots. [StorageCapacity tracking](https://kubernetes.io/docs/concepts/storage/st
 Pods with an ODF-LVM PVC are not scheduled if the requested storage is greater than the volume group's remaining free
 storage.
 
+> **IMPORTANT!** StorageCapacity tracking poses a race condition in which the node's storage capacity is not up-to-date
```
dhellmann, line 22: Is this problem unique to MicroShift, or does it apply to LVMO on SNO, too? If there is no unique MicroShift issue, we should refer the user to LVMO docs that talk about the problem.

copejon: It's a known issue with TopoLVM when deployed in this topology (without the extended scheduler).  LVMO does deploy these extended scheduler, and so doesn't encounter this.  IDK what SNO's deployment mode is for topolvm.  If it's managed via the LVM Operator, then it'll also have the extended scheduler running.  We opted not to run the scheduler on MicroShift to save space and because it's a single node deployment.

fzdarsky: The main reason we've chosen TopoLVM over HPP is so it correctly isolates PVs from each other and guarantees allocated storage. If it doesn't do that we need to consider this a bug.

ggiguash:
@copejon , please fix the linked JIRA issue and let's merge this doc fix.

PR #1094, opened 2022-11-07
USHIFT-292: Streamline the Ansible documentation for devenv and edge benchmark flows

ggiguash:
The following enhancement were implemented as part of testing:
* Streamlined the documentation flow to let users get to running the benchmark with default params and command line arguments
* Flow tested on devenv and R4E hosts
* Fixed minor problems in the roles to work properly in the devenv / R4E cases
* Added variables for some configuration values to streamline benchmark invocation 

Closes [USHIFT-292](https://issues.redhat.com//browse/USHIFT-292)

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,7 +1,7 @@
 # Ansible MicroShift
 
 The purpose of this Ansible automation is to help gain insight into the footprint and start times of MicroShift.
-At this stage, we are capturing the first start, the second start (with cached images), disk usage data from a
+At present, we are capturing the first start, the second start (with cached images), disk usage data from a
 number of points in time as well as performance data for each of the two starts.
 
 ## Prerequisites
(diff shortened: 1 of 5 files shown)


ggiguash:
/assign @sjug 
Can you review the changes? We'll wait for your approval before merging them.

ansible/README.md
```diff
@@ -33,13 +99,16 @@ There are a few variables that may need user configuration and they are found in
...
+### Inventory Configuration
 
-There are a few places to configure before execution.
+The inventory configuration file is located at `inventory/inventory`.
 
-As with any Ansible playbook, the first file to configure is the inventory located at `inventory/inventory`.
+The MicroShift host name or IP address must be configured to match the current environment. If the user account available on the MicroShift server is not `microshift`, it should be changed as well.
 
-### Inventory Configuration
+> The following Ansible command line options can be used instead of editing the `inventory/inventory` file.
```
sjug, line 108: I don't think we should move away from the inventory file as I mentioned above.

ggiguash: I'm trying to avoid editing files in the repository for the main benchmark flows. It makes everything much easier.

ansible/README.md
```diff
@@ -17,9 +17,75 @@ Run the following command on the Ansible server to install the Ansible package.
...
+If the user has a fresh RHEL host to be used for running MicroShift performance benchmarks, it is necessary to first follow the instructions in [Clean Host Example Variables](#clean-host-example-variables) for editing the `vars/all.yml` file.
+
+Run the playbook using the following commands, making sure that the values of the MicroShift server host and user variables are appropriate for your environment.
+```bash
+USHIFT_HOST=microshift-clean
+USHIFT_USER=microshift
+
+ssh-copy-id ${USHIFT_USER}@${USHIFT_HOST}
+time ansible-playbook -v \
+    -e ansible_host=${USHIFT_HOST} -e ansible_user=${USHIFT_USER} \
```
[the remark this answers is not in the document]
ggiguash: It works for me in the devenv and R4E configurations. I think it's because logging group uses local connection only, right?

ansible/README.md
```diff
@@ -1,7 +1,7 @@
 # Ansible MicroShift
 
 The purpose of this Ansible automation is to help gain insight into the footprint and start times of MicroShift.
-At this stage, we are capturing the first start, the second start (with cached images), disk usage data from a
+At present, we are capturing the first start, the second start (with cached images), disk usage data from a
```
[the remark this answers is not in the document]
ggiguash: I have not seen this file. Could it be because we're collecting this information using vnstat command, which is not installed on the host by default? 
I created USHIFT-576 to address this issue.

ansible/README.md
```diff
@@ -17,9 +17,75 @@ Run the following command on the Ansible server to install the Ansible package.
...
+```
 
-There are a few variables that may need user configuration and they are found in `vars/all.yml`.
+## Output
+
+The following text files will be created on the Ansible server running the playbook:
+- The `boot.txt` will have the time that it took the microshift service to start and for all the pods to enter the `Running` state
+- The `disk0.txt, disk1.txt, disk2.txt` files will have a snapshot of the disk usage at different stages of installation and execution.
+
+If Prometheus was enabled, the performance metrics will be uploaded to the Ansible server, where the user can view all the captured performance data using the `http://<ansible-server-ip>:9090` URL.
```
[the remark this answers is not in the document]
ggiguash: I'd suggest we add Grafana documentation consistently when we implement the dashboards.

ansible/README.md
ggiguash: If there's a problem that I'm missing, let's try to make it work.

ansible/README.md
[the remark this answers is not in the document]
sjug: Yes, it's being collected with `vnstat`.

ansible/README.md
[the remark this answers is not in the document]
sjug: The dashboards have been implemented in https://github.com/openshift/microshift/pull/1099

ansible/README.md
sjug: The problem is there's group/host specific variables that we need to set.

ansible/README.md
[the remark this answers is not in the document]
ggiguash: vnstat seems to be installed when MicroShift host is configured, so it does not fall under devenv or R4H flows. We may need to install vnstat the same way we do for Prometheus.

ansible/README.md
ggiguash: I see. I'll change the code to set group variable only.

ggiguash: I changed this to ansible_host_var and ansible_user_var - is this better now?

ansible/vars/all.yml
```diff
@@ -5,8 +5,8 @@ rhel_pool_id:
 manage_repos: false
 
 setup_microshift_host: false
-prometheus_logging: true
-install_microshift: false
+prometheus_logging: "{{ prometheus_logging_arg | default('true')  }}"
```
sjug, line 8: I think it should just be `prometheus_logging: "{{ prometheus_logging | default('true')  }}"`
Referencing the same variable.

ggiguash: I tried that before. Unfortunately, it did not work to use the same name. It breaks somewhere down the execution.

PR #1096, opened 2022-11-08
NP-633: Add gateway interface options in ovn cni config

zshi-redhat:
https://issues.redhat.com/browse/NP-633

### DIFF
--- a/docs/default_cni_plugin.md
+++ b/docs/default_cni_plugin.md
@@ -63,17 +63,22 @@ MicroShift will assume default ovn-kubernetes config values if ovn-kubernetes co
 
 The following configs are supported in ovn-kubernetes config file:
 
-|Field          |Required |Type    |Default |Description                                                       |Example|
-|:--------------|:--------|:-------|:-------|:-----------------------------------------------------------------|:------|
-|disableOVSInit |N        |bool    |false   |Skip configuring OVS bridge "br-ex" in microshift-ovs-init.service|true   |
-|mtu            |N        |uint32  |1400    |MTU value to be used for the Pods                                 |1300   |
+|Field                    |Required |Type    |Default |Description                                                       |Example|
+|:------------------------|:--------|:-------|:-------|:-----------------------------------------------------------------|:------|
+|disableOVSInit           |N        |bool    |false   |Skip configuring OVS bridge "br-ex" in microshift-ovs-init.service|true   |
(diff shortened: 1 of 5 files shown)


docs/default_cni_plugin.md
```diff
@@ -63,17 +63,22 @@ MicroShift will assume default ovn-kubernetes config values if ovn-kubernetes co
...
+|mtu                      |N        |uint32  |1400    |MTU value to be used for the Pods                                 |1300   |
 
 > When `disableOVSInit` is true, OVS bridge "br-ex" needs to be configured manually. This OVS bridge is required by ovn-kubernetes CNI. See section [OVS bridge](#ovs-bridge) for guidance on configuring the OVS gateway bridge manually.
 
 Below is an example of `ovn.yaml`:
 
 ```yaml
-disableOVSInit: true
+ovsInit:
+  disableOVSInit: true
```
mangelajo, line 79:
```suggestion
  disable: true
```
This would be better if we had a proper yaml parser, but I guess it would complicate the ovs init script and it's probably not worth it.

zshi-redhat: yep, installing `yq` is not straightforward on rhel, it may require additional pkg manager. one idea is to use json format config and jq to parse it, but it will differ from microshift yaml config.
I didn't use disable since it may cause mis-interpret with current parser which doesn't check parant key.

mangelajo: let´s leave it as disableOVSInit.

dhellmann: This name is really from the perspective of the implementation. How would you rename it from the perspective of the user, so they understand the impact without having to understand the implementation?

mangelajo (commented):
/lgtm

docs/default_cni_plugin.md
dhellmann: That might be a better name. It does at least point to the effect, rather than the implementation.

Why do we want this to be configurable at all? What would happen if we said users always need to create the bridge for the CNI system to use?

zshi-redhat: We had an issue that ovs init service randomly picks up the gateway interface based on readiness of host interfaces: https://issues.redhat.com/browse/USHIFT-490.  This provides an option to disable auto-selection of gateway interface and use the hint provided in cni config file. The default behavior is not changed - ovs init service selects the default route interface (or first available interface if default route doesn't exist).

ggiguash:
Also fixes USHIFT-490

PR #1098, opened 2022-11-08
USHIFT-540: Add openshift-user-critical PriorityClass

pmtk:


pkg/assets/scheduling.go
```diff
@@ -0,0 +1,105 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+var (
```
mangelajo, line 22: I would put everything (vars, initialization', etc. under ApplyPriorityClasses)

Otherwise, the memory for the scheme and schedulingCodecs will remain allocated for the whole life of MicroShift, while it's not necessary anymore.

If we keep the initialization/setup/etc, under ApplyPriorityClasses, that memory will be freed.

mangelajo (changes requested):
Small comment, memory saving

pkg/assets/scheduling.go
pmtk: Changed

mangelajo (commented):
/lgtm

PR #1099, opened 2022-11-08
USHIFT-448: Add prometheus datasource & dashboard to grafana

sjug:
Adds support for visualizing  Prometheus metrics that we gather from MicroShift in Grafana.

### DIFF
--- a/ansible/roles/install-logging/files/prometheus_datasource.json
+++ b/ansible/roles/install-logging/files/prometheus_datasource.json
@@ -1,0 +1,14 @@
+{
+  "name": "Prometheus",
+  "type": "prometheus",
+  "access": "proxy",
+  "url": "http://localhost:9090",
+  "basicAuth": false,
+  "withCredentials": false,
+  "isDefault": true,
+  "jsonData": {
+    "httpMethod": "POST"
+  },
+  "version": 4,
+  "readOnly": false
+}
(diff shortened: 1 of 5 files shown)


ansible/roles/install-logging/files/prometheus_datasource.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "Prometheus",
+  "type": "prometheus",
+  "access": "proxy",
+  "url": "http://localhost:9090",
```
ggiguash, line 5: The 9090 port is taken by Cockpit server, which we routinely use. Is it easy to change this and use other port, like 9999 for instance?

ggiguash (changes requested):

ansible/roles/install-logging/files/prometheus_datasource.json
ggiguash: I created USHIFT-577 to address this issue separately.

sjug: Would people be installing the logging components on the devenv VM?

ggiguash: I think so, otherwise, the benchmark collection is not complete. 
For R4E, we implemented a special option to embed the collection tool in the image.

sjug: What do you mean the benchmark collection is not complete?
The prometheus config will scrape other nodes all the same.

ggiguash: Moving the discussion to USHIFT-577 issue.

PR #1101, opened 2022-11-08
[USHIFT-573] Regenerate client certificate when subject changes

mangelajo:
The subject part of the certificate is used to authenticate the kubelet and limit the permissions on what can access. On a host name, the client certificate was not being regenerated, so the kubelet could not create a new entry for the new hostname, and blocked on a loop.

This commit forces the regeneration of the client certificate when the subject field on the disk file differs.

Closes #<Issue Number>

### DIFF
--- a/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go
+++ b/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go
@@ -20,6 +20,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


dhellmann:
@mangelajo that file is vendored from library-go. Is there a PR to fix this upstream so we don't have to carry a patch? The rebase automation is likely to wipe this change out, otherwise.

stlaz:
/hold
you can't merge a vendor change

mangelajo:
Lol, sorry 

This is what happens when you click on "follow reference" in VSCode, .. edit and push a commit before the day ends.

I'll go send a PR to that library.

mangelajo:
> The subject part of the certificate is used to authenticate the kubelet and limit the permissions on what can access. On a host name, the client certificate was not being regenerated, so the kubelet could not create a new entry for the new hostname, and blocked on a loop.
> 
> This commit forces the regeneration of the client certificate when the subject field on the disk file differs.

Work moved here

mangelajo:
https://github.com/openshift/library-go/pull/1422

PR #1102, opened 2022-11-08
WRKLDS-604: Enable the Openshift flavor of the kube-controller-manager

tkashem:

Closes #<Issue Number>

### DIFF
--- a/assets/components/kube-controller-manager/defaultconfig.yaml
+++ b/assets/components/kube-controller-manager/defaultconfig.yaml
@@ -1,0 +1,40 @@
+apiVersion: kubecontrolplane.config.openshift.io/v1
+kind: KubeControllerManagerConfig
+extendedArguments:
+  enable-dynamic-provisioning:
+  - "true"
+  allocate-node-cidrs:
+  - "false"
+  configure-cloud-routes:
+  - "false"
+  use-service-account-credentials:
+  - "true"
+  leader-elect:
+  - "true"
+  leader-elect-retry-period:
+  - 3s
+  leader-elect-resource-lock:
+  - configmapsleases
+  controllers:
+  - '*'
+  - -ttl
+  - -bootstrapsigner
+  - -tokencleaner
+  cluster-signing-duration:
+  - 720h
+  secure-port:
+  - "10257"
+  cert-dir:
+  - /var/run/kubernetes
+  root-ca-file:
+  - /etc/kubernetes/static-pod-resources/configmaps/serviceaccount-ca/ca-bundle.crt
+  service-account-private-key-file:
+  - /etc/kubernetes/static-pod-resources/secrets/service-account-private-key/service-account.key
+  cluster-signing-cert-file:
+  - /etc/kubernetes/static-pod-certs/secrets/csr-signer/tls.crt
+  cluster-signing-key-file:
+  - /etc/kubernetes/static-pod-certs/secrets/csr-signer/tls.key
+  kube-api-qps:
+  - "150"
+  kube-api-burst:
+  - "300"
--- a/pkg/controllers/kube-controller-manager.go
+++ b/pkg/controllers/kube-controller-manager.go
@@ -17,88 +17,99 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
+	"sort"
 
-	"github.com/spf13/cobra"
-
+	embedded "github.com/openshift/microshift/assets"
 	"github.com/openshift/microshift/pkg/assets"
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 
-	"k8s.io/component-base/cli/globalflag"
-	"k8s.io/component-base/version/verflag"
+	kubecontrolplanev1 "github.com/openshift/api/kubecontrolplane/v1"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	kubecm "k8s.io/kubernetes/cmd/kube-controller-manager/app"
-	kubecmoptions "k8s.io/kubernetes/cmd/kube-controller-manager/app/options"
 
+	"github.com/ghodss/yaml"
 	klog "k8s.io/klog/v2"
 )
 
+const (
+	kcmDefaultConfigAsset = "components/kube-controller-manager/defaultconfig.yaml"
+)
+
 type KubeControllerManager struct {
-	kubecmOptions *kubecmoptions.KubeControllerManagerOptions
-	kubeconfig    string
-	kubeadmConfig string
+	args    []string
+	applyFn func() error
+
+	// TODO: report configuration errors immediately
+	configureErr error
 }
 
 func NewKubeControllerManager(cfg *config.MicroshiftConfig) *KubeControllerManager {
 	s := &KubeControllerManager{}
-	s.configure(cfg)
+	// TODO: manage and invoke the configure bits independently outside of this.
+	s.args, s.applyFn, s.configureErr = configure(cfg)
 	return s
 }
 
 func (s *KubeControllerManager) Name() string           { return "kube-controller-manager" }
 func (s *KubeControllerManager) Dependencies() []string { return []string{"kube-apiserver"} }
 
-func (s *KubeControllerManager) configure(cfg *config.MicroshiftConfig) {
+func kcmRootCAFile() string {
+	certsDir := cryptomaterial.CertsDirectory(microshiftDataDir)
+	return cryptomaterial.ServiceAccountTokenCABundlePath(certsDir)
+}
+
+func kcmClusterSigningCertKeyAndFile() (string, string) {
 	certsDir := cryptomaterial.CertsDirectory(microshiftDataDir)
 	csrSignerDir := cryptomaterial.CSRSignerCertDir(certsDir)
-	kubeconfig := cfg.KubeConfigPath(config.KubeControllerManager)
-	kubeadmConfig := cfg.KubeConfigPath(config.KubeAdmin)
+	return cryptomaterial.CAKeyPath(csrSignerDir), cryptomaterial.CACertPath(csrSignerDir)
+}
 
-	opts, err := kubecmoptions.NewKubeControllerManagerOptions()
-	if err != nil {
-		klog.Fatalf("%s initialization error command options: %v", s.Name(), err)
-	}
-	s.kubecmOptions = opts
-	s.kubeconfig = kubeconfig
-	s.kubeadmConfig = kubeadmConfig
+func kcmServiceAccountPrivateKeyFile() string {
+	return microshiftDataDir + "/resources/kube-apiserver/secrets/service-account-key/service-account.key"
+}
 
-	args := []string{
-		"--kubeconfig=" + kubeconfig,
-		"--service-account-private-key-file=" + microshiftDataDir + "/resources/kube-apiserver/secrets/service-account-key/service-account.key",
-		"--allocate-node-cidrs=true",
-		"--cluster-cidr=" + cfg.Cluster.ClusterCIDR,
-		"--authorization-kubeconfig=" + kubeconfig,
-		"--authentication-kubeconfig=" + kubeconfig,
-		"--root-ca-file=" + cryptomaterial.ServiceAccountTokenCABundlePath(certsDir),
-		"--bind-address=127.0.0.1",
-		"--secure-port=10257",
-		"--leader-elect=false",
-		"--use-service-account-credentials=true",
-		"--cluster-signing-cert-file=" + cryptomaterial.CACertPath(csrSignerDir),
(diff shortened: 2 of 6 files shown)


assets/components/kube-controller-manager/defaultconfig.yaml
```diff
@@ -0,0 +1,46 @@
...
+  configure-cloud-routes:
+  - "false"
+  use-service-account-credentials:
+  - "true"
+  flex-volume-plugin-dir:
+  - "/etc/kubernetes/kubelet-plugins/volume/exec" # created by machine-config-operator, owned by storage team/hekumar@redhat.com
+  pv-recycler-pod-template-filepath-nfs: # owned by storage team/fbertina@redhat.com
+  - "/etc/kubernetes/static-pod-resources/configmaps/recycler-config/recycler-pod.yaml"
+  pv-recycler-pod-template-filepath-hostpath: # owned by storage team/fbertina@redhat.com
+  - "/etc/kubernetes/static-pod-resources/configmaps/recycler-config/recycler-pod.yaml"
```
tkashem, line 17: @benluddy  i think we should remove the above keys, up to `flex-volume-plugin-dir`

assets/components/kube-controller-manager/defaultconfig.yaml
```diff
@@ -0,0 +1,46 @@
...
+  leader-elect:
+  - "true"
```
tkashem, line 19: should be `false`

assets/components/kube-controller-manager/defaultconfig.yaml
dgrisonnet: I would agree.

Flex volumes plugins are deprecated in both k8s and OCP in favor of CSI drivers:
- https://kubernetes.io/docs/concepts/storage/volumes/#flexvolume
- https://docs.openshift.com/container-platform/4.11/storage/persistent_storage/persistent-storage-flexvolume.html

Same for custom recyclers: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#recycle, they are replaced by dynamic provisioning.

dhellmann: The pattern we've established is to have the [scripts/auto-rebase/rebase.sh](https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase.sh#L395) script copy the asset from the upstream component and use yq to modify it in place with microshift-specific changes. That way the file checked into git here reflects what we want at runtime and we have automation to keep those changes in place each time we update the dependencies.

tkashem:
progress, I do see the message
```
Nov 10 13:32:37 release-ci-ci-op-m3vn0jgm-68c08 microshift[62414]: kube-controller-manager I1110 13:32:37.286556   62414 kube-controller-manager.go:145] kube-controller-manager is ready
```

benluddy:
/retitle [WIP] WRKLDS-604: Enable the Openshift flavor of the kube-controller-manager

assets/components/kube-controller-manager/defaultconfig.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - "*"
+  - "-ttl" # TODO: this is excluded in kube-core, but not in #21092
+  - "-bootstrapsigner"
+  - "-tokencleaner"
+  cluster-signing-duration:
+  - "720h"
+  secure-port:
+  - "10257"
+  cert-dir:
+  - "/var/run/kubernetes"
```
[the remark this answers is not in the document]
atiratree: what is the KCM serving cert going to be?

pkg/controllers/kube-controller-manager.go
```diff
@@ -114,15 +146,39 @@ func (s *KubeControllerManager) Run(ctx context.Context, ready chan<- struct{},
...
+	// Carrying a patch for NewControllerManagerCommand to use cmd.Context().Done()
+	// as the stop channel instead of the channel returned by SetupSignalHandler,
+	// which expects to be called at most once in a process.
+	cmd := kubecm.NewControllerManagerCommand()
+	cmd.SetArgs([]string{
+		"--kubeconfig", s.kubeconfig,
+		"--authentication-kubeconfig", s.kubeconfig,
+		"--authorization-kubeconfig", s.kubeconfig,
+		"--openshift-config", fd.Name(),
+	})
```
atiratree, line 178: is not setting`tls-min-version` and `tls-cipher-suites` any concern for us?

pkg/controllers/kube-controller-manager.go
```diff
@@ -16,89 +16,121 @@ limitations under the License.
...
+				"false",
+			},
+			"use-service-account-credentials": {
+				"true",
+			},
+			"cluster-signing-cert-file": {
+				cryptomaterial.CACertPath(csrSignerDir),
+			},
+			"cluster-signing-key-file": {
+				cryptomaterial.CAKeyPath(csrSignerDir),
```
atiratree, line 109: seeing under this `kubelet-csr-signer-signer` - this should be used by both apiserver and kubelet right? Maybe just the naming is a it off?

pkg/controllers/kube-controller-manager.go
```diff
@@ -16,89 +16,121 @@ limitations under the License.
...
+			},
+			"enable-dynamic-provisioning": {
```
atiratree, line 111: isn't this a bit limiting to have this turned off?

assets/components/kube-controller-manager/defaultconfig.yaml
tkashem: yes, the rebase process will pull this file from kcm-o. overriding in a separate layer makes the rebase easier and also less error prone.
I also added a unit test to verify that microshift doesn't automatically opt in to configuration changes propagating from opeshift. In such a case, the unit test will fail after rebase and we can be aware of the configuration change and then decide whether to opt in or out

benluddy:
/approve

Looks reasonable once the pv-recycler template and flex volume plugin flags are fixed. I will need to eventually revisit my earlier O/K signal carry patch so that we can avoid the carry in MicroShift.

pkg/controllers/kube-controller-manager.go
tkashem: I opened this issue to track these options: https://github.com/openshift/microshift/issues/1113. This PR makes an attempt to retain the current configuration in effect.

tkashem: and here is the jira bug: https://issues.redhat.com/browse/USHIFT-588

fzdarsky:
@tkashem overall this LGTM. I'd just like to get clarification on the purpose of the `kcmPathResolver`, please: It seems that you're copying paths from the MicroShift config and the cryptomaterial into a single struct, just to immediately copy it out again. So doesn't seem useful, lowers readability, and reduces consistency with other embedded components. Is there a deeper plan behind that would justify this?

scripts/auto-rebase/rebase_patches/0030-kube-controller-manager-command-context.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/vendor/k8s.io/kubernetes/cmd/kube-controller-manager/app/controllermanager.go b/vendor/k8s.io/kubernetes/cmd/kube-controller-manager/app/controllermanager.go
+index 28caa49e..9045347f 100644
+--- a/vendor/k8s.io/kubernetes/cmd/kube-controller-manager/app/controllermanager.go
```
dhellmann, line 3: What are the chances of being able to upstream a change so we don't need to continue to carry this patch?

tkashem:
> @tkashem overall this LGTM. I'd just like to get clarification on the purpose of the `kcmPathResolver`, please: It seems that you're copying paths from the MicroShift config and the cryptomaterial into a single struct, just to immediately copy it out again. So doesn't seem useful, lowers readability, and reduces consistency with other embedded components. Is there a deeper plan behind that would justify this?

It is primarily driven by the need to have a useful test to verify that the `configure` method generates the options as expected
```
func TestConfigure(t *testing.T) {
	cfg := config.NewMicroshiftConfig()
	configBuilder := &kcmConfigDoer{
		pathResolverFn: kcmPathResolver,
	}
	kcm := newKubeControllerManager(cfg, configBuilder)

	paths := kcmPathResolver(cfg)
	argsWant := []string{
		"--allocate-node-cidrs=true",
		fmt.Sprintf("--authentication-kubeconfig=%s", paths.KubeConfig),
		fmt.Sprintf("--authorization-kubeconfig=%s", paths.KubeConfig),
...
```

Yes, the plan is we need to have strict unit test that validates configuration values in use by microshift, so in the future when we rebase i expect the test to fail if openshift is adding new configuration, or updating an existing one. Basically we want explicit opt in or out on the part of the microshift product owner/team when the configuration drifts.

I understand that the actual path resolution is no longer inline with the overrides, it's a trade off, but i can move the path resolution function just near the overrides if that makes it easier to read.

Or if there is a better way to achieve the same goal by keeping the path resolution inline with the overrides, I am happy to make the changes, thoughts?

fzdarsky:
> Or if there is a better way to achieve the same goal by keeping the path resolution inline with the overrides, I am happy to make the changes, thoughts?

I understand the goal of this PR as "ensure we start with OpenShift KCM's args and merge in MicroShift's overrides", to which you add "ensure a unit test exists that fails when OpenShift's config (and thus the merged config) drifts". Both goals make total sense to me.

You could achieve that by substituting back the `kcmPath` fields (e.g. `paths.KubeConfig` with `cfg.KubeConfigPath(config.KubeControllerManager` in both the `configure()` and `TestConfigure()` methods and you'd be testing the same thing / achieve the same goals. Next, you'd remove the `kcmPathResolver` and `kcmConfigDoer` part and you wouldn't lose a thing.

This is why I'm confused why all these are introduced in the PR. It seems to add bloat without adding (obvious) value?

benluddy:
> the goal of this PR as "ensure we start with OpenShift KCM's args and merge in MicroShift's overrides"

The goal is to use the same KCM entrypoint used by OCP (https://github.com/openshift/microshift/blob/9125d0c133915144b3737bb828a9a048acef6816/vendor/k8s.io/kubernetes/cmd/kube-controller-manager/app/controllermanager.go#L133-L171), since OCP carries patches there. I understand the args-related churn to be incidental to that.

tkashem:
> You could achieve that by substituting back the kcmPath fields (e.g. paths.KubeConfig with cfg.KubeConfigPath(config.KubeControllerManager in both the configure() and TestConfigure() methods and you'd be testing the same thing / achieve the same goals. Next, you'd remove the kcmPathResolver and kcmConfigDoer part and you wouldn't lose a thing.

I wanted to avoid the test having an unnecessarily bloated setup where it copy pastes a bunch of hard coded kubeconfig  and cert paths. The test setup should be agnostic of how the paths are resolved. 
I think I can achieve both goals by replacing `kcmPathResolver` with a set of internal functions that returns resolved paths

Not in the scope of this PR, but we would like to decouple the configuration bits from the `Run` constructs. This way, we can have more graceful error handling and more focused tests on the config bits.

scripts/auto-rebase/rebase_patches/0030-kube-controller-manager-command-context.patch
dhellmann: OK, I was thinking k/k. If this isn't really useful in o/k it probably makes as much sense to carry it here as there. I don't have a preference.

pkg/controllers/kube-controller-manager.go
```diff
@@ -17,88 +17,99 @@ package controllers
...
+	applyFn func() error
+
+	// TODO: report configuration errors immediately
+	configureErr error
 }
 
 func NewKubeControllerManager(cfg *config.MicroshiftConfig) *KubeControllerManager {
 	s := &KubeControllerManager{}
-	s.configure(cfg)
+	// TODO: manage and invoke the configure bits independently outside of this.
```
dhellmann, line 53: It does feel like we have a gap in the API for the controller manager if configuration can return an error now. Let's work on that separately from these changes, though.

pkg/controllers/kube-controller-manager_test.go
```diff
@@ -0,0 +1,76 @@
...
+	defaultConfigBytes, err := embedded.Asset(kcmDefaultConfigAsset)
+	if err != nil {
+		t.Errorf("expected valid default config, error: %v", err)
+	}
+	if len(defaultConfigBytes) == 0 {
+		t.Errorf("expected non-empty byte array")
+	}
+}
+
+func TestConfigure(t *testing.T) {
```
dhellmann, line 39: If this test is expected to fail when new configuration options appear so that we ack them and make an explicit decision about them for MicroShift, it would be good to have a comment explaining that here at the top. I don't want to hold up this work, so if add a comment like that the only reason to update this PR, it can come in another PR.

dhellmann (commented):
I'm happy with this version. I had a couple of nits, but those can be handled separately.

I'll leave it for @fzdarsky to lgtm, since he also had earlier review comments.

fzdarsky:
/lgtm

Thank you.

PR #1105, opened 2022-11-09
USHIFT-579: eliminate old replicaset of router deployment

zshi-redhat:
revisionHistoryLimit limits the number of old replicaset to 0,
so that there will be no more than one instance of router
replicaset after microshift restart.

### DIFF
--- a/assets/components/openshift-router/deployment.yaml
+++ b/assets/components/openshift-router/deployment.yaml
@@ -124,6 +124,8 @@ spec:
     matchLabels:
       ingresscontroller.operator.openshift.io/deployment-ingresscontroller: default
   replicas: 1
+  strategy:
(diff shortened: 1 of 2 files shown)


zshi-redhat:
/hold
until this is verified.

fzdarsky:
Is the problem limited to MicroShift restarts or does it also exist on _any_ update to the openshift-ingress Deployment?

I believe the underlying issue is that with the current deployment strategy `.spec.strategy.type==RollingUpdate`, a new ReplicaSet (and thus ingress pod) is created first before the old is deleted and that create won't ever succeed because the host network ports aren't available. In that case, would it make more sense to set `.spec.strategy.type==Recreate` instead?

PR #1108, opened 2022-11-10
USHIFT-432: CSI image / manifest rebase script

copejon:

[USHIFT-432](https://issues.redhat.com//browse/USHIFT-432)

### DIFF
--- a/assets/components/lvms/topolvm-controller_deployment.yaml
+++ b/assets/components/lvms/topolvm-controller_deployment.yaml
@@ -26,7 +26,7 @@ spec:
       - command:
         - /topolvm-controller
         - --cert-dir=/certs
-        image: {{ .ReleaseImage.topolvm-csi }}
+        image: {{ .ReleaseImage.topolvm_csi }}
         imagePullPolicy: IfNotPresent
         livenessProbe:
           failureThreshold: 3
@@ -80,7 +80,7 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: metadata.namespace
-        image: {{ .ReleaseImage.topolvm-csi-provisioner }}
+        image: {{ .ReleaseImage.topolvm_csi_provisioner }}
         imagePullPolicy: IfNotPresent
         name: csi-provisioner
         resources:
@@ -94,7 +94,7 @@ spec:
           name: socket-dir
       - args:
         - --csi-address=/run/topolvm/csi-topolvm.sock
-        image: {{ .ReleaseImage.topolvm-csi-resizer }}
+        image: {{ .ReleaseImage.topolvm_csi_resizer }}
         imagePullPolicy: IfNotPresent
         name: csi-resizer
         resources:
@@ -108,7 +108,7 @@ spec:
           name: socket-dir
       - args:
         - --csi-address=/run/topolvm/csi-topolvm.sock
-        image: {{ .ReleaseImage.topolvm-csi-livenessprobe }}
+        image: {{ .ReleaseImage.topolvm_csi_livenessprobe }}
         imagePullPolicy: IfNotPresent
         name: liveness-probe
         resources:
--- a/assets/components/lvms/topolvm-controller_deployment.yaml
+++ b/assets/components/lvms/topolvm-controller_deployment.yaml
@@ -4,21 +4,21 @@ metadata:
   name: topolvm-controller
   namespace: openshift-storage
 spec:
-  progressDeadlineSeconds: 600
   replicas: 1
-  revisionHistoryLimit: 10
   selector:
     matchLabels:
-      app.kubernetes.io/name: topolvm-controller
-  strategy:
-    rollingUpdate:
-      maxSurge: 25%
-      maxUnavailable: 25%
-    type: RollingUpdate
+      app.kubernetes.io/component: topolvm-controller
+      app.kubernetes.io/managed-by: lvms-operator
+      app.kubernetes.io/name: topolvm-csi-driver
+      app.kubernetes.io/part-of: lvms-provisioner
+  strategy: {}
   template:
     metadata:
       labels:
-        app.kubernetes.io/name: topolvm-controller
+        app.kubernetes.io/component: topolvm-controller
+        app.kubernetes.io/managed-by: lvms-operator
+        app.kubernetes.io/name: topolvm-csi-driver
+        app.kubernetes.io/part-of: lvms-provisioner
       name: topolvm-controller
       namespace: openshift-storage
     spec:
@@ -27,16 +27,13 @@ spec:
         - /topolvm-controller
         - --cert-dir=/certs
         image: {{ .ReleaseImage.topolvm_csi }}
-        imagePullPolicy: IfNotPresent
         livenessProbe:
           failureThreshold: 3
           httpGet:
             path: /healthz
             port: healthz
-            scheme: HTTP
           initialDelaySeconds: 10
           periodSeconds: 60
-          successThreshold: 1
           timeoutSeconds: 3
         name: topolvm-controller
         ports:
@@ -44,20 +41,14 @@ spec:
           name: healthz
           protocol: TCP
         readinessProbe:
-          failureThreshold: 3
           httpGet:
             path: /metrics
             port: 8080
             scheme: HTTP
-          periodSeconds: 10
-          successThreshold: 1
-          timeoutSeconds: 1
         resources:
           requests:
             cpu: 2m
             memory: 31Mi
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /run/topolvm
           name: socket-dir
@@ -73,54 +64,42 @@ spec:
         - name: POD_NAME
           valueFrom:
             fieldRef:
-              apiVersion: v1
               fieldPath: metadata.name
         - name: NAMESPACE
           valueFrom:
             fieldRef:
-              apiVersion: v1
               fieldPath: metadata.namespace
         image: {{ .ReleaseImage.topolvm_csi_provisioner }}
-        imagePullPolicy: IfNotPresent
         name: csi-provisioner
         resources:
           requests:
             cpu: 2m
             memory: 35Mi
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /run/topolvm
           name: socket-dir
       - args:
         - --csi-address=/run/topolvm/csi-topolvm.sock
         image: {{ .ReleaseImage.topolvm_csi_resizer }}
-        imagePullPolicy: IfNotPresent
         name: csi-resizer
         resources:
           requests:
             cpu: 1m
             memory: 23Mi
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /run/topolvm
           name: socket-dir
       - args:
         - --csi-address=/run/topolvm/csi-topolvm.sock
         image: {{ .ReleaseImage.topolvm_csi_livenessprobe }}
-        imagePullPolicy: IfNotPresent
         name: liveness-probe
         resources:
           requests:
             cpu: 1m
             memory: 9Mi
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /run/topolvm
           name: socket-dir
-      dnsPolicy: ClusterFirst
       initContainers:
       - command:
         - /usr/bin/bash
@@ -128,20 +107,12 @@ spec:
         - openssl req -nodes -x509 -newkey rsa:4096 -subj '/DC=self_signed_certificate'
           -keyout /certs/tls.key -out /certs/tls.crt -days 3650
         image: {{ .ReleaseImage.openssl }}
-        imagePullPolicy: IfNotPresent
         name: self-signed-cert-generator
         resources: {}
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /certs
           name: certs
-      priorityClassName: system-cluster-critical
-      restartPolicy: Always
-      schedulerName: default-scheduler
-      securityContext: {}
       serviceAccountName: topolvm-controller
-      terminationGracePeriodSeconds: 30
       volumes:
       - emptyDir: {}
         name: socket-dir
--- a/assets/components/lvms/topolvm-controller_deployment.yaml
+++ b/assets/components/lvms/topolvm-controller_deployment.yaml
@@ -26,7 +26,7 @@ spec:
       - command:
         - /topolvm-controller
         - --cert-dir=/certs
-        image: {{ .ReleaseImage.lvms_topolvm }}
+        image: {{ .ReleaseImage.topolvm-csi }}
         imagePullPolicy: IfNotPresent
         livenessProbe:
           failureThreshold: 3
@@ -80,7 +80,7 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: metadata.namespace
-        image: {{ .ReleaseImage.csi_external_provisioner }}
+        image: {{ .ReleaseImage.topolvm-csi-provisioner }}
         imagePullPolicy: IfNotPresent
         name: csi-provisioner
         resources:
@@ -94,7 +94,7 @@ spec:
           name: socket-dir
       - args:
         - --csi-address=/run/topolvm/csi-topolvm.sock
-        image: {{ .ReleaseImage.csi_external_resizer }}
+        image: {{ .ReleaseImage.topolvm-csi-resizer }}
         imagePullPolicy: IfNotPresent
         name: csi-resizer
         resources:
@@ -108,7 +108,7 @@ spec:
           name: socket-dir
       - args:
         - --csi-address=/run/topolvm/csi-topolvm.sock
-        image: {{ .ReleaseImage.csi_livenessprobe }}
+        image: {{ .ReleaseImage.topolvm-csi-livenessprobe }}
         imagePullPolicy: IfNotPresent
         name: liveness-probe
         resources:
--- a/assets/components/lvms/topolvm-node-metrics_v1_service.yaml
+++ b/assets/components/lvms/topolvm-node-metrics_v1_service.yaml
@@ -14,6 +14,6 @@ spec:
     protocol: TCP
     targetPort: 8080
   selector:
-    app.kubernetes.io/name: topolvm-node
+    app.kubernetes.io/component: topolvm-node
 status:
   loadBalancer: {}
--- a/assets/release/release-aarch64.json
+++ b/assets/release/release-aarch64.json
@@ -11,10 +11,10 @@
     "ovn-kubernetes-microshift": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:534258847105e616f856f6a703b9e1ec622dbc2d3541efbb069e7f656d5a23a4",
     "pod": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:467984e3d3f19740278e95b3928ba2239571ca134896944dc160039451f262c2",
     "service-ca-operator": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2ddea0c30e6facf46d52015482a7f0914a8ec0182cff3f0d5ec4a4f6035114e8",
-    "topolvm-csi": "registry.redhat.io/lvms4/topolvm-rhel8@sha256:10bffded5317da9de6c45ba74f0bb10e0a08ddb2bfef23b11ac61287a37f10a1",
-    "topolvm-csi-registrar": "registry.redhat.io/openshift4/ose-csi-node-driver-registrar@sha256:a4319ff7c736ca9fe20500dc3e5862d6bb446f2428ea2eadfb5f042195f4f860",
-    "topolvm-csi-livenessprobe": "registry.redhat.io/openshift4/ose-csi-livenessprobe@sha256:9df24be671271f5ea9414bfd08e58bc2fa3dc4bc68075002f3db0fd020b58be0",
-    "topolvm-csi-resizer": "registry.redhat.io/openshift4/ose-csi-external-resizer@sha256:9d486daffd348664c00d8b80bd0da973b902f3650acdef37e1b813278ed6c107",
-    "topolvm-csi-provisioner": "registry.redhat.io/openshift4/ose-csi-external-provisioner@sha256:199eac2ba4c8390daa511b040315e415cfbcfa80aa7af978a33624445b96c17c"
+    "topolvm_csi": "registry.redhat.io/lvms4/topolvm-rhel8@sha256:10bffded5317da9de6c45ba74f0bb10e0a08ddb2bfef23b11ac61287a37f10a1",
+    "topolvm_csi_registrar": "registry.redhat.io/openshift4/ose-csi-node-driver-registrar@sha256:a4319ff7c736ca9fe20500dc3e5862d6bb446f2428ea2eadfb5f042195f4f860",
+    "topolvm_csi_livenessprobe": "registry.redhat.io/openshift4/ose-csi-livenessprobe@sha256:9df24be671271f5ea9414bfd08e58bc2fa3dc4bc68075002f3db0fd020b58be0",
+    "topolvm_csi_resizer": "registry.redhat.io/openshift4/ose-csi-external-resizer@sha256:9d486daffd348664c00d8b80bd0da973b902f3650acdef37e1b813278ed6c107",
+    "topolvm_csi_provisioner": "registry.redhat.io/openshift4/ose-csi-external-provisioner@sha256:199eac2ba4c8390daa511b040315e415cfbcfa80aa7af978a33624445b96c17c"
   }
 }
--- a/assets/release/release-aarch64.json
+++ b/assets/release/release-aarch64.json
@@ -7,14 +7,14 @@
     "coredns": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:4af574e5ff2866d99c98b579ef3fdef0511c6a80343cf09da80ba7bf460ab30e",
     "haproxy-router": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2746cf3272b28f3479aace4f9ede7f54ada64dd3168fdd708239937f10ac933f",
     "kube-rbac-proxy": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:a16663e3c4547af28738a8690bf3f9eb5b7e41bde92a0fd2008b9254c87b09b9",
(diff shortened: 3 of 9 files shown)


pkg/release/release.go
```diff
@@ -19,12 +19,13 @@ package release
...
-	"kube_rbac_proxy":           "quay.io/microshift/kube-rbac-proxy:" + Base,
-	"odf_topolvm":               "quay.io/microshift/odf-topolvm-rhel8" + Base,
-	"openssl":                   "quay.io/microshift/openssl" + Base,
+	"cli":             "quay.io/microshift/cli:" + Base,
+	"coredns":         "quay.io/microshift/coredns:" + Base,
+	"haproxy_router":  "quay.io/microshift/haproxy-router:" + Base,
+	"kube_rbac_proxy": "quay.io/microshift/kube-rbac-proxy:" + Base,
+	"odf_topolvm":     "quay.io/microshift/odf-topolvm-rhel8" + Base,
+	"openssl":         "quay.io/microshift/openssl" + Base,
+	// TODO (copejon) add snapshotter
```
dhellmann, line 28: I think most of the whitespace changes in this file are triggered by this comment. Maybe Jira is a better place to track the todo?

scripts/auto-rebase/rebase_odf-lvm.sh
```diff
@@ -0,0 +1,144 @@
...
+        ;;
+    esac
+
+    local staged_release
+    staged_release="$(mktemp -d)/$(basename "$release_file")"
+    cp "$release_file" "${staged_release}"
+    # shellcheck disable=SC2155
+    local images="$(parse_images "$src")"
+    # ensure that if a sed substitute fails, we bail out rather than clobber a release file
+    if ! (
```
dhellmann, line 73: We go to a lot of trouble to parse and update the release.go file here and in the main rebase script. What if we wrote a YAML file to the assets directory instead, and had release.go load and parse it? We would want to make the same change for all images, so we should not hold up this PR over the discussion.

copejon: Yeah, I think you're right.  We go through a lot of trouble to do hacky string replacements in the release*.go files.  We'd probably be better off reusing our bindata pipeline to embed the release image data in the binary from a yaml source, rather than all the shell-driven string substitutions we do now.

fzdarsky: In general, though, I agree this script should align more with the other rebase script in style and reuse code.

copejon:
Failing due to missing namespace fields in operator bundles' manifests.  Will fix.

assets/components/lvms/topolvm-node-metrics_v1_service.yaml
```diff
@@ -14,6 +14,6 @@ spec:
     protocol: TCP
     targetPort: 8080
   selector:
-    app.kubernetes.io/name: topolvm-node
+    app.kubernetes.io/component: topolvm-node
```
ggiguash, line 17: Is this something we had missed in manual rebase that was found by the automation?

assets/release/release-aarch64.json
```diff
@@ -15,6 +15,11 @@
...
-    "service-ca-operator": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2ddea0c30e6facf46d52015482a7f0914a8ec0182cff3f0d5ec4a4f6035114e8"
+    "service-ca-operator": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2ddea0c30e6facf46d52015482a7f0914a8ec0182cff3f0d5ec4a4f6035114e8",
+    "topolvm-csi": "registry.redhat.io/lvms4/topolvm-rhel8@sha256:10bffded5317da9de6c45ba74f0bb10e0a08ddb2bfef23b11ac61287a37f10a1",
+    "topolvm-csi-registrar": "registry.redhat.io/openshift4/ose-csi-node-driver-registrar@sha256:a4319ff7c736ca9fe20500dc3e5862d6bb446f2428ea2eadfb5f042195f4f860",
+    "topolvm-csi-livenessprobe": "registry.redhat.io/openshift4/ose-csi-livenessprobe@sha256:9df24be671271f5ea9414bfd08e58bc2fa3dc4bc68075002f3db0fd020b58be0",
+    "topolvm-csi-resizer": "registry.redhat.io/openshift4/ose-csi-external-resizer@sha256:9d486daffd348664c00d8b80bd0da973b902f3650acdef37e1b813278ed6c107",
+    "topolvm-csi-provisioner": "registry.redhat.io/openshift4/ose-csi-external-provisioner@sha256:199eac2ba4c8390daa511b040315e415cfbcfa80aa7af978a33624445b96c17c"
```
ggiguash, line 23: We currently have the same images without "topolvm-" prefix. There seems to be some duplication here. Should the old names be removed?

assets/release/release-x86_64.json
```diff
@@ -15,6 +15,11 @@
...
-    "service-ca-operator": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:b2527145d783ab9ac0428e878e8d318d8f9b1627c707ff02f15fd70ccef7487c"
+    "service-ca-operator": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:b2527145d783ab9ac0428e878e8d318d8f9b1627c707ff02f15fd70ccef7487c",
+    "topolvm-csi": "registry.redhat.io/lvms4/topolvm-rhel8@sha256:10bffded5317da9de6c45ba74f0bb10e0a08ddb2bfef23b11ac61287a37f10a1",
+    "topolvm-csi-registrar": "registry.redhat.io/openshift4/ose-csi-node-driver-registrar@sha256:a4319ff7c736ca9fe20500dc3e5862d6bb446f2428ea2eadfb5f042195f4f860",
+    "topolvm-csi-livenessprobe": "registry.redhat.io/openshift4/ose-csi-livenessprobe@sha256:9df24be671271f5ea9414bfd08e58bc2fa3dc4bc68075002f3db0fd020b58be0",
+    "topolvm-csi-resizer": "registry.redhat.io/openshift4/ose-csi-external-resizer@sha256:9d486daffd348664c00d8b80bd0da973b902f3650acdef37e1b813278ed6c107",
+    "topolvm-csi-provisioner": "registry.redhat.io/openshift4/ose-csi-external-provisioner@sha256:199eac2ba4c8390daa511b040315e415cfbcfa80aa7af978a33624445b96c17c"
```
ggiguash, line 23: Same as above. There seems to be some duplication for csi- and topolvm-csi- images.

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
 }
 
+# LVMS is not integrated into the ocp release image, so the work flow does not fit with core component rebase.  LVMS'
+# operator bundle is the authoritative source for manifest and image digests.
+download_lvms_operator_bundle(){
+    title 'downloading LVMS operator bundles'
+    declare -A images=(["amd64"]="${1}" ["arm64"]="${2:-}")
+    local LVMS_STAGING="${STAGING_DIR}/lvms"
+    for arch in "${!images[@]}"; do
+        # TODO special case arm64 builds of topolvm since they are not available in 4.12
```
ggiguash, line 52: ARM images are available as part of the same manifest. 
See https://catalog.redhat.com/software/containers/lvms4/topolvm-rhel8/63972de3adcb55263891b983?architecture=arm64&container-tabs=gti

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
+                    popd
+                    return 1
+                }
+        popd || return 1
+      done
+}
+
+parse_images() {
+    local src=$1
+    yq '.spec.relatedImages[]? | [.name, .image] | @csv' $src > "$(dirname "$src")/images"
```
ggiguash, line 73: Can we also pass the output file as an argument, so that it's not defined in two places?

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
+    [ -d "$arch" ] || {
+        echo "dir $arch not found"
+        return 1
+    }
+    local csv_manifest="${arch}/lvms-operator.clusterserviceversion.yaml"
+
+    parse_images "$csv_manifest"
+    image_file="./${arch}/images"
+    if ! [ -e $image_file ]; then
+        echo "image file \"$image_file\" not found"
```
ggiguash, line 96: The file will "always" be created. Should we check that the file is not empty?

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
+        name=${LINE%,*}
+        img=${LINE#*,}
+        found=false
+        for included in "${include_images[@]}"; do
+            if [[ "$name" == $included ]]; then
+                found=true
+                break;
+            fi
+        done
+        $found || continue # name not ’found in $REPOROOT/assets/release/release-$arch.json, skip
```
ggiguash, line 109: Should we log a warning message here? Is this a "normal" situation?

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
+
+update_lvms_images(){
+    local workdir="$STAGING_DIR/lvms"
+    [ -d "$workdir" ] || {
+        >&2 echo 'lvms staging dir not found, aborting image update'
+        return 1
+    }
+    pushd "$workdir"
+    local archs
+    set -x
```
ggiguash, line 122: Let's remove this?

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
+    archs="$(ls -d */ | sed 's,/$,,')"
```
ggiguash, line 123: This can be done by `$(ls -1 .)` without the need for `sed`

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
+        write_lvms_images_for_arch "$a"
+    done
+    popd
+}
+
+set_openshift_storage_ns(){
+        local manifest="$1"
+        local tmp="$(oc create --dry-run=client --namespace=openshift-storage -o yaml -f "${manifest}")"
+        # assume that if the returned string is nil, it's because the resource is not namespaces.  Failures
+        # like file-not-found will still result in the script exiting with a non-zero code.
```
ggiguash, line 134: Can you explain what we're trying to do here? We are taking a manifest and overwriting it eventually?

scripts/auto-rebase/rebase.sh
```diff
@@ -43,6 +42,159 @@ title() {
...
+          "topolvm-csi-resizer_rbac.authorization.k8s.io_v1_rolebinding.yaml"
+          "topolvm-metrics_rbac.authorization.k8s.io_v1_role.yaml"
+          "topolvm-metrics_rbac.authorization.k8s.io_v1_rolebinding.yaml"
+          "topolvm-node-metrics_v1_service.yaml"
+          "topolvm-node_rbac.authorization.k8s.io_v1_clusterrole.yaml"
+          "topolvm-node_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml"
+          "topolvm-node_v1_serviceaccount.yaml"
+          "topolvm.io_logicalvolumes.yaml"
+    )
+    set -x
```
ggiguash, line 164: Should we remove this?

scripts/auto-rebase/rebase.sh
```diff
@@ -901,18 +1057,34 @@ usage() {
...
-    manifests) update_manifests;;
+    images)
+        update_images
+        update_lvms_images
+        ;;
+    manifests)
+        update_manifests
+        update_lvms_manifests
+        ;;
+    csi)
```
ggiguash, line 1080: This may have been a debugging option, but would we ever want to update just csi images? We do not have special option for ovn, etc., so why csi is special?

ggiguash (changes requested):

assets/components/lvms/topolvm-node-metrics_v1_service.yaml
ggiguash: Is it necessary to backport this fix to 4.12 release?

assets/release/release-aarch64.json
copejon: The topolvm name here reflects the container names in the upstream OCP manifests and means we don't have to maintain a mapping of OCP -> uShift images.

scripts/auto-rebase/rebase.sh
copejon: Ah, yes.  This is stale, will remove.

scripts/auto-rebase/rebase.sh
copejon: +1

assets/components/lvms/topolvm-node-metrics_v1_service.yaml
copejon: no, the previous version was also pulled from the operator bundle, so this change will have come from upstream topolvm

scripts/auto-rebase/rebase.sh
copejon: This check exists to filter out topolvm images used in OCP (like snapshotter) that are not used in microshift.  It's using the current release-*.json images list as the source of truth for this.

scripts/auto-rebase/rebase.sh
copejon: Good catch. I'll adjust it to check for content rather than file presence.

scripts/auto-rebase/rebase.sh
copejon: We're free to remove this.  It was initially for debugging, but b/c LVMS isn't tethered to OCP z-streams (i.e. isn't tracked by the ocp-release image), I figured it'd be useful to have should we need to pull in OCP y-stream compatible versions ad hoc.

assets/release/release-aarch64.json
ggiguash: So, we should probably remove all the csi-* images from the JSON files?

scripts/auto-rebase/rebase.sh
ggiguash: I think I'd rather remove this option to keep it similar with the other components.

scripts/auto-rebase/rebase.sh
copejon: The manifests in the operator bundle don't specify namespace, and MicroShift doesn't inherently assume a namespace for LVMS workloads.  This command feeds the downloaded manifests through `oc` to inject the namespace and outputs new yamls.  Using `oc` is nice b/c it won't specify a namespace on a non-namespaced object.

assets/release/release-aarch64.json
copejon: My mistake, I hadn't caught what you meant.  Will remove the old image tags.

scripts/auto-rebase/rebase.sh
```diff
@@ -904,18 +1054,25 @@ usage() {
 command=${1:-help}
 case "$command" in
     to)
-        [[ $# -ne 3 ]] && usage
-        rebase_to "$2" "$3"
+        [[ $# -ne 4 ]] && usage
+        rebase_to "$2" "$3" "$4"
```
ggiguash, line 1058: If we're changing the usage, let's update the "usage" function by adding something like LVMS_IMAGE_MULTI_ARCH?
We would also need to update the rebase automation calling this script.

scripts/auto-rebase/rebase.sh
```diff
@@ -36,13 +36,160 @@ RELEASE_TAG_RX="(.+)-([0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]{6})"
...
+    images="$(parse_images "$src")" || return 1
+    cat images
+    # ensure that if a sed substitute fails, we bail out rather than clobber a release file
+    if ! (
+        while IFS=' ' read -r line; do
+          yq '.images['""']'
+        done <"$images"
+    ); then
+        return 1
+    fi
```
ggiguash, line 190: All this construct looks a bit complex, cat images, if with nested while, etc. Can we please simplify this code?

ggiguash (changes requested):

ggiguash:
@copejon, please update the rebase documentation in this PR to reflect the changes to the script.

scripts/auto-rebase/rebase.sh
dhellmann: Is `oc` available in the rebase script when it runs in CI?

scripts/auto-rebase/rebase.sh
```diff
@@ -891,31 +1035,38 @@ rebase_to() {
...
 case "$command" in
     to)
-        [[ $# -ne 3 ]] && usage
-        rebase_to "$2" "$3"
+        [[ $# -ne 4 ]] && usage
+        rebase_to "$2" "$3" "$4"
         ;;
     download)
-        [[ $# -ne 3 ]] && usage
+        [[ $# -ne 4 ]] && usage
```
dhellmann, line 1055: Does the automated job get this new argument?

assets/release/release-x86_64.json
```diff
@@ -7,12 +7,7 @@
...
     "openssl": "registry.access.redhat.com/ubi8/openssl@sha256:9e743d947be073808f7f1750a791a3dbd81e694e37161e8c6c6057c2c342d671",
-    "csi-external-provisioner": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:1e865ed6300ea2095ed8d5b8ea0426112790409d09457453d992586809df49de",
-    "csi-external-resizer": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:0c797c755c5435b1cd12b09f7f0996041f62b80ef578f7306846261f88a2d6cd",
-    "csi-node-driver-registrar": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:4f70795d058996e0c4244582177416569a89df9281df9f35c0d58bb89a78aaa4",
-    "csi-livenessprobe": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7a3729aa6aafa40dd14f7f200ac79728f37b6e5863407e8d6e2aba8a7afacaf2",
     "ovn-kubernetes-microshift": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5e583d857aca165fa524bc6486f70aec7572b483e12f539a60e6f5f60566c79e",
     "pod": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:566c0838998b1dd8464b43dbe2150b78fcbb83262ea7d25d41c85acb1a33e31c",
     "service-ca-operator": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:b2527145d783ab9ac0428e878e8d318d8f9b1627c707ff02f15fd70ccef7487c"
```
ggiguash, line 13: Looks like we're missing the new images here

ggiguash (changes requested):

assets/release/release-aarch64.json
```diff
@@ -1,20 +1,20 @@
 {
   "release": {
-    "base": "4.13.0-0.nightly-arm64-2023-01-21-233324"
+    "base": "4.13.0-0.nightly-arm64-2023-01-17-171314"
```
pmtk, line 3: Could you use images from `last_rebase.sh`? To have this changes in sync with what we already have

scripts/auto-rebase/rebase.sh
```diff
@@ -891,31 +1035,38 @@ rebase_to() {
...
-    echo "$(basename "$0") download RELEASE_IMAGE_INTEL RELEASE_IMAGE_ARM   Downloads the content of a release image to disk in preparation for rebasing. Specify both amd64 and arm64."
-    echo "$(basename "$0") buildfiles                                       Updates build files (Makefile, Dockerfile, .spec)"
-    echo "$(basename "$0") go.mod                                           Updates the go.mod file to the downloaded release"
-    echo "$(basename "$0") generated-apis                                   Regenerates OpenAPIs"
-    echo "$(basename "$0") images                                           Rebases the component images to the downloaded release"
-    echo "$(basename "$0") manifests                                        Rebases the component manifests to the downloaded release"
+    echo "$(basename "$0") to RELEASE_IMAGE_INTEL RELEASE_IMAGE_ARM LVMS_RELEASE_MULTI_ARCH        Performs all the steps to rebase to a release image. Specify both amd64 and arm64 ocp releases, and multi-arch lvms operator bundle image."
```
pmtk, line 1038: So, job will have this argument hardcoded to `registry.redhat.io/lvms4/lvms-operator-bundle:v4.YY` depending on branch it runs for?  Like `release-4.12` -> `v4.12`, `main` -> `v4.13` (where it's released)?

scripts/auto-rebase/rebase.sh
copejon: Yes, it's an existing prereq for the script.  This PR does not add new dependencies.

scripts/auto-rebase/rebase.sh
copejon: Not yet.  To prevent breaking the job in CI, I intend to update CI after this is merged to update the env vars that are consumed by the wrapper script (rebase-script-automation.sh), and then glue it all together with an update to wrapper script to pass the env var thru. Hopefully, this will mean the rebase job doesn't see any downtime.

pmtk: This line changes expectation that script will receive one more argument, so shouldn't we change job config so we have this available in container and then merge this change + changes in `rebase_job_entrypoint.sh`?
If you merge this first and not change `rebase_job_entrypoint.sh` then the script will fail expecting 4 args when given 3 or I missed something?

scripts/auto-rebase/rebase_job_entrypoint.sh
```diff
@@ -18,14 +18,23 @@ release_amd64="$(oc get configmap/release-release-images-latest -o yaml \
 release_arm64="$(oc get configmap/release-release-images-arm64-latest -o yaml \
     | yq '.data."release-images-arm64-latest.yaml"' \
     | jq -r '.metadata.name')"
+# LVMS is not tracked in the OCP release image.  Instead, rely on the latest from the same X.Y stream as the release image.
+# Input:
+#   registry.ci.openshift.org/ocp/release:4.13.0-0.nightly-2023-01-17-152326
+# Output:
+#    "4.13"
+release_lvms=$( echo "$release_amd64" | sed -E 's,.*:(v?[0-9]+\.[0-9]+).*,\1,')
```
ggiguash, line 26: How is this going to work in main branch? I mean, we have no 4.13 release of LVMS yet, don't we?

pmtk, line 26: I think, in CI, release_amd64 is already a tag:
```
++ oc get configmap/release-release-images-latest -o yaml
++ yq '.data."release-images-latest.yaml"'
++ jq -r .metadata.name
+ release_amd64=4.13.0-0.nightly-2023-01-27-165107
```
so I guess you'd need to drop `.*:` from regexp.
Also, Giga is right - we might just hardcode 4.12 for now

assets/components/lvms/topolvm-controller_deployment.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   revisionHistoryLimit: 10
   selector:
     matchLabels:
-      app.kubernetes.io/name: topolvm-controller
+      app.kubernetes.io/name: topolvm_controller
```
pmtk, line 12: I don't see this file in:
```
update_lvms_manifests() {
    local src="${STAGING_DIR}/lvms/amd64"
    local dest="${REPOROOT}/assets/components/lvms"
    local manifests=(
          "topolvm-controller_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml"
          "topolvm-controller_rbac.authorization.k8s.io_v1_clusterrole.yaml"
          "topolvm-controller_v1_serviceaccount.yaml"
          "topolvm-csi-provisioner_rbac.authorization.k8s.io_v1_clusterrole.yaml"
          "topolvm-csi-provisioner_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml"
          "topolvm-csi-provisioner_rbac.authorization.k8s.io_v1_role.yaml"
          "topolvm-csi-provisioner_rbac.authorization.k8s.io_v1_rolebinding.yaml"
          "topolvm-csi-resizer_rbac.authorization.k8s.io_v1_clusterrole.yaml"
          "topolvm-csi-resizer_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml"
          "topolvm-csi-resizer_rbac.authorization.k8s.io_v1_role.yaml"
          "topolvm-csi-resizer_rbac.authorization.k8s.io_v1_rolebinding.yaml"
          "topolvm-metrics_rbac.authorization.k8s.io_v1_role.yaml"
          "topolvm-metrics_rbac.authorization.k8s.io_v1_rolebinding.yaml"
          "topolvm-node-metrics_v1_service.yaml"
          "topolvm-node_rbac.authorization.k8s.io_v1_clusterrole.yaml"
          "topolvm-node_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml"
          "topolvm-node_v1_serviceaccount.yaml"
          "topolvm.io_logicalvolumes.yaml"
    )
```
How that line changed? Shouldn't we have this file in the `manifests` array?

copejon: unfortunately this file doesn't exist as part of the manifest bundle.  This deployment (and the daemonset) are both programmatically defined in the LVMS operator code.  We have a ticket open with LVMS team to add a render type behavior for these.

pmtk:
This look good to me. We also tested changes using pj-rehearse with my fork and the rebase PR looks good as well.

/lgtm

assets/components/lvms/topolvm-node-metrics_v1_service.yaml
pmtk, line 17: Sorry for digging this out, but it got me thinking...

We updated this file as it's not generated in runtime by the operator, but we're missing other part of the equation - labels in daemonset.
Looking at https://github.com/red-hat-storage/lvm-operator/blob/08b37d20e0711b72dabf884eeb77b228cde7c7fe/controllers/topolvm_node.go#L178-L182 it looks we also need to update labels in our topolvm-node_daemonset.yaml. 
Currently, we create Service, but it has nothing to point to?

pmtk: While we wait for lvms team to provide us yamls, I think it could be faster to ask them to export ` getNodeDaemonSet()` and other functions so we could introduce yet another language to our rebase automation and generate these manifests ourselves :see_no_evil:

copejon: As suggested, I generated the deployment and daemonset manifests (the only 2 not available in the operator bundle) and selectively merged them onto our current manifests.  I ignored upstream values for resource requests, since MicroShift runs with reduced values. Also ignored image refs since inject those those via go template.  Lastly, I left off the snapshotter container spec.

copejon: > While we wait for lvms team to provide us yamls, I think it could be faster to ask them to export getNodeDaemonSet() and other functions so we could introduce yet another language to our rebase automation and generate these manifests ourselves 🙈

I think taking it one step further and adding a "render" command to the operator would be easier to manager from our perspective.

pmtk:
Looking at updated manifests, I think we can drop labels with `lvms-operator` values, but we can do it in another PR

PR #1112, opened 2022-11-11
USHIFT-278: Add specific kubeconfig files per component

oglok:
Missing components are:
  - Cluster Policy Controller
  - Route Controller Manager

Closes [USHIFT-278](https://issues.redhat.com//browse/USHIFT-278)


oglok:
It's still WIP

oglok:
I'm still  seeing some errors like this:

```
route-controller-manager F1111 15:15:06.448116   95450 client_builder_dynamic.go:138] serviceaccounts "ingress-to-route-controller" is forbidden: User "system:route-controller-manager" cannot get resource "serviceaccounts" in API group "" in the namespace "openshift-infra"
```

I've patched all the roles taken from here: 
https://github.com/openshift/cluster-openshift-controller-manager-operator/tree/master/bindata/v3.11.0/openshift-controller-manager

to replace the subject from the ServiceAccount with the User `system:route-controller-manager` that is been used in the certificate for the route controller manager.
It'd be ideal to reuse the same manifests without patching them. Would it be reasonable to use the user subject in OCP (RCM)? @stlaz @atiratree

oglok:
And I'm not completely sure if I need all these new manifests to make it work...

atiratree:
> route-controller-manager F1111 15:15:06.448116   95450 client_builder_dynamic.go:138] serviceaccounts "ingress-to-route-controller" is forbidden: User "system:route-controller-manager" cannot get resource "serviceaccounts" in API group "" in the namespace "openshift-infra"

to fix this,  the following role should be present https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/route-controller-separate-sa-role.yaml

assets/core/ingress-to-route-controller-clusterrolebinding.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: system:openshift:openshift-controller-manager:ingress-to-route-controller
+roleRef:
+  kind: ClusterRole
+  name: system:openshift:openshift-controller-manager:ingress-to-route-controller
+subjects:
+- kind: User
+  namespace: openshift-infra
+  name: system:route-controller-manager
```
atiratree, line 11: this should not be changed from the original and should keep binding to the `ingress-to-route-controller`

atiratree: wrt to the informers this should be handled by `route-controller-informer-clusterrole` for route-controller-manager

assets/core/leader-ingress-to-route-controller-role.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: system:openshift:openshift-controller-manager:leader-locking-ingress-to-route-controller
```
atiratree, line 4: this role + binding should not be needed anymore and will be removed from OCM-o as well

assets/core/route-controller-separate-sa-rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  namespace: openshift-infra
+  name: system:openshift:sa-creating-route-controller-manager
+roleRef:
+  kind: Role
+  name: system:openshift:sa-creating-route-controller-manager
+subjects:
+- kind: User
+  namespace: openshift-route-controller-manager
+  name: system:route-controller-manager
```
atiratree, line 13: also `system:openshift:tokenreview-openshift-route-controller-manager` cluster role + binding should be present

assets/core/ingress-to-route-controller-clusterrolebinding.yaml
oglok: I've deleted this manifest, because it was already applied as part of the openshift-router

assets/core/route-controller-separate-sa-rolebinding.yaml
oglok: Done

scripts/auto-rebase/rebase.sh
```diff
@@ -405,6 +405,15 @@ update_manifests() {
     # 1) Adopt resource manifests
     #    Selectively copy in only those core manifests that MicroShift is already using
     cp "${STAGING_DIR}/cluster-openshift-controller-manager-operator/bindata/v3.11.0/openshift-controller-manager/ns.yaml" "${REPOROOT}"/assets/core/0000_50_cluster-openshift-controller-manager_00_namespace.yaml
+    cp "${STAGING_DIR}/cluster-openshift-controller-manager-operator/bindata/v3.11.0/openshift-controller-manager/route-controller-informer-clusterrole.yaml" "${REPOROOT}"/assets/core/
```
atiratree, line 408: can we have the management of `ingress-to-route-controller-clusterrole.yaml` and `ingress-to-route-controller-clusterrolebinding.yaml` resources in the same place? And move also the initialization from the` openshift-router` to the route-controller-manager since it should be collocated there?

atiratree (commented):
minus the collocation remark, the changes look good

scripts/auto-rebase/rebase.sh
oglok: There is a specific section for components in the rebase script. The openshift-router is handled here:

https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase.sh#L491-L499

As for the initialization, it's also part of the service that starts the components such as `service-ca` deployment, `dns` deployment, etc.

https://github.com/openshift/microshift/blob/main/pkg/components/components.go#L23-L25

If we move the router outside this service, I believe it would look inconsistent.

pkg/cmd/init.go
```diff
@@ -95,19 +95,19 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 				},
 				UserInfo: &user.DefaultInfo{Name: "system:kube-scheduler"},
 			},
-			&cryptomaterial.ClientCertificateSigningRequestInfo{
-				CertificateSigningRequestInfo: cryptomaterial.CertificateSigningRequestInfo{
+			&certchains.ClientCertificateSigningRequestInfo{
+				CSRMeta: certchains.CSRMeta{
 					Name:         "cluster-policy-controller",
-					ValidityDays: cryptomaterial.ClientCertValidityDays,
+					ValidityDays: cryptomaterial.ShortLivedCertificateValidityDays,
```
stlaz, line 101: this seems like a fixup of the first commit (should be squashed)

pkg/cmd/init.go
```diff
@@ -95,19 +95,19 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
...
-			&cryptomaterial.ClientCertificateSigningRequestInfo{
-				CertificateSigningRequestInfo: cryptomaterial.CertificateSigningRequestInfo{
+			&certchains.ClientCertificateSigningRequestInfo{
+				CSRMeta: certchains.CSRMeta{
 					Name:         "route-controller-manager",
-					ValidityDays: cryptomaterial.ClientCertValidityDays,
+					ValidityDays: cryptomaterial.ShortLivedCertificateValidityDays,
 				},
-				UserInfo: &user.DefaultInfo{Name: "system:route-controller-manager"},
+				UserInfo: &user.DefaultInfo{Name: "system:serviceaccount:openshift-route-controller-manager:route-controller-manager-sa"},
```
stlaz, line 110: see `k8s.io/apiserver/pkg/authentication/serviceaccount` for name/groups setting up.

Note that you should add the `system:authenticated` group to the groups `MakeGroupNames()` function returns

scripts/auto-rebase/rebase.sh
atiratree: We should keep the router as is. I am just saying we should move the RBAC files for `ingress-to-route-controller` out. Since these files are only relevant for route-controller-manager and not for openshift-router.

atiratree: also, it would be good to have a comment above all these manifests from OCM-o that they belong to route-controller-manager

oglok: Ok, understood! I'll move them together then. Thakns Filip!

assets/core/route-controller-informer-clusterrole.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
atiratree, line 1: shouldn't these files be under its own subfolder like the other components?

atiratree: ^ this is just cosmetic, we can return to this later if needed

oglok: Let's do a follow up PR for assets arrangement.

PR #1114, opened 2022-11-11
WIP: Configure route host assignment admission plugin.

benluddy:


benluddy:
Rebase PR with dependencies opened: https://github.com/openshift/microshift/pull/1129

assets/crd/route.crd.yaml
```diff
@@ -12,381 +12,287 @@ spec:
...
-                      description: name of the service/target that is being referred
-                        to. e.g. name of the service
+                      description: name of the service/target that is being referred to. e.g. name of the service
                       minLength: 1
                       type: string
                     weight:
-                      description: weight as an integer between 0 and 256, default
-                        100, that specifies the target's relative weight against other
-                        target reference objects. 0 suppresses requests to this backend.
+                      default: 100
```
benluddy, line 222: The schema diff looks scary due to https://github.com/openshift/api/commit/c53e08e0561c14644db0f22b5b4e3574ea746e2b, but this default value for "weight" is the thing that is relevant to route/v1 API compatibility.

PR #1122, opened 2022-11-16
WIP: disable ovs gateway

zshi-redhat:


mangelajo:
How does this work?

zshi-redhat:
> How does this work?

@mangelajo The same steps as how we run microshift (rpm or iso), the PR skips the configure-ovs.sh script and the cluster shall be brought up correctly (all the traffic and features shall work as usual). Once the cluster is up, there will only be one ovs bridge `br-int` in the output of cmd `ovs-vsctl show` , uplink physical interface is left untouched.

PR #1123, opened 2022-11-16
Update GO_LD_FLAGS

fzdarsky:
Updates the GO_LD_FLAGS in the Makefile and extends the rebase script so it keeps those flags up-to-date.

Closes [USHIFT-604](https://issues.redhat.com//browse/USHIFT-604)


scripts/auto-rebase/rebase.sh
```diff
@@ -586,6 +586,37 @@ update_manifests() {
...
+    KUBE_GIT_VERSION=$(git describe --tags --match='v*' --abbrev=0)
+    if [[ "${KUBE_GIT_VERSION}" =~ ^v([0-9]+)\.([0-9]+)(\.[0-9]+)?([-].*)?([+].*)?$ ]]; then
+        KUBE_GIT_MAJOR=${BASH_REMATCH[1]}
+        KUBE_GIT_MINOR=${BASH_REMATCH[2]}
+        if [[ -n "${BASH_REMATCH[4]}" ]]; then
+          KUBE_GIT_MINOR+="+"
+        fi
+    fi
+    KUBE_GIT_COMMIT=$(git rev-parse --short=7 HEAD)
+    sed -i -E "s|k8s\.io/component-base/version\.gitMajor=[0-9]+|k8s.io/component-base/version.gitMajor=${KUBE_GIT_MAJOR}|g" "${REPOROOT}/Makefile"
```
dhellmann, line 607: This looks like it does the job, but it also looks pretty complicated to maintain over time.

What do you think about generating a completely separate file with some variables in a format that make can read and use in the build command line? That way the actual Makefile never changes, just the variable file, and we don't have to edit it we sed, we can just regenerate it as part of the rebase.

dhellmann (commented):
/lgtm

scripts/auto-rebase/rebase.sh
fzdarsky: k8s itself uses `kube::version::save_version_vars` and `kube::version::ldflags` functions in `hack/lib/version.sh` to generate such a file resp. immediately the LD_FLAGS for those components. The version they produce `git describe`s without abbreviating, so looks like `v1.25.0-2573+f33d98e342f30f` as the o/k repo has many changes over k/k's tag.

I could write either output to a file and load into the Makefile at runtime, though I'd need to find out a way to do that without making this even more complicated than it already is.

dhellmann:
/lgtm

This is much easier for me to follow, thanks.

PR #1125, opened 2022-11-16
USHIFT-434: Set storageCapacity explicitly in csi manifest

copejon:


ggiguash:
/retitle USHIFT-434: Set storageCapacity explicitly in csi manifest

PR #1130, opened 2022-11-21
USHIFT-463: Update requests for topolvm to reflect measured usage

jerpeter1:
https://docs.google.com/spreadsheets/d/1eLBkG4HhlKlxRlB9H8kjuXs5F3zzVOPfkOOhFIOrfzU/edit#gid=0

### DIFF
--- a/assets/components/odf-lvm/topolvm-controller_deployment.yaml
+++ b/assets/components/odf-lvm/topolvm-controller_deployment.yaml
@@ -54,8 +54,8 @@ spec:
           timeoutSeconds: 1
         resources:
           requests:
-            cpu: 250m
-            memory: 250Mi
+            cpu: 2m
+            memory: 31Mi
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
         volumeMounts:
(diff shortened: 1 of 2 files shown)


assets/components/odf-lvm/topolvm-controller_deployment.yaml
```diff
@@ -54,8 +54,8 @@ spec:
           timeoutSeconds: 1
         resources:
           requests:
-            cpu: 250m
-            memory: 250Mi
+            cpu: 2m
+            memory: 31Mi
```
ggiguash, line 58: Why are we so "precise" in resource requests? Should we not give it a bit of an extra by rounding the numbers up?

fzdarsky: AFAICT the numbers _are_ already rounded up. While safety margins make sense, let’s also consider these are minimum _requests_ used for scheduling and containers can burst higher (as no limit is set) if resources are available.

We may consider rounding to numbers that “feel less precise” (31Mi —> 32MB, 9Mi —> 10Mi), but then let’s work with those for now.

We also want to run those by the ODF team, of course.

ggiguash: I had a conversation with @jerpeter1 about this and it looks like we're all on the same page. 
Let's try working with these values for now and follow-up with the ODF team separately.

PR #1136, opened 2022-11-22
USHIFT-623: Fix CSI and OpenSSL images to be consistent

ggiguash:
* CSI images should come from the quay.io/openshift-release-dev repo
* OpenSSL image manifest should be used so that it works on both platforms

Closes [USHIFT-623](https://issues.redhat.com//browse/USHIFT-623)


mangelajo (commented):
/lgtm

PR #1140, opened 2022-11-24
USHIFT-626: Install sos utility with MicroShift and document its usage

ggiguash:
Ensure the `sos` presence on MicroShift hosts.

Also created [USHIFT-628](https://issues.redhat.com//browse/USHIFT-628) task to follow-up on embedding MicroShift-specific information in the reports generated by the `sos` utility.

Closes [USHIFT-626](https://issues.redhat.com//browse/USHIFT-626)

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -62,6 +62,7 @@ Requires: iptables
 Requires: microshift-selinux
 Requires: microshift-networking
 Requires: conntrack-tools
+Requires: sos
 
 %{?systemd_requires}
 
@@ -268,6 +269,9 @@ systemctl enable --now --quiet openvswitch || true
 # Use Git command to generate the log and replace the VERSION string
 # LANG=C git log --date="format:%a %b %d %Y" --pretty="tformat:* %cd %an <%ae> VERSION%n- %s%n" microshift.spec
 %changelog
+* Fri Nov 25 2022 Gregory Giguashvili <ggiguash@redhat.com> 4.12.0
+- Install sos utility with MicroShift and document its usage
+
 * Mon Oct 24 2022 Zenghui Shi <zshi@redhat.com> 4.12.0
 - Add arch specific crio conf
 
--- a/docs/debugging_tips.md
+++ b/docs/debugging_tips.md
@@ -27,3 +27,33 @@ metadata:
   name: microshift-version
   namespace: kube-public
 ```
+
+## Generating an SOS Report
+
+The MicroShift RPMs have an explicit dependency on the `sos` utility allowing to collect
+configuration, diagnostic, and troubleshooting data to be provided to Red Hat Technical Support.
+
+> See [Generating sos reports for technical support](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/generating_sos_reports_for_technical_support/index) for more information on the `sos` utility usage.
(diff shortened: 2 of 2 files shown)


packaging/rpm/microshift.spec
```diff
@@ -62,6 +62,7 @@ Requires: iptables
 Requires: microshift-selinux
 Requires: microshift-networking
 Requires: conntrack-tools
+Requires: sos
```
mangelajo, line 65: Do we have numbers on the size of sos + dependencies? (On a bare try here on my RHEL it showed 3.2MB which is probably fine everywhere and we get the guarantee if having sos always)

If we wanted to make this "optional" install in some cases we can use a weak dependency:
```
Recommends: sos
```

This way sos could be avoided during installation, but I suspect the specific flag to avoid weak dependencies is not even available in the image builder.




I lean more towards what you just proposed, but wanted to share the thought.

mangelajo (commented):
/approve

packaging/rpm/microshift.spec
fzdarsky: IMO, the sos reporting tool is something that belongs into the base R4E installation, not into MicroShift. Unlike in OCP, we do not "own" the OS. *If* we want to make it a dependency, we should definitely make it a weak dependency, so users can opt out of it.

ggiguash: The recommended dependency works nicely - thank you for the suggestion.
I'm making sure that RHEL for Edge images pull the recommended dependency by default before uploading the fixed code for review.

ggiguash: I checked with the Image Builder team and there's no option to automatically pull in weak RPM dependencies. So, unless we create a strong dependency, we cannot guarantee sos utility presence.

@fzdarsky do you thik it's a good enough reason to have a strong dependency on sos?

mangelajo: No, a weak dependency would be pulled in always, but you can opt-out.

With the "Recommends" users would be getting it in image-builder. And maybe at some point, we could get them to allow opt-outs, that's probably something that would need implementation on ostree.

In dnf you would do:
`dnf --setopt=install_weak_deps=False` which I only noticed in the container world where you want a very thin container most of the time.

ggiguash: I tested the "Recommends" dependency and Image Builder does not pull the "sos" package when weak dependency is specified. I double checked with the Image Builder team and they confirmed this is the current implementation - the only way is to either add strong dependency, or specify the package in the blueprint.

mangelajo: Oh, maybe rpmostree interprets --setopt=install_weak_deps=False by default.

Thanks for testing it in image builder @ggiguash

mangelajo:
I would propose the following strategy:

1) Having this as a hard requirement for TechPreview/Limited availability, specially since most of the issues could happen at the initial phases of the project.
2) Later on, decide based on customer feedback if the dependency needs to be removed. It seems to be less than 5MB.

ggiguash:
Long term, we should also work with the RHEL team to include `sos` utility in the base image. Then, the need for the explicit dependency should become obsolete.

fzdarsky:
We've agreed to add a strong dep for now, which we'll remove when sos has been added to the base OS.

/lgtm

PR #1144, opened 2022-11-25
USHIFT-631: Rearrangement of assets directory

oglok:

Closes [USHIFT-631](https://issues.redhat.com//browse/USHIFT-631)


pkg/controllers/openshift-route-controller-manager.go
```diff
@@ -114,37 +114,37 @@ func (s *OCPRouteControllerManager) Run(ctx context.Context, ready chan<- struct
 		klog.Fatalf("failed to apply openshift namespaces %v", err)
 	}
 	if err := assets.ApplyClusterRoles([]string{
-		"core/ingress-to-route-controller-clusterrole.yaml",
-		"core/route-controller-informer-clusterrole.yaml",
-		"core/route-controller-tokenreview-clusterrole.yaml",
+		"rbac/ingress-to-route-controller-clusterrole.yaml",
```
atiratree, line 117: since we do not always have a prefix `route-controller`, I would be in favor of having even more granular approach like `rbac/route-controller-manager/` to distinguish the files. But we are not doing this in our operators and is probably okay in a current state with not as many RBACs

scripts/auto-rebase/rebase.sh
```diff
@@ -405,39 +405,39 @@ update_manifests() {
...
+    cp "${STAGING_DIR}/cluster-kube-controller-manager-operator/bindata/assets/kube-controller-manager/podsecurity-admission-label-syncer-controller-clusterrolebinding.yaml" "${REPOROOT}"/assets/rbac
     #    Selectively copy in only those CRD manifests that MicroShift is already using
     cp "${REPOROOT}"/vendor/github.com/openshift/api/route/v1/route.crd.yaml "${REPOROOT}"/assets/crd
     cp "${STAGING_DIR}/release-manifests/0000_03_security-openshift_01_scc.crd.yaml" "${REPOROOT}"/assets/crd
     cp "${STAGING_DIR}/release-manifests/0000_03_securityinternal-openshift_02_rangeallocation.crd.yaml" "${REPOROOT}"/assets/crd
     # The following manifests are just MicroShift specific and are not present in any other OpenShift repo.
-    # - assets/crd/authorizationv1-local-apiservice.yaml (local API service for authorization API group, needed if OpenShift API server is not present)
-    # - assets/crd/securityv1-local-apiservice.yaml (local API service for security API group, needed if OpenShift API server is not present)
+    # - assets/apiservice/securityv1-local-apiservice.yaml (local API service for security API group, needed if OpenShift API server is not present)
```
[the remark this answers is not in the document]
oglok: It was removed because it caused garbage collection issues when no available API from that API group was present.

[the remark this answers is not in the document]
atiratree: hm interesting, is there an issue posted somewhere regarding this GC behaviour? This should not happen and might be fixable upstream.

scripts/auto-rebase/rebase.sh
```diff
@@ -405,49 +405,49 @@ update_manifests() {
     # 1) Adopt resource manifests
     #    Selectively copy in only those core manifests that MicroShift is already using
     cp "${STAGING_DIR}/cluster-openshift-controller-manager-operator/bindata/v3.11.0/openshift-controller-manager/ns.yaml" "${REPOROOT}"/assets/core/0000_50_cluster-openshift-controller-manager_00_namespace.yaml
+    cp "${STAGING_DIR}/cluster-openshift-controller-manager-operator/bindata/v3.11.0/openshift-controller-manager/route-controller-ns.yaml" "${REPOROOT}"/assets/core/0000_50_cluster-openshift-route-controller-manager_00_namespace.yaml
```
mangelajo, line 408: what is the diffeerence between those two and why wouldn't  all this go in `"${REPOROOT}"/assets/controllers/route-controller-manager/` as the following ones?

mangelajo: in here I see we are using actually route-controller-manager dir ? 

https://github.com/openshift/microshift/pull/1144/files#diff-5beee82adbac20a721984bd736ca88df45a3934f5225bd9e3e61c37881992e9fR112

oglok: good catch! I'll force update the PR.

The difference between those two files are just different ns: `openshift-controller-manager` and `openshift-route-controller-manager`.

PR #1145, opened 2022-11-25
USHIFT-574: Add subjectAltNames and rework external kubeconfig

pacevedom:

Closes #<Issue Number>


pacevedom:
/hold
Adding unit tests

ggiguash:
@pacevedom , please also update https://github.com/openshift/microshift/blob/main/docs/howto_config.md doc

pkg/config/config.go
```diff
@@ -92,6 +97,25 @@ func (cfg *MicroshiftConfig) KubeConfigPath(id KubeConfigID) string {
...
+	cmd := exec.Command("/bin/hostname", "-A")
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		return nil, fmt.Errorf("Error when executing 'hostname -A': %v", err)
+	}
+	outString := out.String()
+	outString = strings.Trim(outString[:len(outString)-1], " ")
+	names := strings.Split(outString, " ")
```
mangelajo, line 114:
```suggestion
	// remove any duplicate names
	names := strings.Split(outString, " ")
```

I was scratching my head for 10 seconds of why list->set->list

pkg/config/config.go
```diff
@@ -249,10 +281,50 @@ func (c *MicroshiftConfig) ReadAndValidate(configFile string, flags *pflag.FlagS
...
+		if err != nil {
+			return fmt.Errorf("failed to parse cluster URL: %v", err)
+		}
+		if u.Hostname() == "localhost" || u.Hostname() == "127.0.0.1" {
+			if stringSliceContains(c.SubjectAltNames, "localhost", "127.0.0.1") {
+				return fmt.Errorf("subjectAltNames must not contain localhost, 127.0.0.1")
+			}
+		} else {
+			if stringSliceContains(c.SubjectAltNames, c.NodeIP) {
+				return fmt.Errorf("subjectAltNames must not contain node IP")
```
mangelajo, line 299:
```suggestion
				return fmt.Errorf("subjectAltNames must not contain node IP")
```
 
We should probably document why this should not be done (the SNI/client-go issues)

pacevedom: Good point. Let me add the comment

pkg/config/config.go
pacevedom: That's it. The intention is to avoid duplicates in the SAN which would be confusing (and useless).

pkg/cmd/init.go
```diff
@@ -346,6 +345,33 @@ func initKubeconfigs(
...
+	}
+	apiServerPort, err := cfg.Cluster.ApiServerPort()
+	if err != nil {
+		return fmt.Errorf("failed to get apiserver port: %v", err)
+	}
+
+	// If using localhost this is part of the default URL and it needs external access.
+	// Generate all other kubeconfigs, one per name.
+	if u.Hostname() == "localhost" || u.Hostname() == "127.0.0.1" {
+		for _, name := range append(cfg.SubjectAltNames, cfg.NodeName, "localhost") {
```
mangelajo, line 361: Ok, to fully end the issue we also need another patch to split NodeName & HostName

But that makes me think, what will happen with URL, which contains a hostname.

Should we remove the URL parameter?, and construct URL From protocol + hostname + port ?

mangelajo: Or should URL be something different that will simply differ across kubeconfig files?

This is the only part were I believe we need more consistence and agreement, but it's not specific to this patch.

pacevedom: > Ok, to fully end the issue we also need another patch to split NodeName & HostName
> 
> But that makes me think, what will happen with URL, which contains a hostname.
> 
> Should we remove the URL parameter?, and construct URL From protocol + hostname + port ?

It is going to have a different [name](https://github.com/openshift/microshift/pull/1030/files#diff-e279d89e0daa9e6b083e46b5c0f2b9d18015b12fb264fd7caf6fb3a5d4c3ee52R37) but it serves the same purpose, as a URL. I have to say I see the URL more as a hostname+port thing (or even only port, having the hostname/nodename and SAN already there). Is there any use case where the apiserver is going to require a path in that URL?



> Or should URL be something different that will simply differ across kubeconfig files?
> 
> This is the only part were I believe we need more consistence and agreement, but it's not specific to this patch.

Your first sentence is what this PR is doing, generate a set of kubeconfigs only with different `server` value.

mangelajo: Sorry for my previous comments; I was trying to grasp the whole picture.

I believe the proposed approach makes sense. The Hostname is derived from the URL instead of NodeName, so we already have the two different parameters we wanted, and there is no need for more config parameters.

pkg/cmd/init.go
```diff
@@ -346,6 +345,33 @@ func initKubeconfigs(
...
+		return fmt.Errorf("failed to parse cluster URL: %v", err)
+	}
+	apiServerPort, err := cfg.Cluster.ApiServerPort()
+	if err != nil {
+		return fmt.Errorf("failed to get apiserver port: %v", err)
+	}
+
+	// If using localhost this is part of the default URL and it needs external access.
+	// Generate all other kubeconfigs, one per name.
+	if u.Hostname() == "localhost" || u.Hostname() == "127.0.0.1" {
```
mangelajo, line 360: Hmm, I propose that we generate the alt kubeconfigs always, regardless of the URL/host we configure.
Unless I am missing something and this is also part of the SNI/TLS issues we have and it cannot be done.

mangelajo:
/approve

I also ran some manual testing to test the different kubeconfig based on the available alt names from the hosts and it works:

```
[root@rhel9 kubeadmin]# export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/rhel9.local/kubeconfig
[root@rhel9 kubeadmin]# oc get nodes -o wide
NAME          STATUS   ROLES                         AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                              KERNEL-VERSION                 CONTAINER-RUNTIME
rhel9.local   Ready    control-plane,master,worker   4m20s   v1.25.0   192.168.1.173   <none>        Red Hat Enterprise Linux 9.1 (Plow)   5.14.0-162.6.1.el9_1.aarch64   cri-o://1.25.1-2.rhaos4.12.gitafa0c57.el8

[root@rhel9 kubeadmin]# export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/rhel9.casa/kubeconfig
[root@rhel9 kubeadmin]# oc get nodes -o wide
NAME          STATUS   ROLES                         AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                              KERNEL-VERSION                 CONTAINER-RUNTIME
rhel9.local   Ready    control-plane,master,worker   4m29s   v1.25.0   192.168.1.173   <none>        Red Hat Enterprise Linux 9.1 (Plow)   5.14.0-162.6.1.el9_1.aarch64   cri-o://1.25.1-2.rhaos4.12.gitafa0c57.el8

[root@rhel9 kubeadmin]# export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/localhost/kubeconfig
[root@rhel9 kubeadmin]# oc get nodes -o wide
NAME          STATUS   ROLES                         AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                              KERNEL-VERSION                 CONTAINER-RUNTIME
rhel9.local   Ready    control-plane,master,worker   4m40s   v1.25.0   192.168.1.173   <none>        Red Hat Enterprise Linux 9.1 (Plow)   5.14.0-162.6.1.el9_1.aarch64   cri-o://1.25.1-2.rhaos4.12.gitafa0c57.el8

[root@rhel9 kubeadmin]# export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig
[root@rhel9 kubeadmin]# oc get nodes -o wide
NAME          STATUS   ROLES                         AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                              KERNEL-VERSION                 CONTAINER-RUNTIME
rhel9.local   Ready    control-plane,master,worker   4m47s   v1.25.0   192.168.1.173   <none>        Red Hat Enterprise Linux 9.1 (Plow)   5.14.0-162.6.1.el9_1.aarch64   cri-o://1.25.1-2.rhaos4.12.gitafa0c57.el8

[root@rhel9 kubeadmin]# export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/test-microshift.local/kubeconfig
[root@rhel9 kubeadmin]# oc get pods -A
NAMESPACE                  NAME                                  READY   STATUS    RESTARTS       AGE
openshift-dns              dns-default-6hdvh                     2/2     Running   0              11m
openshift-dns              node-resolver-rc6qr                   1/1     Running   0              12m
openshift-ingress          router-default-59d75d5779-s2dsh       1/1     Running   0              12m
openshift-ovn-kubernetes   ovnkube-master-2tfzj                  4/4     Running   0              12m
openshift-ovn-kubernetes   ovnkube-node-4tpvp                    1/1     Running   0              12m
openshift-service-ca       service-ca-68d88b596-8fj57            1/1     Running   0              12m
openshift-storage          topolvm-controller-5797766b59-pjfm6   4/4     Running   2 (3m4s ago)   12m
openshift-storage          topolvm-node-pj7qp                    4/4     Running   0              11m
```

pkg/cmd/init.go
pacevedom: You are right. I was assuming that a non-default configuration should dictate a single kubeconfig and disregard the rest of the alternative names in the `server` field, but there might be use cases where you need to refer to the same cluster with different names coming from different DNS servers.

pacevedom: Done

mangelajo:
Ok this looks great to me. 

/lgtm

PR #1149, opened 2022-11-28
USHIFT-573 avoid NodeName changes in running systems

mangelajo:
NodeName changes are dangerous to running systems (attached storage, running pods, etc., need to be relocated).
This change persists the NodeName in /var/lib/microshift/nodename and makes sure it's not modified across runs.
If the NodeName was auto-generated from the hostname it will write a Warning and continue with the old NodeName.
If the NodeName is explicitly configured, it will fail with an error.

[Closes [USHIFT-573](https://issues.redhat.com//browse/USHIFT-573)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -148,6 +148,53 @@ func NewMicroshiftConfig() *MicroshiftConfig {
 	}
 }
 
+// Determine if the config file specified a NodeName (by default it's assigned the hostname)
+func (c *MicroshiftConfig) isDefaultNodeName() bool {
+	hostname, err := os.Hostname()
+	if err != nil {
+		klog.Fatalf("Failed to get hostname %v", err)
+	}
+	return c.NodeName == hostname
+}
+
+// Read or set the NodeName that will be used for this MicroShift instance
+func (c *MicroshiftConfig) establishNodeName() (string, error) {
+	filePath := filepath.Join(GetDataDir(), ".nodename")
+	contents, err := os.ReadFile(filePath)
+	if os.IsNotExist(err) {
+		// ensure that dataDir exists
+		os.MkdirAll(GetDataDir(), 0700)
+		if err := os.WriteFile(filePath, []byte(c.NodeName), 0444); err != nil {
+			return "", fmt.Errorf("failed to write nodename file %q: %v", filePath, err)
+		}
+		return c.NodeName, nil
+	} else if err != nil {
+		return "", err
+	}
+	return string(contents), nil
+}
+
+// Validate the NodeName to be used for this MicroShift instances
+func (c *MicroshiftConfig) validateNodeName(isDefaultNodeName bool) error {
+	establishedNodeName, err := c.establishNodeName()
+	if err != nil {
+		return fmt.Errorf("failed to establish NodeName: %v", err)
+	}
+
+	if establishedNodeName != c.NodeName {
+		if !isDefaultNodeName {
+			return fmt.Errorf("configured NodeName %q does not match previous NodeName %q , NodeName cannot be changed for a device once established",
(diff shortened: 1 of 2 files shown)


pkg/cmd/run.go
```diff
@@ -81,6 +81,12 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 
 	os.MkdirAll(microshiftDataDir, 0700)
 
+	// Validate NodeName in config file, node-name should not be changed for an already
+	// initialized MicroShift instance (see https://issues.redhat.com/browse/USHIFT-573)
```
pmtk, line 85: I'm not sure if we want to have jira urls in code (was once told to remove them) - we can check the git blame, get PR's name and we'll have jira ticket

pmtk (commented):
2 small things, otherwise looks good to me but will for other's review

pkg/config/config.go
```diff
@@ -118,6 +118,56 @@ func NewMicroshiftConfig() *MicroshiftConfig {
...
+	hostname, err := os.Hostname()
+	if err != nil {
+		klog.Fatalf("Failed to get hostname %v", err)
+	}
+	return c.NodeName == hostname
+}
+
+// Read or set the NodeName that will be used for this MicroShift instance
+func (c *MicroshiftConfig) establishNodeName() (string, error) {
+	filePath := filepath.Join(GetDataDir(), "nodename")
```
pacevedom, line 132: Hide this file, maybe? Users should not edit it.
It would be possible to stop the service, edit the file, change the config (in order to not use the default), start the service and have a broken node name, right?

pkg/config/config.go
```diff
@@ -118,6 +118,56 @@ func NewMicroshiftConfig() *MicroshiftConfig {
...
+	} else if err != nil {
+		return "", err
+	}
+	return string(contents), nil
+}
+
+const (
+	IS_DEFAULT_NODENAME     = true
+	IS_NOT_DEFAULT_NODENAME = false
+)
```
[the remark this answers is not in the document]
mangelajo: yes, will do. It was added mainly for testability.

pkg/cmd/run.go
mangelajo: Ack.

pkg/config/config.go
mangelajo: It's written with read-only mode for all (root/users)

May be we can add a first line as comment explaining that change is not supported, and that changing that will lead to problems.

Hiding can lead to the same issues, as soon as they find where to change it.

pacevedom: You are right, missed the read only. Maybe add this to docs instead of the file? You would need to parse it.

pkg/config/config.go
```diff
@@ -118,6 +118,56 @@ func NewMicroshiftConfig() *MicroshiftConfig {
...
+
+// Validate the NodeName to be used for this MicroShift instances
+func (c *MicroshiftConfig) ValidateNodeName(isDefaultNodeName bool) error {
```
fzdarsky, line 151: No need to have this method exported if only called from config package.
Unless we have a reason ever to call this method without `IsDefaultNodeName()`, I'd simply initialise that var from within the method.

pkg/config/config.go
mangelajo: Right, I made it hidden, and I will explain in the docs that NodeName cannot be changed once established.

pkg/config/config.go
mangelajo: no, that is for testability of the function, but I will make it private.

mangelajo: python monkey patching for testing made things easier, but here we cannot do that.

PR #1151, opened 2022-11-29
USHIFT-638: Change the default config.yaml file name to allow its overwrite

ggiguash:
Closes USHIFT-638


mangelajo:
/hold

There is a more standard way to handle this, I am re-reading the rpm/spec docs, brb.

packaging/rpm/microshift.spec
```diff
@@ -173,7 +173,7 @@ install -d -m755 %{buildroot}/%{_unitdir}
 install -p -m644 packaging/systemd/microshift.service %{buildroot}%{_unitdir}/microshift.service
 
 install -d -m755 %{buildroot}/%{_sysconfdir}/microshift
-install -p -m644 packaging/microshift/config.yaml %{buildroot}%{_sysconfdir}/microshift/config.yaml
+install -p -m644 packaging/microshift/config.yaml %{buildroot}%{_sysconfdir}/microshift/config.yaml.default
```
mangelajo, line 176:
```suggestion
install -p -m644 packaging/microshift/config.yaml %{buildroot}%{_sysconfdir}/microshift/config.yaml
```

packaging/rpm/microshift.spec
```diff
@@ -242,7 +242,7 @@ systemctl enable --now --quiet openvswitch || true
 %{_bindir}/cleanup-all-microshift-data
 %{_unitdir}/microshift.service
 %{_sysconfdir}/crio/crio.conf.d/microshift.conf
-%{_sysconfdir}/microshift/config.yaml
+%{_sysconfdir}/microshift/config.yaml.default
```
mangelajo, line 245:
```suggestion
%config(noreplace) %{_sysconfdir}/microshift/config.yaml
```

mangelajo: see:

https://docs.fedoraproject.org/en-US/packaging-guidelines/#_configuration_files
https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html

mangelajo: When we make changes to the config file /etc/microshift/config.yaml.rpmnew

Will be written into disk for user reference, but the user changes will be kept in disk

mangelajo (changes requested):

ggiguash:
Thank you @mangelajo. Let me test these changes.

ggiguash:
@mangelajo, I tested the changes and I confirm that the installation is behaving as expected. When `/etc/microshift/config.yaml` file exists, it is not overwritten and .rpmnew file is created instead. 
```
$ ls -l
total 8
-rw-r--r--. 1 root root 587 Nov 29 13:14 config.yaml
-rw-r--r--. 1 root root 588 Nov 28 18:11 config.yaml.rpmnew
```

ggiguash:
@mangelajo I got more information about the flow we're trying to solve. It's not possible to copy single files to a RH4E image, so one would need to use RPMs to install config.yaml to ostree.

This means that we must change our `config.yaml` file name to avoid conflicts between 2 different RPMs, not within the same RPM.

Putting this on hold until you comment further.
/hold

mangelajo:
Ah ok, because they are trying to package the config.yaml inside an RPM, 

That makes it more difficult, yes. You cannot install two rpms containing the same files. If we need to support that, because I guess image builder won't let us... we need to follow your initial approach.

Btw, rpm-ostree allows to write custom files on the file system when building an ostree image (as part of the treefile json) but probably this is not exposed in image builder?.

mangelajo:
> Ah ok, because they are trying to package the config.yaml inside an RPM,
> 
> That makes it more difficult, yes. You cannot install two rpms containing the same files. If we need to support that, because I guess image builder won't let us... we need to follow your initial approach.
> 
> Btw, rpm-ostree allows to write custom files on the file system when building an ostree image (as part of the treefile json) but probably this is not exposed in image builder?.

it's not.. https://www.osbuild.org/guides/blueprint-reference/blueprint-reference.html

I guess if we want to support them packaging the RPM config while there is no other option we will need to use your initial approach. Sorry for the back an forth, and thanks for all the double-checking @ggiguash

mangelajo (commented):
/lgtm

PR #1152, opened 2022-11-29
USHIFT-194: Add kubeconfig to Scheduler's authz and authn

pmtk:


oglok:
Good catch Patryk!

/lgtm

PR #1154, opened 2022-11-29
fix ApplyConfigMapWithData

pmtk:


dhellmann:
What sort of test should we have in place to have caught this before?

/lgtm

pmtk:
> What sort of test should we have in place to have caught this before?

Linter could be a start (unused parameter). For e2e: routing for https service (see USHIFT-541)

PR #1155, opened 2022-11-30
pass log level to KCM

atiratree:
needed for testing the KCM https://issues.redhat.com/browse/WRKLDS-604

It seems to me that log level is passed correctly to the other controllers

### DIFF
--- a/pkg/controllers/kube-controller-manager.go
+++ b/pkg/controllers/kube-controller-manager.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strconv"
 
 	embedded "github.com/openshift/microshift/assets"
 	"github.com/openshift/microshift/pkg/assets"
@@ -92,6 +93,7 @@ func configure(cfg *config.MicroshiftConfig) (args []string, applyFn func() erro
 			"use-service-account-credentials":  {"true"},
 			"cluster-signing-cert-file":        {clusterSigningCert},
 			"cluster-signing-key-file":         {clusterSigningKey},
+			"v":                                {strconv.Itoa(cfg.LogVLevel)},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/kube-controller-manager.go
```diff
@@ -96,6 +96,8 @@ func configure(cfg *config.MicroshiftConfig) (args []string, applyFn func() erro
 	}
 
 	args, err = mergeAndConvertToArgs(overrides)
+	args = append(args, fmt.Sprintf("-v=%d", cfg.LogVLevel))
```
pacevedom, line 99: Does including
```
	"v": {cfg.LogVLevel},
```
To `ExtendedArguments` in `overrides` do the same? You wouldnt need this line.

atiratree: @pacevedom  Yup, looks better to use the same approach for both short and long arguments. I have modified `GetKubeControllerManagerArgs` to support that.

pkg/controllers/kube-controller-manager.go
```diff
@@ -171,7 +173,11 @@ func GetKubeControllerManagerArgs(config map[string]interface{}) []string {
 	args := []string{}
 	for key, value := range extendedArguments.(map[string]interface{}) {
 		for _, arrayValue := range value.([]interface{}) {
-			args = append(args, fmt.Sprintf("--%s=%s", key, arrayValue.(string)))
+			if len(key) == 1 {
+				args = append(args, fmt.Sprintf("-%s=%s", key, arrayValue.(string)))
```
pacevedom, line 177: This should work with [double dash](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/#options) too.

atiratree: yes, you are right. But for example `-h` doesn't, so it shouldn't hurt anything to keep this here.

atiratree: also a minor plus for consistency is that we are using `-v` in KCM operator as well when generating the args.

Nevertheless, I don't have a strong preference if you want to remove this.

oglok: log level verbosity flag in KCM supports simple or double dash. So this sounds good to me.

PR #1157, opened 2022-11-30
USHIFT-645: Pass version from spec to Makefile

pmtk:
As a result, `%{version}` given to .spec will be printed out when running `microshift version` command and will match version part of RPM filename.

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -134,9 +134,9 @@ GOARCH=amd64
 
 # if we have git commit/tag/state to be embedded in the binary pass it down to the makefile
 %if 0%{?embedded_git_commit:1}
-make _build_local GOOS=${GOOS} GOARCH=${GOARCH} EMBEDDED_GIT_COMMIT=%{embedded_git_commit} EMBEDDED_GIT_TAG=%{embedded_git_tag} EMBEDDED_GIT_TREE_STATE=%{embedded_git_tree_state}
+make _build_local GOOS=${GOOS} GOARCH=${GOARCH} EMBEDDED_GIT_COMMIT=%{embedded_git_commit} EMBEDDED_GIT_TAG=%{embedded_git_tag} EMBEDDED_GIT_TREE_STATE=%{embedded_git_tree_state} MICROSHIFT_VERSION=%{version}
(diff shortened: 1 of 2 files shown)


packaging/rpm/microshift.spec
```diff
@@ -134,9 +134,9 @@ GOARCH=amd64
 
 # if we have git commit/tag/state to be embedded in the binary pass it down to the makefile
 %if 0%{?embedded_git_commit:1}
-make _build_local GOOS=${GOOS} GOARCH=${GOARCH} EMBEDDED_GIT_COMMIT=%{embedded_git_commit} EMBEDDED_GIT_TAG=%{embedded_git_tag} EMBEDDED_GIT_TREE_STATE=%{embedded_git_tree_state}
+make _build_local GOOS=${GOOS} GOARCH=${GOARCH} EMBEDDED_GIT_COMMIT=%{embedded_git_commit} EMBEDDED_GIT_TAG=%{embedded_git_tag} EMBEDDED_GIT_TREE_STATE=%{embedded_git_tree_state} MICROSHIFT_VERSION=%{version}
 %else
-make _build_local GOOS=${GOOS} GOARCH=${GOARCH}
+make _build_local GOOS=${GOOS} GOARCH=${GOARCH} MICROSHIFT_VERSION=%{version}
```
mangelajo, line 139: I believe we should also add %{release} and %{dist}

the release is an indicator of changes in the version too, and we could have builds for different distros (EL8/EL9)

mangelajo: I am trying the PR to see how it looks like now...

mangelajo: Nah, ignore my comment, I believe what you proposed makes sense, the since we use the same variable from git on a local build... we need to keep it coherent.

/lgtm

pmtk: Well... tbh that MICROSHIFT_VERSION is ends up different when building locally (`commit_timestamp-commit_sha[-dirty]`) and from ART (eg `4.12-ec2` or similar).
But according to the RPM guidelines the release is more for package maintainer so we don't use is (even though it includes this repo's commit SHA)

PR #1160, opened 2022-12-01
[BUG] USHIFT-657 MicroShift does not find lvmd config if no microshift config file exists.

copejon:

Closes # [USHIFT-657](https://issues.redhat.com//browse/USHIFT-657)

This PR removes the assumption that the microshift config and the lvmd config are always co-located.  This created a situation where, if a user does not create a microshift config, the search for the lvmd config would short circuit.  This isn't a safe assumption because the microshift default values, used when a user hasn't created a config file, are generally safe for common use cases.  The lvmd config however must usually be defined to describe the host's storage environment, which is probably less uniform with arbitrary volume group names.

This became and issue when the name of the microshift default/example config was changed in the RPM spec, which provided an anchor for the lvmd config logic to rely on when no user config was specified.

### DIFF
--- a/pkg/components/storage.go
+++ b/pkg/components/storage.go
@@ -13,19 +13,13 @@ import (
 	"github.com/openshift/microshift/pkg/config/lvmd"
 )
 
-// getCSIPluginConfig searches for a user-defined lvmd configuration file in predefined locations are returns the first
-// found, else a default-value lvmd config. Returns an error if the file exists and cannot be read, or exists and cannot
-// be deserialized (indicating a malformed config).  User's home directory is checked first.  If not found, global
-// MicroShift config directory is checked.
+// getCSIPluginConfig searches for a user-defined lvmd configuration file in /etc/microshift.  If found, returns
+// the lvmd config.  If not found, returns a default-value lvmd config.  If an unmarshalling errors, returns nil
+// and the error.
 func getCSIPluginConfig() (*lvmd.Lvmd, error) {
-	userCfg := filepath.Dir(config.DefaultUserConfigFile)
-	globalCfg := filepath.Dir(config.DefaultGlobalConfigFile)
-
-	if _, err := os.Stat(userCfg); errors.Is(err, os.ErrNotExist) {
-		return lvmd.NewLvmdConfigFromFile(filepath.Join(lvmd.LvmdConfigFileName))
-	}
-	if _, err := os.Stat(globalCfg); errors.Is(err, os.ErrNotExist) {
-		return lvmd.NewLvmdConfigFromFile(filepath.Join(globalCfg, lvmd.LvmdConfigFileName))
+	lvmdConfig := filepath.Join(filepath.Dir(config.DefaultGlobalConfigFile), lvmd.LvmdConfigFileName)
+	if _, err := os.Stat(lvmdConfig); errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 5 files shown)


copejon:
I'm unable to access the jira board at the moment.  Will try again tomorrow and append the ticket link asap.

pkg/components/storage.go
```diff
@@ -11,6 +13,23 @@ import (
...
+func getCSIPluginConfig() (*lvmd.Lvmd, error) {
+	userCfg := filepath.Dir(config.DefaultUserConfigFile)
+	globalCfg := filepath.Dir(config.DefaultGlobalConfigFile)
+
+	if _, err := os.Stat(userCfg); errors.Is(err, os.ErrNotExist) {
+		return lvmd.NewLvmdConfigFromFile(filepath.Join(lvmd.LvmdConfigFileName))
+	}
+	if _, err := os.Stat(globalCfg); errors.Is(err, os.ErrNotExist) {
+		return lvmd.NewLvmdConfigFromFile(filepath.Join(globalCfg, lvmd.LvmdConfigFileName))
+	}
```
pmtk, line 29:
```suggestion
	userCfg := filepath.Join(filepath.Dir(config.DefaultUserConfigFile), lvmd.LvmdConfigFileName)
	globalCfg := filepath.Join(filepath.Dir(config.DefaultGlobalConfigFile), lvmd.LvmdConfigFileName)

	if _, err := os.Stat(userCfg); !errors.Is(err, os.ErrNotExist) {
		return lvmd.NewLvmdConfigFromFile(userCfg)
	}
	if _, err := os.Stat(globalCfg); !errors.Is(err, os.ErrNotExist) {
		return lvmd.NewLvmdConfigFromFile(globalCfg)
	}
```

pmtk: Issue here was that code checked if config dir existed and then tried to load nonexistent file from that dir. Changed logic to stat the lvmd file instead and in case of existence use it to create the config

copejon:
pre-steps are failing during system dep installs
/retest

pkg/components/storage.go
```diff
@@ -11,6 +13,17 @@ import (
 	"github.com/openshift/microshift/pkg/config/lvmd"
 )
 
+// getCSIPluginConfig searches for a user-defined lvmd configuration file in /etc/microshift.  If found, returns
+// the lvmd config.  If not found, returns a default-value lvmd config.  If an unmarshalling errors, returns nil
+// and the error.
+func getCSIPluginConfig() (*lvmd.Lvmd, error) {
+	lvmdConfig := filepath.Join(filepath.Dir(config.DefaultGlobalConfigFile), lvmd.LvmdConfigFileName)
+	if _, err := os.Stat(lvmdConfig); errors.Is(err, os.ErrNotExist) {
+		return lvmd.NewLvmdConfigFromFile(filepath.Join(lvmdConfig, lvmd.LvmdConfigFileName))
```
pmtk, line 22: `lvmdConfig` already points to `/etc/microshift/lvmd.yaml`

```suggestion
		return lvmd.NewLvmdConfigFromFile(lvmdConfig)
```

pmtk, line 22: `lvmdConfig` already points to `/etc/microshift/lvmd.yaml` and we need to assert that file exists (so `not ErrNotExist`)

```suggestion
	if _, err := os.Stat(lvmdConfig); !errors.Is(err, os.ErrNotExist) {
		return lvmd.NewLvmdConfigFromFile(lvmdConfig)
```

PR #1162, opened 2022-12-02
USHIFT-447: Add os_id label to kubelet node registration

pacevedom:

Closes https://issues.redhat.com/browse/USHIFT-447

### DIFF
--- a/pkg/node/kubelet.go
+++ b/pkg/node/kubelet.go
@@ -16,12 +16,14 @@ limitations under the License.
 package node
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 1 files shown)


pkg/node/kubelet.go
```diff
@@ -180,3 +187,21 @@ func loadConfigFile(name string) (*kubeletconfig.KubeletConfiguration, error) {
 	}
 	return kc, err
 }
+
+func loadOSID() (string, error) {
+	readFile, err := os.Open("/etc/os-release")
+	if err != nil {
+		return "", err
```
mangelajo, line 194: This will break the environment of customers testing on non-rhel os's for GPU reasons. May be we should print a warning and use "unknown"? 

It'd be easy for them to workaround though, but I wanted to highlight it.

fzdarsky: Any system relying on `systemd` will support `/etc/os-release`, including Debian-derived ones, but I agree I don't see a reason to fail fatally if the OS info cannot be read.

pacevedom: Done

PR #1165, opened 2022-12-05
USHIFT-640: Add ovn-kubernetes traffic flows doc

zshi-redhat:

Related-Issue: https://issues.redhat.com/browse/USHIFT-640


docs/network/ovn_kubernetes_traffic_flows.md
```diff
@@ -0,0 +1,457 @@
+# OVNKubernetes Traffic Flows in MicroShift
+
+## Table of Contents
```
ggiguash, line 3: Isn't the TOC generated automatically? Can we avoid hardcoding it?

ggiguash:
@zshi-redhat, could you add a reference to this doc from the top-level README.md?

ggiguash:
/retitle USHIFT-640: Add ovn-kubernetes traffic flows doc

docs/network/ovn_kubernetes_traffic_flows.md
```diff
@@ -0,0 +1,445 @@
...
+openshift-dns       dns-default               ClusterIP   10.43.0.10      <none>        53/UDP,53/TCP,9154/TCP    6d6h
+openshift-ingress   router-internal-default   ClusterIP   10.43.179.188   <none>        80/TCP,443/TCP,1936/TCP   6d6h
+```
+
+## Pod Network
+
+- All the pods (non hostnetwork) are connected to an OVN logical switch named after the node name.
+- All the pods (non hostnetwork) are assigned with an overlay IP address from cluster network subnet (default: 10.42.0.0/16).
+
+### pod to pod
```
dhellmann, line 41: I love all of these flow descriptions. As someone who isn't an expert in this area, it would be useful for me to understand why each flow might be relevant. Some are obvious (like pod to pod). For the others, is there a common example of what sort of traffic would flow down that path? For example, why would the pod communicate directly with the host, or vice versa?

dhellmann (commented):
This looks great. I have a few questions, but overall this is very helpful.

docs/network/ovn_kubernetes_traffic_flows.md
```diff
@@ -0,0 +1,445 @@
+# OVNKubernetes Traffic Flows in MicroShift
+
+## OVN Topology
+
+The following diagram captures the OVN network topology for the MicroShift deployment:
+
+![picture_alt](ovn_topology.svg "OVN Topology")
```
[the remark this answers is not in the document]
zshi-redhat: > Nice diagram! That's exactly what I was hoping to get.
> 
> There are a couple of IPs around the "ovs join switch" that start `100.64` but I think they should probably be `10.64` based on the CIDR.

Good catch!  This is a typo, should be in the range of 100.64, I will update the diagram to use right IPs.

> 
> We have had some requests to handle network features like VLANs and bonding, and to generally provide better user control for which NICs can or cannot communicate with the cluster. Based on the diagram you provided, I wonder if we could achieve some of those goals by asking the user to create br-ex however they want, and then just having the CNI system attach to it. Would something like that work?

Yes, it will work.  We have similar use cases in openshift, where bond/vlan were created by MCO or nmstate, and br-ex is created on top of bond/vlan interfaces. In microshift, user can do:
1) manually create bond/vlan, then start microshfit (microshift-ovs-init) service
2) disable microshift-ovs-init, manually create bond/vlan and br-ex
Both shall work, user has to make sure the manual configurations persist across reboots.

docs/network/ovn_kubernetes_traffic_flows.md
zshi-redhat: Yes, there are examples of what traffic would flow though different path. I will list them in the flow section.
For example, we support pod using host network, in this case, pod with cluster network IP talks to pod with host network IP. Some system daemons and kubelet running in the host network, need to talk with pod in cluster network.

docs/network/ovn_kubernetes_traffic_flows.md
```diff
@@ -0,0 +1,445 @@
...
+NAMESPACE           NAME                      TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                   AGE
+default             kubernetes                ClusterIP   10.43.0.1       <none>        443/TCP                   6d6h
+default             svc-nodeport              NodePort    10.43.19.182    <none>        7000:30700/TCP            30m
+openshift-dns       dns-default               ClusterIP   10.43.0.10      <none>        53/UDP,53/TCP,9154/TCP    6d6h
+openshift-ingress   router-internal-default   ClusterIP   10.43.179.188   <none>        80/TCP,443/TCP,1936/TCP   6d6h
+```
+
+## Pod Network
+
+- All the pods (non hostnetwork) are connected to an OVN logical switch named after the node name.
```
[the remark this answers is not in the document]
zshi-redhat: That is a good question. 
Ovnk builds its database based on k8s API data, in case the database is lost, it can recover from the API data saved in etcd. So when the node name changed, I presume ovnk will rebuild the logcial networks with the updated node name. However, I didn't try if the rebuilt logical network will work as expected or not in microshift.

docs/network/ovn_kubernetes_traffic_flows.md
zshi-redhat: added a comment in each flow section.

zshi-redhat:
Added more desciption about ovn logical components in the ovn topology section, including which components serve which network etc.

docs/network/ovn_kubernetes_traffic_flows.md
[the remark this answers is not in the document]
zshi-redhat: added instructions for manual vlan/bond configuration in default_cni_plugin.md doc.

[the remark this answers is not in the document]
zshi-redhat: I will send another PR to cleanup the default_cni_plugin doc: remove content that is duplicate with the ovn_kubernetes_traffic_flow.md or host_networking.md.

zshi-redhat:
Rebased and fixed some comments, I think it is ready for another review.

docs/default_cni_plugin.md
```diff
@@ -93,14 +93,50 @@ When `disableOVSInit` is set to true in ovn-kubernetes CNI config file, OVS brid
...
 sudo ip link set br-ex up
 ```
 
-Replace `<physical-interface-name>` with the network interface name where node IP address is assigned to.
-Once br-ex up is up, assign the node IP address to br-ex bridge.
+Replace `enps10` with the network interface name in your own environment where node IP address is assigned to.<br>
+Once br-ex is up, assign the node IP address to br-ex bridge.<br>
 
-> Adding physical interface to br-ex bridge will disconnect ssh connection on node IP address.
+**NOTE:* Adding physical interface `enp1s0` to br-ex bridge will disconnect ssh connection on node IP address.
```
mangelajo, line 104: I would suggest we leave this part out of the PR.

If we give instructions to do this, we should probably do things via the network manager to make sure the changes are persistent across reboots.

WDYT?

mangelajo (changes requested):

docs/default_cni_plugin.md
zshi-redhat: Agree, removed from this PR.

mangelajo (approved):
/lgtm

PR #1167, opened 2022-12-05
Better cleanup of MicroShift resources for developers

mangelajo:
Without this, sometimes part of the container executables
stay running on the system under the kubepods.slice systemd
slice.


dhellmann:
Should we add something about this to our troubleshooting guide, too?

PR #1170, opened 2022-12-06
USHIFT-666: Add %{commit} and embed it into binary

pmtk:
ART's build system will substitute commit value with actual commit from o/microshift repo, so `microshift version -o yaml` will report it instead of printing empty string.

Also removed no longer used support for building RPM based on `github_tag `.

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -1,9 +1,12 @@
 # parameters that must be provided via --define , or fixed into the spec file:
 # global version 4.7.0
 # global release 2021_08_31_224727
-# global github_tag 4.7.0-0.microshift-2021-08-31-224727
-# global git_commit 81264d0ebb17fef06eff9ec7d4f2a81631c6b34a
+# global commit 81264d0ebb17fef06eff9ec7d4f2a81631c6b34a
 
+%{!?commit:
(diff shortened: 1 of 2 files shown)


ggiguash:
@pmtk, please update the `%changelog` section of microshift.spec.

packaging/rpm/microshift.spec
```diff
@@ -269,9 +255,18 @@ systemctl enable --now --quiet openvswitch || true
...
+
+* Wed Nov 30 2022 Patryk Matuszak <305846+pmtk@users.noreply.github.com> 4.12.0
+- Pass %{version} to Makefile
+
 * Wed Nov 30 2022 Gregory Giguashvili <ggiguash@redhat.com> 4.12.0
 - Change the config.yaml file name to allow its overwrite by users
 
+* Mon Nov 28 2022 Patryk Matuszak <305846+pmtk@users.noreply.github.com> 4.12.0
+- Use commit time & sha for RPM and exec
+
```
ggiguash, line 269: Can you fix your email address?

ggiguash (changes requested):

packaging/rpm/microshift.spec
pmtk: You won't let me be anonymous :) Fixed

PR #1175, opened 2022-12-07
USHIFT-660: yet another cluster debug info improvement

pmtk:
changed execution flow to:
* first register commands to run
* print a "table of contents" so it's obvious what command will be executed and easily searched for
* print a command with `>` prefix and execute the command

By replacing `set -x` with printing the cmd it is now visually clearer and suitable for inclusion in bug tickets (`./validate-microshift/cluster-debug-info.sh > /tmp/cluster.log`)

### DIFF
--- a/validate-microshift/cluster-debug-info.sh
+++ b/validate-microshift/cluster-debug-info.sh
@@ -1,28 +1,40 @@
 #!/usr/bin/env bash
 
-set -x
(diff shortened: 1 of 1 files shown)


ggiguash:
/retitle USHIFT-660: yet another cluster debug info improvement

/lgtm

PR #1176, opened 2022-12-07
Re-introduce PR#1030 baseDomain patch and add backend logic

mangelajo:
- Re-introduce PR#1030 baseDomain patch
- [USHIFT-664](https://issues.redhat.com//browse/USHIFT-664) Backend logic to baseDomain

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -56,7 +56,7 @@ network:
     - 10.43.0.0/16
   serviceNodePortRange: 30000-32767
 dns:
-  baseDomain: cluster.local
+  baseDomain: microshift.example.com
 debugging:
   logLevel: "Normal"
 ```
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -34,7 +34,7 @@ The configuration settings alongside with the supported command line arguments a
 | cidr (clusterNetwork) | --cluster-cidr            | MICROSHIFT_CLUSTER_CLUSTERCIDR          | A block of IP addresses from which Pod IP addresses are allocated
 | serviceNetwork        | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
 | serviceNodePortRange  | --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort
-| baseDomain            | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
+| baseDomain            | --base-domain             | MICROSHIFT_BASEDOMAIN                   | Base DNS domain used to construct fully qualified pod and service domain names
 | url                   | --url                     | MICROSHIFT_CLUSTER_URL                  | URL of the API server for the cluster.
 | nodeIP                | --node-ip                 | MICROSHIFT_NODEIP                       | The IP address of the node, defaults to IP of the default route
 | nodeName              | --node-name               | MICROSHIFT_NODENAME                     | The name of the node, defaults to hostname
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -34,7 +34,7 @@ The configuration settings alongside with the supported command line arguments a
 | cidr (clusterNetwork) | --cluster-cidr            | MICROSHIFT_CLUSTER_CLUSTERCIDR          | A block of IP addresses from which Pod IP addresses are allocated
 | serviceNetwork        | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
 | serviceNodePortRange  | --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort
-| baseDomain            | --base-domain             | MICROSHIFT_BASEDOMAIN                   | Base DNS domain used to construct fully qualified pod and service domain names
+| baseDomain            | --base-domain             | MICROSHIFT_BASEDOMAIN                   | Base DNS domain used to construct fully qualified router and API domain names.
 | url                   | --url                     | MICROSHIFT_CLUSTER_URL                  | URL of the API server for the cluster.
 | nodeIP                | --node-ip                 | MICROSHIFT_NODEIP                       | The IP address of the node, defaults to IP of the default route
 | nodeName              | --node-name               | MICROSHIFT_NODENAME                     | The name of the node, defaults to hostname
@@ -56,7 +56,7 @@ network:
     - 10.43.0.0/16
   serviceNodePortRange: 30000-32767
 dns:
-  baseDomain: microshift.example.com
+  baseDomain: example.com
 debugging:
   logLevel: "Normal"
 ```
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -15,13 +15,14 @@ subjectAltNames:
 nodeName: ""
 nodeIP: ""
 url: ""
-clusterDomain: ""
 network:
   clusterNetwork:
     - cidr: ""
   serviceNetwork:
     - ""
   serviceNodePortRange: ""
+dns:
+  baseDomain: ""
 debugging:
   logLevel: ""
 ```
@@ -33,7 +34,7 @@ The configuration settings alongside with the supported command line arguments a
 | cidr (clusterNetwork) | --cluster-cidr            | MICROSHIFT_CLUSTER_CLUSTERCIDR          | A block of IP addresses from which Pod IP addresses are allocated
 | serviceNetwork        | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
 | serviceNodePortRange  | --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort
-| clusterDomain         | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
+| baseDomain            | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
 | url                   | --url                     | MICROSHIFT_CLUSTER_URL                  | URL of the API server for the cluster.
 | nodeIP                | --node-ip                 | MICROSHIFT_NODEIP                       | The IP address of the node, defaults to IP of the default route
 | nodeName              | --node-name               | MICROSHIFT_NODENAME                     | The name of the node, defaults to hostname
@@ -48,13 +49,14 @@ In case `config.yaml` is not provided, the following default settings will be us
 nodeName: ""
 nodeIP: ""
 url: https://127.0.0.1:6443
-clusterDomain: cluster.local
 network:
   clusterNetwork:
     - cidr: 10.42.0.0/16
   serviceNetwork:
     - 10.43.0.0/16
   serviceNodePortRange: 30000-32767
+dns:
+  baseDomain: cluster.local
 debugging:
   logLevel: "Normal"
 ```
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -234,7 +234,7 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 				Hostnames: append(
 					cfg.SubjectAltNames,
 					cfg.NodeName,
-					"api.microshift."+cfg.BaseDomain,
+					"api."+config.DefaultClusterName+"."+cfg.BaseDomain,
 					// TODO: OpenShift actually uses  api.$ClusterName.$BaseDomain
 					// but we don't have a ClusterName parameter yet, using microshift
 					// for now
@@ -278,8 +278,8 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 					"openshift.default",
 					"openshift.default.svc",
 					"openshift.default.svc.cluster.local",
-					"api.microshift." + cfg.BaseDomain,     // TODO: OpenShift actually uses  api.$ClusterName.$BaseDomain
-					"api-int.microshift." + cfg.BaseDomain, // TODO: OpenShift actually uses  api.$ClusterName.$BaseDomain
+					"api." + config.DefaultClusterName + "." + cfg.BaseDomain,     // TODO: OpenShift actually uses  api.$ClusterName.$BaseDomain
+					"api-int." + config.DefaultClusterName + "." + cfg.BaseDomain, // TODO: OpenShift actually uses  api.$ClusterName.$BaseDomain
 					apiServerServiceIP.String(),
 				},
 			},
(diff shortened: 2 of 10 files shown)


mangelajo:
@fzdarsky @dinhxuanvu , please see the two commit structure, on first commit, I introduce @dinhxuanvu original patch, on second commit I handle the logic to match OpenShift behavior, also introducing the fixed "cluster.local" setting.

pkg/controllers/kube-apiserver.go
```diff
@@ -165,7 +165,7 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 									APIVersion: "route.openshift.io/v1",
 									Kind:       "HostAssignmentAdmissionConfig",
 								},
-								Domain: cfg.Cluster.Domain,
+								Domain: cfg.BaseDomain, // @benluddy, would this be the right thing?
```
mangelajo, line 168: @benluddy ^ would this be the right thing ?

docs/howto_config.md
```diff
@@ -33,7 +34,7 @@ The configuration settings alongside with the supported command line arguments a
 | cidr (clusterNetwork) | --cluster-cidr            | MICROSHIFT_CLUSTER_CLUSTERCIDR          | A block of IP addresses from which Pod IP addresses are allocated
 | serviceNetwork        | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
 | serviceNodePortRange  | --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort
-| clusterDomain         | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
+| baseDomain            | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
```
mangelajo, line 37:
```suggestion
| baseDomain            | --cluster-domain          | MICROSHIFT_BASEDOMAIN                 | Base DNS domain used to construct fully qualified pod and service domain names
```

pkg/controllers/kube-apiserver.go
```diff
@@ -165,7 +165,7 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 									APIVersion: "route.openshift.io/v1",
 									Kind:       "HostAssignmentAdmissionConfig",
 								},
-								Domain: cfg.Cluster.Domain,
+								Domain: cfg.BaseDomain,
```
mangelajo, line 168: @benluddy would this be the right thing to use here ^ ?  before we were using cluster.local but I wonder if this is more related to  https://github.com/openshift/microshift/pull/1176/files#diff-c70ba87d9ac52efa40c2eb55913d76e6c263c9690a9f3e48b074865cbd366738L65  ?

pkg/cmd/init.go
```diff
@@ -214,7 +214,8 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 					ValidityDays: cryptomaterial.ShortLivedCertificateValidityDays,
 				},
 				Hostnames: []string{
-					"router-default.apps." + cfg.Cluster.Domain,
+					"router-default.apps." + cfg.BaseDomain,
+					"*.apps." + cfg.BaseDomain, // wildcard for any additional auto-generated domains
```
mangelajo, line 218: I checked on an full OpenShift:

[majopela@rhel9 microshift]$ openssl s_client -connect oauth-openshift.apps.ci-ln-n3z62kb-76ef8.aws-2.ci.openshift.org:443  </dev/null 2>/dev/null | openssl x509 -noout -text | grep DNS
                DNS:*.apps.ci-ln-n3z62kb-76ef8.aws-2.ci.openshift.org

Probably we can remove 217 as it is redundant,

docs/howto_config.md
```diff
@@ -48,13 +49,14 @@ In case `config.yaml` is not provided, the following default settings will be us
...
 url: https://127.0.0.1:6443
-clusterDomain: cluster.local
 network:
   clusterNetwork:
     - cidr: 10.42.0.0/16
   serviceNetwork:
     - 10.43.0.0/16
   serviceNodePortRange: 30000-32767
+dns:
+  baseDomain: cluster.local
```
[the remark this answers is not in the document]
dhellmann: Maybe the example should be `microshift.example.com`?

pkg/cmd/init.go
dhellmann: @stlaz and @oglok are our cert experts, let's get them to confirm.

pkg/cmd/init.go
```diff
@@ -271,6 +278,7 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 					"openshift.default",
 					"openshift.default.svc",
 					"openshift.default.svc.cluster.local",
+					"api-int." + cfg.BaseDomain, // TODO: OpenShift actually uses  api.$ClusterName.$BaseDomain
```
dhellmann, line 281: Don't we have both for OCP? Do we need both here?

pkg/config/config.go
```diff
@@ -61,22 +60,23 @@ type IngressConfig struct {
...
 
-// Top level config
+// Top level config file
 type Config struct {
 	NodeName        string    `json:"nodeName"`
 	NodeIP          string    `json:"nodeIP"`
 	URL             string    `json:"url"`
-	ClusterDomain   string    `json:"clusterDomain"`
 	Network         Network   `json:"network"`
+	ClusterName     string    `json:"clusterName"`
```
dhellmann, line 78: I don't see `ClusterName` being used in this PR. We had a discussion on slack about what the default for the cluster name should be, how to report it in the cluster, how to keep it from being changed, etc. so I think adding it will be a big enough change to deserve its own PR.

pkg/config/config.go
```diff
@@ -106,6 +106,17 @@ type ClusterNetworkEntry struct {
...
 }
 
+type DNS struct {
+	// baseDomain is the base domain of the cluster. All managed DNS records will
+	// be sub-domains of this base.
+	//
+	// For example, given the base domain `openshift.example.com`, an API server
+	// DNS record may be created for `cluster-api.openshift.example.com`.
+	//
+	// Once set, this field cannot be changed.
```
dhellmann, line 116: Do we need another ticket to protect against the value changing?

dinhxuanvu (changes requested):
Just a few minor nits. Overall, the PR looks good.

pkg/config/config.go
dinhxuanvu: Just as Doug pointed out, I do not see the usage of this `ClusterName` field. Perhaps, this is unintentional. I recommend removing it given it is not relevant to the changes in this PR.

dinhxuanvu (changes requested):

pkg/config/config.go
dgrisonnet: I believe that this is a comment inherited from the OCP config API: https://github.com/openshift/api/blob/master/config/v1/types_dns.go#L27-L34

We might want to update it if it doesn't apply to MicroShift

pkg/config/config.go
dgrisonnet: it seems to be needed in https://github.com/openshift/microshift/pull/1176/commits/5147e3fac8b87e1096d3fc0c51bc9788c003f87c#diff-a53aedd081647bc5a1bc5c1a969ad049f4ff08c9b4c5ba9408ea647488849234R281, but it is not used yet. I am also leaning towards adding that in a separate effort.

pkg/config/config.go
fzdarsky: It does apply to MicroShift, too, I think @dhellmann 's concern was more where/how this is enforced.

pkg/config/config.go
fzdarsky: We're not just replacing a name here, though. What used to be "ClusterDomain" will become "ClusterName.BaseDomain".

Considering this, is your suggestion then that BaseDomain should change its semantic between this PR and the "separate effort"? If not, the minimum we have to dl is to at least update all locations where this PR replaces ClusterDomain with BaseDomain to add a hardcoded ClusterName.

pkg/cmd/init.go
mangelajo: no, SNO / OpenShift uses a wildcard cert *.apps.clustername.basedomain

that means that any other applications exposed under that domain will have a "valid" (self signed, but OK), certificate...

mangelajo: See my check above @fzdarsky :

```
[majopela@rhel9 microshift]$ openssl s_client -connect oauth-openshift.apps.ci-ln-n3z62kb-76ef8.aws-2.ci.openshift.org:443 </dev/null 2>/dev/null | openssl x509 -noout -text | grep DNS
DNS:*.apps.ci-ln-n3z62kb-76ef8.aws-2.ci.openshift.org
```

pkg/config/config.go
mangelajo: Sorry it's unintentional, it was intended to be introduced on a PR after this one , which we needed to discuss.

pkg/cmd/init.go
mangelajo: Let me add both. I thought api was for the externally serving cert, and api-int for the internally serving. But it would harm nobody.

pkg/config/config.go
mangelajo: I am adding "microshift" for now as the clustername part, 
on a 2nd PR we can make that configurable.

pkg/config/config.go
mangelajo: We may want to enforce against changes, I agree.

pkg/cmd/init.go
fzdarsky: Sorry for not being clear: I was trying to point out the missing `cfg.ClusterName` part (i.e. `ci-ln-n3z62kb-76ef8`). You're right, of course, that this is not only for the `router-default.aps.` line but also for the `*.apps.` line.

pkg/controllers/kube-apiserver.go
fzdarsky: Missing `cfg.ClusterName`

pkg/controllers/kube-apiserver.go
fzdarsky: Missing `cfg.ClusterName`, I assume.

pkg/config/config.go
```diff
@@ -61,22 +59,22 @@ type IngressConfig struct {
...
 
-// Top level config
+// Top level config file
 type Config struct {
 	NodeName        string    `json:"nodeName"`
 	NodeIP          string    `json:"nodeIP"`
 	URL             string    `json:"url"`
-	ClusterDomain   string    `json:"clusterDomain"`
 	Network         Network   `json:"network"`
+	DNS             DNS       `json:"dns"`
```
fzdarsky, line 77: Can we call this `cluster`, please? The `baseDomain` (and possibly also the `clusterName` or whatever we call it) are properties of the cluster itself and used in more places than just DNS (which suggests CoreDNS), e.g. Ingress.

pkg/config/config.go
```diff
@@ -180,12 +189,12 @@ func NewMicroshiftConfig() *MicroshiftConfig {
 		SubjectAltNames: subjectAltNames,
 		NodeName:        nodeName,
 		NodeIP:          nodeIP,
+		BaseDomain:      "microshift.example.com",
```
fzdarsky, line 192: Shouldn't this be "example.com", as "microshift" is also the clusterName?
(i.e. we'd end up getting`*.apps.microshift.microshift.example.com`)

pkg/cmd/init.go
```diff
@@ -214,7 +214,7 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 					ValidityDays: cryptomaterial.ShortLivedCertificateValidityDays,
 				},
 				Hostnames: []string{
-					"router-default.apps." + cfg.Cluster.Domain,
+					"*.apps.microshift." + cfg.BaseDomain, // wildcard for any additional auto-generated domains
```
fzdarsky, line 217: Is this change from `router-default` to `*` wildcard intentional?

docs/howto_config.md
```diff
@@ -33,7 +34,7 @@ The configuration settings alongside with the supported command line arguments a
 | cidr (clusterNetwork) | --cluster-cidr            | MICROSHIFT_CLUSTER_CLUSTERCIDR          | A block of IP addresses from which Pod IP addresses are allocated
 | serviceNetwork        | --service-cidr            | MICROSHIFT_CLUSTER_SERVICECIDR          | A block of virtual IP addresses for Kubernetes services
 | serviceNodePortRange  | --service-node-port-range | MICROSHIFT_CLUSTER_SERVICENODEPORTRANGE | The port range allowed for Kubernetes services of type NodePort
-| clusterDomain         | --cluster-domain          | MICROSHIFT_CLUSTER_DOMAIN               | Base DNS domain used to construct fully qualified pod and service domain names
+| baseDomain            | --base-domain             | MICROSHIFT_BASEDOMAIN                   | Base DNS domain used to construct fully qualified pod and service domain names
```
fzdarsky, line 37: Documentation string needs update, as its semantic changed. (maybe just: "Base domain of the cluster.")

pkg/cmd/init.go
mangelajo: Yes, see my comments; it is what OCP does. Providing a wildcard domain in this certificate.

In this way both router-default and any other router generated domains will be covered by this serving cert.

mangelajo: It is true that may be it does not exactly belong in this PR, but I was trying to align with what OCP did regarding the basedomain/clustername on the certs, and that's what I found.

```
[majopela@rhel9 microshift]$ openssl s_client -connect oauth-openshift.apps.ci-ln-n3z62kb-76ef8.aws-2.ci.openshift.org:443 </dev/null 2>/dev/null | openssl x509 -noout -text | grep DNS
DNS:*.apps.ci-ln-n3z62kb-76ef8.aws-2.ci.openshift.org
```

pkg/config/config.go
mangelajo: That was part of the alignment @dinhxuanvu tried to do regarding openshift-install config, so I added it as-is. I would need him to weigh in. To me, your suggestion makes sense but would like to hear his opinion.

pkg/config/config.go
mangelajo: +1 true, at least until we resolve/allow adding a cluster name

pkg/controllers/kube-apiserver.go
mangelajo: I believe it probably makes sense to add a constant "ClusterName" to the config module that we can use instead of just adding "microshift." everywhere. Then the intent will be clearer and also, it will be simpler to convert into cfg.ClusterName afterwards.

mangelajo: Or even apps. + ClusterName + BaseDomain? 

I am not sure of the intent of this admission controller config. @benluddy ^

oglok:
Squash at least last two commits, so we avoid having two with the same title in history.

pkg/cmd/init.go
fzdarsky: Thanks for investigating this. Makes sense to me then,

pkg/config/config.go
fzdarsky: Ok, I'd suggest to get this merged and do a follow-up PR, as this is not immediately related.

mangelajo:
> Squash at least last two commits, so we avoid having two with the same title in history.

I added the merge method "squash" for tide to handle.

PR #1178, opened 2022-12-07
ETCD-356: Changing etcd data directory to /var/lib/etcd

dusk125:
This change aligns the etcd data directory with that of OpenShift so that existing scripts, documentation, and customer/support knowledge can be more easily applied to MicroShift for etcd.


dusk125:
/hold
Waiting for 4.12 to branch so this can land in early 4.13

mangelajo:
Doing it in 4.13 could potentially need a migration path?, for example customers upgrading from 4.12 to 4.13?, I propose we do it for 4.12 if possible. But I know we are already trying to squeeze too much for release. @ggiguash @dhellmann

dhellmann:
> Doing it in 4.13 could potentially need a migration path?, for example customers upgrading from 4.12 to 4.13?, I propose we do it for 4.12 if possible. But I know we are already trying to squeeze too much for release. @ggiguash @dhellmann

We do not have a requirement to be able to upgrade from 4.12 to 4.13, so it should be safe to do this work in 4.13.

dusk125:
This is planned to land after #976 the hold is still valid until that merges

dusk125:
This is ready for consideration @dhellmann @pmtk 
/unhold

dhellmann:
I'm a little torn on this. On the one hand, I see the value of being consistent. On the other hand, we don't want to support etcd running on its own, outside of MicroShift and this move makes that product decision a little fuzzier.

If we do want to do this, the PR needs to update `hack/cleanup.sh` as well.

ggiguash:
I agree with @dhellmann 's comment. I'm not sure we'd want to handle `etcd` database out of MicroShift context.

ggiguash:
If we implement this change, the cleanup script also needs to be updated to delete the directory out of /var/lib/microshift

ggiguash:
/hold 
Pending a discussion and decision whether to proceed with this change

PR #1184, opened 2022-12-10
USHIFT-717: Remove IPs from certificates

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/etcd/cmd/microshift-etcd/run.go
+++ b/etcd/cmd/microshift-etcd/run.go
@@ -69,16 +69,16 @@ func (s *EtcdService) configure(cfg *config.MicroshiftConfig) {
 	//s.etcdCfg.ForceNewCluster = true //TODO
 	s.etcdCfg.Logger = "zap"
 	s.etcdCfg.Dir = dataDir
-	url2380 := setURL([]string{"127.0.0.1"}, "2380")
-	url2379 := setURL([]string{"127.0.0.1"}, "2379")
+	url2380 := setURL([]string{"localhost"}, "2380")
+	url2379 := setURL([]string{"localhost"}, "2379")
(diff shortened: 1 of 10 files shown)


dhellmann:
@pacevedom could you add some detail about why we want to do this?

pacevedom:
> @pacevedom could you add some detail about why we want to do this?

IPs in certificates were not being used and this PR enables the removal of the MicroShift automatic reset when the node IP changes (but this is not included here).
It also provides an easier handling/reasoning of certificates relying only on hostnames and DNS.
Need to create a JIRA ticket for all this if the PR is accepted, as there will be (at least) another one.

dhellmann:
> > @pacevedom could you add some detail about why we want to do this?
> 
> IPs in certificates were not being used and this PR enables the removal of the MicroShift automatic reset when the node IP changes (but this is not included here). It also provides an easier handling/reasoning of certificates relying only on hostnames and DNS. Need to create a JIRA ticket for all this if the PR is accepted, as there will be (at least) another one.

Does this add an expectation that DNS is always configured? Or pre-configured? I'm not sure we can count on that in all settings.

ggiguash:
This is a 4.13 task, so putting it on hold until 4.12 branch is created.
/hold

mangelajo (commented):
/approve

I like this. We completely decouple from NodeIP everywhere.

dhellmann:
@pacevedom based on https://github.com/openshift/enhancements/pull/1320 we no longer want this, right?

dhellmann:
The changes look OK, but CI doesn't seem very happy.

pmtk:
Maybe changing https://github.com/openshift/microshift/blob/main/pkg/controllers/etcd.go#L152 will help?

PR #1185, opened 2022-12-11
Watch firewall reload in sysconfigwatcher

zshi-redhat:
Related-Issue: https://issues.redhat.com/browse/NP-641


pkg/sysconfwatch/sysconfwatch_linux.go
```diff
@@ -92,6 +104,17 @@ func (c *SysConfWatchController) Run(ctx context.Context, ready chan<- struct{},
 	// Take a snapshot of the system and monototic clocks as a base reference
 	stimeRef, mtimeRef := getSysMonTimes()
 
+	for i := range c.iptClients {
+		iptClient := c.iptClients[i]
+		go iptClient.Monitor(
+			utiliptables.Chain("MICROSHIFT-SYSCONF-CANARY"),
+			[]utiliptables.Table{utiliptables.TableMangle, utiliptables.TableNAT, utiliptables.TableFilter},
+			func() { klog.Warningf("iptables flush is detected, restarting MicroShift"); os.Exit(0) },
+			1*time.Minute,
```
ggiguash, line 113: Should we be using the same sysConfigCheckInterval here?

pkg/sysconfwatch/sysconfwatch_linux.go
```diff
@@ -92,6 +104,17 @@ func (c *SysConfWatchController) Run(ctx context.Context, ready chan<- struct{},
 	// Take a snapshot of the system and monototic clocks as a base reference
 	stimeRef, mtimeRef := getSysMonTimes()
 
+	for i := range c.iptClients {
+		iptClient := c.iptClients[i]
+		go iptClient.Monitor(
+			utiliptables.Chain("MICROSHIFT-SYSCONF-CANARY"),
+			[]utiliptables.Table{utiliptables.TableMangle, utiliptables.TableNAT, utiliptables.TableFilter},
+			func() { klog.Warningf("iptables flush is detected, restarting MicroShift"); os.Exit(0) },
```
ggiguash, line 112: Need to update [this document](https://github.com/openshift/microshift/blob/main/docs/howto_sysconf_watch.md) with the new restart case.

ggiguash (changes requested):

pkg/sysconfwatch/sysconfwatch_linux.go
zshi-redhat: My concern was it might be too frequent to query the iptable rules, but having a short interval shall work I think.
Updated to use the sysConfigCheckInterval.

pkg/sysconfwatch/sysconfwatch_linux.go
zshi-redhat: Updated.

zshi-redhat:
resolved merge conflict.

docs/howto_sysconf_watch.md
```diff
@@ -138,4 +142,30 @@ The below (non-proportional!) graph shows when certificates are rotated.
...
+Dec 11 08:23:30 localhost.localdomain microshift[168331]: sysconfwatch-controller W1211 08:23:30.914772  168331 sysconfwatch_linux.go:112] iptables flush is detected, restarting MicroShift
+Dec 11 08:23:30 localhost.localdomain systemd[1]: microshift.service: Succeeded.
+Dec 11 08:23:30 localhost.localdomain systemd[1]: microshift.service: Consumed 28.612s CPU time
+Dec 11 08:23:31 localhost.localdomain systemd[1]: microshift.service: Service RestartSec=100ms expired, scheduling restart.
+Dec 11 08:23:31 localhost.localdomain systemd[1]: microshift.service: Scheduled restart job, restart counter is at 1.
+Dec 11 08:23:31 localhost.localdomain systemd[1]: Stopped MicroShift.
+Dec 11 08:23:31 localhost.localdomain systemd[1]: microshift.service: Consumed 28.612s CPU time
+Dec 11 08:23:31 localhost.localdomain systemd[1]: Starting MicroShift...
+Dec 11 08:23:42 localhost.localdomain microshift[171333]: ??? I1211 08:23:42.114965  171333 run.go:118] Starting MicroShift
+```
```
ggiguash, line 171: Can we present the output of `sudo journalctl -xu microshift | egrep 'Starting MicroShift|restarting MicroShift'` command to be consistent with other restart cases?

zshi-redhat: sure, done.

ggiguash:
This change is LGTM, but we need to wait until 4.12 branch is created.
/hold

pkg/config/config.go
```diff
@@ -65,6 +65,8 @@ type MicroshiftConfig struct {
 	NodeIP          string        `json:"nodeIP"`
 	BaseDomain      string        `json:"baseDomain"`
 	Cluster         ClusterConfig `json:"cluster"`
+	// Used to trigger a force rollout on network components upon MicroShift restart
+	CreationTimestamp string `json:"creationTimestamp"`
```
dhellmann, line 69: Do we need to expose this field to users or can it be an internal value?

dhellmann (commented):
For CSI, we drew a boundary so there was a clear line between the system configuration the user had to provide (volume groups and the lvmd.conf file) and the setup that MicroShift did (a default lvmd.conf if none is provided). Can we create a similar boundary for CNI so that we do not need to monitor the firewall rules, IP, or other networking changes?

pkg/config/config.go
zshi-redhat: It will be for internal use only. Captured when microshift main process starts.

zshi-redhat:
> For CSI, we drew a boundary so there was a clear line between the system configuration the user had to provide (volume groups and the lvmd.conf file) and the setup that MicroShift did (a default lvmd.conf if none is provided). Can we create a similar boundary for CNI so that we do not need to monitor the firewall rules, IP, or other networking changes?

Agree on the boundary statement between microshift and plugins, I thought about adding this in CNI implementation only.
One consideration of adding this monitor logic in sysconfWatcher is that iptables could be used by other components.
For example:  with hostPort in pod spec, iptables rules are used to map host ports to container ports, they will be affected by firewall reload as well. These hostPort iptables are not managed by CNI.

dhellmann:
> > For CSI, we drew a boundary so there was a clear line between the system configuration the user had to provide (volume groups and the lvmd.conf file) and the setup that MicroShift did (a default lvmd.conf if none is provided). Can we create a similar boundary for CNI so that we do not need to monitor the firewall rules, IP, or other networking changes?
> 
> Agree on the boundary statement between microshift and plugins, I thought about adding this in CNI implementation only. One consideration of adding this monitor logic in sysconfWatcher is that iptables could be used by other components. For example: with hostPort in pod spec, iptables rules are used to map host ports to container ports, they will be affected by firewall reload as well. These hostPort iptables are not managed by CNI.

Is there a way to organize the system so that we do not need to restart MicroShift when firewall rules change at all? Can we isolate the rule sets that users manage from the ones that MicroShift manages somehow?

zshi-redhat:
> > > For CSI, we drew a boundary so there was a clear line between the system configuration the user had to provide (volume groups and the lvmd.conf file) and the setup that MicroShift did (a default lvmd.conf if none is provided). Can we create a similar boundary for CNI so that we do not need to monitor the firewall rules, IP, or other networking changes?
> > 
> > 
> > Agree on the boundary statement between microshift and plugins, I thought about adding this in CNI implementation only. One consideration of adding this monitor logic in sysconfWatcher is that iptables could be used by other components. For example: with hostPort in pod spec, iptables rules are used to map host ports to container ports, they will be affected by firewall reload as well. These hostPort iptables are not managed by CNI.
> 
> Is there a way to organize the system so that we do not need to restart MicroShift when firewall rules change at all? Can we isolate the rule sets that users manage from the ones that MicroShift manages somehow?

@dhellmann  I moved the iptable detection logic to infrastructure manager service  which only reloads affected microshift components (router and ovnk). This avoids microshift restart. For application workload that uses hostPort, this won't help since the iptable rules for hostPort are injected by cri-o during pod sandbox creation, these rules won't be reconciled periodically unless recreating the workload. So current patch takes care of microshift components when firewall reload happens, we probably want to document that user workload with hostPort could be impacted by user action (reload firewall). wdyt?

pkg/components/controllers.go
```diff
@@ -162,7 +163,10 @@ func startIngressController(cfg *config.MicroshiftConfig, kubeconfigPath string)
 		return err
 	}
 
-	if err := assets.ApplyDeployments(apps, renderTemplate, renderParamsFromConfig(cfg, nil), kubeconfigPath); err != nil {
+	extraParams := assets.RenderParams{
```
mangelajo, line 166: I guess that for a decoupled CNI-plugin we may need to call the CNI plugin (know it's a CNI plugin) and call it to re-create the deployment.

We could make that part of a plugin interface i.e. have a "restart" call to the plugin which would add creation timestamps to the objects that need re-creation in the event of necessity.

Hmmmmm

mangelajo:
@dhellmann I left a comment inside about could we make something like this work with pluggable CNI.

pkg/components/controllers.go
dhellmann: This is one reason to have MicroShift run the plugin scripts, instead of relying solely on systemd. @fzdarsky

mangelajo:
/hold

I believe I found an alternative to make firewalld leave iptables alone (not flushing them)

dhellmann:
I think since we took #1306 we no longer need this. Please reopen the PR if I misunderstood.

/close

PR #1189, opened 2022-12-12
Add rhel volume group creation to dev env docs

pacevedom:

Closes #<Issue Number>

/cc @ggiguash


docs/devenv_rhel8.md
```diff
@@ -40,6 +40,7 @@ In the OS installation wizard, set the following options:
     - Under "New Red Hat Enterprise Linux 8.x Installation", click "Click here to create them automatically"
     - Select the root partition (`/`)
         - On the right side of the menu, set "Desired Capacity" to `40 GiB`
+        - On the right side of the menu, create a new volume group called `rhel`.
```
fzdarsky, line 43: The default VG to be created is already `rhel`. Did you mean to _verify_ it's set to `rhel` or actually to create a new one?

pacevedom: Had a vg called `microshift_root00` in my case. Maybe wording this to _verify/ensure_ is better.

ggiguash: Yes, this needs to be fixed. There's no need to create a new volume. We just need to make sure that the root volume is called `rhel` and it has unallocated space.

ggiguash (changes requested):

PR #1191, opened 2022-12-13
USHIFT-677: Remove _output on successful rebase

pmtk:


scripts/auto-rebase/rebase.sh
```diff
@@ -691,6 +691,9 @@ rebase_to() {
     else
         echo "No changes to buildfiles."
     fi
+
+    title "# Removing staging directory"
+    rm -rf "$REPOROOT/_output"
```
dhellmann, line 696: Does this mean the directory is only cleaned up in the full "to" workflow? But if a developer uses the individual commands to fetch images, update manifests, etc. then the temporary directory is not cleaned up? I think that's what we want, but I want to make sure I understand the changes.

pmtk: Exactly. And if anything along the way fails, script won't reach the removal, so (I hope) this addition won't be an annoyance for devs, but will address source RPM issue on our side rather than ART's.

PR #1196, opened 2022-12-15
USHIFT-247: Generate release-info

fzdarsky:
Adds a new RPM package "microshift-release-info" that contains JSON files containing the container image digests of the respective MicroShift release, so users can use these digests when creating osbuild-composer blueprints.

As side-effect, moves MicroShift's own release info from the pkg/release/release_*.go into JSON files in the assets/release directory.

Closes [USHIFT-247](https://issues.redhat.com//browse/USHIFT-247)


ggiguash:
@fzdarsky, can you add a short doc for devs explaing how this information can be used out of the new RPM?

packaging/rpm/microshift.spec
```diff
@@ -237,6 +255,9 @@ systemctl enable --now --quiet openvswitch || true
 # Use Git command to generate the log and replace the VERSION string
 # LANG=C git log --date="format:%a %b %d %Y" --pretty="tformat:* %cd %an <%ae> VERSION%n- %s%n" packaging/rpm/microshift.spec
 %changelog
+* Wed Dec 14 2022 Frank A. Zdarsky <fzdarsky@redhat.com> 4.12.0
```
ggiguash, line 258: Is it not 4.13 change?

docs/rhel4edge_iso.md
```diff
@@ -167,7 +166,7 @@ sudo scp microshift@microshift-dev:/home/microshift/microshift/_output/image-bui
 Run the following commands to create a virtual machine using the installer image.
 ```bash
 VMNAME="microshift-edge"
-VERSION=$(~/microshift/pkg/release/get.sh base)
+VERSION=$(grep "var Base" ~/microshift/pkg/release/release.go | cut -d\" -f 2)
```
ggiguash, line 169:
```suggestion
VERSION=$(grep "^var Base =" ~/microshift/pkg/release/release.go | cut -d\" -f 2)
```

pkg/release/release.go
```diff
@@ -33,3 +42,28 @@ var Image = map[string]string{
...
 }
+
+func init() {
+	arch_replacer := strings.NewReplacer("amd64", "x86_64", "arm64", "aarch64")
+	arch := arch_replacer.Replace(runtime.GOARCH)
+
+	release_file := "release/release-" + arch + ".json"
+	data, err := embedded.Asset(release_file)
+	if err != nil {
+		// If there is no release file for this architecture, work with the generic specs
```
ggiguash, line 53: Is it a good time to remove generic specs? Are they any good?

pkg/release/release.go
```diff
@@ -33,3 +42,28 @@ var Image = map[string]string{
...
+		panic(fmt.Errorf("unmarshaling %s: %v", release_file, err))
+	}
+
+	images := release["images"].(map[string]any)
+	for name, pullspec := range images {
+		name := strings.Replace(name, "-", "_", -1)
+		if _, ok := Image[name]; ok {
+			Image[name] = pullspec.(string)
+		}
+	}
```
ggiguash, line 68: Can we add some comments in this section on what we're trying to do and why?

scripts/image-builder/build.sh
```diff
@@ -4,7 +4,7 @@ set -e -o pipefail
 ROOTDIR=$(git rev-parse --show-toplevel)
 SCRIPTDIR=${ROOTDIR}/scripts/image-builder
 IMGNAME=microshift
-IMAGE_VERSION=$(${ROOTDIR}/pkg/release/get.sh base)
+IMAGE_VERSION=$(grep "var Base" "${ROOTDIR}/pkg/release/release.go" | cut -d\" -f 2)
```
ggiguash, line 7:
```suggestion
IMAGE_VERSION=$(grep "^var Base =" "${ROOTDIR}/pkg/release/release.go" | cut -d\" -f 2)
```

ggiguash (changes requested):

ggiguash:
@fzdarsky, please update [this doc section](https://github.com/openshift/microshift/blob/409d4e6e42f3c69e8a45b68c099f2b7966aa864a/docs/devenv_rhel8.md?plain=1#L93) with the new RPM

packaging/rpm/microshift.spec
```diff
@@ -61,7 +61,16 @@ Requires: sos
 %{?systemd_requires}
 
 %description
-The MicroShift package provides an OpenShift Kubernetes distribution optimized for small form factor and edge computing.
+The microshift package provides an OpenShift Kubernetes distribution optimized for small form factor and edge computing.
```
ggiguash, line 64: Should we add a dependency on the new RPM so that it's automatically deployed when "dnf install -y microshift" command is run?

ggiguash (changes requested):

packaging/rpm/microshift.spec
fzdarsky: Well, the problem is this information isn't needed on the machine where MicroShift is run, it is needed on the machine where osbuild-composer blueprints are updated (which may not even be an RPM-based machine, so this solution is probably just temporary). So I don't think adding this a required dependency (even if small) is relevant.

pkg/release/release.go
fzdarsky: In the upstream MicroShift project, this has been our fallback for cases when no equivalent images from OCP exist that we could reference. My assumption is we'll needed when we revive the upstream community.

packaging/rpm/microshift.spec
ggiguash: I would like to see a document on the usage scenarios before I make up my mind about this.

pkg/release/release.go
fzdarsky: Done.

fzdarsky:
> @fzdarsky, please update [this doc section](https://github.com/openshift/microshift/blob/409d4e6e42f3c69e8a45b68c099f2b7966aa864a/docs/devenv_rhel8.md?plain=1#L93) with the new RPM

Done.

pkg/release/release.go
```diff
@@ -33,3 +42,30 @@ var Image = map[string]string{
 	"pod":                       "quay.io/microshift/pause:" + Base,
 	"service_ca_operator":       "quay.io/microshift/service-ca-operator:" + Base,
 }
+
+func init() {
+	arch_replacer := strings.NewReplacer("amd64", "x86_64", "arm64", "aarch64")
```
oglok, line 47: I wonder why arch names have been replaced: `amd64` to `x86_64` and `arm64` to `aarch64`

oglok:
I think it's a very smart approach to solve this problem.

packaging/rpm/microshift.spec
dhellmann: The contents of the files going into this new RPM are also compiled into the microshift binary like the other assets, so microshift does not need the files to be installed. It would not be wrong to install them on the host running microshift, but nothing would necessarily consume them.

dhellmann (commented):
This is pretty much exactly what I had in mind. I had one or two nits, but nothing to hold it up. I know there were some other commenters, so I'll leave a hold along with the approval and you can remove that when everyone else is satisfied.

/lgtm
/hold

pkg/release/release.go
fzdarsky: My motivation for this was user experience. The x86_64/aarch64 architecture names are more idiomatic on Linux systems in general and RPM-based distros in particular.

For example, this way users can use `$(uname -i)` in scripts and avoid conditional logic and substitutions as with the Golang names. It’s also consistent with the RPM package naming scheme.

packaging/rpm/microshift.spec
ggiguash: Still, a document is missing in this PR. We need to document in GitHub what a dev can do with this RPM. A mostly copy-paste from the proposal doc?

ggiguash:
@fzdarsky, can we please implement 2 changes before we merge this PR?
- Add a document in GitHub (referenced from the README.md) explaining the RPM usage scenarios
- Restore the get.sh script and fix it to support the new parsing.

scripts/image-builder/build.sh
```diff
@@ -4,7 +4,7 @@ set -e -o pipefail
 ROOTDIR=$(git rev-parse --show-toplevel)
 SCRIPTDIR=${ROOTDIR}/scripts/image-builder
 IMGNAME=microshift
-IMAGE_VERSION=$(${ROOTDIR}/pkg/release/get.sh base)
+IMAGE_VERSION=$(grep "^var Base =" "${ROOTDIR}/pkg/release/release.go" | cut -d\" -f 2)
```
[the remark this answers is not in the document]
ggiguash: I agree. I think removing get.sh script forced us to duplicate and spread parsing of this information in other scripts. Let's return the get.sh file

pkg/release/release.go
```diff
@@ -33,3 +42,30 @@ var Image = map[string]string{
...
+	arch := arch_replacer.Replace(runtime.GOARCH)
+
+	release_file := "release/release-" + arch + ".json"
```
mangelajo, line 50: It would be great if we could also produce a toml snippet so customers just can append that to their blueprints without additional processing if they wanted.

I suggest that as follow up, not now.

packaging/rpm/microshift.spec
mangelajo: Yes, we should not add a dependency, but documentation is always good.

packaging/rpm/microshift.spec
```diff
@@ -61,7 +61,16 @@ Requires: sos
...
+
+
+%package release-info
+Summary: Release information for MicroShift
+BuildArch: noarch
+
+%description release-info
+The microshift-release package provides release information files for this release.
```
mangelajo, line 72: May be we should explain the purpose of this; i.e.

This package can be used to find the images consumed by MicroShift in runtime
and embed those into blueprints/ostree images.

fzdarsky:
> @fzdarsky, can you add a short doc for devs explaing how this information can be used out of the new RPM?

Added new `docs/howto_offline_containers.md` and linked it.

scripts/image-builder/build.sh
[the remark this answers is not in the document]
fzdarsky: Moved the base into the JSON, too, so now parsing is uniform and simple.

packaging/rpm/microshift.spec
fzdarsky: Added.

mangelajo:
/lgtm

Still get.sh is not re-added as talked in slack. But we can add that back if necessary.

PR #1203, opened 2022-12-19
NP-660: use systemd-resolved in CoreDNS configmap

zshi-redhat:
systemd-resolved conf is used by kubelet as alternative
dns resolver when exists. This commit aligns the DNS
Corefile conf with kubelet resolvConf when default
systemd-resolved config is detected.

### DIFF
--- a/assets/components/openshift-dns/dns/configmap.yaml
+++ b/assets/components/openshift-dns/dns/configmap.yaml
@@ -16,7 +16,7 @@ data:
             fallthrough in-addr.arpa ip6.arpa
         }
         prometheus 127.0.0.1:9153
-        forward . /etc/resolv.conf {
+        forward . {{if eq .UpstreamResolver ""}}/etc/resolv.conf{{else}}{{.UpstreamResolver}}{{end}} {
             policy sequential
         }
         cache 900 {
--- a/assets/components/openshift-dns/dns/configmap.yaml
+++ b/assets/components/openshift-dns/dns/configmap.yaml
@@ -16,7 +16,7 @@ data:
             fallthrough in-addr.arpa ip6.arpa
         }
         prometheus 127.0.0.1:9153
-        forward . {{if eq .UpstreamResolver ""}}/etc/resolv.conf{{else}}{{.UpstreamResolver}}{{end}} {
+        forward . {{with .UpstreamResolver}}{{.}}{{else}}/etc/resolv.conf{{end}} {
             policy sequential
         }
         cache 900 {
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -199,6 +199,7 @@ func startDNSController(cfg *config.MicroshiftConfig, kubeconfigPath string) err
 		klog.Warningf("Failed to apply", "namespace", ns, "err", err)
 		return err
 	}
+
 	extraParams := assets.RenderParams{
 		"ClusterIP": cfg.Cluster.DNS,
 	}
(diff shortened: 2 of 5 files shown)


zshi-redhat:
@Miciah Could you please help to check if this is a right use of dns forward plugin?

pkg/components/controllers.go
```diff
@@ -219,7 +221,14 @@ func startDNSController(cfg *config.MicroshiftConfig, kubeconfigPath string) err
 		klog.Warningf("Failed to apply serviceAccount %v %v", sa, err)
 		return err
 	}
-	if err := assets.ApplyConfigMaps(cm, nil, nil, kubeconfigPath); err != nil {
+	// set DNS forward to systemd-resolved resolver if exists
+	// https://github.com/coredns/coredns/blob/master/plugin/loop/README.md#troubleshooting-loops-in-kubernetes-clusters
+	if _, err := os.Stat(config.DefaultSystemdResolvedFile); !errors.Is(err, os.ErrNotExist) {
+		extraParams["UpstreamResolver"] = config.DefaultSystemdResolvedFile
```
Miciah, line 227: Checking only for `os.ErrNotExist` seems too specific.  For example, `err` could be a "permission denied" error, in which case it wouldn't make sense to use the path.  Why not just check `err != nil`?

assets/components/openshift-dns/dns/configmap.yaml
```diff
@@ -16,7 +16,7 @@ data:
             fallthrough in-addr.arpa ip6.arpa
         }
         prometheus 127.0.0.1:9153
-        forward . /etc/resolv.conf {
+        forward . {{if eq .UpstreamResolver ""}}/etc/resolv.conf{{else}}{{.UpstreamResolver}}{{end}} {
```
Miciah, line 19: This is slightly simpler:
```suggestion
        forward . {{with .UpstreamResolver}}{{.}}{{else}}/etc/resolv.conf{{end}} {
```
But functionally, this looks fine.

pkg/components/controllers.go
zshi-redhat: I think it tries to ignore other errors and continue to use systemd-resolve file. But I see your point there could be other errors preventing the file be read by microshift. I will use the the generic err check.

zshi-redhat: Fixed.

pkg/node/kubelet.go
```diff
@@ -132,8 +131,8 @@ serverTLSBootstrap: false #TODO`)
 
 	// Load real resolv.conf in case systemd-resolved is used
 	// https://github.com/coredns/coredns/blob/master/plugin/loop/README.md#troubleshooting-loops-in-kubernetes-clusters
-	if _, err := os.Stat("/run/systemd/resolve/resolv.conf"); !errors.Is(err, os.ErrNotExist) {
```
dhellmann, line 135: Are there other errors we need to handle differently?

scripts/auto-rebase/rebase.sh
```diff
@@ -532,6 +532,7 @@ update_manifests() {
     sed -i 's|REPLACE_COREDNS_IMAGE|{{ .ReleaseImage.coredns }}|' "${REPOROOT}"/assets/components/openshift-dns/dns/daemonset.yaml
     sed -i 's|REPLACE_RBAC_PROXY_IMAGE|{{ .ReleaseImage.kube_rbac_proxy }}|' "${REPOROOT}"/assets/components/openshift-dns/dns/daemonset.yaml
     sed -i 's|REPLACE_CLUSTER_IP|{{.ClusterIP}}|' "${REPOROOT}"/assets/components/openshift-dns/dns/service.yaml
+    sed -i 's|forward.*|forward . {{if eq .UpstreamResolver ""}}/etc/resolv.conf{{else}}{{.UpstreamResolver}}{{end}} {|' "${REPOROOT}"/assets/components/openshift-dns/dns/configmap.yaml
```
dhellmann, line 535: The value in this expression doesn't match the value in the asset file at the top of the PR.

zshi-redhat: Good catch!  I forgot to update the rebase script when fixing the comment.
Updated.

mangelajo (commented):
/lgtm

pkg/node/kubelet.go
dhellmann: I agree. I don't see that error handling here, yet.

PR #1210, opened 2022-12-22
NP-646: Prefer OVN-K gateway IP as the nodeIP

pliurh:
In OCP, we always use the same IP as OVN-K gateway and kubelet node IP. We shall do the same thing for microshift.

### DIFF
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -38,17 +38,18 @@ const (
 )
 
 func GetHostIP() (string, error) {
+	// Prefer OVN-K gateway IP if it is the CNI
+	gatewayIP, err := getOVNGatewayIP()
+	if err != nil && !strings.Contains(err.Error(), "no such network interface") {
+		return "", err
+	}
+	klog.V(2).Infof("ovn gateway IP address: %s", gatewayIP)
+
 	ip, err := net.ChooseHostInterface()
 	if err == nil {
 		return ip.String(), nil
 	}
 	klog.V(2).Infof("failed to find default route IP address: %v", err)
-
-	gatewayIP, err := getOVNGatewayIP()
-	if err != nil {
-		return "", err
-	}
-	klog.V(2).Infof("ovn gateway IP address: %s", gatewayIP)
 
 	return gatewayIP, nil
 }


zshi-redhat:
/title NP-646: always use OVN-K gateway IP as node IP by default

zshi-redhat:
/retitle NP-646: always use OVN-K gateway IP as node IP by default

zshi-redhat:
@pliurh could you fix the unit test error?

pkg/util/net.go
```diff
@@ -38,12 +37,6 @@ const (
 )
 
 func GetHostIP() (string, error) {
-	ip, err := net.ChooseHostInterface()
-	if err == nil {
-		return ip.String(), nil
-	}
-	klog.V(2).Infof("failed to find default route IP address: %v", err)
-
```
zshi-redhat, line 46: We may want to keep this logic when getOVNGatewayIP fails to support other CNIs.

zshi-redhat: also note that we are discussing to remove plugin specific code from microshift main process through the [pluggable plugin enhancement](https://github.com/openshift/enhancements/pull/1304/),  I wonder if we should let microshift to detect the node IP here and move the configure-ovs.sh execution plugin specific code.
/cc @mangelajo

pkg/components/render_test.go
```diff
@@ -151,8 +151,20 @@ func Test_renderTopolvmDaemonsetTemplate(t *testing.T) {
 		{
 			name: "renders lvmd-socket-name path",
 			args: args{
-				tb:   tb,
-				data: renderParamsFromConfig(config.NewMicroshiftConfig(), assets.RenderParams{"SocketName": "/run/lvmd/lvmd.socket", "lvmd": "foobar"}),
+				tb: tb,
+				data: renderParamsFromConfig(&config.MicroshiftConfig{
+					LogVLevel:       0,
```
mangelajo, line 156: Why do we need to fill this additional info here (Test_renderTopolvmDaemonsetTemplate) ?

pkg/util/net.go
mangelajo: I don't understand why this change forces the testing into having a fully populated config structure. How is it related?

dhellmann:
Does this have any impact on https://github.com/openshift/enhancements/pull/1320?

zshi-redhat:
> Does this have any impact on [openshift/enhancements#1320](https://github.com/openshift/enhancements/pull/1320)?

It won't.
This PR is to fix a bug where microshift-ovs-init service and microshift main process may choose different node IPs.
It won't impact the certificate since both ovnk and APIServer are guaranteed to use the same IP address after the fix.

pkg/util/net.go
pliurh: When we run the test, there's no OVN-K, thus bridge `br-ex` doesn't exist. So `GetHostIP` will fail.

pliurh:
@zshi-redhat @mangelajo I've added the `ChooseHostInterface` back. PTAL

mangelajo:
hey, it looks good to me now :)

Thanks for updating the patch.

/lgtm

PR #1220, opened 2022-12-26
NP-648: single pod ovnk

zshi-redhat:


ggiguash:
This is a 4.13 release functionality. Holding the PR until the 4.12 branch is created.
/hold

mangelajo:
@ggiguash @zshi-redhat I have observed a stable drop of 50MBi of RAM when setting it up together in one pod tested on my system.

@zshi-redhat Not sure if it's feasible, but... using a single container, in addition, could bring around an extra 20MBi drop. (all the associated conmon and bash processes...) I don't know if it's worth though.

dhellmann:
/retest

We don't have to do it this time, but this seems like an example of where we might need upgrade logic to clean up old settings (the ClusterRole, for example). Something to keep in mind for when we get to the point where upgrades are fully supported.

zshi-redhat:
> /retest
> 
> We don't have to do it this time, but this seems like an example of where we might need upgrade logic to clean up old settings (the ClusterRole, for example). Something to keep in mind for when we get to the point where upgrades are fully supported.

added this comment to the ovnk upgrade ticket.

zshi-redhat:
> @ggiguash @zshi-redhat I have observed a stable drop of 50MBi of RAM when setting it up together in one pod tested on my system.
> 
> @zshi-redhat Not sure if it's feasible, but... using a single container, in addition, could bring around an extra 20MBi drop. (all the associated conmon and bash processes...) I don't know if it's worth though.

@mangelajo how did you measure the memory?

zshi-redhat:
/unhold

remove the hold flag added by openshift bot due to 3 failed retest.

PR #1221, opened 2022-12-27
Store ovn databases to hostpath

zshi-redhat:
ovn databases (nb/sb) were stored in containers and recreated every time when ovnkube-master container restarts. This commit saves container databases to host directory (/var/lib/ovn) which avoids recreation of ovn databases after ovnkube-master restarts. It optimizes around 0~10M memory for each database container from a snapshot metrics with metrics-server. Also add a command in cleanup script to remove the ovn databases.

Related-Issue: https://issues.redhat.com/browse/NP-648

### DIFF
--- a/assets/components/ovn/master/daemonset.yaml
+++ b/assets/components/ovn/master/daemonset.yaml
@@ -74,9 +74,9 @@ spec:
         - mountPath: /run/ovn/
           name: run-ovn
         - mountPath: /etc/openvswitch
-          name: etc-openvswitch-node
+          name: etc-ovn
         - mountPath: /etc/ovn/
-          name: etc-openvswitch-node
+          name: etc-ovn
         - mountPath: /var/lib/openvswitch
           name: var-lib-openvswitch
         - mountPath: /var/log/ovn
@@ -129,6 +129,10 @@ spec:
           name: run-openvswitch
         - mountPath: /run/ovn/
           name: run-ovn
+        - mountPath: /etc/openvswitch/
+          name: etc-ovn
+        - mountPath: /var/lib/openvswitch/
+          name: data-ovn
         resources:
           requests:
             cpu: 10m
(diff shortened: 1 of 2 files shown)


zshi-redhat:
/hold
4.13 task

zshi-redhat:
> > ovn databases (nb/sb) were stored in containers and recreated every time when ovnkube-master container restarts. This commit saves container databases to host directory (/etc/ovn/) which avoids recreation of ovn databases after ovnkube-master restarts. It optimizes around 0~10M memory for each database container from a snapshot metrics with metrics-server. Also add a command in cleanup script to remove the ovn databases.
> > Related-Issue: https://issues.redhat.com/browse/NP-648
> 
> It looks like the files are going to /var/lib not /etc. Some of the mount points are /var/lib/openvswitch and some are /var/lib/ovn. Do those all need to be the same?

I forgot to change the commit message, it shall be `/var/lib/ovn`.
Yes, the container mount points are different for ovn and openvswitch inside pod, for example, the ovn databases `etc-ovn` are mounted to `/etc/ovn` and `/etc/openvswitch` separately. `data-ovn` is mounted to `/var/lib/openvswitch`. 

I compared and updated the mount paths according to how they are used in openshift:
https://docs.google.com/spreadsheets/d/1Hj23VEV4zGpKp0PgvVbioZhUA3CYZrECsw8UZSqZlyk/edit#gid=1509429642

> 
> This change moves more state outside of MicroShift, which means more to consider when upgrading or rolling back. Is the database still rebuilt when the system starts?

Databases will be rebuilt if they are removed or don't exist, in other words, they can be removed and rebuilt entirely from k8s apiserver.  For upgrading or rolling back, we could either delete the databases or keep them. If the databases are kept during upgrade, the database schema will be updated if needed every time when ovnkube-master pod starts.

dhellmann:
> > This change moves more state outside of MicroShift, which means more to consider when upgrading or rolling back. Is the database still rebuilt when the system starts?
> 
> Databases will be rebuilt if they are removed or don't exist, in other words, they can be removed and rebuilt entirely from k8s apiserver. For upgrading or rolling back, we could either delete the databases or keep them. If the databases are kept during upgrade, the database schema will be updated if needed every time when ovnkube-master pod starts.

OK. It sounds like we should think about what approach we want to standardize on so we can pick one thing to support. Should that be part of @oglok 's upgrade/rollback enhancement or should we do it separately?

mangelajo:
The change / improvement looks good, as we said we only need to make sure it will work with ugprades/rollbacks on greenboot.

/approve

zshi-redhat:
Closing the PR so that we have less thing to concern for the upgrade/rollback.

PR #1224, opened 2022-12-27
Delete openshift-tests' static analysis code

pmtk:


zshi-redhat:
Will the tests be moved to somewhere else?

PR #1227, opened 2023-01-01
Initial implementation of MicroShift integration with greenboot

ggiguash:
Closes:
* [USHIFT-696](https://issues.redhat.com//browse/USHIFT-696): Greenboot script implementation
* [USHIFT-697](https://issues.redhat.com//browse/USHIFT-697): Include Greenboot script in RPM
* [USHIFT-698](https://issues.redhat.com//browse/USHIFT-698): Document MicroShift Greenboot specifics

### DIFF
--- a/docs/devenv_setup.md
+++ b/docs/devenv_setup.md
@@ -93,13 +93,18 @@ make srpm
 
 The artifacts of the build are located in the `_output/rpmbuild` directory.
 ```bash
-$ cd ~/microshift/ && find _output -name \*.rpm
-_output/rpmbuild/RPMS/x86_64/microshift-4.13.0_20221215165847_758c8b61-1.el8.x86_64.rpm
-_output/rpmbuild/RPMS/x86_64/microshift-networking-4.13.0_20221215165847_758c8b61-1.el8.x86_64.rpm
-_output/rpmbuild/RPMS/noarch/microshift-release-info-4.13.0_20221215165847_758c8b61-1.el8.noarch.rpm
-_output/rpmbuild/RPMS/noarch/microshift-selinux-4.13.0_20221215165847_758c8b61-1.el8.noarch.rpm
-_output/rpmbuild/SRPMS/microshift-4.13.0_20221215165847_758c8b61-1.el8.src.rpm
+$ cd ~/microshift/_output/rpmbuild && find . -name \*.rpm 
+./RPMS/x86_64/microshift-4.13.0_0.nightly_2023_01_17_152326_20230124054037_b67f6bc3_dirty-1.el8.x86_64.rpm
+./RPMS/x86_64/microshift-networking-4.13.0_0.nightly_2023_01_17_152326_20230124054037_b67f6bc3_dirty-1.el8.x86_64.rpm
+./RPMS/noarch/microshift-release-info-4.13.0_0.nightly_2023_01_17_152326_20230124054037_b67f6bc3_dirty-1.el8.noarch.rpm
+./RPMS/noarch/microshift-selinux-4.13.0_0.nightly_2023_01_17_152326_20230124054037_b67f6bc3_dirty-1.el8.noarch.rpm
+./RPMS/noarch/microshift-greenboot-4.13.0_0.nightly_2023_01_17_152326_20230124054037_b67f6bc3_dirty-1.el8.noarch.rpm
+./SRPMS/microshift-4.13.0_0.nightly_2023_01_17_152326_20230124083515_fad43b98_dirty-1.el8.src.rpm
 ```
+
+> The `microshift-release-info` and `microshift-greenboot` RPM packages are optional.
+> See [Embedding MicroShift Container Images for Offline Deployments](./howto_offline_containers.md) and
+> [Integrating MicroShift with Greenboot](./greenboot.md) for more information.
 
 ## Run MicroShift Executable
 Log into the development virtual machine with the `microshift` user credentials.
--- a/docs/greenboot.md
+++ b/docs/greenboot.md
@@ -1,0 +1,81 @@
+# Integrating MicroShift with Greenboot
+
+## Motivation
+
+Serviceability of Edge Devices is often limited or non-existent, which makes it
+challenging to troubleshoot device problems following a failed software or
+operating system upgrade. 
+
+To mitigate these problems, MicroShift uses [greenboot](https://github.com/fedora-iot/greenboot),
+the Generic Health Check Framework for `systemd` on `rpm-ostree` based systems.
+If a failure is detected, the system is expected to boot into the last known
+working configuration using `rpm-ostree` rollback facilities.
+
+This functionality benefits the users by reducing the risk of being locked out
+of an Edge Device when upgrades take place. Users should not experience significant
+interruption of service in case of a failed upgrade.
+
+## Implementation Details
+
+### MicroShift Validation
+
+MicroShift includes the `40_microshift_running_check.sh` health check script to
+validate that all the required MicroShift services are up and running. There is
+no explicit dependency on the `greenboot` RPM from MicroShift core packages, or
+a requirement to use health check procedures. The health check script is packaged
+in the separate `microshift-greenboot` RPM, which may be installed on the system
+if the `greenboot` facilities are to be used.
+
+The health check script is installed into the `/etc/greenboot/check/required.d`
+directory and it is not executed during the system boot in case the `greenboot`
+package is not present.
+
+In addition, if the `greenboot-default-health-check` RPM subpackage is installed,
+it already includes
+[health check scripts](https://github.com/fedora-iot/greenboot#health-checks-included-with-subpackage-greenboot-default-health-checks)
+verifying that DNS and `ostree` services can be accessed.
+
+Exiting the script with a non-zero status will have the boot declared as failed.
+Greenboot redirects all the script output to the system log, accessible via the
+`journalctl -u greenboot-healthcheck.service` command.
+
+|Validation                                           |Pass  |Fail  |
+|-----------------------------------------------------|------|------|
+|Check the script runs with 'root' permissions        |Next  |exit 0|
+|Check microshift.service is enabled                  |Next  |exit 0|
+|Wait for microshift.service to be active (!failed)   |Next  |exit 1|
+|Wait for Kubernetes API health endpoints to be OK    |Next  |exit 1|
+|Wait for any Pod to start                            |Next  |exit 1|
+|For each core namespace, wait for images to be pulled|Next  |exit 1|
+|For each core namespace, wait for Pods to be ready   |Next  |exit 1|
+|For each core namespace, check Pods not restarting   |exit 0|exit 1|
+
+> If the system is not booted using the `ostree` file system, the health check
+> procedures still run, but no rollback would be possible in case of an upgrade
+> failure.
+
+The wait period in each validation starts from 5 minutes base time and it is
+incremented by the base wait period after each boot in the verification loop.
+It is possible to override the base time wait period setting with the
+`MICROSHIFT_WAIT_TIMEOUT_SEC` environment variable in the `/etc/greenboot/greenboot.conf`
+configuration file alongside other [Greenboot Configuration](https://github.com/fedora-iot/greenboot#configuration)
+settings.
+
+### User Workloads Validation
+
+Some 3rd party user workloads may become operational before the upgrade is
+declared valid and potentially create or update data on the device. If a
+rollback is performed subsequentially, there is a risk of data loss because the
+file system is reverted to its state before the upgrade. One of the ways to
+mitigate this problem is to have 3rd party workloads wait until a boot is
+declared successful.
+
+```bash
+$ sudo grub2-editenv - list | grep ^boot_success
+boot_success=1
+```
+
+> Note that the MicroShift health check script only performs validation of the
+> core MicroShift services. Users should install their own workload validation
+> scripts using `greenboot` facilities to ensure the successful operation after
+> system upgrades.
--- a/packaging/greenboot/microshift-running-check.sh
+++ b/packaging/greenboot/microshift-running-check.sh
@@ -1,0 +1,241 @@
+#!/bin/bash
+set -e
+
+SCRIPT_NAME=$(basename $0)
+SCRIPT_PID=$$
+OCGET_CMD="microshift get"
+OCGET_OPT="--kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig --no-headers"
+PODS_NS_LIST=(openshift-ovn-kubernetes openshift-service-ca openshift-ingress openshift-dns openshift-storage)
+PODS_CT_LIST=(2                        1                    1                 2             2)
+
+# Source Greenboot configuration file if it exists
+GREENBOOT_CONF_FILE=/etc/greenboot/greenboot.conf
+[ -f "${GREENBOOT_CONF_FILE}" ] && source ${GREENBOOT_CONF_FILE}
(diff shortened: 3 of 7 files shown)


ggiguash:
This is a 4.13 task - putting on hold until the branch is created. It is ready to be reviewed, though
/hold

packaging/rpm/microshift.spec
```diff
@@ -57,6 +56,7 @@ Requires: microshift-selinux
 Requires: microshift-networking
 Requires: conntrack-tools
 Requires: sos
+Requires: greenboot
```
dhellmann, line 59: MicroShift works fine without greenboot, and I don't think we want to add it as a hard requirement for the base RPM. Would it make sense to have it be a passive integration (install our script, but nothing would use it if greenboot is not installed) or to package the script in a separate RPM ("microshift-greenboot") that does depend on greenboot? The passive approach would let us use the script as a general health check tool, so I think I would prefer that if it works.

docs/greenboot.md
```diff
@@ -0,0 +1,207 @@
+# MicroShift Greenboot
```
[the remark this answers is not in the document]
ggiguash: Agreed

docs/greenboot.md
```diff
@@ -0,0 +1,207 @@
...
+`MICROSHIFT_WAIT_TIMEOUT_SEC` environment variable.
+
+### User Workloads Validation
+
+Some 3rd party user workloads may become operational before the upgrade is
+declared valid and potentially create or update data on the device. If a
+rollback is performed subsequentially, there is a risk of data loss because the
+file system is reverted to its state before the upgrade. To mitigate this
+problem, 3rd party workloads should never become operational until a boot is
+declared successful.
```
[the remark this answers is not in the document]
ggiguash: I rephrased to make the statement less imposing. Is it better now?

docs/greenboot.md
```diff
@@ -0,0 +1,207 @@
...
+$ sudo grub2-editenv - list | grep ^boot_success
+boot_success=1
+```
+
+> Note that the MicroShift health check script only performs validation of the
+> core MicroShift services. Users should install their own workload validation
+> scripts using `greenboot` facilities to ensure the successful operation after
+> system upgrades.
+
+## Testing Failed Updates
```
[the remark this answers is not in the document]
ggiguash: Good idea. Created greenboot_dev.md file.

packaging/greenboot/microshift-greenboot-required.sh
```diff
@@ -0,0 +1,209 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ggiguash: Yes, here's the logic in the naming of the script:
- Greenboot has "required" and "wanted" checks, so using "required" in the script name denotes mandatory checks
- For the "greenboot" in the name, the script is integrated in the Greenboot upgrade flow. However, it can be run in a standalone fashion and the checks will work.

We can rename the script to something more generic, but it's intended to be used with Greenboot. 
Any further suggestions for changing the name?

packaging/greenboot/microshift-greenboot-required.sh
```diff
@@ -0,0 +1,209 @@
...
+[ -f "${GREENBOOT_CONF_FILE}" ] && source ${GREENBOOT_CONF_FILE}
+WAIT_TIMEOUT_SECS_BASE=${MICROSHIFT_WAIT_TIMEOUT_SEC:-300}
+
+# Check the current user before the output redirect
+if [ $(id -u) -ne 0 ] ; then
+    echo "The '${SCRIPT_NAME}' script must be run with the 'root' user privileges"
+    exit 0
+fi
+
+# Redirect all the output to the log file and set the exit handler
```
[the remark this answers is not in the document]
ggiguash: The output is redirected to system logs by Greenboot. 
I removed the script redirection to avoid custom log file handling in the future - great point.

packaging/greenboot/microshift-greenboot-required.sh
```diff
@@ -0,0 +1,209 @@
...
+
+# Dump GRUB boot and Greenboot variables if any
+grub_vars=$(grub2-editenv - list | grep ^boot_ || true)
+boot_vars=$(set | egrep '^GREENBOOT_|^MICROSHIFT_' || true)
+
+[ -z "${grub_vars}" ] && grub_vars=None
+[ -z "${boot_vars}" ] && boot_vars=None
+
+log "GRUB boot variables:\n${grub_vars}"
+log "Greenboot variables:\n${boot_vars}"
```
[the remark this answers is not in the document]
ggiguash: These variables determine Greenboot behavior and the boot validation status. I could not find another place where these are logged, so I think they will be important for our greeenboot validation troubleshooting.

packaging/greenboot/microshift-greenboot-required.sh
```diff
@@ -0,0 +1,209 @@
...
+
+# Exit if the ostree filesystem is not used
```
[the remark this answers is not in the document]
ggiguash: In general, greenboot RPM can be installed on a non-ostree file system as well. It would run the checks, without the ability to rollback. I'm removing this check so that the script can be used regardless of the ostree presence.

packaging/rpm/microshift.spec
ggiguash: Updated the implementation, docs and ER according to this proposal.

docs/greenboot.md
```diff
@@ -0,0 +1,74 @@
...
+Greenboot redirects all the script output to the system log, accessible via the
+`journalctl -u greenboot-healthcheck.service` command.
+
+|Validation                                        |Pass  |Fail  |
+|--------------------------------------------------|------|------|
+|Check microshift.service is enabled               |Next  |exit 0|
+|Wait for microshift.service to be active (!failed)|Next  |exit 1|
+|Wait for any pods to start                        |Next  |exit 1|
+|For each core namespace, wait for pods to be ready|Next  |exit 1|
+|For each core namespace, check pods not restarting|exit 0|exit 1|
```
mangelajo, line 44:
```suggestion
|For each core namespace, check pods not restarting|exit 0|exit 1|
```

This check can be dangerous, some pods will restart, for example, pods that depend on microshift, will miss microshift during a shutdown, etc... and restart..

May be we will need to disable this at some point, or leave it out for the time being, maybe even white list some pods?

docs/greenboot.md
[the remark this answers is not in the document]
mangelajo: Sounds less restrictive, but how could you handle that?

docs/greenboot.md
ggiguash: @mangelajo, please look at the implementation PR for this check. It is done by comparing samples over time, so there should be more non-restarting than restarting samples

docs/greenboot.md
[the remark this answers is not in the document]
ggiguash: I think the point is that we need to "warn" 3rd party developers, but we cannot enforce this requirement on them. When they need to implement the check, the boot_success variable is a way to go.

packaging/greenboot/microshift-greenboot-required.sh
```diff
@@ -0,0 +1,194 @@
...
+#
+# args: None
+# env1: 'CHECK_PODS_NS' environment variable for the namespace to check
+# env2: 'CHECK_PODS_CT' environment variable for the pod count to check
+# return: 0 if the expected number of pods are ready, or 1 otherwise
+function namespace_pods_ready() {
+    local ns=${CHECK_PODS_NS}
+    local ct=${CHECK_PODS_CT}
+
+    local status=$(oc get pods ${OCGET_OPT} -n ${ns} -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}' 2>/dev/null)
```
mangelajo, line 93: Something to keep in mind is that we are adding a dependency to the openshift-client with this patch. I am not sure how desirable it is on all situations, but also I am not sure how could we get away without it. One possibility could be to have microshift report the running status via a simple localhost API endpoint, or a file/unixsocket.




I

packaging/greenboot/microshift-greenboot-required.sh
```diff
@@ -0,0 +1,194 @@
...
+    local restarts=0
+
+    local count1=$(oc get pods ${OCGET_OPT} -n ${ns} -o 'jsonpath={..status.containerStatuses[].restartCount}' 2>/dev/null)
+    for i in $(seq 10) ; do
+        sleep 5
+        local countS=$(oc get pods ${OCGET_OPT} -n ${ns} -o 'jsonpath={..status.containerStatuses[].started}' 2>/dev/null | grep -vc false)
+        local count2=$(oc get pods ${OCGET_OPT} -n ${ns} -o 'jsonpath={..status.containerStatuses[].restartCount}' 2>/dev/null)
+
+        # If pods started, a restart is detected by comparing the count string between the checks.
+        # The number of pod restarts is incremented when a restart is detected, or decremented otherwise.
```
mangelajo, line 127: I am not sure if this would work, for example pod restarts are delayed over time (backoff behavior). At some point the delay is bigger than 5secs, the restart counter will keep decrementing.

Also I am not sure about a safe way to check this. For example some pods also restart naturally when you reboot which I guess is what you were trying to avoid detecting as a failure with this behavior...

ggiguash: The pod restarts check is performed after making sure pods are ready. If pods are restarting at this point of time, it means there is a problem. That was my thinking at least.

mangelajo: But sometimes when you restart pods are reported as ready, even though not even the containers are around yet (checked via crio). ':-)

It's a tricky area.

ggiguash: I'm not sure how that's going to happen. This is implemented [here](https://github.com/openshift/microshift/pull/1227/files#diff-dc4eae6d665efc45c9b98bf8a612cbdb19b1fc7e9313affdabd1d158d1d70d8aR93)

Regardless, we're adding image download check, before readiness, so we ensure that all images are there.

packaging/greenboot/microshift-greenboot-required.sh
ggiguash: Created the USHIFT-772 issue as a follow-up.

packaging/greenboot/microshift-greenboot-required.sh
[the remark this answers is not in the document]
fzdarsky: Here are the names of the scripts in the optional `greenboot-default-health-checks` RPM:
```
/usr/lib/greenboot/check/required.d/01_repository_dns_check.sh
/usr/lib/greenboot/check/required.d/02_watchdog.sh
/usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh
```

I notice we may want to prefix a numbering to order this check after the base OS checks in case both are installed. Other than that, there doesn't seem to be a naming convention. I would have assumed names express the property being checked (e.g. "microshift_up").

fzdarsky (changes requested):

oglok:
After previous comments, the PR looks good to me. Apparently there's still a requested change from @fzdarsky .

Otherwise, I'm ready to tag it.

packaging/greenboot/microshift-running-check.sh
```diff
@@ -0,0 +1,239 @@
...
+    local ns=${CHECK_PODS_NS}
+    local ct=${CHECK_PODS_CT}
+
+    local status=$(${OCGET_CMD} pods ${OCGET_OPT} -n ${ns} -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}' 2>/dev/null)
+    local tcount=$(echo $status | grep -o True  | wc -l)
+    local fcount=$(echo $status | grep -o False | wc -l)
+
+    # Terminate the script in case more pods are ready than expected - nothing to wait for
+    if [ "${tcount}" -gt "${ct}" ] ; then
+        echo "The number of ready pods in the '${ns}' namespace is greater than the expected '${ct}' count. Terminating..."
```
pmtk, line 125: Do we get any debug output?

packaging/greenboot/microshift-running-check.sh
```diff
@@ -0,0 +1,239 @@
...
+trap 'script_exit' EXIT
+
+# The script exit handler logging the FAILURE or FINISHED message depending
+# on the exit status of the last command
+#
+# args: None
+# return: None
+function script_exit() {
+    [ "$?" -ne 0 ] && status=FAILURE || status=FINISHED
+    echo $status
```
pmtk, line 26: How that works? Scripts sends `kill -TERM` for itself and this gets executed, greenboot makes a decision based on output `FAILURE` or `FINISHED`?

ggiguash: The way bash EXIT handler works is that it's called on the script exit and "$?" contains the status of the last command. So, by adding a handler, we should always have FAILURE or SUCCESS printed in the end of the check, unless the script is killed with SIGKILL.

packaging/greenboot/microshift-running-check.sh
ggiguash: Not at this time. We could add some code in the EXIT handler to print info on FAILURE.
I would do it in a separate PR though because we also need to define what we'd like to see there.

docs/greenboot.md
```diff
@@ -0,0 +1,80 @@
...
+it already includes
+[health check scripts](https://github.com/fedora-iot/greenboot#health-checks-included-with-subpackage-greenboot-default-health-checks)
+verifying that DNS and `ostree` services can be accessed.
+
+Exiting the script with a non-zero status will have the boot declared as failed.
+Greenboot redirects all the script output to the system log, accessible via the
+`journalctl -u greenboot-healthcheck.service` command.
+
+|Validation                                           |Pass  |Fail  |
+|-----------------------------------------------------|------|------|
```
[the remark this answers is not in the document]
ggiguash: It was already implemented, so I added it in the docs.

packaging/greenboot/microshift-running-check.sh
ggiguash: Greenboot only checks the script exit code, never the output. This one is for user-friendly "visualization" of the result in the logs.

docs/devenv_setup.md
```diff
@@ -93,13 +93,18 @@ make srpm
...
 The artifacts of the build are located in the `_output/rpmbuild` directory.
 ```bash
-$ cd ~/microshift/ && find _output -name \*.rpm
-_output/rpmbuild/RPMS/x86_64/microshift-4.13.0_20221215165847_758c8b61-1.el8.x86_64.rpm
-_output/rpmbuild/RPMS/x86_64/microshift-networking-4.13.0_20221215165847_758c8b61-1.el8.x86_64.rpm
-_output/rpmbuild/RPMS/noarch/microshift-release-info-4.13.0_20221215165847_758c8b61-1.el8.noarch.rpm
-_output/rpmbuild/RPMS/noarch/microshift-selinux-4.13.0_20221215165847_758c8b61-1.el8.noarch.rpm
-_output/rpmbuild/SRPMS/microshift-4.13.0_20221215165847_758c8b61-1.el8.src.rpm
-```
+$ cd ~/microshift/_output/rpmbuild && find . -name \*.rpm
```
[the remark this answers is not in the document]
pmtk: Sure, no problem :)

packaging/greenboot/microshift-running-check.sh
pmtk: No prob

packaging/greenboot/microshift-running-check.sh
pmtk: And `kill -TERM` causes the script to exit with `1`?

Fine, I was kind of going over a incredible situation when we'd report status based on `$?` of echo but it's not an issue

PR #1232, opened 2023-01-05
Add arch info for assets/release/release-x86_64.json base section

praveenkumar:
As of now assets/release/release-aarch64.json file have arch info as part of base section `"base": "4.12.0-0.nightly-arm64-2023-01-03-161334"` but it is not present for assets/release/release-x86_64.json. This PR puts the arch info in base same as aarch64 asset file.

### DIFF
--- a/assets/release/release-x86_64.json
+++ b/assets/release/release-x86_64.json
@@ -1,6 +1,6 @@
 {
   "release": {
-    "base": "4.12.0-0.nightly-2023-01-03-161331"
+    "base": "4.12.0-0.nightly-amd64-2023-01-03-161331"
   },
   "images": {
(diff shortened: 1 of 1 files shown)


dhellmann:
This file is managed as part of the automatic rebase process.

https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase.sh#L390

/hold

pmtk:
x86 is missing amd64 because that's how the releases are named, see: https://amd64.ocp.releases.ci.openshift.org/#4.12.0-0.nightly and https://arm64.ocp.releases.ci.openshift.org/#4.12.0-0.nightly-arm64
I don't think we want to mess with this

praveenkumar:
> x86 is missing amd64 because that's how the releases are named, see: https://amd64.ocp.releases.ci.openshift.org/#4.12.0-0.nightly and https://arm64.ocp.releases.ci.openshift.org/#4.12.0-0.nightly-arm64 I don't think we want to mess with this

@pmtk ack I think in our script we have to trim it.

/close

PR #1236, opened 2023-01-09
NP-603: default to system mtu in ovnk

zshi-redhat:
OVNKubernetes builds overlay networking implementation based on geneve
tunnel which adds extra header for packets coming from overlay (Pod) to
underlay network. However, the geneve tunnel is only useful for
cross-nodes communication and not required for microshift single node
deployment. This PR leverages the single-node change[1] in OVNKubernetes
and updates the default overlay MTU to the same size as default uplink
MTU (1500).

[1]: https://github.com/ovn-org/ovn-kubernetes/pull/3290

### DIFF
--- a/docs/network/default_cni_plugin.md
+++ b/docs/network/default_cni_plugin.md
@@ -63,14 +63,16 @@ MicroShift will assume default ovn-kubernetes config values if ovn-kubernetes co
 
 The following configs are supported in ovn-kubernetes config file:
 
-|Field                            |Required |Type    |Default |Description                                                       |Example|
-|:--------------------------------|:--------|:-------|:-------|:-----------------------------------------------------------------|:------|
-|ovsInit.disableOVSInit           |N        |bool    |false   |Skip configuring OVS bridge "br-ex" in microshift-ovs-init.service|true   |
-|ovsInit.gatewayInterface         |N        |string  |""      |Interface to be added in OVS gateway bridge "br-ex"               |eth0   |
-|ovsInit.externalGatewayInterface |N        |string  |""      |Interface to be added in external OVS gateway bridge "br-ex1"     |eth1   |
-|mtu                              |N        |uint32  |1400    |MTU value to be used for the Pods                                 |1300   |
+|Field                            |Required |Type    |Default |Description                                                                  |Example|
+|:--------------------------------|:--------|:-------|:-------|:----------------------------------------------------------------------------|:------|
+|ovsInit.disableOVSInit           |N        |bool    |false   |Skip configuring OVS bridge "br-ex" in microshift-ovs-init.service           |true   |
+|ovsInit.gatewayInterface         |N        |string  |""      |Interface to be added in OVS gateway bridge "br-ex"                          |eth0   |
+|ovsInit.externalGatewayInterface |N        |string  |""      |Interface to be added in external OVS gateway bridge "br-ex1"                |eth1   |
+|mtu                              |N        |int     |*auto*  |MTU value to be used for the Pods, must be less than or equal to the MTU of default route interface|1500|
 
 > When `disableOVSInit` is true, OVS bridge "br-ex" needs to be configured manually. This OVS bridge is required by ovn-kubernetes CNI. See section [OVS bridge](#ovs-bridge) for guidance on configuring the OVS gateway bridge manually.
+> When `gatewayInterface` is not provided, it defaults to the default route interface.
+> When `mtu` is not provided, it defaults to the MTU of `gatewayInterface` interface. In the case that `gatewayInterface` is not specified, it is set to the default route MTU.
 
 Below is an example of `ovn.yaml`:
 
@@ -79,7 +81,7 @@ ovsInit:
   disableOVSInit: true
   gatewayInterface: eth0
   externalGatewayInterface: eth1
-mtu: 1300
+mtu: 1500
 ```
 **NOTE:* The change of `mtu` configuration in `ovn.yaml` requires node reboot to take effect. <br>
 
--- a/pkg/config/ovn/ovn.go
+++ b/pkg/config/ovn/ovn.go
@@ -7,27 +7,25 @@ import (
 	"os"
 	"path/filepath"
 
-	"gopkg.in/yaml.v2"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
 )
 
 const (
 	ovnConfigFileName           = "ovn.yaml"
-	GeneveHeaderLengthIPv4      = 58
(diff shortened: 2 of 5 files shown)


zshi-redhat:
/hold
4.13 task

zshi-redhat:
depends on the ovnk change: https://github.com/ovn-org/ovn-kubernetes/pull/3290, not in openshift yet.

assets/components/ovn/configmap.yaml
```diff
@@ -29,6 +29,7 @@ data:
     [gateway]
     mode=local
     nodeport=true
+    single-node=true
```
dhellmann, line 32: Is this asset managed by the rebase script?

pkg/config/ovn/ovn.go
```diff
@@ -43,7 +41,7 @@ func (o *OVNKubernetesConfig) ValidateOVSBridge(bridge string) error {
 
 func (o *OVNKubernetesConfig) withDefaults() *OVNKubernetesConfig {
 	o.OVSInit.DisableOVSInit = false
-	o.MTU = 1400
+	o.MTU = 1500
```
dhellmann, line 44: The value is defaulted here and down on line 61. Should we have a const for what the default should be to avoid duplicating the literal value?

dhellmann (commented):
This PR would benefit from a description explaining why we want the change. I have 1 nit on implementation and one question inline.

assets/components/ovn/configmap.yaml
zshi-redhat: Not yet.
The ovnk assets commands are currently commented out in the rebase script. 
Adding microshift ovnk manifests to openshift repo doesn't add any extra value for us now since there is no auto-generation of manifests nor CI tests in openshift repo.

pkg/config/ovn/ovn.go
zshi-redhat: Added a const for the default mtu size.

zshi-redhat:
> This PR would benefit from a description explaining why we want the change. I have 1 nit on implementation and one question inline.

Added description in the commit message and updated the PR description.

pmtk:
Looks like infra problems
/retest

zshi-redhat:
@pmtk @pacevedom what's the default VM mtu in CI? I suspect it is less than 1500, can you help to confirm?

zshi-redhat:
> @pmtk @pacevedom what's the default VM mtu in CI? I suspect it is less than 1500, can you help to confirm?

I checked the gcp VM, the default mtu is 1460, which is smaller than 1500.
If I changed the mtu to 1460 in the live CI cluster, the ovnkube pod started successfully.
@pmtk @pacevedom Is there a way to use VM with large mtu 1500? or shall we change the CI config to use 1460 in ovn.yaml?

pacevedom:
> > @pmtk @pacevedom what's the default VM mtu in CI? I suspect it is less than 1500, can you help to confirm?
> 
> I checked the gcp VM, the default mtu is 1460, which is smaller than 1500. If I changed the mtu to 1460 in the live CI cluster, the ovnkube pod started successfully. @pmtk @pacevedom Is there a way to use VM with large mtu 1500? or shall we change the CI config to use 1460 in ovn.yaml?

We can change the MTU to any of these values: https://cloud.google.com/vpc/docs/mtu#valid_mtus
It only takes an option in one of the commands we issue when creating the VM.

I dont have a strong opinion to either change, which one do you think its best? Change OVN config or VM config?

zshi-redhat:
> > > @pmtk @pacevedom what's the default VM mtu in CI? I suspect it is less than 1500, can you help to confirm?
> > 
> > 
> > I checked the gcp VM, the default mtu is 1460, which is smaller than 1500. If I changed the mtu to 1460 in the live CI cluster, the ovnkube pod started successfully. @pmtk @pacevedom Is there a way to use VM with large mtu 1500? or shall we change the CI config to use 1460 in ovn.yaml?
> 
> We can change the MTU to any of these values: https://cloud.google.com/vpc/docs/mtu#valid_mtus It only takes an option in one of the commands we issue when creating the VM.
> 
> I dont have a strong opinion to either change, which one do you think its best? Change OVN config or VM config?

As discussed, we would change the mtu of VM config in CI infra: https://github.com/openshift/release/pull/36067

pacevedom (approved):
/lgtm

zshi-redhat:
/hold cancel
PR was held due to 4.13 master not open.

pacevedom:
/lgtm cancel
Pending detailed conversation

zshi-redhat:
Changed the PR to first detect system mtu (when mtu is not provided by user in ovn.yaml), and falls back to 1500 in case system mtu is unable to get.

docs/network/default_cni_plugin.md
```diff
@@ -68,7 +68,7 @@ The following configs are supported in ovn-kubernetes config file:
 |ovsInit.disableOVSInit           |N        |bool    |false   |Skip configuring OVS bridge "br-ex" in microshift-ovs-init.service|true   |
 |ovsInit.gatewayInterface         |N        |string  |""      |Interface to be added in OVS gateway bridge "br-ex"               |eth0   |
 |ovsInit.externalGatewayInterface |N        |string  |""      |Interface to be added in external OVS gateway bridge "br-ex1"     |eth1   |
-|mtu                              |N        |uint32  |1400    |MTU value to be used for the Pods                                 |1300   |
+|mtu                              |N        |uint32  |1500    |MTU value to be used for the Pods                                 |1300   |
```
ggiguash, line 71: Do we need to update the description?
Also, the example of 1300, can we change it to something that one would actually set?

zshi-redhat: updated

pkg/config/ovn/ovn.go
```diff
@@ -6,24 +6,24 @@ import (
 	"net"
 	"os"
 
-	"gopkg.in/yaml.v2"
+	"github.com/ghodss/yaml"
```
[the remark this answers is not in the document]
zshi-redhat: `gopkg.in/yaml.v2` fails to parse the Interface fields in ovn.yaml (this is actually a bug in 4.12 and 4.13), the current one works. I will change it to `sigs.k8s.io/yaml` in a separate PR.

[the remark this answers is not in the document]
pmtk: Don't worry about that, I was just surprised to see `ghodss/yaml`. It's something very low prio - if we could drop that dependency (less is better?) and use something we must vendor, then why not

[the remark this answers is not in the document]
zshi-redhat: done

PR #1241, opened 2023-01-10
NP-604: Add loadbalancer service controller

pliurh:
With this new controller and OVN-K, microshift is able to support k8s services of loadblancer type. The controller is responsible for updating the status.loadBalancer with the node ip and hostname of the local host. OVN-K is responsible for adding the iptables rules accordingly.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -11,6 +11,7 @@ import (
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/controllers"
 	"github.com/openshift/microshift/pkg/kustomize"
+	"github.com/openshift/microshift/pkg/loadbalancerservice"
 	"github.com/openshift/microshift/pkg/mdns"
 	"github.com/openshift/microshift/pkg/node"
 	"github.com/openshift/microshift/pkg/servicemanager"
@@ -94,6 +95,7 @@ func RunMicroshift(cfg *config.MicroshiftConfig) error {
 	util.Must(m.AddService((controllers.NewVersionManager((cfg)))))
 	util.Must(m.AddService(kustomize.NewKustomizer(cfg)))
 	util.Must(m.AddService(node.NewKubeletServer(cfg)))
+	util.Must(m.AddService(loadbalancerservice.NewLoadbalancerServiceController(cfg)))
 
 	// Storing and clearing the env, so other components don't send the READY=1 until MicroShift is fully ready
 	notifySocket := os.Getenv("NOTIFY_SOCKET")
--- a/pkg/loadbalancerservice/controller.go
+++ b/pkg/loadbalancerservice/controller.go
@@ -1,0 +1,205 @@
+package loadbalancerservice
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/cloud-provider/service/helpers"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/servicemanager"
+)
+
+const defaultInformerResyncPeriod = 10 * time.Minute
+
+type LoadbalancerServiceController struct {
+	NodeIP     string
+	KubeConfig string
+	client     *kubernetes.Clientset
+	indexer    cache.Indexer
+	queue      workqueue.RateLimitingInterface
+	informer   cache.SharedIndexInformer
+}
+
+var _ servicemanager.Service = &LoadbalancerServiceController{}
+
+func NewLoadbalancerServiceController(cfg *config.MicroshiftConfig) *LoadbalancerServiceController {
+	return &LoadbalancerServiceController{
+		NodeIP:     cfg.NodeIP,
+		KubeConfig: cfg.KubeConfigPath(config.KubeAdmin),
+	}
+}
+
+func (c *LoadbalancerServiceController) Name() string {
+	return "microshift-loadbalancer-service-controller"
+}
+
+func (c *LoadbalancerServiceController) Dependencies() []string {
+	return []string{}
+}
+
+func (c *LoadbalancerServiceController) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+	defer close(stopped)
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	config, err := c.restConfig()
+	if err != nil {
+		return errors.Wrap(err, "error creating rest config for service controller")
+	}
+	c.client, err = kubernetes.NewForConfig(config)
+	if err != nil {
+		return errors.Wrap(err, "failed to create clientset for service controller")
+	}
+
+	klog.Infof("Starting service controller")
+
+	factory := informers.NewSharedInformerFactory(c.client, defaultInformerResyncPeriod)
+	serviceInformer := factory.Core().V1().Services()
+	c.informer = serviceInformer.Informer()
+	c.queue = workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	c.indexer = c.informer.GetIndexer()
+	c.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			if err == nil {
+				c.queue.Add(key)
+			}
+		},
+		UpdateFunc: func(old interface{}, new interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(new)
+			if err == nil {
+				c.queue.Add(key)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			if err == nil {
+				c.queue.Add(key)
+			}
+		},
+	})
+
+	factory.Start(stopCh)
+
+	if !cache.WaitForCacheSync(stopCh, c.informer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+
+	go wait.Until(c.runWorker, time.Second, stopCh)
+
+	close(ready)
+
+	<-ctx.Done()
+
+	return ctx.Err()
+}
+
+func (c *LoadbalancerServiceController) runWorker() {
+	for c.processNextItem() {
+	}
+}
+
+func (c *LoadbalancerServiceController) processNextItem() bool {
+	key, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	defer c.queue.Done(key)
+
+	err := c.updateServiceStatus(key.(string))
+	c.handleErr(err, key)
+	return true
+}
+
+// handleErr checks if an error happened and makes sure we will retry later.
+func (c *LoadbalancerServiceController) handleErr(err error, key interface{}) {
+	if err == nil {
+		c.queue.Forget(key)
+		return
+	}
+
+	klog.Infof("Error syncing service %v: %v", key, err)
+
+	// Re-enqueue the key rate limited. Based on the rate limiter on the
+	// queue and the re-enqueue history, the key will be processed later again.
+	c.queue.AddRateLimited(key)
+}
+
+func (c *LoadbalancerServiceController) updateServiceStatus(key string) error {
+	obj, exists, err := c.indexer.GetByKey(key)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


pliurh:
@zshi-redhat @fzdarsky PTAL.
I know we would like to have a generic solution that can work with different CNIs. But I think this can be our first step.

pkg/cmd/run.go
```diff
@@ -107,6 +108,7 @@ func RunMicroshift(cfg *config.MicroshiftConfig, flags *pflag.FlagSet) error {
 	util.Must(m.AddService((controllers.NewVersionManager((cfg)))))
 	util.Must(m.AddService(kustomize.NewKustomizer(cfg)))
 	util.Must(m.AddService(node.NewKubeletServer(cfg)))
+	util.Must(m.AddService(loadbalancerservice.NewLoadbalancerServiceController(cfg)))
```
zshi-redhat, line 111: Should the loadbalancer service start before CNI (infrastructureService)?

pkg/loadbalancerservice/controller.go
```diff
@@ -0,0 +1,154 @@
...
+
+const defaultInformerResyncPeriod = 10 * time.Minute
+
+type LoadbalancerServiceController struct {
+	sync.Mutex
+	NodeName   string
+	NodeIP     string
+	KubeConfig string
+	client     *kubernetes.Clientset
+	myIPs      []string
```
zshi-redhat, line 31: I don't see `myIPs` or `hostCount` used in the loadbalancer controller. why do we need them?

pkg/loadbalancerservice/controller.go
```diff
@@ -0,0 +1,154 @@
...
+					return nil, fmt.Errorf("node port %d occupied", ep.Port)
+				}
+			}
+		}
+	}
+
+	newStatus := &corev1.LoadBalancerStatus{}
+	newStatus.Ingress = append(newStatus.Ingress, corev1.LoadBalancerIngress{
+		IP:       c.NodeIP,
+		Hostname: c.NodeName,
```
zshi-redhat, line 139: I think we record the hostname as the node name the first time microshift starts, after that, the hostname may change, but the recorded node name stay unchanged. I wonder if we shall use the actual hostname here for the loadbalancer service.  @mangelajo wdyt?

pkg/loadbalancerservice/controller.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (c *LoadbalancerServiceController) patchStatus(svc *corev1.Service, newStatus *corev1.LoadBalancerStatus) error {
+	if helpers.LoadBalancerStatusEqual(&svc.Status.LoadBalancer, newStatus) {
+		return nil
+	}
+
+	updated := svc.DeepCopy()
+	updated.Status.LoadBalancer = *newStatus
+	_, err := helpers.PatchService(c.client.CoreV1(), svc, updated)
```
zshi-redhat, line 151: This err is not logged anywhere, shall it be logged here or in the parent function (c.patchStatus) ?

zshi-redhat (commented):
@pliurh I know this is the first step adding generic load balancer for microshift, do you envision how this approach can be used for other CNI implementations, for example tge `bridge-cni`? 
We will make the CNI pluggable in 4.13 which allows the use of other CNIs with upstream microshift, I wonder how this loadbalancer controller can be used by other CNIs, does it need to know the CNI type?

pkg/loadbalancerservice/controller.go
pliurh: removed

pliurh:
> do you envision how this approach can be used for other CNI implementations, for example the bridge-cni?

We can take a similar approach to what K3S does. In my opinion, there're 2 things we need to do for supporting the LB service in microshift: 1) update the 'status.loadBalancer' field of the service as we do in this patch; 2) do the plumbing job to steer the ingress traffic to the pods. For OVN-K, we only need 1), whereas for other CNIs we may need to do both. In K3S's implementation, it does 2) with a daemonset that can insert iptables rules to the host.

> We will make the CNI pluggable in 4.13 which allows the use of other CNIs with upstream microshift, I wonder how this loadbalancer controller can be used by other CNIs, does it need to know the CNI type?

Yes, I suppose we will add a new field in microshift config to specify the CNI type. Then we can add a check in this controller. So that it can know what it needs to do for different CNIs.

pkg/loadbalancerservice/controller.go
pliurh: I've added an error-handling mechanism. The error will be logged in the function `handleErr`.

fzdarsky:
Note our goal in this release cycle is to _remove all CNI-specific logic_ from the MicroShift binary. That is, we cannot add this controller to MicroShift unless it works independently of OVNk / the CNI. If it uses OVNk, we'll need to bundle it with the OVNk CNI RPM package (or think of a separate load balancer plug-in).

pkg/loadbalancerservice/controller.go
```diff
@@ -0,0 +1,206 @@
...
+	KubeConfig string
+	client     *kubernetes.Clientset
+	indexer    cache.Indexer
+	queue      workqueue.RateLimitingInterface
+	informer   cache.SharedIndexInformer
+}
+
+func NewLoadbalancerServiceController(cfg *config.MicroshiftConfig) *LoadbalancerServiceController {
+	return &LoadbalancerServiceController{
+		NodeIP:     cfg.NodeIP,
```
pmtk, line 37: This is a follow up question from our meeting and relates to pluggability of CNI (I didn't had time to look at this PR yet, so take it with grain of salt):
Can we (and should we) use some K8s object to pass information from the CNI to the MicroShift?
Just like CNI is setting annotations on Pods about their iface, CIDR and stuff, we could use Node's annotation or label, or maybe some configmap if we're concerned about too open RBAC

pkg/cmd/run.go
mangelajo: Why/where does it depend on OVN-k?

I believe we should make an agnostic implementation. In the end this would just work by having an nft table / iptable rediecting from the LB port into the NodePort for the host network, no matter what CNI we have.

Otherwise it will complicate the history of CNI pluggability.

mangelajo: @pliurh btw, very cool to see work on this :-) we have been missing it for a long time!

pkg/loadbalancerservice/controller.go
mangelajo: Yes, we record the hostname, and we don't allow subsequent changes. Providing the old one when that does, or throwing an error if the host name was config provided and different from previous ones.

mangelajo:
> > do you envision how this approach can be used for other CNI implementations, for example the bridge-cni?
> 
> We can take a similar approach to what K3S does. In my opinion, there're 2 things we need to do for supporting the LB service in microshift: 1) update the 'status.loadBalancer' field of the service as we do in this patch; 2) do the plumbing job to steer the ingress traffic to the pods. For OVN-K, we only need 1), whereas for other CNIs we may need to do both. In K3S's implementation, it does 2) with a daemonset that can insert iptables rules to the host.
> 
> > We will make the CNI pluggable in 4.13 which allows the use of other CNIs with upstream microshift, I wonder how this loadbalancer controller can be used by other CNIs, does it need to know the CNI type?
> 
> Yes, I suppose we will add a new field in microshift config to specify the CNI type. Then we can add a check in this controller. So that it can know what it needs to do for different CNIs.

Where does the plumbing for this happen in OVNk ? 

My understanding is that for OVNK on regular openshift it was part of the cloud provider redirecting the LB port to the NodePort port on the nodes. But we still need a LB-port -> NodePort-port in the host. We should make this happen in general, not only OVN-k

pliurh:
In OVN-K, the plumbing part has already been implemented. It will watch the change in the status of the LB service, then add the iptables rules. https://github.com/openshift/ovn-kubernetes/blob/master/docs/external-ip-and-loadbalancer-ingress.md.

pkg/cmd/run.go
pliurh: OVN-K will add the iptable rules. It will add a DNAT rule to redirect the LB port to the clusterIP.

I don't think a complete CNI agnostic way is realistic. We cannot assume what a CNI will do with the LB service. An agnostic way may conflict with the per-CNI behavior.

pliurh:
> Note our goal in this release cycle is to _remove all CNI-specific logic_ from the MicroShift binary. That is, we cannot add this controller to MicroShift unless it works independently of OVNk / the CNI. If it uses OVNk, we'll need to bundle it with the OVNk CNI RPM package (or think of a separate load balancer plug-in package).

The logic of the controller is common and required for all the CNIs to support the LB service. It can work with OVN-K out-of-box as OVN-K has already got the plumbing function. For other CNIs, we may need to implement the plumbing logic. It could be adding iptables rule or any CNI-specific ways.

pkg/loadbalancerservice/controller.go
pliurh: I checked the API description. It looks like we don't need to set the hostName. It shall be used by a DNS based LB provided by cloud providers.

pliurh:
I created an enhancement PR at https://github.com/openshift/enhancements/pull/1316. Maybe we can continue our discussion there.

pkg/loadbalancerservice/controller.go
mangelajo: We already pass this down to CNI from configuration of MicroShift, so it should be equal


Also, some other components before CNI need this IP, and we pass it from MicroShift down to the components initialized before CNI... so it's MicroShift -> CNI what we have to do (as far as I understood).

pkg/cmd/run.go
mangelajo: Yes, that's partly true. May be for other CNIs we will have to add something that does the same handling in terms of iptables.

pliurh:
/hold
Waiting until the enhancement get accepted.

pmtk (commented):
I'd like to understand a little bit more how that works:
- Admin deploys an app that include Service with type: LoadBalancer
- From K8s perspective LoadBalancer builds on top of NodePort and NodePorts builds on top of ClusterIP: requesting LB causes a NodePort to be set up.
- In cloud environment external LB needs an information about where to direct the traffic which would be the "imaginary NodePort that is part of LoadBalancer svc", right?

Looking at the code it looks like we're leveraging fact that NodePort gets created and just adding NodeIP (from ushift config) to Service's `status.loadBalancer.ingress.IP`?
Why would this not work with other CNIs? They handle NodePorts, and we just fill in the ingress IP.

----------------------
## EDIT (also put in separate comment)

> Looking at the code it looks like we're leveraging fact that NodePort gets created and just adding NodeIP (from ushift config) to Service's `status.loadBalancer.ingress.IP`? Why would this not work with other CNIs? They handle NodePorts, and we just fill in the ingress IP.

Okay, so I see that beside having NodePort (30xxxx port) it also exposes the port as NodeIP:port (e.g. 8080).
And this is the part that OVN-K handles by itself, but for others we would need something like service/klipper-lb in K3s to set up 192.168.122.X:8080 -> 10.42.0.Y:8080 (Pod), right?

pmtk:
> Looking at the code it looks like we're leveraging fact that NodePort gets created and just adding NodeIP (from ushift config) to Service's `status.loadBalancer.ingress.IP`? Why would this not work with other CNIs? They handle NodePorts, and we just fill in the ingress IP.

Okay, so I see that beside having NodePort (30xxxx port) it also exposes the port as NodeIP:port (e.g. 8080).
And this is the part that OVN-K handles by itself, but for others we would need something like service/klipper-lb in K3s to set up 192.168.122.X:8080 -> 10.42.0.Y:8080 (Pod), right?

pkg/loadbalancerservice/controller.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+	return nil
+}
+
+func (c *LoadbalancerServiceController) restConfig() (*rest.Config, error) {
+	return clientcmd.BuildConfigFromFlags("", c.KubeConfig)
+}
+
+func (c *LoadbalancerServiceController) getNewStatus(svc *corev1.Service) *corev1.LoadBalancerStatus {
```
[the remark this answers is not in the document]
pliurh: You got it right. I made the change

pliurh:
> And this is the part that OVN-K handles by itself, but for others we would need something like service/klipper-lb in K3s to set up 192.168.122.X:8080 -> 10.42.0.Y:8080 (Pod), right?

Correct.

pkg/loadbalancerservice/controller.go
```diff
@@ -0,0 +1,205 @@
...
+		klog.Infof("Service %s does not exist anymore", key)
+	} else {
+		svc := obj.(*corev1.Service)
+		if svc.Spec.Type != corev1.ServiceTypeLoadBalancer {
+			return nil
+		}
+		klog.Infof("Process service %s/%s", svc.Namespace, svc.Name)
+
+		newStatus := c.getNewStatus(svc)
+		if len(newStatus.Ingress) == 0 {
```
pmtk, line 157: I feel like this is too implicit. Can `getNewStatus()` return `(*corev1.LoadBalancerStatus, error)` and error would happen in this place:
```go
				if ep.Port == np.Port {
					klog.Infof("Node port %d occupied", ep.Port)
					return newStatus
				}
```
new:
```go
				if ep.Port == np.Port {
					return nil, fmt.Error("Node port %d occupied", ep.Port)
				}
```

pliurh: done

PR #1252, opened 2023-01-12
Updated devevn_rhel8 doc with

vishnoianil:
*.* update instruction to install go-toolset
*.* remove make clean in RPM Packages section, because it will delete Microshift executable.
*.* Added troubleshooting section

Signed-off-by: Anil Kumar Vishnoi <avishnoi@redhat.com>

### DIFF
--- a/docs/devenv_rhel8.md
+++ b/docs/devenv_rhel8.md
@@ -246,15 +246,13 @@ go tool pprof heap.pprof
 To view all the available profiles, run `oc get --raw /debug/pprof`.
 
 ## Troubleshooting
-### No valid repository ID for fast-datapath
-If you encounter following error message while enabling fast-datapath repo:
+### No Valid Repository ID for OpenShift RPM Channels
+If you encounter following error message while enabling fast-datapath repo (You can encounter similar error for other OpenShift related repos as well)
 
 ```
-# sudo subscription-manager repos \
->     --enable fast-datapath-for-rhel-8-$(uname -i)-rpms
 Error: 'fast-datapath-for-rhel-8-x86_64-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.
 ```
-Please make sure that your system is registered and attached to the `Red Hat Openshift Container Platform` subscription. Once the repo is enabled, you must see atleast following repo enabled.
+Make sure that your system is registered and attached to the `Red Hat Openshift Container Platform` pr equivalent subscription. Once the repo is enabled, you must see at least the following repos enabled.
 
 ```
 # sudo subscription-manager repos --list-enabled
--- a/docs/devenv_rhel8.md
+++ b/docs/devenv_rhel8.md
@@ -60,7 +60,7 @@ Run the following commands to configure SUDO, upgrade the system, install basic 
 echo -e 'microshift\tALL=(ALL)\tNOPASSWD: ALL' | sudo tee /etc/sudoers.d/microshift
 sudo dnf clean all -y
 sudo dnf update -y
-sudo dnf install -y git cockpit make gcc selinux-policy-devel rpm-build bash-completion go-toolset
+sudo dnf install -y git cockpit make golang selinux-policy-devel rpm-build bash-completion
 sudo systemctl enable --now cockpit.socket
 ```
 You should now be able to access the VM Cockpit console using `https://<vm_ip>:9090` URL.
@@ -76,6 +76,7 @@ cd ~/microshift
 ### Executable
 Run `make` command in the top-level directory. If necessary, add `DEBUG=true` argument to the `make` command for building a binary with debug symbols.
 ```bash
+make clean
 make
 ```
 The artifact of the build is the `microshift` executable file located in the `_output/bin` directory.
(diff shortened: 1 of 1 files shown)


docs/devenv_rhel8.md
```diff
@@ -83,7 +83,6 @@ The artifact of the build is the `microshift` executable file located in the `_o
 ### RPM Packages
 Run make command with the `rpm` or `srpm` argument in the top-level directory. 
 ```bash
-make clean
```
ggiguash, line 86: We need clean because these instructions are reused in the automation script and it should be idempotent. If clean is not present, the RPM install command may fail because it'll pick up RPMs from more than one build.

docs/devenv_rhel8.md
```diff
@@ -60,7 +60,7 @@ Run the following commands to configure SUDO, upgrade the system, install basic
 echo -e 'microshift\tALL=(ALL)\tNOPASSWD: ALL' | sudo tee /etc/sudoers.d/microshift
 sudo dnf clean all -y
 sudo dnf update -y
-sudo dnf install -y git cockpit make gcc selinux-policy-devel rpm-build bash-completion
+sudo dnf install -y git cockpit make gcc selinux-policy-devel rpm-build bash-completion go-toolset
```
ggiguash, line 63: Why is this package needed? Also, if we change anything here, automation script need to be updated.

docs/devenv_rhel8.md
```diff
@@ -245,3 +244,35 @@ go tool pprof heap.pprof
 ```
 
 To view all the available profiles, run `oc get --raw /debug/pprof`.
+
+## Troubleshooting
+### No valid repository ID for fast-datapath
+If you encounter following error message while enabling fast-datapath repo:
+
+```
+# sudo subscription-manager repos \
```
ggiguash, line 253: Let's remove the command and leave the message alone. In general, the same problem may occur for rhocp channel enablement

docs/devenv_rhel8.md
```diff
@@ -245,3 +244,35 @@ go tool pprof heap.pprof
 ```
 
 To view all the available profiles, run `oc get --raw /debug/pprof`.
+
+## Troubleshooting
+### No valid repository ID for fast-datapath
```
ggiguash, line 249:
```suggestion
### No Valid Repository ID for OpenShift RPM Channels
```

docs/devenv_rhel8.md
```diff
@@ -245,3 +244,35 @@ go tool pprof heap.pprof
...
+If you encounter following error message while enabling fast-datapath repo:
+
+```
+# sudo subscription-manager repos \
+>     --enable fast-datapath-for-rhel-8-$(uname -i)-rpms
+Error: 'fast-datapath-for-rhel-8-x86_64-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.
+```
+Please make sure that your system is registered and attached to the `Red Hat Openshift Container Platform` subscription. Once the repo is enabled, you must see atleast following repo enabled.
```
ggiguash, line 257:
```suggestion
Make sure that your system is registered and attached to the `Red Hat Openshift Container Platform` or equivalent subscription. Once the repo is enabled, you must see at least the following repos enabled.
```

docs/devenv_rhel8.md
vishnoianil: I was following the instructions in the doc manually (as i believe most of the folks trying to setup their local environment). In the `Executable` section above, make builds the microshift executable, but `make clean` in this section deletes that, and you need to make it again. I think having a specific make target for cleaning executable and RPMS would be better i think. what do you think?

docs/devenv_rhel8.md
vishnoianil: It installs golang, otherwise `make` won't build the microshift executable, it throws error about `go` not present on the system (atleast in RHEL 8.7)

docs/devenv_rhel8.md
vishnoianil: done.

docs/devenv_rhel8.md
vishnoianil: done

docs/devenv_rhel8.md
vishnoianil: done

docs/devenv_rhel8.md
ggiguash: Right. We missed reverting this line when we moved to go installation from RPM. The command used in the automation is the correct one (see https://github.com/openshift/microshift/blob/main/scripts/devenv-builder/configure-vm.sh#L47) 

```suggestion
sudo dnf install -y git cockpit make golang selinux-policy-devel rpm-build bash-completion
```

docs/devenv_rhel8.md
ggiguash: I would not add a separate target just for deleting microshift executable. Note that RPM build rebuilds its own executable in a different location.
What about moving the 'make clean' before 'make' in the previous section?

ggiguash (changes requested):

docs/devenv_rhel8.md
vishnoianil: Done.

docs/devenv_rhel8.md
vishnoianil: yeah, that's reasonable. Moved make clean to Executable section.

ggiguash:
tide/merge-method-squash

PR #1256, opened 2023-01-14
USHIFT-758: Include arm release in rebase branch name

pmtk:
Currently only an amd64 release tag is used for branch name. This means that if new amd64 nightly is not produced, but there's a new ARM64 nightly we won't update images for arm until next amd64 nightly.

### DIFF
--- a/scripts/auto-rebase/rebase.sh
+++ b/scripts/auto-rebase/rebase.sh
@@ -31,6 +31,7 @@ trap 'echo "Script exited with error."' ERR
 REPOROOT="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/../..")"
 STAGING_DIR="$REPOROOT/_output/staging"
 PULL_SECRET_FILE="${HOME}/.pull-secret.json"
+RELEASE_TAG_RX="(.+)-([0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]{6})"
 
 EMBEDDED_COMPONENTS="route-controller-manager cluster-policy-controller hyperkube etcd"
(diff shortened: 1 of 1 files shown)


scripts/auto-rebase/rebase.sh
```diff
@@ -698,7 +713,7 @@ rebase_to() {
     fi
 
     update_images
-    if [[ -n "$(git status -s pkg/release)" ]]; then
+    if [[ -n "$(git status -s pkg/release packaging/crio.conf.d)" ]]; then
```
dhellmann, line 716: What's going on here?

pmtk: Something I discovered testing manually. If no changes are in `pkg/release` we won't `git add packaging/crio.conf.d` resulting in not committing it - I think it's just pause image but still

pmtk:
added rationale to the PR's description
/cc @ggiguash

scripts/auto-rebase/rebase.sh
```diff
@@ -672,8 +673,22 @@ rebase_to() {
...
+    fi
+
+    if [[ "${release_image_arm64#*:}" =~ ${RELEASE_TAG_RX} ]]; then
+        arm64_date="${BASH_REMATCH[2]}"
+    else
+        echo "Failed to match regex against arm64 image tag"
+        exit 1
+    fi
+
+    rebase_branch="rebase-${ver_stream}+amd64-${amd64_date}+arm64-${arm64_date}"
```
ggiguash, line 691: Is this necessary for creating a branch name? I mean, why not to simply have rebase-${branch}-${timestamp}?

Since the name is transient, it may simplify the code, I think

pmtk: We have this flow in create_pr.py that if no new nightly was produced, but we merged something to microshift's main it'll update already existing branch (and PR) with "rebased" (on newer main) changes

ggiguash: Understood. If that's what we need, the code seems to be fine.

PR #1272, opened 2023-01-19
NOISSUE: look up rebase version and date in the image if needed

dhellmann:
By default we use version and date information from the image pull
spec. If that information is not available, fall back to looking
inside the release info file. This allows rebasing onto public
versions using pullspecs like
quay.io/openshift-release-dev/ocp-release:4.12.0-x86_64

/assign @pmtk
/cc @jogeo


scripts/auto-rebase/rebase.sh
```diff
@@ -816,15 +816,17 @@ rebase_to() {
...
+        amd64_date="$(cat ${STAGING_DIR}/release_amd64.json | jq -r .config.created | cut -f1 -dT)"
     fi
 
     if [[ "${release_image_arm64#*:}" =~ ${RELEASE_TAG_RX} ]]; then
         arm64_date="${BASH_REMATCH[2]}"
     else
-        echo "Failed to match regex against arm64 image tag"
-        exit 1
+        echo "Failed to match regex against arm64 image tag, using release file"
+        ver_stream="$(cat ${STAGING_DIR}/release_arm64.json | jq -r '.config.config.Labels["io.openshift.release"]')"
```
pmtk, line 828:
```suggestion
```
We already have it from amd

dhellmann: Good point, fixed.

I wonder if there's any reason to keep trying to parse the image name at all? Would it make sense to just look at the metadata inside the image all the time?

dhellmann:
Retroactively linking to USHIFT-792

PR #1273, opened 2023-01-19
WIP: feat: add staticly defined workload partitioning

eggfoobar:
Signed-off-by: ehila <ehila@redhat.com>

### DIFF
--- a/packaging/crio.conf.d/microshift-ovn.conf
+++ b/packaging/crio.conf.d/microshift-ovn.conf
@@ -2,8 +2,3 @@
 # ovn-kubernetes is the name configured by OVN-K in /etc/cni/net.d/ config file
 # by declaring this CRI-O will wait until that network is configured.
 cni_default_network = "ovn-kubernetes"
-
-[crio.runtime.workloads.management]
-activation_annotation = "target.workload.openshift.io/management"
-annotation_prefix = "resources.workload.openshift.io"
-resources = { "cpushares" = 0, "cpuset" = "0" }
--- a/packaging/kubernetes/openshift-workload-pinning
+++ b/packaging/kubernetes/openshift-workload-pinning
@@ -1,0 +1,5 @@
+{
+    "management": {
+        "cpuset": "0"
+    }
+}
--- a/packaging/crio.conf.d/microshift-workloads.conf
+++ b/packaging/crio.conf.d/microshift-workloads.conf
@@ -1,0 +1,4 @@
+[crio.runtime.workloads.management]
+activation_annotation = "target.workload.openshift.io/management"
+annotation_prefix = "resources.workload.openshift.io"
+resources = { "cpushares" = 0, "cpuset" = "0" }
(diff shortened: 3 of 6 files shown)


packaging/crio.conf.d/microshift-ovn.conf
```diff
@@ -2,8 +2,3 @@
 # ovn-kubernetes is the name configured by OVN-K in /etc/cni/net.d/ config file
 # by declaring this CRI-O will wait until that network is configured.
 cni_default_network = "ovn-kubernetes"
-
```
dhellmann, line 5: I think these settings were a key part of managing the size of the CNI implementation, so I'm worried that moving them to a separate file is going to cause problems when the new optional RPM is not installed.

cc @zshi-redhat @mangelajo

dhellmann (commented):
When I talked to @ggiguash about this, I said if we could do it entirely with user documentation we should do that but if we needed code changes we should have an enhancement to ensure the whole team (including docs and QE) understands the goal, we can decide if it should be optional or always in place, and we can avoid splitting the discussions of those topics across multiple slack threads.

I have 1 comment on the implementation inline. Let me know how you feel about the other questions that could lead to us wanting an enhancement.

packaging/crio.conf.d/microshift-ovn.conf
eggfoobar: I see. The only thing we need to clarify is if the desire is that the default behavior is single CPUSet for microshift as a whole. If that's the case then I think encapsulating that configuration in it's own file seems easier to maintain. Since CRIO will merge these values and the last file to be read in updates the value then the workload file will always override this value. However, if the optional CNI config has it's own limits, then it makes perfect sense for it to be delivered with those limits, either way though, there should be no issue with adding this back in.

dhellmann: OK, let's see what Zenghui and Miguel think when they have a chance to take a look.

zshi-redhat: It makes sense to move the workload crio settings to its own file. ovnk can still use it w/o any changes (the annotation was already added in ovnk daemonset), other microshift pods can benefit from the settings as well.

zshi-redhat: Regarding the default behavior to pin a single cpu, I think it is reasonable in most cases since our target is small form factor device. It may become useful to make the cpuset configurable if the target changes in the future.

packaging/kubernetes/openshift-workload-pinning
```diff
@@ -0,0 +1,5 @@
+{
```
zshi-redhat, line 1: How is it configured in openshift?

eggfoobar: On versions prior to 4.12 and earlier, this was configured by MCO with the customer defining these configurations or we would via things like ZTP. Going forward from 4.13, this is now controlled by NTO.

ggiguash:
We will implement this feature in 4.14 release. Closing the PR for now - to be reviewed and revived when the 4.14 development cycle starts.
/close

PR #1278, opened 2023-01-20
USHIFT-377: verify if assets/** are covered in rebase.sh

pmtk:


scripts/auto-rebase/presubmit.py
```diff
@@ -0,0 +1,174 @@
...
+    def pred(line):
+        return (line.find("cp ") != -1 or
+                line.find("git restore") != -1 or
+                line.find("envsubst <") != -1)
+
+    with open(REBASE_SCRIPT, "r") as f:
+        contents = f.read()
+        start = contents.find(UPDATE_MANIFESTS_START)
+        end = contents.find(UPDATE_MANIFESTS_END, start) + len(UPDATE_MANIFESTS_END)
+        return [line for line in contents[start:end].splitlines() if pred(line)]
```
dhellmann, line 52: Is there a way for us to restructure the rebase command to make this need to parse it less complex (and possibly fragile)? For example, what if we had a separate text file of all of the files that need to be copied into the assets tree from a release payload? The rebase script could read that file, and so could this function.

pmtk: If we'd put all `cp` in the text file, file by file (no globs) - that would definitely simplify validation.
But I'm not sure if that would be so simple - there're some exceptions and quirks like clearing dirs, `git restore`-ing certain files.

We would have to tear apart current organization of the function, i.e.:
```
    #-- OpenShift control plane ---------------------------
    #-- openshift-dns -------------------------------------
    #-- openshift-router ----------------------------------
    #-- service-ca ----------------------------------------
    #-- ovn-kubernetes -----------------------------------
```
and each has following steps:
```
    # 1) Adopt resource manifests
    # 2) Render operand manifest templates like the operator would
    # 3) Make MicroShift-specific changes
    # 4) Replace MicroShift templating vars (do this last, as yq trips over Go templates)
```

If we'd run `rm`s first, then process that list with files to copy and then go back to `yq`-ing, `sed`-ing, and `git restore`-ing then this structure falls apart.
One thing that might be misleading is having to edit two files: list to copy and rebase.sh to make modifications, but other than that I think it would make the process (both adding new entries and validating) more robust.
We could go even further and move everything from `update_manifests()` to one file (json? yaml? let say a CR of some sort :D)  but not sure how time consuming would that be and would have to be tested quite hard

dhellmann: OK, I didn't realize the existing logic was so complicated.

I'll push a little bit again on making it easier to parse: Would it help to have comments or other markers in the script? Something that could be an explicit interface to this code, instead of relying on code style conventions?

If you're confident of keeping the 2 things in sync as they are, we can move ahead, but I have to ask. :-)

PR #1279, opened 2023-01-20
USHIFT-543: Delete old rebase branches

pmtk:


scripts/auto-rebase/rebase.py
```diff
@@ -308,13 +327,16 @@ def main():
...
     pr = try_get_pr(gh_repo, org, base_branch, rebase_branch_name)
     if pr == None:
         pr = create_pr(gh_repo, base_branch, rebase_branch_name, pr_title, desc)
-        post_comment(pr)
     else:
         update_pr(pr, pr_title, desc)
-        post_comment(pr, f"Rebase job updated the branch\n{desc}")
+        comment = f"Rebase job updated the branch\n{desc}"
+
+    cleanup_branches(gh_repo)
```
dhellmann, line 338: Now that we run this script for each branch, is there any concern about a race condition if we do this in every job? Should it be standalone?

ggiguash: Or, maybe, delete only the stale branches in the current release where it is run.

pmtk: Or this logic could execute only when job runs for `main` branch (i.e. current release under development) but scan delete branches for all releases. This way we won't have to backport this immediately to 4.12 (which still uses old `create_pr.py` - at least until we think that `rebase.py` is good enough to backport it and don't touch anymour)

ggiguash: Yes, I'm for limiting this functionality to the main branch - good idea

dhellmann: I like the idea of having the special logic only in the main branch job.

pmtk: Changed

scripts/auto-rebase/rebase.py
```diff
@@ -258,6 +258,25 @@ def get_expected_branch_name(amd, arm):
...
+        prs = gh_repo.get_pulls(head=f"{gh_repo.owner.login}:{branch.name}", state="all")
+        all_prs_are_closed = all([pr.state == "closed" for pr in prs])
+        logging.info(f"'{branch.name}' is referenced in following PRs: " + ", ".join([f"#{pr.number} ({pr.state})" for pr in prs]))
+        if all_prs_are_closed:
+            ref = gh_repo.get_git_ref(f"heads/{branch.name}")
+            deleted_branches.append(branch.name)
+            if REMOTE_DRY_RUN:
+                logging.info(f"[DRY RUN] Delete '{ref.ref}'")
+            else:
+                ref.delete()
```
dhellmann, line 275: What sorts of conditions might cause this to throw an exception?

pmtk: Right... it's http resp >= 400. Added a try-except. Thanks!

PR #1280, opened 2023-01-20
USHIFT-781: add 'get' command line client

dhellmann:
Expose the embedded 'kubectl get' command for use in greenboot and
sos.


ggiguash:
Tested in private workspace. Very cool! 
/lgtm

PR #1291, opened 2023-01-24
Update MetalLB how to with instructions for k8s 1.25+

### LINKED ISSUE #1290: [BUG] MetalLB instructions fail on k8s 1.25+
#### What happened?

I followed the documentation on [Deploying a TCP Load Balancer for User Workloads](https://github.com/openshift/microshift/blob/main/docs/howto_load_balancer.md) on my MicroShift cluster running on OpenShift 4.12.  The instructions to deploy resource manifests from the metallb project do not work because `PodSecurityPolicy` which was previously deprecated, has now been removed from OpenShift 4.12 / Kubernetes 1.25.

Applying the YAML manifests to deploy the project should work.

1. Deploy MicroShift on OpenShift 4.12.
2. Run the instructions to apply the resource manifests from metallb project:
```
oc apply -f https://raw.githubusercontent.com/metallb/metallb/v0.13.4/config/manifests/metallb-native.yaml
```

It should fail when attempting to create objects of type `PodSecurityPolicy`.

#### Anything else we need to know?

I've already found a solution, which is to use updated manifests from that project:
```
oc apply -f https://raw.githubusercontent.com/metallb/metallb/v0.13.7/config/manifests/metallb-native.yaml
```

Afterwards, I still needed to apply the privileged scc to the controller and speaker serviceaccounts as follows:
```
oc adm policy add-scc-to-user privileged -z controller -n metallb-system
oc adm policy add-scc-to-user privileged -z speaker -n metallb-system
```

#### Environment
- MicroShift version (use `microshift version`):
$ microshift version
MicroShift Version: 4.12.0-0.microshift-2022-11-17-084702-untagged
Base OCP Version: 4.12.0-0.nightly-2022-11-07-181244

- Hardware configuration:
Intel NUC with 2 cores and 16 GB RAM

$ cat /etc/os-release
VERSION="8.7 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.7 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/8/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.7
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"

- Kernel (e.g. `uname -a`):
$ uname -a
Linux [hostname redacted] 4.18.0-425.10.1.el8_7.x86_64 #1 SMP Wed Dec 14 16:00:01 EST 2022 x86_64 x86_64 x86_64 GNU/Linux

- Others:

kevchu3:
**Which issue(s) this PR addresses**: #1290 

Closes #1290

### DIFF
--- a/docs/config/nginx-IP-header.yaml
+++ b/docs/config/nginx-IP-header.yaml
@@ -30,6 +30,8 @@ spec:
         - name: nginx-configs
           subPath: headers.conf
           mountPath: /etc/nginx/conf.d/headers.conf
+      securityContext:
+        runAsUser: 0
       volumes:
         - name: nginx-configs
           configMap:
--- a/docs/howto_load_balancer.md
+++ b/docs/howto_load_balancer.md
@@ -18,7 +18,9 @@ oc get pods -A
 Log into the virtual machine and run the following commands to create the `MetalLB` namespace and deployment.
 
 ```
-oc apply -f https://raw.githubusercontent.com/metallb/metallb/v0.13.4/config/manifests/metallb-native.yaml
+oc apply -f https://raw.githubusercontent.com/metallb/metallb/v0.13.7/config/manifests/metallb-native.yaml
+oc adm policy add-scc-to-user privileged -z controller -n metallb-system
+oc adm policy add-scc-to-user privileged -z speaker -n metallb-system
 ```
 
 Verify that the `MetalLB` pods are up and running in the `metallb-system` namespace.
--- a/docs/howto_load_balancer.md
+++ b/docs/howto_load_balancer.md
@@ -48,11 +48,12 @@ EOF
 ```
 
 ## Install User Workload
-Log into the virtual machine and create the namespace to be used for deploying the test application.
+Log into the virtual machine and create the namespace to be used for deploying the test application.  Be aware that this test application will run as root user, which is not a security best practice.
 
 ```bash
 NAMESPACE=nginx-lb-test
 oc create ns $NAMESPACE
+oc adm policy add-scc-to-user privileged -z default -n $NAMESPACE
 ```
 
 Run the following command to deploy **3 replicas** of a test `nginx` application in the specified namespace.


ggiguash:
Thank you, @kevchu3. 
If we are fixing this document, can we also look at the `nginx` installation and test? It should have the same problem, shouldn't it?

kevchu3:
Hi @ggiguash,

Here are my thoughts on applying this to the nginx example application.  As a best practices, end user OpenShift (and subsequently MicroShift) applications should not be running as privileged pods (i.e. root).  The recommendations for the metallb-system load balancers are different as they are part of the infrastructure and may require root.  I looked into the docker.io/library/nginx:1.18 image that was being used, and it is indeed intended to be run as root, which may be fine with other k8s distros but is not best practices for MicroShift.  In fact, a quick traversal of the image showed mostly all of the filesystem including configurations needed by nginx are owned by root user and group.  That being said, the proper approach would likely be to select another base image that adheres to OpenShift standards of running non-privileged containers.  However, I couldn't immediately find one that was consumable for nginx which didn't require us to build from source.

The other item to call out for our nginx application is that there is no deprecated/removed PodSecurityPolicy being applied in the manifests.  Thus, the issues affecting this nginx application when deploying in OpenShift are entirely different, and have persisted well before k8s 1.25+ removed PodSecurityPolicy.

Anyways, I'll walk through a hack workaround to resolve the issue for purposes of testing.  Here are my results:

First, I ran the [original instructions](https://github.com/openshift/microshift/blob/main/docs/howto_load_balancer.md#install-user-workload):

```
NAMESPACE=nginx-lb-test
oc create ns $NAMESPACE
oc apply -n $NAMESPACE -f https://raw.githubusercontent.com/openshift/microshift/main/docs/config/nginx-IP-header.yaml
```

Which subsequently runs into the issue you've observed where the pod can't modify files on the filesystem.  Note that `whoami` confirms the pod runs as the high UID in MicroShift.

```
$ oc get pods
NAME                     READY   STATUS             RESTARTS      AGE
nginx-65fd5cf854-cnvhw   0/1     CrashLoopBackOff   3 (52s ago)   90s
nginx-65fd5cf854-gnjz2   0/1     CrashLoopBackOff   3 (46s ago)   90s
nginx-65fd5cf854-trpt2   0/1     CrashLoopBackOff   3 (50s ago)   90s

$ oc logs nginx-65fd5cf854-cnvhw
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: error: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2023/01/25 17:37:50 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2023/01/25 17:37:50 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
[root@microshift-starter ~]# oc debug pod/nginx-65fd5cf854-cnvhw
Starting pod/nginx-65fd5cf854-cnvhw-debug ...
Pod IP: 10.42.0.33
If you don't see a command prompt, try pressing enter.

$ cd /etc/nginx/conf.d	
$ ls -l
total 8
-rw-r--r--. 1 root root       1093 Oct 29  2020 default.conf
-rw-r--r--. 1 root 1000160000   44 Jan 25 17:38 headers.conf
$ touch newfile
touch: cannot touch 'newfile': Permission denied
$ whoami
1000160000
```

I then apply the privileged annotation to the namespace.  Note that `pod-security.kubernetes.io/enforce: privileged` is now set.

```
$ oc adm policy add-scc-to-user privileged -z default -n $NAMESPACE
$ oc get ns $NAMESPACE -o yaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/sa.scc.mcs: s0:c13,c2
    openshift.io/sa.scc.supplemental-groups: 1000160000/10000
    openshift.io/sa.scc.uid-range: 1000160000/10000
  creationTimestamp: "2023-01-25T17:37:45Z"
  labels:
    kubernetes.io/metadata.name: nginx-lb-test
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/enforce-version: v1.24
  name: nginx-lb-test
  resourceVersion: "109157"
  uid: a11ba70c-6ae8-4913-8b88-9e2dd784f82f
spec:
  finalizers:
  - kubernetes
status:
```

However, this isn't sufficient because the pod still expects to run as the high UID (in my case 1000160000).  What we need to do is set `spec.securityContext.runAsUser: 0` in the definition of the container inside the Deployment to explicitly tell the pod to run as root.  My disclaimer still applies as this is NOT a security best practice.  Here is the complete YAML manifest and should probably replace the existing [nginx-IP-header.yaml](https://raw.githubusercontent.com/openshift/microshift/main/docs/config/nginx-IP-header.yaml) example:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx
data:
  headers.conf: |
    add_header X-Server-IP $server_addr always;
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      securityContext:
        runAsUser: 0
      containers:
      - image: nginx:1.18
        imagePullPolicy: Always
        name: nginx
        ports:
        - containerPort: 80
        volumeMounts:
        - name: nginx-configs
          subPath: headers.conf
          mountPath: /etc/nginx/conf.d/headers.conf
      volumes:
        - name: nginx-configs
          configMap:
            name: nginx
            items:
              - key: headers.conf
                path: headers.conf
```

I'd like to hear your thoughts on whether we should move forward with this example for purposes of testing, before I go ahead and update the PR to include these changes.

ggiguash:
@kevchu3, I think your suggestion makes a lot of sense. The purpose of this example is to demonstrate MetalLB load balancing, so having a simple-enough test is preferrable. 

Please, update the PR with your proposal. Is only running as root user sufficient, or we also need to apply scc?
If scc change is necessary, let's do it right after creating the namespace.

kevchu3:
Looks like the SCC is required.  Here's what happens when I create the namespace and workload but don't apply the SCC.  I'll create a PR with your latest recommendations.
```
$ oc get events
LAST SEEN   TYPE      REASON              OBJECT                       MESSAGE
5s          Warning   FailedCreate        replicaset/nginx-9d96b8987   Error creating: pods "nginx-9d96b8987-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, spec.containers[0].securityContext.runAsUser: Invalid value: 0: must be in the ranges: [1000130000, 1000139999], provider "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "topolvm-node": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```

ggiguash:
@kevchu3 , thank you for your help.
/lgtm

PR #1295, opened 2023-01-26
USHIFT-791: Rebase: don't change Corefile in DNS's ConfigMap

pmtk:
DNS's configmap.yaml is owned by MicroShift (it's a file that would be generated by the operator)


dhellmann (commented):
I think this is going to undo the work from https://github.com/openshift/microshift/pull/1203

pmtk:
> I think this is going to undo the work from #1203

That is correct. This `sed` command is not necessary:
- we own the file (it's not from `cluster-dns-operator`, it's `git restore`d), so we can keep it exactly as needed directly in repo
- sed eval is deleting `forward` command leaving only it's argument (the template), which renders Corefile invalid resulting in coredns failing to start (https://github.com/openshift/microshift/pull/1293/files#r1087533434)

dhellmann:
Thanks for explaining all of that. It looks good, do we need to link this PR to a bug?

pmtk:
Let's try keeping process clean :)

dhellmann:
Thanks, yeah, we're on a path to start requiring jira tickets on PRs across the org, so it's good to get into the habit. I honestly wasn't sure if there was already a ticket and it just wasn't linked or if there was no ticket, yet.

/lgtm

zshi-redhat:
@pmtk thanks for the fix!

PR #1296, opened 2023-01-26
rebase 1185

mangelajo:
This is a rebase of https://github.com/openshift/microshift/pull/1185 with a branch on the openshift/microshift repo to allow @jogeo to test on CI.

- Watch firewall reload in sysconfigwatcher
- Move firewall watch logic to infrastructure manager


mangelajo:
We have noticed that sometimes ovnkube-master fails to become ready after this restart.

PR #1298, opened 2023-01-27
USHIFT-716: add configurable KAS advertise address

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -39,7 +39,7 @@ The configuration settings alongside with the supported command line arguments a
 | nodeIP                | --node-ip                 | MICROSHIFT_NODEIP                       | The IP address of the node, defaults to IP of the default route
 | hostnameOverride      | --hostname-override       | MICROSHIFT_HOSTNAMEOVERRIDE             | The name of the node, defaults to hostname
 | logLevel              | --v                       | MICROSHIFT_LOGVLEVEL                    | Log verbosity (Normal, Debug, Trace, TraceAll)
-| subjectAltNames     | --subject-alt-names       | MICROSHIFT_SUBJECTALTNAMES              | Subject Alternative Names for apiserver certificates
+| subjectAltNames       | --subject-alt-names       | MICROSHIFT_SUBJECTALTNAMES              | Subject Alternative Names for apiserver certificates
 
 ## Default Settings
 
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -70,6 +70,21 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 		return nil, err
 	}
 
+	externalCertNames := []string{
+		cfg.NodeName,
+		"api." + cfg.BaseDomain,
+	}
+	externalCertNames = append(externalCertNames, cfg.SubjectAltNames...)
+	// When Kube apiserver advertise address matches the node IP we can not add
+	// it to the certificates or else the internal pod access to apiserver is
+	// broken. Because of client-go not using SNI and the way apiserver handles
+	// which certificate to serve which destination IP, internal pods start
+	// getting the external certificate, which is signed by a different CA and
+	// does not match the hostname.
+	if cfg.KASAdvertiseAddress != cfg.NodeIP {
+		externalCertNames = append(externalCertNames, cfg.NodeIP)
+	}
+
 	certsDir := cryptomaterial.CertsDirectory(microshiftDataDir)
 
 	certChains, err := certchains.NewCertificateChains(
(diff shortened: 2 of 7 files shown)


docs/howto_config.md
```diff
@@ -24,6 +24,7 @@ node:
 apiServer:
   subjectAltNames:
     - ""
+  advertiseAddress: ""
```
pacevedom, line 27: This is exposed so as to enable CNCF certification tests where we need more than one node. Users shouldnt use it, as it defaults to the api server service IP internally.

dhellmann: If we put this in the documentation, someone is going to use it. Do we need to include it in this document?

packaging/microshift/config.yaml
```diff
@@ -26,6 +26,9 @@ apiServer:
   # The Subject Alternative Names for the external certificates in API server (defaults to hostname -A)
   #subjectAltNames: []
 
+  # The Kube API Server advertise address for pods to connect to it (defaults to 10.43.0.1)
+  #advertiseAddress: ''
```
dhellmann, line 30: Can we remove it from here, too?

pkg/cmd/init.go
```diff
@@ -70,6 +70,15 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 		return nil, err
 	}
 
+	externalCertNames := []string{
+		cfg.NodeName,
+		"api." + cfg.BaseDomain,
+	}
+	externalCertNames = append(externalCertNames, cfg.SubjectAltNames...)
+	if cfg.KASAdvertiseAddress != cfg.NodeIP {
```
dhellmann, line 78: A comment here explaining why we only include the NodeIP if it does not match the kube API server address will help maintainers who are looking at this code later.

pkg/cmd/run.go
```diff
@@ -36,6 +36,7 @@ func addRunFlags(cmd *cobra.Command, cfg *config.MicroshiftConfig) {
 	flags.String("service-cidr", cfg.Cluster.ServiceCIDR, "The IP range in CIDR notation for services in the cluster.")
 	flags.String("service-node-port-range", cfg.Cluster.ServiceNodePortRange, "The port range to reserve for services with NodePort visibility. This must not overlap with the ephemeral port range on nodes.")
 	flags.String("base-domain", cfg.BaseDomain, "The base domain for this cluster.")
+	flags.String("kas-advertise-address", cfg.KASAdvertiseAddress, "The KAS advertise address for pods to connect.")
```
dhellmann, line 39: Do we need a command line flag?

pkg/config/config.go
```diff
@@ -59,11 +59,13 @@ type IngressConfig struct {
...
-	SubjectAltNames []string      `json:"subjectAltNames"`
-	NodeName        string        `json:"nodeName"`
-	NodeIP          string        `json:"nodeIP"`
-	BaseDomain      string        `json:"baseDomain"`
-	Cluster         ClusterConfig `json:"cluster"`
+	SubjectAltNames     []string      `json:"subjectAltNames"`
+	NodeName            string        `json:"nodeName"`
+	NodeIP              string        `json:"nodeIP"`
+	KASAdvertiseAddress string        `json:"kasAdvertiseAddress"`
+	SkipKASInterface    bool          `json:"-"`
```
dhellmann, line 66: What is SkipKASInterface?

At some point we should come back and add comments to all of these fields. You don't have to do all of them in this PR, but let's at least add good docs for the new things.

pkg/controllers/kube-apiserver.go
```diff
@@ -298,6 +304,20 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	if !s.skipLoInterface {
```
dhellmann, line 307: It's not clear where to draw a line for the single responsibility principle here. Should this IP management be done inside the KubeAPIServer service or can it be its own thing that KubeAPIServer depends on?

dhellmann:
Is there a way to add a test case to verify that this solves the problem and so we do not regress with future changes?

pkg/controllers/kube-apiserver.go
pacevedom: Turned this into a new service. Makes more sense, thanks for the suggestion.

docs/howto_config.md
dhellmann: That's a good question. @jerpeter1 has been looking at the certification rules, maybe he came across something that would help answer it?

pkg/cmd/run.go
dhellmann: I would like to standardize on just using the configuration file, and dropping the environment variables and command line flags. We can always add this later if we find a reason not to do that.

pacevedom: Done

pacevedom:
/override ci/prow/e2e-openshift-conformance-sig-scheduling ci/prow/e2e-openshift-conformance-sig-instrumentation ci/prow/e2e-openshift-conformance-sig-api-machinery

PR #1300, opened 2023-01-27
NP-641: Avoid firewalld flush of iptables

mangelajo:
Disables firewalld access to iptables related nftables translator,
this avoids firewalld flush of iptables under any circumstance,
firewalld will work at nftables level, while MicroShift/OVN can
continue to work at iptables level (translated to nft) under the hood.

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -174,6 +174,17 @@ install -p -m644 packaging/systemd/microshift-ovs-init.service %{buildroot}%{_un
 install -p -m755 packaging/systemd/configure-ovs.sh %{buildroot}%{_bindir}/configure-ovs.sh
 install -p -m755 packaging/systemd/configure-ovs-microshift.sh %{buildroot}%{_bindir}/configure-ovs-microshift.sh
 
+# Avoid firewalld manipulation and flushing of iptable rules,
+# this is a workaround for https://issues.redhat.com/browse/NP-641
+# It will trigger some warnings on the selinux audit log when restarting firewalld.
+# In the future firewalld should stop flushing iptables unless we use any firewalld rule with direct
+# iptables rules, once that's available in RHEL we can remove this workaround
+# see https://github.com/firewalld/firewalld/issues/863#issuecomment-1407059938
+
+mkdir -p -m755 %{buildroot}%{_sysconfdir}/systemd/system/firewalld.service.d
+install -p -m644 packaging/systemd/firewalld-no-iptables.conf %{buildroot}%{_sysconfdir}/systemd/system/firewalld.service.d/firewalld-no-iptables.conf
+
+
 mkdir -p -m755 %{buildroot}/var/run/kubelet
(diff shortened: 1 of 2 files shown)


packaging/systemd/firewalld-no-iptables.conf
```diff
@@ -0,0 +1,10 @@
+[Service]
+InaccessiblePaths=/usr/sbin/xtables-nft-multi
+
+# This is a workaround for https://issues.redhat.com/browse/NP-641
+# it will trigger some warnings on the selinux audit log, but it's much
+# better than the failure mode of having all iptables flushed by
+# firewalld
+
+# hopefully in the future we can have a setting in firewalld to avoid
+# iptables flushing when using nft.
```
ggiguash, line 10: We need to rephrase this explanation. I would not include a reference to the issue, which is currently private. Also, we should only explain what we're trying to solve without future statements.

packaging/rpm/microshift.spec
```diff
@@ -261,6 +266,9 @@ systemctl enable --now --quiet openvswitch || true
 # Use Git command to generate the log and replace the VERSION string
 # LANG=C git log --date="format:%a %b %d %Y" --pretty="tformat:* %cd %an <%ae> VERSION%n- %s%n" packaging/rpm/microshift.spec
 %changelog
+* Fri Jan 27 2023 Miguel Angel Ajo Pelayo <majopela@rehat.com> 4.12.0
+- Add firewalld systemd service override configuration to avoid access to iptables NP-641
```
ggiguash, line 270: I would not include NP-641 issue reference in the comment.

ggiguash:
/retitle NP-641: Avoid firewalld flush of iptables

packaging/systemd/firewalld-no-iptables.conf
mangelajo: We have talked about this in the past, but IMHO, as a project maintainer, it always helped me to figure out why certain things were done in projects, it is true that you can always look at the code/commit and find the original reference. But for example it gets harder when the file has moved around and the git reference becomes harder to track.

packaging/rpm/microshift.spec
mangelajo: It was standard practice in rpm packaging to include references to bugs, in fact at some point in history I remember you could not push a packaging change without the reference to a bz#  has this changed?

ggiguash: The current history in the file does not include any issue references. If we decide to reference any issues, we should probably trace the history back and add it all along. For now, I would stick out.

packaging/systemd/firewalld-no-iptables.conf
ggiguash: This file will end up on user systems, so we need to review the text in the comments, like documentation. Let's discuss it further and reach a conclusion backed up by our content experts.

mangelajo:
/unhold

It seems to work well on my tests.

PR #1303, opened 2023-01-30
OCPBUGS-6785: Use systemd-resolved in CoreDNS configmap

zshi-redhat:
systemd-resolved conf is used by kubelet as alternative dns resolver when exists. This commit aligns the DNS Corefile conf with kubelet resolvConf when default systemd-resolved config is detected.

Co-authored-by: Miciah Dashiel Butler Masters <mmasters@redhat.com>
(cherry picked from commit 24ae57ee616e7ccdb4298c6f159ae5442bc87abc)

### DIFF
--- a/assets/components/openshift-dns/dns/configmap.yaml
+++ b/assets/components/openshift-dns/dns/configmap.yaml
@@ -16,7 +16,7 @@ data:
             fallthrough in-addr.arpa ip6.arpa
         }
         prometheus 127.0.0.1:9153
-        forward . /etc/resolv.conf {
+        forward . {{if eq .UpstreamResolver ""}}/etc/resolv.conf{{else}}{{.UpstreamResolver}}{{end}} {
             policy sequential
         }
         cache 900 {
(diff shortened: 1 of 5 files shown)


zshi-redhat:
cc @pmtk

assets/components/openshift-dns/dns/configmap.yaml
```diff
@@ -16,7 +16,7 @@ data:
             fallthrough in-addr.arpa ip6.arpa
         }
         prometheus 127.0.0.1:9153
-        forward . /etc/resolv.conf {
+        forward . {{if eq .UpstreamResolver ""}}/etc/resolv.conf{{else}}{{.UpstreamResolver}}{{end}} {
```
pmtk, line 19: I see this changed compared to what was cherry-picked, is there a functional difference?

zshi-redhat: Good catch! I missed backporting the other commit.
fixed.

Issue #1307, opened 2023-01-31
[BUG] Install MicroShift on RHEL for Edge script needed fast-datapath-beta repo instead for 8.7


WIP42:
#### What happened?
1. Followed along https://github.com/openshift/microshift/blob/main/docs/rhel4edge_iso.md
2. At `Building Installer` step: `~/microshift/scripts/image-builder/build.sh -pull_secret_file ~/.pull-secret.json`
3. I got the error ```Error: Unknown repo: 'fast-datapath-for-rhel-8-x86_64-rpms'```, so had to update the script to refer to `fast-datapath-beta` repo instead.

#### What did you expect to happen?
To follow documentation for installing from new 

#### How to reproduce it (as minimally and precisely as possible)?
1. See `What happened?` section above

#### Environment
- MicroShift version (use `microshift version`): ```MicroShift Version: 4.13.0_0.nightly_2023_01_27_165107_20230130151612_17696e4e
Base OCP Version: 4.13.0-0.nightly-2023-01-27-165107```
- Hardware configuration:
- OS (e.g: `cat /etc/os-release`):
```
NAME="Red Hat Enterprise Linux"
VERSION="8.7 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.7 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/8/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.7
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"
```
- Kernel (e.g. `uname -a`): `Linux microshift-dev.localdomain 4.18.0-425.10.1.el8_7.x86_64 #1 SMP Wed Dec 14 16:00:01 EST 2022 x86_64 x86_64 x86_64 GNU/Linux`
- Others:

ggiguash:
[...]

PR #1308, opened 2023-01-31
Install MicroShift on RHEL for Edge script needed fast-datapath-beta …

WIP42:
…repo instead for 8.7

See https://github.com/openshift/microshift/issues/1307

Closes #<Issue Number>


ggiguash:
We'd like to avoid using beta channels unless there is a special need for it. 
As I commented in the #1307 issue, you must use the right subscription allowing access to a regular channel

/close

PR #1322, opened 2023-02-05
OCPBUGS-7056: fix ovn config file path

zshi-redhat:
GetConfigFile() may return empty path or user defined path, but ovn config file is only supported under /etc/microshift


ggiguash:
@zshi-redhat, can we link this PR to a JIRA issue/bug?

zshi-redhat:
> @zshi-redhat, can we link this PR to a JIRA issue/bug?

Done.

PR #1323, opened 2023-02-05
USHIFT-819: MicroShift devenv to support RHEL 9 and CentOS 9

ggiguash:
The procedure was tested on RHEL 9.1 and CentOS 9 Stream
Closes [USHIFT-819](https://issues.redhat.com//browse/USHIFT-819)

### DIFF
--- a/docs/devenv_setup_auto.md
+++ b/docs/devenv_setup_auto.md
@@ -1,8 +1,8 @@
 # Automating Development Environment Setup
 This document describes how to automate the setup of MicroShift development environment per instructions in the following sections:
-* [Create Development Virtual Machine](./devenv_rhel8.md#create-development-virtual-machine)
-* [Build MicroShift](./devenv_rhel8.md#build-microshift)
-* [Run MicroShift Executable](./devenv_rhel8.md#run-microshift-executable)
+* [Create Development Virtual Machine](./devenv_setup.md#create-development-virtual-machine)
+* [Build MicroShift](./devenv_setup.md#build-microshift)
+* [Run MicroShift Executable](./devenv_setup.md#run-microshift-executable)
 
 It is recommended to review the above mentioned documentation sections before proceeding with the automation instructions in the current document.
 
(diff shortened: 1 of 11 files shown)


scripts/devenv-builder/configure-vm.sh
```diff
@@ -35,16 +41,18 @@ if [ "$(whoami)" != "microshift" ] ; then
 fi
 
 # Check the subscription status and register if necessary
-if ! sudo subscription-manager status >& /dev/null ; then
-   sudo subscription-manager register
+if $RHEL_SUBSCRIPTION ; then
```
fzdarsky, line 44: Better to test explicity (`[ "$RHEL_SUBSCRIPTION" = true ]`), because as written the expression would evaluate to true even if the var is unset or empty.

docs/devenv_setup_auto.md
```diff
@@ -1,41 +1,45 @@
...
-Usage: create-vm.sh [<VMNAME> <VMDISKDIR> <RHELISO> <NCPUS> <RAMSIZE> <DISKSIZE> <SWAPSIZE> <DATAVOLSIZE>]
+Usage: create-vm.sh [<VMNAME> <VMDISKDIR> <ISOFILE> <NCPUS> <RAMSIZE> <DISKSIZE> <SWAPSIZE> <DATAVOLSIZE>]
 INFO: Specify 0 swap size to disable swap partition
 INFO: Positional arguments also can be specified using environment variables
 INFO: All sizes in GB
 
 ERROR: Invalid VM name: ''
 ```
-> It is mandatory to use the RHEL 8 DVD (not boot) image for installing the virtual machine.
+> When configuring a RHEL OS, it is mandatory to use the DVD (not boot) image for installing the virtual machine.
```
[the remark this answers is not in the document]
ggiguash: Since we are not registering the OS, pulling packages remotely is not going to work. I'll add a clarification.

fzdarsky (approved):

scripts/devenv-builder/configure-vm.sh
```diff
@@ -35,16 +41,18 @@ if [ "$(whoami)" != "microshift" ] ; then
 fi
 
 # Check the subscription status and register if necessary
-if ! sudo subscription-manager status >& /dev/null ; then
-   sudo subscription-manager register
+if [ $RHEL_SUBSCRIPTION = true ] ; then
```
dhellmann, line 44: "true" and "false" are linux commands, so this can just be 
```suggestion
if $RHEL_SUBSCRIPTION ; then
```

dhellmann (commented):
This looks great, I'm looking forward to testing it out this week.

PR #1331, opened 2023-02-07
NP-669: validate ovnk prerequisites and configurations

zshi-redhat:


pkg/config/ovn/ovn.go
```diff
@@ -77,3 +130,23 @@ func NewOVNKubernetesConfigFromFileOrDefault(path string) (*OVNKubernetesConfig,
...
+	if err != nil {
+		return "", err
+	}
+	addrs, err := iface.Addrs()
+	if err != nil {
+		return "", err
+	}
+	for _, addr := range addrs {
+		ip := addr.(*net.IPNet).IP
+		if ip.To4() != nil {
```
pmtk, line 145: Do we need `ip.To4()`? Looks like `ip.String()` already does that under the hood: https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/net/ip.go;l=321

pmtk (commented):
Two minor comments, otherwise looks good to me

PR #1337, opened 2023-02-07
USHIFT-130: Add inspect command

pacevedom:
Bring the inspect command from oc adm to microshift. Inspect requires kubeconfig to be provided either by flags or environment variables, as it lacks kubectl package functionality to pass flags.

Closes #<Issue Number>


dhellmann:
After a quick scan, it looks like this just updates our existing dependencies, but doesn't add anything to go.mod. Is that right? 

It does bring in a lot of new vendored code, though. How much difference is there in the binary size? How complex is `oc adm inspect`? Could we reproduce the parts we care about in a shell script, for example?

dhellmann:
> After a quick scan, it looks like this just updates our existing dependencies, but doesn't add anything to go.mod. Is that right?
> 
> It does bring in a lot of new vendored code, though. How much difference is there in the binary size? How complex is `oc adm inspect`? Could we reproduce the parts we care about in a shell script, for example?

Or in sos directly, I suppose.

pacevedom:
> After a quick scan, it looks like this just updates our existing dependencies, but doesn't add anything to go.mod. Is that right?

It adds only the oc/pkg/cli/admin/inspect package as a new require. This brings a few more dependencies, none of them new.

> It does bring in a lot of new vendored code, though. How much difference is there in the binary size? How complex is `oc adm inspect`? Could we reproduce the parts we care about in a shell script, for example?

Binary size difference (in K):
112648 with inspect
112328 without inspect

The tricky part about reproducing this in a shell script is the output format. You would be implementing inspect in another language. Also reducing the amount of get calls to make it fast requires a significant amount of yaml handling.

dhellmann:
We want to settle the questions in https://github.com/openshift/enhancements/pull/1318 before taking this change.

/hold

go.mod
```diff
@@ -188,28 +188,28 @@ require (
...
-	google.golang.org/genproto v0.0.0-20220502173005-c8bf987b8c21 // indirect
+	google.golang.org/genproto v0.0.0-20220624142145-8cd45d7dbd1f // indirect
 	google.golang.org/grpc v1.49.0 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/gcfg.v1 v1.2.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
-	gopkg.in/square/go-jose.v2 v2.2.2 // indirect
-	gopkg.in/warnings.v0 v0.1.1 // indirect
+	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
```
pmtk, line 205: I looked if we could not do this, but `o/oc` imports `containers/image` and this bump comes from somewhere there (it's also indirect for `c/i`).

A little context:
This was `replace`d before K8s 1.26, but they dropped a lot of replaces in k8s' `go.mod` and so did we.
Even if we'd rework that procedure to not look for replace version, but direct/indirect ones it wouldn't do us any good because this bump comes from somewhere else and forces us to patch vendored code...
Guess this is the price for vendoring binaries

pacevedom:
Closing in favor of using `oc` instead.

PR #1348, opened 2023-02-10
OCPBUGS-7205: don't advertise ovn-kubernetes local addresses in mdns

zshi-redhat:

### DIFF
--- a/pkg/mdns/controller.go
+++ b/pkg/mdns/controller.go
@@ -3,11 +3,11 @@ package mdns
 import (
 	"context"
 	"net"
-	"regexp"
 	"strings"
 	"sync"
 
 	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/config/ovn"
 	"github.com/openshift/microshift/pkg/mdns/server"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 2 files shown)


pkg/mdns/controller.go
```diff
@@ -3,11 +3,11 @@ package mdns
 import (
 	"context"
 	"net"
-	"regexp"
 	"strings"
 	"sync"
 
 	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/config/ovn"
```
dhellmann, line 10: This introduces a dependency between the MDNS and OVN components, which is going to make it harder to make OVN pluggable. We can do this for now, but we need a plan to decouple these again in the future. What can we do to keep them loosely coupled?

zshi-redhat: I'm actually doing some code preparation work in this PR to move ovnk related code in its own files under  `pkg/config/ovn`. Before this PR, the ovnk interfaces and special IP address are hardcoded in mdns controller. For fully decoupling ovnk from mdns controller (assuming mdns will stay in microshift main), we would need an interface between microshift and  plugins to exchange the mdns related information (this is where the binary/script method in pluggable enhancement would help).

dhellmann:
Let's move ahead with this fix. We will have to think about the MDNS requirements when we split CNI out of the MicroShift binary.

/lgtm

PR #1353, opened 2023-02-10
USHIFT-790: always take image version from content when rebasing

dhellmann:
Instead of parsing the image name to find the version and date, use
the image data we have already extracted.


dhellmann:
I was able to manually run the script far enough to get it to create the branch with the right name and apply several changes to it. It eventually failed trying to rebase the lvms stuff because I didn't have a local cluster running so the `oc create` commands failed.

PR #1354, opened 2023-02-12
USHIFT-148: Implement MicroShift pre-rollback Greenboot procedure

ggiguash:
Implement pre-rollback cleanup of OVN data and MicroShift pods to allow a rollback. 
Move `hack/clean.sh` script to `scripts/microshift-cleanup-data.sh` and add option to clean OVN-only data
Add documentation, including the recommendation on configuring the `systemd` journal service for persisting data across boots.
Updated image builder and getting started kickstart files to perform the journal service configuration

Closes [USHIFT-148](https://issues.redhat.com//browse/USHIFT-148)


scripts/microshift-cleanup-data.sh
```diff
@@ -0,0 +1,102 @@
...
+            fi
+            crictl rmp -f ${ocp_pods} &>/dev/null
+            retries=$(( retries - 1 ))
+        done
+    done
+fi
+
+echo Killing conmon, pause and OVN processes
+ovs-vsctl del-br br-int 2>/dev/null
+for pname in conmon pause ovn-controller ovn-northd ovsdb-server ; do
```
zshi-redhat, line 87: `pkill conmon` would kill all conmon process for all microshift containers.

from crio doc:
```
Each container is monitored by a separate conmon process. The conmon process holds the pty of the PID1 of the container process. It handles logging for the container and records the exit code for the container process
```

scripts/microshift-cleanup-data.sh
```diff
@@ -0,0 +1,102 @@
...
+    for i in ${!PODS_NS_LIST[@]}; do
+        ns=${PODS_NS_LIST[$i]}    
+
+        retries=5
+        while [ ${retries} -gt 0 ] ; do
+            ocp_pods=$(crictl pods --namespace ${ns} -q)
+            if [ "$(echo ${ocp_pods} | wc -w)" -eq 0 ] ; then
+                break
+            fi
+            crictl rmp -f ${ocp_pods} &>/dev/null
```
zshi-redhat, line 79: what would happen if red.d scripts fail?

docs/greenboot_dev.md
```diff
@@ -41,6 +41,9 @@ FAILURE
 ...
 ```
 
+> Run the `sudo journalctl -o cat -u redboot-task-runner.service` command
```
zshi-redhat, line 44: where is the `redboot-task-runner.service` added?

docs/greenboot.md
```diff
@@ -30,14 +30,26 @@ The health check script is installed into the `/etc/greenboot/check/required.d`
 directory and it is not executed during the system boot in case the `greenboot`
 package is not present.
 
+The `40_microshift_pre_rollback.sh` pre-rollback script is installed into the
+`/etc/greenboot/red.d` directory, to be executed right before the system rollback
+takes place. The script performs MicroShift pod and OVN database cleanup to avoid
```
zshi-redhat, line 35:
```suggestion
takes place. The script performs MicroShift pod and OVN cleanup to avoid
```

docs/greenboot.md
```diff
@@ -30,14 +30,26 @@ The health check script is installed into the `/etc/greenboot/check/required.d`
...
+potential conflicts with previous versions of the software.
```
zshi-redhat, line 36: Previous version may sound confusing, what about `conflicts with the version to be upgraded to`?

scripts/microshift-cleanup-data.sh
ggiguash: The event will be logged and the system will continue with the rollback.

scripts/microshift-cleanup-data.sh
ggiguash: Are you saying we should not kill conmon processes? At this time of the script, either all the crio pods/images have been wiped off, or every pod has been stopped. Does it make a difference?

docs/greenboot_dev.md
ggiguash: This service is added by Greenboot when its RPM is installed.

docs/greenboot.md
ggiguash: Rephrased to make it sound less confusing.

dhellmann:
This is related to https://github.com/openshift/enhancements/pull/1312, right?

ggiguash:
> This is related to [openshift/enhancements#1312](https://github.com/openshift/enhancements/pull/1312), right?
Not directly.
The mentioned ER is about backup/restore of MicroShift data, i.e. backup before upgrade and restore after downgrade.
The current PR is about cleaning the OVN cache before downgrade.

The ER implementation would also use greenboot facilities, but with different hooks.

scripts/microshift-cleanup-data.sh
zshi-redhat: I see, I thought the non-ovn pods are still running at this stage, but it is not.

scripts/microshift-cleanup-data.sh
zshi-redhat: ack+

scripts/microshift-cleanup-data.sh
```diff
@@ -0,0 +1,102 @@
...
+            break
+            ;;
+        *)
+            echo "Aborting cleanup" 
+            exit 0
+            ;;
+        esac
+    done
+fi
+
```
dhellmann, line 50: Some of the rest of this script could benefit from being organized into functions. That would avoid having to check FULL_CLEAN so many times, and would make it clear to a reader what steps are going to happen in the different cleaning modes.

scripts/microshift-cleanup-data.sh
```diff
@@ -0,0 +1,102 @@
...
+fi
+pkill -9 --exact microshift
+pkill -9 --exact microshift-etcd
+
+if ${FULL_CLEAN} ; then
+    echo Removing crio container and image storage
+    crio wipe -f &>/dev/null
+    systemctl restart crio
+else
+    echo Removing MicroShift pods
```
dhellmann, line 69: Why do we need to kill the pods? Is this part of the greenboot handling, or is that for some other use?

scripts/microshift-cleanup-data.sh
```diff
@@ -0,0 +1,102 @@
...
+                break
+            fi
+            crictl rmp -f ${ocp_pods} &>/dev/null
+            retries=$(( retries - 1 ))
+        done
+    done
+fi
+
+echo Killing conmon, pause and OVN processes
+ovs-vsctl del-br br-int 2>/dev/null
```
dhellmann, line 86: Why don't we care about error messages here?

scripts/microshift-cleanup-data.sh
```diff
@@ -0,0 +1,102 @@
...
+    systemctl stop --now   ${service} 2>/dev/null
+    systemctl reset-failed ${service} 2>/dev/null
+done
+
+if ${FULL_CLEAN} ; then
+    echo Disabling MicroShift services
+    systemctl disable microshift 2>/dev/null
+fi
+pkill -9 --exact microshift
+pkill -9 --exact microshift-etcd
```
dhellmann, line 62: It seems like kill -9 etcd if we aren't deleting its database could lead to data corruption. Is there a safer way to ensure it stops cleanly?

packaging/greenboot/microshift-pre-rollback.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#
+# Main
+#
+
+# Exit if the current user is not 'root'
+if [ $(id -u) -ne 0 ] ; then
+    echo "The '${SCRIPT_NAME}' script must be run with the 'root' user privileges"
+    exit 0
```
dhellmann, line 26: This seems like a case where we want to report an error.

packaging/rpm/microshift.spec
```diff
@@ -140,7 +140,7 @@ make
 install -d %{buildroot}%{_bindir}
 install -p -m755 ./_output/microshift %{buildroot}%{_bindir}/microshift
 install -p -m755 ./_output/microshift-etcd %{buildroot}%{_bindir}/microshift-etcd
-install -p -m755 hack/cleanup.sh %{buildroot}%{_bindir}/cleanup-all-microshift-data
+install -p -m755 scripts/microshift-cleanup-data.sh %{buildroot}%{_bindir}/microshift-cleanup-data
```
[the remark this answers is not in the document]
ggiguash: I was just renaming the existing script by removing "all" from its name. We currently ship few scripts, like this and OVN-specific into the /usr/bin directory. 
I think that if we want to move them elsewhere, we might want to do this in a dedicated task/PR.

packaging/greenboot/microshift-pre-rollback.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+echo "STARTED"
+
+# Dump GRUB boot variables and ostree status
```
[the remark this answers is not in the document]
ggiguash: The grub settings affect the execution of the script. It only works when "boot_counter=0", so I thought it would be good to have these settings in the logs. 
In general, both greenboot-related scripts dump these settings because this info is critical when troubleshooting

scripts/microshift-cleanup-data.sh
ggiguash: It is necessary to kill the pods (ovn last) to allow for further killing of conmon, etc. processes. Otherwise, they hold the /var/run/ovn files open and it's not possible to clean up ovn databases.

scripts/microshift-cleanup-data.sh
ggiguash: We're performing a clean-up operation, so some utilities report errors/warnings that are confusing to the caller. The idea is to get rid of these devices and processes so that `rm -rf /var/run/ovn` command succeeds. The main goal is to clean the ovn database...

scripts/microshift-cleanup-data.sh
ggiguash: We are performing service stop before executing the SIGKILL. These lines are the last resort in case the `systemctl stop` failed to kill the process. Usually, it's only effective in the devenv where processes are not started from systemd.

packaging/greenboot/microshift-pre-rollback.sh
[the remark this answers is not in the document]
dhellmann: OK, a more detailed comment about that would be good.

packaging/rpm/microshift.spec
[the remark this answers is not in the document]
dhellmann: Sure, we can talk about whether we need to do that at all separately.

scripts/microshift-cleanup-data.sh
dhellmann: The script will be more maintainable if we include that sort of information in comments.

scripts/microshift-cleanup-data.sh
dhellmann: I'm worried about debugging a situation where the command fails, the error is discarded, and then some other error happens later because the original action could not be taken. Can we structure the logging with "Starting X action" and "Completed X action" messages around the commands to make clear what errors go with what commands and still keep the stderr output?

scripts/microshift-cleanup-data.sh
dhellmann: OK, this is another place where comments inline would help.

scripts/microshift-cleanup-data.sh
ggiguash: That's a fair point. I changed to code to list the interface with output redirection. Then, we either execute the delete operation on the bridge or get a message that the bridge does not exist.

Please, review the two example outputs and let me know if this is better now.

```
$ sudo ./scripts/microshift-cleanup-data.sh --ovn
Stopping MicroShift services
Removing MicroShift pods
Deleting the br-int interface
Killing conmon, pause and OVN processes
Removing OVN configuration
Cleanup succeeded
```

```
$ sudo ./scripts/microshift-cleanup-data.sh --ovn
Stopping MicroShift services
Removing MicroShift pods
The br-int bridge interface does not exist
Killing conmon, pause and OVN processes
Removing OVN configuration
Cleanup succeeded
```

scripts/microshift-cleanup-data.sh
```diff
@@ -0,0 +1,134 @@
...
+            retries=$(( retries - 1 ))
+        done
+    done
+}
+
+function clean_processes() {
+    if ${FULL_CLEAN} ; then
+        # This operation can only be run in the full clean mode as the removal
+        # of host interfaces is not detected by kubelet and pods cannot be
+        # recovered following the service restart
```
ggiguash, line 68: @zshi-redhat , can you confirm that this is correct?

zshi-redhat:
LGTM

scripts/microshift-cleanup-data.sh
zshi-redhat: Yes, it looks good to me.

docs/greenboot.md
```diff
@@ -79,3 +94,19 @@ boot_success=1
...
+
+## The `systemd` Journal Service Configuration
+
+The default configuration of the `systemd` journal service stores the data in
+the volatile `/run/log/journal` directory, which does not persist after a system
+boot. To monitor `greenboot` activities across system boots, it is recommended to
+enable the journal data persistency by creating the `/var/log/journal` directory
+and setting limits on the maximal journal data size.
+
+Run the following commands to configure the journal data persistency and limits.
```
dhellmann, line 106: It's also possible to include the settings in a new file in /etc/systemd/journald.conf.d/ so I wonder if that's a better approach for someone building their own images.

dhellmann (commented):
I had one small suggestion for the docs, but it's not worth holding up progress.

/lgtm

docs/greenboot.md
ggiguash: I have included this configuration in kickstart. I'm not sure we would want to impose system logging configuration from MicroShift RPM.

dhellmann: I just meant in our docs, not in our RPMs.

dhellmann (commented):
/lgtm

PR #1357, opened 2023-02-13
NO-ISSUE: Gather more debug info

pmtk:


validate-microshift/cluster-debug-info.sh
```diff
@@ -7,34 +7,62 @@ function to_run() {
 }
 
 to_run oc get cm -n kube-public microshift-version -o=jsonpath='{.data}'
-to_run microshift version 
-to_run microshift version -o yaml 
+to_run microshift version
+to_run microshift version -o yaml
```
pacevedom, line 11: I am missing the show-config, wdyt?

pmtk: added

PR #1371, opened 2023-02-14
NO-ISSUE: bring back tabular `get pods` to debug info

pmtk:


pacevedom (approved):
This is only changing the order for readability, right? Like summary, then details.

pacevedom (commented):
/lgtm

pmtk:
> This is only changing the order for readability, right? Like summary, then details.

That's right

PR #1373, opened 2023-02-14
OCPBUGS-6859: Create empty manifest directory when installing MicroShift

ggiguash:
The /etc/microshift/manifests empty directory is created when installing RPMs

Closes OCPBUGS-6859

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -165,6 +165,7 @@ install -d -m755 %{buildroot}/%{_unitdir}
 install -p -m644 packaging/systemd/microshift.service %{buildroot}%{_unitdir}/microshift.service
 
 install -d -m755 %{buildroot}/%{_sysconfdir}/microshift
+install -d -m755 %{buildroot}/%{_sysconfdir}/microshift/manifests
 install -p -m644 packaging/microshift/config.yaml %{buildroot}%{_sysconfdir}/microshift/config.yaml.default
 
 # release-info files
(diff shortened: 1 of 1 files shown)


pmtk:
[`pkg/config/config.go`](https://github.com/openshift/microshift/blob/main/pkg/config/config.go#L37) still points to `/usr/lib/microshift/manifests/` and not `/usr/etc/microshift/manifests/`. This should be changed as well, right?

ggiguash:
@pmtk , I missed a few places to replace the manifests directory. More files added now - can you look at it again?

dhellmann (commented):
I don't know enough about the FHS standard to know why the original path may have been selected. @fzdarsky can you comment?

dhellmann:
This is going to trigger documentation updates, too.

What is the motivation for changing `/usr/lib/` to `/usr/etc/`? How are those directories expected to be used on an ostree-based system?

/hold

ggiguash:
> What is the motivation for changing `/usr/lib/` to `/usr/etc/`? How are those directories expected to be used on an ostree-based system?

The reason is due to symmetry with config.yaml file, which is located in the /etc or /usr/etc pair. 
So, the situation I was trying to fix is the following
* /etc/microshift/config.yaml.default -> /usr/etc/
* /etc/microshift/mamifests -> /usr/lib/microshift/manifests
We had 2 configuration directories under /usr.

ggiguash:
@dhellmann, I reverted the manifest directory change, as discussed.
/unhold

packaging/rpm/microshift.spec
```diff
@@ -165,6 +165,7 @@ install -d -m755 %{buildroot}/%{_unitdir}
 install -p -m644 packaging/systemd/microshift.service %{buildroot}%{_unitdir}/microshift.service
 
 install -d -m755 %{buildroot}/%{_sysconfdir}/microshift
+install -d -m755 %{buildroot}/%{_sysconfdir}/microshift/manifests
```
dhellmann, line 168: Do we need /usr/lib/microshift and /usr/lib/microshift/manifests, too? Maybe we should add those in another PR, in case we change the names?

dhellmann (commented):
One open question but we can deal with it separately

/lgtm

packaging/rpm/microshift.spec
ggiguash: The /usr/lib/microshift directory is not created at present.
Since we're talking about read-only directory structure that can only be updated from RPM, I thought it does not make sense to create an empty hierarchy.

dhellmann: OK, I thought we might need to declare that we own that directory so that other RPMs can just put files into it.

PR #1378, opened 2023-02-15
OCPBUGS-6861: Allow TopoLVM initial grace period while service-ca attempts to start

copejon:
TopoLVM on MicroShift suffers flakey behavior that stems in part from a race window with the service-ca.  This PR pads TopoLVM's startup time to give the service-ca a chance to issue certs.  This PR sets the initial startupProbe values to a 5 minute timeout (30 retries at 10 second intervals).

The race window was exposed by flakey behavior with service-ca startup that is being addressed separately.  The work there will reduce the scale of the race window, however it will still be present.  The window appears during (re)boot because topolvm depends on the service-ca to issue cert. In this scenario, topolvm and service-ca are started concurrently; their boot order is not controlled by microshift. This means that both services' retry loops are initiated at the same time, so the diff between the service-ca issuing a cert and topolvm reaching it's retry max creates this race window.  By providing topolvm a startup grace period, we allow the service ca to stabilize and improve the user experience since topolvm will not enter a crashloopbackoff as a result of timing out.  

_Note that this grace period does not delay topolvm's startup, it only delays the point at which topolvm startup failures are considered true failures._

### DIFF
--- a/assets/components/lvms/topolvm-controller_deployment.yaml
+++ b/assets/components/lvms/topolvm-controller_deployment.yaml
@@ -23,6 +23,12 @@ spec:
         - /topolvm-controller
         - --cert-dir=/certs
         image: '{{ .ReleaseImage.topolvm_csi }}'
+        startupProbe:
+          failureThreshold: 60
+          periodSeconds: 2
+          httpGet:
+            port: healthz
+            path: /healthz
         livenessProbe:
           failureThreshold: 3
           httpGet:
--- a/assets/components/lvms/topolvm-node_daemonset.yaml
+++ b/assets/components/lvms/topolvm-node_daemonset.yaml
@@ -64,6 +64,13 @@ spec:
           initialDelaySeconds: 10
           periodSeconds: 60
           timeoutSeconds: 3
+        startupProbe:
+          failureThreshold: 60
+          successThreshold: 1
+          periodSeconds: 2
+          httpGet:
+            port: healthz
+            path: /healthz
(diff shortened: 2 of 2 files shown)


pacevedom:
Do you have some more details on the race condition? In the logs from the bug I see there is some trouble here and there in both pods, but not clear to me.

scripts/auto-rebase/rebase.sh
```diff
@@ -178,6 +178,16 @@ update_lvms_manifests() {
     for m in "${namespace_scoped_manifests[@]}"; do
         yq -i '.metadata.namespace = "openshift-storage"' "${dest}/${m}" || return 1
     done
+
+    # These manifests are not published and have to be gotten manually from the LVMS operator.
+    # See https://issues.redhat.com/browse/OCPVE-251 request to support MicroShift
```
pmtk, line 183: Take a look at https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/assets.yaml#L60. We have a field `ignore: "reason"` and it documents why we don't touch some topolvm manifests during rebase.

Also we don't want to include jira urls in source code

copejon:
Hey @pacevedom I've updated the description to provide a bit more context.

I've also remove the jira reference as it's determined that the cause of that bug is not related to this patch.

dhellmann:
This looks OK, except for the messy history.

@pmtk and @pacevedom , we could set the merge method to squash to take the changes. Thoughts?

PR #1384, opened 2023-02-17
USHIFT-387: simplify configuration mechanisms

pacevedom:

Closes #<Issue Number>


docs/howto_config.md
```diff
@@ -1,11 +1,8 @@
 # Configuration
 
-MicroShift can be configured in the following ways, in order of precedence:
-* Command line arguments
-* Environment variables
-* Configuration file
+MicroShift can be configured via a configuration file only.
```
dhellmann, line 3: Saying "only" is relevant to someone who read the old docs, but doesn't feel necessary if we're just documenting how it works. Can we just remove this line, too, and not change the wording on what is now line 5?

dhellmann:
I had one nit, but otherwise this looks good.

docs/howto_config.md
pacevedom: Done

zshi-redhat:
This change has an impact on the proposed pluggable enhancement which uses the env var mechnism for passing microshift config to plugin: https://github.com/openshift/enhancements/pull/1304/files#diff-3f76abf49288e379a81f4c6ba7e0724a6195dca9e5b90ee62663e29c9fa01242R180. 
Added a comment in the enhancement to update the design.

PR #1391, opened 2023-02-21
OCPBUGS-7779: apply service ca-bundle directly to ingress configmap

zshi-redhat:
The data of openshift-ingress service-ca-bundle configmap is erased everytime when microshift starts since the configmap is applied without data. This results in delay in starting router-default deployment pod after reboot as it has to wait for service-ca controller to inject the ca-bundle again. This commit applies service-ca-bundle configmap directly with data generated in service-ca dir.


zshi-redhat:
cc @ggiguash

dhellmann:
This looks OK to me.

We might want @Miciah to take a look.

PR #1395, opened 2023-02-21
ETCD-352: Adding etcd defrag goroutine

dusk125:
This is the implementation for the 'defrag goroutine' approach from [this enhancement](https://github.com/openshift/enhancements/pull/1350)

### DIFF
--- a/etcd/cmd/microshift-etcd/run.go
+++ b/etcd/cmd/microshift-etcd/run.go
@@ -1,19 +1,23 @@
 package main
 
 import (
+	"context"
 	"fmt"
+	"math"
 	"net"
 	"net/url"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"syscall"
+	"time"
 
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 
 	"github.com/spf13/cobra"
 	etcd "go.etcd.io/etcd/server/v3/embed"
+	"go.etcd.io/etcd/server/v3/mvcc/backend"
 	"k8s.io/klog/v2"
 )
 
@@ -30,6 +34,7 @@ func NewRunEtcdCommand() *cobra.Command {
 			return e.Run()
 		},
 	}
+
 	return cmd
 }
 
@@ -43,7 +48,11 @@ var tlsCipherSuites = []string{
 }
 
 type EtcdService struct {
-	etcdCfg *etcd.Config
+	etcdCfg                 *etcd.Config
+	minDefragBytes          int64
+	maxFragmentedPercentage float64
+	defragCheckFreq         time.Duration
+	doStartupDefrag         bool
 }
 
 func NewEtcd(cfg *config.MicroshiftConfig) *EtcdService {
@@ -55,6 +64,11 @@ func NewEtcd(cfg *config.MicroshiftConfig) *EtcdService {
 func (s *EtcdService) Name() string { return "etcd" }
 
 func (s *EtcdService) configure(cfg *config.MicroshiftConfig) {
+	s.minDefragBytes = cfg.Etcd.MinDefragBytes
+	s.maxFragmentedPercentage = cfg.Etcd.MaxFragmentedPercentage
+	s.defragCheckFreq = cfg.Etcd.DefragCheckFreq
+	s.doStartupDefrag = cfg.Etcd.DoStartupDefrag
+
 	microshiftDataDir := config.GetDataDir()
 	certsDir := cryptomaterial.CertsDirectory(microshiftDataDir)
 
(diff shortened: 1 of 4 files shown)


etcd/cmd/microshift-etcd/run.go
```diff
@@ -97,17 +113,67 @@ func (s *EtcdService) Run() error {
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-timer.C:
+			if isBackendFragmented(be) {
+				klog.Info("attempting to defragment backend")
+				start := time.Now()
+				if err := be.Defrag(); err != nil {
+					klog.Errorf("defragmentation failed: %v", err)
+					// TODO handle error
```
dusk125, line 166: In Openshift, the operator will retry a failing defrag 3 times before marking the etcd node degraded.
We could retry (configurable a default of) 3 times before throwing an error and stopping things, or maybe even allow the user to specify a value that just turns it off? Thoughts?

etcd/cmd/microshift-etcd/run.go
```diff
@@ -1,22 +1,38 @@
...
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 
 	"github.com/spf13/cobra"
 	etcd "go.etcd.io/etcd/server/v3/embed"
+	"go.etcd.io/etcd/server/v3/mvcc/backend"
 	"k8s.io/klog/v2"
 )
 
+// TODO these need to be configurable (with an option to completely turn off)
```
dusk125, line 24: I think these should be user configurable so if they don't want/need defrag, they can control how and when it runs.

dusk125: I added the configuration to the Microshift config so the microshift-etcd process itself consumes the config.

etcd/cmd/microshift-etcd/run.go
```diff
@@ -97,17 +112,66 @@ func (s *EtcdService) Run() error {
...
+	}
+
+	// Stop the controller if defrags are disabled.
+	if s.defragCheckFreq == 0 {
+		klog.Warning("defragmentation has been disabled")
+		return nil
+	}
+
+	// This timer will check the fragmented conditions periodically.
+	timer := time.NewTimer(s.defragCheckFreq)
```
pmtk, line 150: Would `time.Ticker` be a better fit here? Or do you want to make sure that between `N defrag end` `N+1 defrag start` is almost exactly `defragCheckFreq` (so it takes length of defrag into an account)?

etcd/cmd/microshift-etcd/run.go
```diff
@@ -97,17 +112,66 @@ func (s *EtcdService) Run() error {
...
 	return nil
 }
 
+func (s *EtcdService) defragController(ctx context.Context, be backend.Backend) error {
+	// If we were told to, go ahead and do a defragment now.
+	if s.doStartupDefrag {
+		if err := be.Defrag(); err != nil {
+			err = fmt.Errorf("initial defragmentation failed: %v", err)
+			klog.Error(err)
+			return err
```
pmtk, line 139: This error is not caught because 
```
go s.defragController(defragCtx, e.Server.Backend())
```
Maybe startup defrag should be outside defragController?

etcd/cmd/microshift-etcd/run.go
dusk125: Yeah, using time.Timer was intentional, in the (rare) cause that defrag takes a significant amount of time and/or if the user sets the time to something very low, I don't want to have the potential for multiple defrags going at the same time.
Using either is probably fine, but at least the time.Timer will save us from that edge case.

etcd/cmd/microshift-etcd/run.go
```diff
@@ -97,17 +112,66 @@ func (s *EtcdService) Run() error {
 	}
 	<-e.Server.ReadyNotify()
 
+	// If we were told to, go ahead and do a defragment now.
+	if s.doStartupDefrag {
+		if err := e.Server.Backend().Defrag(); err != nil {
+			err = fmt.Errorf("initial defragmentation failed: %v", err)
+			klog.Error(err)
+			return err
```
pmtk, line 120: In that case, should we put `e.Server.Stop()` and `<-e.Server.StopNotify()` in defer just after starting?

etcd/cmd/microshift-etcd/run.go
```diff
@@ -97,17 +112,66 @@ func (s *EtcdService) Run() error {
...
+	defer func() {
+		if !timer.Stop() {
+			<-timer.C
+		}
+	}()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
```
pmtk, line 160: Totally nit: according to docs: `When the Timer expires, the current time will be sent on C` so maybe we could do:
```suggestion
		case start := <-timer.C:
```

dusk125: Love it!

etcd/cmd/microshift-etcd/run.go
dusk125: Good catch

pmtk:
I have https://github.com/openshift/microshift/pull/1404 ready - it'll fix vendoring of `microshift` in `etcd/vendor/`.
When you rebase onto that (hopefully it'll make fast into the main), changing `MicroShiftConfig` should be as easy as changing `pkg/config/config.go` and running `go mod vendor` in `etcd/` (that file should be just copied from local dir, not downloaded from remote `github.com/openshift/microshift` repo)

pmtk:
looks good to me
/approve

PR #1397, opened 2023-02-21
USHIFT-907: Remove unsupported topolvm manifests

copejon:

This change removes files that contain the rbac / scc configuration for topolvm workloads (resizer, snapshotter) that are not shipped on MicroShift by default.  As such, they don't enable or impede existing MicroShift components.  Removing them is simply a housecleaning task.

Closes #[USHIFT-907](https://issues.redhat.com//browse/USHIFT-907)


dhellmann:
Could you add a little detail to the description and/or commit message explaining why we don't want these files?

There are 7 commits in the PR. Are the all independent, or can they be squashed into fewer?

copejon:
Hey @dhellmann, np.  I've updated the description.

dhellmann:
@pmtk or @pacevedom , we could set the merge method to squash this as it lands, if we want to get it in this week. What do you think?

PR #1399, opened 2023-02-22
USHIFT-910: Remove redundant coreos package deletion from build.sh

ggiguash:
Tested ISO image creation and boot using build.sh script on RHEL 8 x86.
Closes [USHIFT-910](https://issues.redhat.com//browse/USHIFT-910)


zshi-redhat:
LGTM

How do we find these repo changes in the future?

pacevedom (approved):
/lgtm

oglok:
Reposync is no longer needed, we can use image builder sources directly. If you want, it could be a follow up PR.

PR #1404, opened 2023-02-22
go get & vendor microshift in etcd/go.mod

pmtk:


pmtk:
/hold
I need to address this in rebase procedures

pmtk:
Commits are logically separated, but not functional without each other (all commits on main should be buildable)
/label tide/merge-method-squash

PR #1405, opened 2023-02-22
USHIFT-896: Use UUID for enabling automatic network device connections in kickstart

ggiguash:
Use UUID instead of names in the command to make it work for device names with spaces
Closes [USHIFT-896](https://issues.redhat.com//browse/USHIFT-896)


pacevedom (approved):
/lgtm

PR #1411, opened 2023-02-23
NO-ISSUE: add tool for easier management of developer VMs

dhellmann:
Wrap the existing create script with one that also adds SSH keys and
registers the RHEL subscription. Include a delete command and another
to get the VM's IP.


dhellmann:
Rebased

PR #1412, opened 2023-02-23
NO-ISSUE: add developer tool for getting CI pull secret

dhellmann:
The script uses the token from the console web app to login and get
the pull secret value for the CI registry, then prints it to the
screen to be copied and pasted into ~/.pull_secret.json

### DIFF
--- a/scripts/auto-rebase/get-ci-token.sh
+++ b/scripts/auto-rebase/get-ci-token.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+# Taken from utils.sh in openshift-metal3/dev-scripts
+
+CI_TOKEN=${CI_TOKEN:-$1}
+CI_SERVER=${CI_SERVER:-api.ci.l2s4.p1.openshiftapps.com}
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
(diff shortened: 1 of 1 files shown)


ggiguash:
@dhellmann, this is LGTM, except for one question. 
I think we only use CI tokens during auto-rebase, so does the script belong in that directory?

dhellmann:
> @dhellmann, this is LGTM, except for one question.
> I think we only use CI tokens during auto-rebase, so does the script belong in that directory?

I just always get a token as part of setting up an environment. If it's only useful for the rebase, sure, I could move it.

dhellmann:
> > @dhellmann, this is LGTM, except for one question.
> > I think we only use CI tokens during auto-rebase, so does the script belong in that directory?
> 
> I just always get a token as part of setting up an environment. If it's only useful for the rebase, sure, I could move it.

Moved

ggiguash:
> Moved

Sorry, @dhellmann , but the script seem to still be in the `devenv-builder` directory.

dhellmann:
> > > @dhellmann, this is LGTM, except for one question.
> > > I think we only use CI tokens during auto-rebase, so does the script belong in that directory?
> > 
> > 
> > I just always get a token as part of setting up an environment. If it's only useful for the rebase, sure, I could move it.
> 
> Moved

Committed, and pushed again. 😁

PR #1417, opened 2023-02-27
NO-ISSUE: Add routes and kubectl firewall rules to configure-vm.sh

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/scripts/devenv-builder/configure-vm.sh
+++ b/scripts/devenv-builder/configure-vm.sh
@@ -144,6 +144,13 @@ if [ $BUILD_AND_INSTALL = true ] ; then
     sudo systemctl enable firewalld --now
     sudo firewall-cmd --permanent --zone=trusted --add-source=10.42.0.0/16
     sudo firewall-cmd --permanent --zone=trusted --add-source=169.254.169.1
+    sudo firewall-cmd --permanent --zone=public --add-port=80/tcp
+    sudo firewall-cmd --permanent --zone=public --add-port=443/tcp
+    sudo firewall-cmd --permanent --zone=public --add-port=5353/udp
+    sudo firewall-cmd --permanent --zone=public --add-port=30000-32767/tcp
+    sudo firewall-cmd --permanent --zone=public --add-port=30000-32767/udp
+    sudo firewall-cmd --permanent --zone=public --add-port=6443/tcp
+    sudo firewall-cmd --permanent --zone=public --add-service=mdns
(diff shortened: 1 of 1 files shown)


scripts/devenv-builder/configure-vm.sh
```diff
@@ -144,6 +144,9 @@ if [ $BUILD_AND_INSTALL = true ] ; then
     sudo systemctl enable firewalld --now
     sudo firewall-cmd --permanent --zone=trusted --add-source=10.42.0.0/16
     sudo firewall-cmd --permanent --zone=trusted --add-source=169.254.169.1
+    sudo firewall-cmd --permanent --zone=public --add-port=80/tcp
+    sudo firewall-cmd --permanent --zone=public --add-port=443/tcp
+    sudo firewall-cmd --permanent --zone=public --add-port=6443/tcp
```
ggiguash, line 149: Since we're adding optional settings here, do you think we should add all of the documented ones? 
So, basically, should we replicate the documented settings as-is here?

ggiguash (changes requested):

ggiguash:
/retitle NO-ISSUE: Add routes and kubectl firewall rules to configure-vm.sh

scripts/devenv-builder/configure-vm.sh
dhellmann: I like that.

I think we're missing

```
sudo firewall-cmd --permanent --zone=public --add-port=5353/udp
sudo firewall-cmd --permanent --zone=public --add-port=30000-32767/tcp
sudo firewall-cmd --permanent --zone=public --add-port=30000-32767/udp
sudo firewall-cmd --permanent --zone=public --add-port=6443/tcp
sudo firewall-cmd --permanent --zone=public --add-service=mdns
```

from https://github.com/openshift/microshift/blob/main/docs/howto_firewall.md

pacevedom: I added only the bare minimum, but this might be better. Wondering if we should care about udp though.

ggiguash: I think we want to be as conformant to the configuration docs as possible. If adding UDP does not "hurt", let's have it enabled.

pacevedom: Then lets have it! CI shouldnt be impacted because there are additional firewall rules at the compute level.

PR #1421, opened 2023-02-27
OCPBUGS-2869: patch out unused components in openshift-kube-apiserver

dhellmann:
Remove startup code that adds the resource quota controller and informers
for the Group and User APIs, because none of the relevant APIs are
present in MicroShift.

/hold

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
@@ -9,17 +9,12 @@ import (
 	"github.com/openshift/apiserver-library-go/pkg/securitycontextconstraints/sccadmission"
 	securityv1client "github.com/openshift/client-go/security/clientset/versioned"
 	securityv1informer "github.com/openshift/client-go/security/informers/externalversions"
-	userclient "github.com/openshift/client-go/user/clientset/versioned"
-	userinformer "github.com/openshift/client-go/user/informers/externalversions"
 	"github.com/openshift/library-go/pkg/apiserver/admission/admissionrestconfig"
 	"github.com/openshift/library-go/pkg/apiserver/apiserverconfig"
 	"k8s.io/apiserver/pkg/admission"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	clientgoinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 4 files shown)


vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
```diff
@@ -152,11 +135,6 @@ func newInformers(loopbackClientConfig *rest.Config) (*kubeAPIServerInformers, e
 		OpenshiftUserInformers:     userinformer.NewSharedInformerFactory(userClient, defaultInformerResyncPeriod),
 		OpenshiftConfigInformers:   configv1informer.NewSharedInformerFactory(configClient, defaultInformerResyncPeriod),
 	}
-	if err := ret.OpenshiftUserInformers.User().V1().Groups().Informer().AddIndexers(cache.Indexers{
-		usercache.ByUserIndexName: usercache.ByUserIndexKeys,
-	}); err != nil {
-		return nil, err
-	}
```
tkashem, line 138: https://github.com/openshift/microshift/pull/1421/files#diff-ebfa6ec837b26117893888b31b6d9683c24f3d74583db8bc27ef515a4b754c90L184-L189

I believe, we can't start those anymore
```
i.OpenshiftQuotaInformers.Start(stopCh)
i.OpenshiftUserInformers.Start(stopCh)
```

should we try removing these two fields from the struct `kubeAPIServerInformers` since we don't support either in MicroShift
https://github.com/openshift/microshift/pull/1421/files#diff-ebfa6ec837b26117893888b31b6d9683c24f3d74583db8bc27ef515a4b754c90R142-R147

dhellmann: Good idea, I can look at that.

dhellmann: I've pushed updates that remove those informers and restructured the PR to include the patch file and changes to the vendor file in each commit.

@tkashem , please take another look.

oglok:
Long time ago I opened this PR to address this same issue using carried patches: https://github.com/openshift/microshift/pull/968

We thought it'd be a better idea to fix it directly in the OpenShift KAS, but I agree that if that work is not prioritized, we could fix it temporarily like this, and once it's done upstream, we'll remove the carried patches.

tkashem:
/lgtm

/hold
(please feel free to cancel the hold once you are happy with the ci results)

PR #1422, opened 2023-02-27
USHIFT-896: Fix kickstart to edit autoconnect option in the configuration files

ggiguash:
Implement a solution for enabling secondary network interfaces in kickstart. 

Using `ncli connection modify` command does not update the configuration file when run in post install. As a workaround, edit the configuration files by removing `autoconnect=.*` option from them.

Closes [USHIFT-896](https://issues.redhat.com//browse/USHIFT-896)

### DIFF
--- a/scripts/image-builder/config/kickstart.ks.template
+++ b/scripts/image-builder/config/kickstart.ks.template
@@ -72,8 +72,10 @@ sed -i 's/.*SystemMaxUse=.*/SystemMaxUse=1G/g'   /etc/systemd/journald.conf
 sed -i 's/.*RuntimeMaxUse=.*/RuntimeMaxUse=1G/g' /etc/systemd/journald.conf
 
 # Make sure all the Ethernet network interfaces are connected automatically
-for dev in $(nmcli -f uuid,type,autoconnect connection | awk '$2 == "ethernet" && $3 == "no" {print $1}') ; do
-    nmcli connection modify ${dev} connection.autoconnect yes
+for uuid in $(nmcli -f uuid,type,autoconnect connection | awk '$2 == "ethernet" && $3 == "no" {print $1}') ; do
+    # Remove autoconnect option from the configuration file to keep it enabled after reboot
+    file=$(nmcli -f uuid,filename connection | awk -v uuid=${uuid} '$1 == uuid' | sed "s/${uuid}//g" | xargs)
+    sed -i '/autoconnect=.*/d' "${file}"
 done
 
 %end


zshi-redhat:
@ggiguash  I think autoconnect is yes or true if not explicitly set in the .nmconnection file. 
Why do we think it is not persistent after rebooting?

ggiguash:
> @ggiguash I think autoconnect is yes or true if not explicitly set in the .nmconnection file. Why do we think it is not persistent after rebooting?

In post-install stage: the `.nmconnection` file comes with `autoconnect=false` line for the secondary devices. Running `nmcli connection modify` from the kickstart is not reflected in the configuration file. We must remove the `autoconnect` setting from the file, or the device gets disabled on the next boot.
At the first boot: I can run `nmcli connection modify` to manipulate the `autoconnect` flag and it's immediately reflected in the appropriate `.nmconnection` file.

Since `nmcli connection modify` does not update the `.nmconnection` file when run from kickstart post-install, thus the current implementation directly updating the file.

zshi-redhat:
> > @ggiguash I think autoconnect is yes or true if not explicitly set in the .nmconnection file. Why do we think it is not persistent after rebooting?
> 
> The following statements stem from my tests, backed up by QE findings. The current implementation has been confirmed by QE as working.
> 
> In post-install stage: the `.nmconnection` file comes with `autoconnect=false` line for the secondary devices. Running `nmcli connection modify` from the kickstart is not reflected in the configuration file. We must remove the `autoconnect` setting from the file, or the device gets disabled on the next boot. At the first boot: I can run `nmcli connection modify` to manipulate the `autoconnect` flag and it's immediately reflected in the appropriate `.nmconnection` file.
> 
> As `nmcli connection modify` does not update the `.nmconnection` file when run from kickstart post-install, the current implementation directly updates the file.

@ggiguash thanks for the details! 
I came across this similar bug: https://bugzilla.redhat.com/show_bug.cgi?id=2100883#c19, wonder what is the NM version?

ggiguash:
@zshi-redhat , here's the current version of the tools I've been testing.
```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.1 (Plow)

$ rpm -q NetworkManager
NetworkManager-1.40.0-1.el9.x86_64

$ nmcli -v
nmcli tool, version 1.40.0-1.el9
```

zshi-redhat:
> @zshi-redhat , here's the current version of the tools I've been testing.
> 
> ```
> $ cat /etc/redhat-release 
> Red Hat Enterprise Linux release 9.1 (Plow)
> 
> $ rpm -q NetworkManager
> NetworkManager-1.40.0-1.el9.x86_64
> 
> $ nmcli -v
> nmcli tool, version 1.40.0-1.el9
> ```

Thanks!  this version is even higher than the fixed version in the bugzilla.
I cannot think of a better way to persist the NM config in kickstart stage. If changing NM config after first boot is not an option, I'm fine with current change.

LGTM

PR #1429, opened 2023-03-01
USHIFT-915: Add wait flag when creating VMs to make sure they are restarted

ggiguash:
Closes [USHIFT-915](https://issues.redhat.com//browse/USHIFT-915)

### DIFF
--- a/docs/devenv_setup.md
+++ b/docs/devenv_setup.md
@@ -32,6 +32,7 @@ virt-install \
     --os-type generic \
     --events on_reboot=restart \
     --cdrom ./${ISONAME} \
+    --wait \
 "
 ```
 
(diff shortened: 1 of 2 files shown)


docs/devenv_setup.md
```diff
@@ -32,6 +32,7 @@ virt-install \
     --os-type generic \
     --events on_reboot=restart \
     --cdrom ./${ISONAME} \
+    --wait \
```
dhensel-rh, line 35: Would it be ok to add a time value (in minutes) to this ? I do not know what the default is if you do not include a time value. I have been using --wait 5, and it works ok. Thoughts ?

docs/rhel4edge_iso.md
```diff
@@ -168,6 +168,7 @@ virt-install \
     --os-type generic \
     --events on_reboot=restart \
     --cdrom ./microshift-installer-*.$(uname -i).iso \
+    --wait \
```
dhensel-rh, line 171: Would it be ok to add a time value (in minutes) to this ? I do not know what the default is if you do not include a time value. I have been using --wait 5, and it works ok. Thoughts ?

ggiguash: The manual states:
```
--wait [WAIT]         Minutes to wait for install to complete.
```
So, the wait timeout is optional. I think we should just wait indefinitely until the process completes or fails. 
Picking the "right" timeout is always tricky and I'm not sure it's necessary in this case.

docs/devenv_setup.md
ggiguash: I responded below.

PR #1432, opened 2023-03-02
NP-604: Loadbalancer Service document update

pliurh:
Update the document as we have a built-in loadbalancer controller in Microshift.

### DIFF
--- a/docs/config/nginx-IP-header.yaml
+++ b/docs/config/nginx-IP-header.yaml
@@ -21,17 +21,23 @@ spec:
         app: nginx
     spec:
       containers:
-      - image: nginx:1.18
+      - image: quay.io/packit/nginx-unprivileged
         imagePullPolicy: Always
         name: nginx
         ports:
-        - containerPort: 80
+        - containerPort: 8080
         volumeMounts:
         - name: nginx-configs
           subPath: headers.conf
           mountPath: /etc/nginx/conf.d/headers.conf
-      securityContext:
-        runAsUser: 0
+        securityContext:
+          allowPrivilegeEscalation: false
+          seccompProfile:
+            type: RuntimeDefault
+          capabilities:
+            drop: ["ALL"]
+          runAsNonRoot: true
+          runAsUser: 1000
       volumes:
         - name: nginx-configs
           configMap:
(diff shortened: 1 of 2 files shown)


ggiguash:
@pliurh, do you think there's a problem leaving both built-in and native LB procedures in this document?

I mean, can we stress the built-in LB and put the MetalLB instructions in a separate chapter in the end of the doc?

pliurh:
> @pliurh, do you think there's a problem leaving both built-in and MetalLB procedures in this document?
> 
> I mean, can we stress the built-in LB and put the MetalLB instructions in a separate chapter in the end of the doc?

Currently, there is no configuration flag to turn off the built-in LB controller. There will be conflicts if we use Metal LB. So if there is still a requirement to use Metal LB on Microshift. We'll need to allow users to turn off the built-in LB controller.

dhellmann:
> > @pliurh, do you think there's a problem leaving both built-in and MetalLB procedures in this document?
> > I mean, can we stress the built-in LB and put the MetalLB instructions in a separate chapter in the end of the doc?
> 
> Currently, there is no configuration flag to turn off the built-in LB controller. There will be conflicts if we use Metal LB. So if there is still a requirement to use Metal LB on Microshift. We'll need to allow users to turn off the built-in LB controller.

How does that work on OCP? Do we support multiple load balancers there?

@ggiguash do we know we need MetalLB support long term, or was that the approach for providing a load balancer until this work was done?

pliurh:
> How does that work on OCP? Do we support multiple load balancers there?

For cloud platforms, there are built-in service controller implementations for each cloud provider. For on-prem platforms including baremetal, there is no built-in support in OCP. Metal-LB operator can be installed when needed. I don't think we prevent users from using other implementations as long as they are compatible with our CNIs. 

However, Microshift is special. The features supported by other LB implementations, such as VIP pool management, publishing the VIP on L2/L3, HA, etc, may not be useful in Microshift scenario.

zshi-redhat:
> > @pliurh, do you think there's a problem leaving both built-in and MetalLB procedures in this document?
> > I mean, can we stress the built-in LB and put the MetalLB instructions in a separate chapter in the end of the doc?
> 
> Currently, there is no configuration flag to turn off the built-in LB controller. There will be conflicts if we use Metal LB. So if there is still a requirement to use Metal LB on Microshift. We'll need to allow users to turn off the built-in LB controller.

This is a good point. we shall provide an option to disable build-in load balancer.

pliurh:
I proposed a PR to add the flag to turn the built-in loadbalancer controller off https://github.com/openshift/microshift/pull/1491.

dhellmann:
/approve

Thank you for this, the documentation looks good. Maybe we can move the metallb procedure into a separate file and update it as part of #1491, so that knowledge is not lost?

pliurh:
> /approve
> 
> Thank you for this, the documentation looks good. Maybe we can move the metallb procedure into a separate file and update it as part of #1491, so that knowledge is not lost?

Yeah, it makes sense.

PR #1435, opened 2023-03-02
ETCD-401: Adding etcd config to user-facing config

dusk125:
#1395 Added etcd configuration to the internal Microshift config structure, but not the user-facing configuration. This PR adds the etcd config to the user-facing configuration file.

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -20,6 +20,12 @@ apiServer:
     - ""
 debugging:
   logLevel: ""
+etcd:
+  quotaBackendSize: ""
+  defragCheckFreq: ""
+  doStartupDefrag: false
+  minDefragSize: ""
+  maxFragmentedPercentage: 0
 ```
 
 ## Default Settings
@@ -42,6 +48,12 @@ apiServer:
   subjectAltNames: []
 debugging:
   logLevel: "Normal"
+etcd:
+  quotaBackendSize: "2Gi"
+  defragCheckFreq: "5m"
+  doStartupDefrag: true
+  minDefragSize: "100Mi"
+  maxFragmentedPercentage: 45
 ```
 
 ## Service NodePort range
--- a/packaging/microshift/config.yaml
+++ b/packaging/microshift/config.yaml
@@ -29,3 +29,20 @@ apiServer:
 debugging:
   # Log verbosity ('Normal', 'Debug', 'Trace', 'TraceAll'):
   #logLevel: 'Normal'
+
+etcd:
+  # The limit on the size of the etcd database; the etcd server will start failing writes if its size on disk reaches this value. (Default: 2GB)
+  #quotaBackendSize: '2Gi'
+
+  # How often to check the conditions for defragmenting the etcd database (0 means no defrags, except for a single on startup if `doStartupDefrag` is set). (Default: 5 minutes)
+  #defragCheckFreq: '5m'
+
+  # Whether or not to defragment the etcd database when the etcd server finishes starting. (Default: true)
+  #doStartupDefrag: true
+
+  # Defragment conditions: if both of the following are true when the condition check (controlled by the DefragCheckFreq) runs, defragment the etcd database.
+  # The minimum size of the etcd database, if the database is smaller than this value, defragmenting will not occur. (Default: 100MB)
+  #minDefragSize: '100Mi'
+
+  # The maximum allowed fragmented percentage, if the database is fragmented less than this value, defragmenting will not occur. (Default: 45)
+  #maxFragmentedPercentage: 45
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,6 +17,7 @@ import (
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/pflag"
 
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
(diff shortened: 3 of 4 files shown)


ggiguash:
@dusk125 , if we are making this change, we need to also update the following files:
- Docs: https://github.com/openshift/microshift/blob/main/docs/howto_config.md
- Default config file: https://github.com/openshift/microshift/blob/main/packaging/microshift/config.yaml

packaging/microshift/config.yaml
```diff
@@ -29,3 +29,20 @@ apiServer:
 debugging:
   # Log verbosity ('Normal', 'Debug', 'Trace', 'TraceAll'):
   #logLevel: 'Normal'
+
+etcd:
+  # The limit on the size of the etcd database, in bytes; thee etcd server will start failing writes if its size on disk reaches this value. (Default: 2GB)
```
dhellmann, line 34: Can we do something nicer than make this "in bytes"? Would we expect anyone to ever set this lower than 1MB, for example? I've seen some other kube-related API fields that parse strings like `1GB`. Is that code somewhere we could share it?

```suggestion
  # The limit on the size of the etcd database, in bytes; the etcd server will start failing writes if its size on disk reaches this value. (Default: 2GB)
```

packaging/microshift/config.yaml
```diff
@@ -29,3 +29,20 @@ apiServer:
...
+	#quotaBackendBytes: 2147483648
+
+	# How often, in nanoseconds, to check the conditions for defragmenting the etcd database (0 means no defrags, except for a single on startup if `doStartupDefrag` is set). (Default: 5 minutes)
```
dhellmann, line 37: Same here. Do we expect someone to use nanoseconds? We ought to be able to use at least seconds.

packaging/microshift/config.yaml
```diff
@@ -29,3 +29,20 @@ apiServer:
...
+	#defragCheckFreq: 300000000000
+
+	# Whether or not to defragment the etcd database when the etcd server finishes starting. (Default: true)
+	#dkoStartupDefrag: true
+
+  # Defragment conditions: if both of the following are true when the condition check (controlled by the DefragCheckFreq) runs, defragment the etcd database.
+	# The minimum size of the etcd database, in bytes, if the database is smaller than this value, defragmenting will not occur. (Default: 100MB)
+	#minDefragBytes: 104857600
```
dhellmann, line 45: This should use the same units as the quota backend.

docs/howto_config.md
```diff
@@ -20,6 +20,12 @@ apiServer:
     - ""
 debugging:
   logLevel: ""
+etcd:
+	quotaBackendBytes: 0
+	defragCheckFreq: 0
+	doStartupDefrag: false
+	minDefragBytes: 0
+	maxFragmentedPercentage: 0
```
ggiguash, line 28: Can we fix the tabulation to be consistent with the rest of the file?

docs/howto_config.md
```diff
@@ -42,6 +48,12 @@ apiServer:
   subjectAltNames: []
 debugging:
   logLevel: "Normal"
+etcd:
+  quotaBackendBytes: 214748364
+	defragCheckFreq: 300000000000
+	doStartupDefrag: true
+	minDefragBytes: 104857600
+	maxFragmentedPercentage: 45
```
ggiguash, line 56: Can we fix the tabulation to be consistent with the rest of the file?

packaging/microshift/config.yaml
```diff
@@ -29,3 +29,20 @@ apiServer:
...
+
+  # The maximum allowed fragmented percentage, if the database is fragmented less than this value, defragmenting will not occur. (Default: 45)
+	#maxFragmentedPercentage: 45
```
ggiguash, line 48: Can we fix the tabulation to be consistent with the rest of the file?

docs/howto_config.md
pmtk, line 28:
```suggestion
etcd:
  quotaBackendBytes: 0
  defragCheckFreq: 0
  doStartupDefrag: false
  minDefragBytes: 0
  maxFragmentedPercentage: 0
```

docs/howto_config.md
pmtk, line 56:
```suggestion
etcd:
  quotaBackendBytes: 214748364
  defragCheckFreq: 300000000000
  doStartupDefrag: true
  minDefragBytes: 104857600
  maxFragmentedPercentage: 45
```

packaging/microshift/config.yaml
pmtk, line 48:
```suggestion
etcd:
  # The limit on the size of the etcd database, in bytes; thee etcd server will start failing writes if its size on disk reaches this value. (Default: 2GB)
  #quotaBackendBytes: 2147483648

  # How often, in nanoseconds, to check the conditions for defragmenting the etcd database (0 means no defrags, except for a single on startup if `doStartupDefrag` is set). (Default: 5 minutes)
  #defragCheckFreq: 300000000000

  # Whether or not to defragment the etcd database when the etcd server finishes starting. (Default: true)
  #doStartupDefrag: true

  # Defragment conditions: if both of the following are true when the condition check (controlled by the DefragCheckFreq) runs, defragment the etcd database.
  # The minimum size of the etcd database, in bytes, if the database is smaller than this value, defragmenting will not occur. (Default: 100MB)
  #minDefragBytes: 104857600

  # The maximum allowed fragmented percentage, if the database is fragmented less than this value, defragmenting will not occur. (Default: 45)
  #maxFragmentedPercentage: 45
```

PR #1442, opened 2023-03-03
USHIFT-936, USHIFT-942: reconcile the internal and external config data structures

dhellmann:
This PR eliminates some technical debt by merging the internal and
external representation of our configuration. The PR is structured with
several commits that should each work independently to make reviewing
the large number of changes simpler.

/assign @pmtk
/assign @pacevedom
/assign @copejon


dhellmann:
Tests failing with 
```
Mar 04 00:23:38 release-ci-ci-op-swi8jhnb-5da1b microshift[212294]: ??? F0304 00:23:38.868198  212294 run.go:46] Error in reading or validating configuration: Failed to validate configuration file /etc/microshift/config.yaml: network.serviceNetwork not filled in
[96](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/1442/pull-ci-openshift-microshift-main-e2e-router-smoke-test/1631802154794618880#1:build-log.txt%3A96)
```

pkg/config/files.go
```diff
@@ -0,0 +1,121 @@
...
+func parse(contents []byte) (*Config, error) {
+	c := &Config{}
+	fmt.Printf("parsing %s\n", string(contents))
+	if err := yaml.Unmarshal(contents, c); err != nil {
+		return nil, fmt.Errorf("Unable to decode configuration: %v", err)
+	}
+	if err := c.fillDefaults(); err != nil {
+		return nil, fmt.Errorf("Invalid configuration: %v", err)
+	}
+	if err := c.updateComputedValues(); err != nil {
```
pliurh, line 86: This breaks the CI. This function is invoked in `fillDefaults`. And it will set `c.ApiServer.SkipInterface` to true when it's invoked again.

pkg/config/files.go
```diff
@@ -80,31 +80,37 @@ func parse(contents []byte) (*Config, error) {
...
+		return nil, err
+	}
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing config file %q: %v", configFile, err)
 	}
-	if err := c.updateComputedValues(); err != nil {
+
+	results := &Config{}
+	results.userSettings = userSettings
+	results.incorporateUserSettings()
```
pliurh, line 97: Shall we invoke `incorporateUserSettings` after `fillDefaults`? So that user settings can always overwrite the defaults.

pkg/config/config.go
```diff
@@ -148,8 +154,13 @@ func (c *Config) updateComputedValues() error {
 		}
 	}
 
-	// If KAS advertise address is not configured then grab it from the service
-	// CIDR automatically.
+	// If KAS advertise address configured, we do not want to apply
+	// the IP to the internal interface.
+	if c.userSettings != nil && len(c.userSettings.ApiServer.AdvertiseAddress) != 0 {
+		c.ApiServer.SkipInterface = true
```
pliurh, line 160: If a user set `SkipInterface` to false? Shall we return an error here?

pkg/config/files.go
dhellmann: Yes, great point. The fillDefaults function tries to be smart, but it might be better to have that be simple and move the smarts to the other function. I will try to do that tomorrow.

pkg/config/config.go
dhellmann: The SkipInterface value is a computed value, so it is not currently being read from the file at all. c.userSettings.ApiServer.SkipInterface should always be false. 

@pacevedom is working on another PR that exposes the flag, although we haven't had time to talk about why that's necessary. https://github.com/openshift/microshift/pull/1478

pkg/config/files.go
dhellmann: Done

pkg/config/config.go
```diff
@@ -2,492 +2,233 @@ package config
...
 }
 
-func NewMicroshiftConfig() *MicroshiftConfig {
-	nodeName, err := os.Hostname()
+// fillDefaults forcibly sets the configuration to the default
+// values. We do not use a static struct for the defaults because some
+// of them are computed from the environment. If any error occurs
+// probing the environment, the values in the Config instance are not
+// changed.
+func (c *Config) fillDefaults() error {
```
pacevedom, line 58: Should we use constants for the fixed values in this function?

pacevedom (approved):
Two nits, looking good.

pkg/config/etcd.go
```diff
@@ -0,0 +1,29 @@
...
+
+import "time"
+
+const (
+	EtcdStartupDefragEnabled  = "Enabled"
+	EtcdStartupDefragDisabled = "Disabled"
+)
+
+type EtcdConfig struct {
+	// The limit on the size of the etcd database; etcd will start failing writes if its size on disk reaches this value
```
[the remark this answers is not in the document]
dhellmann: I pushed an update for this.

pkg/config/config.go
dhellmann: We could. The main point of having constants is to reuse the same value in multiple places. In this case most of these values only appear the one time, except for tests. I could prepare another PR to see what the changes would look like.

pacevedom: That was my point, yes. In order to reduce cognitive load when changing tests (or even md docs in the repo), have them all in a single place easy to look for. I am ok with doing it in another PR.

pacevedom:
/lgtm
/hold
I am ok with the change, holding just in case you want more reviews. Feel free to remove it.

Makefile
```diff
@@ -244,4 +244,10 @@ vendor:
 		echo "Applying patch $$p"; \
 		git mailinfo /dev/null /dev/stderr 2<&1- < $$p | git apply --reject || exit 1; \
 	done
+	make vendor-etcd
```
pmtk, line 247: PTAL https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase.sh#L929
We can remove `(cd "${dirpath}" && make vendor)` and just run `make vendor` in `microshift/`

pkg/config/files.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func getActiveConfigFromYAML(contents []byte) (*Config, error) {
+	userSettings, err := parse(contents)
+	if err != nil {
+		return nil, err
+	}
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing config file %q: %v", configFile, err)
+	}
```
pmtk, line 92: 1st `if` looks unnecessary

dhellmann:
@pacevedom @pmtk  I have rebased this and it is ready for another review.

Makefile
dhellmann: I worry that just running `make vendor` will have other side effects for the top level vendor dir. I can change that line to run `make vendor-etcd` though.

dhellmann: Actually, looking closer at that script, it updates each directory one at a time and commits each separately. We can't do that if we run `make vendor` and using `make vendor-etcd` assumes that there is only 1 submodule.

I think there's some work to optimize that script, but maybe we can leave it for another PR to give us time to consider how we want it to work?

Makefile
```diff
@@ -244,4 +244,10 @@ vendor:
...
 .PHONY: vendor
+
+# Update the etcd dependencies, including especially MicroShift itself.
+vendor-etcd:
+	(cd etcd && go mod vendor)
```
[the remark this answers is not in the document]
dhellmann: I'll change it to use make recursively.

pkg/config/files.go
dhellmann: Good catch, thanks!

pkg/controllers/kube-apiserver.go
```diff
@@ -110,14 +110,14 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 	}
 
 	// Get the apiserver port so we can set it as an argument
-	apiServerPort, err := cfg.Cluster.ApiServerPort()
+	apiServerPort, err := cfg.ApiServerPort()
```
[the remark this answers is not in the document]
dhellmann: Good idea, done

Makefile
dhellmann: I've removed this line for now. I'm trying to do is get to a point where we could run a job that tries to update all of the vendor directories and fails if any changes are found (because we want developers to update the vendor dir in their PR if they change something that needs to be revendored), but I can do that in another PR.

pmtk:
/lgtm
/hold
@pacevedom please unhold when you review

dhellmann:
I've marked this WIP because @dusk125 is going to be removing the etcd configuration options (more details to follow elsewhere). I will rebase this change on top of that work when it is done.

dhellmann:
Rebased to include #1506

pkg/config/apiserver.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// The URL of the API server
+	URL string `json:"-"`
+}
+
+// extract the api server port from the cluster URL
+func (a *ApiServer) Port() (int, error) {
+	var port string
+
+	parsed, err := url.Parse(a.URL)
```
eggfoobar, line 29: If the `ApiServer.URL` and `Port` value is an internally created one and not changing at runtime, what do you think about changing the type to `URL` and populate a `Port` attribute on instantiation, that way we push out the error check so it's not on the caller to handle?

```go
type ApiServer struct {
        ...
	// The URL of the API server
	URL  *url.URL `json:"-"`
        // The Port of the API server
        Port int      `json:"-"`
}
```

pkg/config/debugging.go
```diff
@@ -0,0 +1,25 @@
...
+type Debugging struct {
+	// Valid values are: "Normal", "Debug", "Trace", "TraceAll".
+	// Defaults to "Normal".
+	LogLevel string `json:"logLevel"`
+}
+
+// GetVerbosity returns the numerical value for LogLevel which is an enum
+func (c *Config) GetVerbosity() int {
+	var verbosity int
+	switch c.Debugging.LogLevel {
```
eggfoobar, line 12: What do you think about doing a `strings.ToLower(c.Debugging.LogLevel)` and match on lowercase to avoid the odd chance a user adds incorrect capitalization?

dhellmann: That makes sense. This PR is already pretty big. Maybe we can do that in a follow-up?

pkg/config/apiserver.go
dhellmann: I think I like that. I can experiment with it when I rebase this on #1506

pkg/config/debugging.go
eggfoobar: Yeah absolutely, I have no issue with it.

dhellmann:
/hold cancel

I believe this is ready for review.

pkg/config/apiserver.go
dhellmann: I decided to stick with a string for the URL, since that's how it's used, and added a separate field for the port.

pkg/config/debugging.go
dhellmann: I opened USHIFT-1043 to remember to do this later.

eggfoobar:
LGTM, I was able to pull this and run it with out any issues.

pkg/config/files.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func GetManifestsDir() []string {
+	return manifestsDir
+}
+
+// Returns the default user config file if that exists, else the default global
+// config file, else the empty string.
+func findConfigFile() string {
+	userConfigFile, _ := homedir.Expand(DefaultUserConfigFile)
```
pmtk, line 44: It just occurred to me that we wanted to remove user dir (microshift needs to be run as root anyway), but given the size of the PR and effort spent on reviewing it we might do it in another PR :)

pmtk:
@eggfoobar gave +1. Also saw @copejon +1 on slack
/lgtm

PR #1463, opened 2023-03-08
OCPBUGS-8425: Fix the kustomization procedure not to exit with fatal error

ggiguash:
Closes OCPBUGS-8425

Example of a failed run, which backs off after 1m of retries, logs an error and proceeds.
```
kustomizer I0312 18:30:20.289488  180399 manager.go:114] Starting kustomizer
kustomizer I0312 18:30:20.289524  180399 apply.go:64] No kustomization found at /usr/lib/microshift/manifests/kustomization.yaml
kustomizer I0312 18:30:20.289532  180399 apply.go:57] Applying kustomization at /etc/microshift/manifests/kustomization.yaml 
kustomizer I0312 18:30:20.738705  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer I0312 18:30:30.790331  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer I0312 18:30:40.806277  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer I0312 18:30:50.811309  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer I0312 18:31:00.794630  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer I0312 18:31:10.800532  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer I0312 18:31:20.785916  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer I0312 18:31:20.857375  180399 apply.go:71] Applying kustomization failed: accumulating resources: accumulating resources from 'busybox.yaml': MalformedYAMLError: yaml: line 20: could not find expected ':' in File: busybox.yaml. Retrying in 10s.
kustomizer E0312 18:31:20.857418  180399 apply.go:59] Applying kustomization at /etc/microshift/manifests/kustomization.yaml failed: timed out waiting for the condition. Giving up.
kustomizer I0312 18:31:20.857431  180399 manager.go:119] kustomizer completed
```

### DIFF
--- a/pkg/kustomize/apply.go
+++ b/pkg/kustomize/apply.go
@@ -56,7 +56,7 @@ func (s *Kustomizer) ApplyKustomizationPath(path string) {
 	if _, err := os.Stat(kustomization); !errors.Is(err, os.ErrNotExist) {
 		klog.Infof("Applying kustomization at %v ", kustomization)
 		if err := ApplyKustomizationWithRetries(path, s.kubeconfig); err != nil {
-			klog.Fatalf("Applying kustomization at %v failed: %s. Giving up.", kustomization, err)
+			klog.Errorf("Applying kustomization at %v failed: %s. Giving up.", kustomization, err)
 		} else {
 			klog.Infof("Kustomization at %v applied successfully.", kustomization)
 		}
@@ -66,7 +66,7 @@ func (s *Kustomizer) ApplyKustomizationPath(path string) {
 }
 
 func ApplyKustomizationWithRetries(kustomization string, kubeconfig string) error {
-	return wait.Poll(retryInterval, retryTimeout, func() (bool, error) {
+	return wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
 		if err := ApplyKustomization(kustomization, kubeconfig); err != nil {
 			klog.Infof("Applying kustomization failed: %s. Retrying in %s.", err, retryInterval)
 			return false, nil


pkg/kustomize/apply.go
```diff
@@ -66,13 +68,29 @@ func (s *Kustomizer) ApplyKustomizationPath(path string) {
...
 func ApplyKustomizationWithRetries(kustomization string, kubeconfig string) error {
-	return wait.Poll(retryInterval, retryTimeout, func() (bool, error) {
+	retryCount := 0
+	retryError := fmt.Errorf("reached maximum of %v retries", retryMaxCount)
+
+	err := wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
+		retryCount++
+		if retryCount > retryMaxCount {
+			klog.Infof("Applying kustomization failed: %v", retryError)
+			return true, nil
```
pacevedom, line 78: Why not just return the `retryError` here and simplify the return at the end?

pkg/kustomize/apply.go
```diff
@@ -66,7 +68,15 @@ func (s *Kustomizer) ApplyKustomizationPath(path string) {
...
 
 func ApplyKustomizationWithRetries(kustomization string, kubeconfig string) error {
-	return wait.Poll(retryInterval, retryTimeout, func() (bool, error) {
+	retryCount := 0
+
+	return wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
+		retryCount++
+		if retryCount > retryMaxCount {
+			retryError := fmt.Errorf("reached maximum of %v retries", retryMaxCount)
+			klog.Infof("Applying kustomization failed: %v", retryError)
```
pacevedom, line 77: Small nit, this is already logged in line 61, but up to you :)

pacevedom (approved):

pkg/kustomize/apply.go
ggiguash: Agreed. Let's make it even simpler.

pkg/kustomize/apply.go
```diff
@@ -21,6 +22,7 @@ import (
 const (
 	retryInterval = 10 * time.Second
 	retryTimeout  = 1 * time.Minute
+	retryMaxCount = 3
```
dhellmann, line 25: Why 3? Should we retry at all?

ggiguash: Can the apply operation take time? We have timeouts and retries to mitigate that, I think.

dhellmann: I'm not sure why any given API call would take very long. The resource might not be updated immediately (a Deployment might not produce a Pod right away, for example), but the API call itself should return quickly.

How complex do we want this to be? Should we have some sort of backoff algorithm for the delay between retries? Should it retry indefinitely (with that extra delay), like a normal controller, in case some agent updates the files?

ggiguash: We can use different versions of Poll function to implement poll techniques according to our choice.
I would leave this with a few retries and 10s delay in between and see if we get more complains from the end-users.

zshi-redhat: @ggiguash  My understanding of the `indefinite retry` is that microshift logs error and still proceeds when kustomize fails, but it comes back after the syncInterval and run kustomize again with the hope that user agent may fix the bad yamls.

ggiguash: I wonder if this scenario is useful on edge devices. If an update is installed affecting the manifests, the reboot needs to take place anyway, right?

ggiguash: @dhellmann / @zshi-redhat 
To make progress with this PR, I need your recommendation on the number of retries. We can set it to 1, 3, or -1 for indefinite.

zshi-redhat: @ggiguash  It may not be just about changing the retry value.
I wonder if we could implement an indefinite retry on the kustomize service controller, rather than inside the kustomize apply function. It means that microshift will proceed to start the rest of service controllers (kubelet, load balancer service controllers) after the first call of kustomize fails, but be able to rerun the kustomize service controller based on the last kustomize return value.

ggiguash: I'm not sure if this approach is necessary for edge devices where the updates actually require rebooting the host. If we're rebooting, the indefinite retry seems to be redundant, doesn't it?

pkg/kustomize/apply.go
```diff
@@ -56,7 +58,7 @@ func (s *Kustomizer) ApplyKustomizationPath(path string) {
 	if _, err := os.Stat(kustomization); !errors.Is(err, os.ErrNotExist) {
 		klog.Infof("Applying kustomization at %v ", kustomization)
 		if err := ApplyKustomizationWithRetries(path, s.kubeconfig); err != nil {
-			klog.Fatalf("Applying kustomization at %v failed: %s. Giving up.", kustomization, err)
+			klog.Errorf("Applying kustomization at %v failed: %s. Giving up.", kustomization, err)
```
dhellmann, line 61: This is the only change needed. The call to `Fatalf` causes microshift to exit, which means the API is unavailable. Systemd eventually restarts microshift, and the cycle resumes. If this call is changed to `Errorf` then the error is logged, the kustomizer eventually times out and MicroShift goes into a ready state. We don't need the changes to the retry loop to fix the bug.

dhellmann (commented):
/lgtm

Issue #1449, opened 2023-03-07
[Bug] Microshift does not come up if the Hostname has an upper case letter


AllenZMC:
```
Mar 07 11:36:57 WHIIPVR00291 microshift[164043]: kubelet E0307 11:36:57.280990  164043 kubelet.go:2471] "Error getting node" err="node \"whiipvr00291\" not found"

Mar 07 11:36:57 WHIIPVR00291 microshift[164043]: kube-apiserver I0307 11:36:57.315820  164043 node_authorizer.go:285] NODE DENY: 'WHIIPVR00291' &authorizer.AttributesRecord{User:(*user.DefaultInfo)(0xc010247380), Verb:"get", Namespace:"", APIGroup:"storage.k8s.io", APIVersion:"v1", Resource:"csinodes", Subresource:"", Name:"whiipvr00291", ResourceRequest:true, Path:"/apis/storage.k8s.io/v1/csinodes/whiipvr00291"}

Mar 07 11:36:57 WHIIPVR00291 microshift[164043]: kubelet I0307 11:36:57.316651  164043 csi_plugin.go:1036] Failed to contact API server when waiting for CSINode publishing: csinodes.storage.k8s.io "whiipvr00291" is forbidden: User "system:node:WHIIPVR00291" cannot get resource "csinodes" in API group "storage.k8s.io" at the cluster scope: can only access CSINode with the same name as the requesting node

Mar 07 11:36:57 WHIIPVR00291 microshift[164043]: kubelet E0307 11:36:57.381809  164043 kubelet.go:2471] "Error getting node" err="node \"whiipvr00291\" not found"
```

What is the cause of this and how to solve it ?

ggiguash:
Thank you for reporting this issue, @AllenZMC.

From the log messages, it looks like your host name is defined with uppercase. We have a known issue about MicroShift not working in this configuration. See [OCPBUGS-8411](https://issues.redhat.com/browse/OCPBUGS-8411)

ggiguash:
@AllenZMC , The fix for this problem has been merged at #1453. Can you rebuild MicroShift binary in the main branch and confirm it works for you now?

ggiguash:
/retitle [Bug] Microshift does not come up if the Hostname has an upper case letter

AllenZMC:
@ggiguash 
How to cherry-pick to the release4.12 branch, I want to use the 4.12 branch to compile？

ggiguash:
[...]

PR #1470, opened 2023-03-08
[release-4.12] Use lower case node names only

AllenZMC:
https://github.com/openshift/microshift/issues/1449

Closes #<Issue Number>

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -215,7 +215,7 @@ func NewMicroshiftConfig() *MicroshiftConfig {
 	return &MicroshiftConfig{
 		LogVLevel:       2,
 		SubjectAltNames: subjectAltNames,
-		NodeName:        nodeName,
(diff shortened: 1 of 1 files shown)


ggiguash:
@AllenZMC , thank you for opening this PR. 
I discussed it with the team and we would like to avoid backporting this functionality into 4.12 release because it's not a critical problem with an easy workaround. We encourage users to utilize the main branch or the 4.13 release targeted in a few weeks.

We already have a PR #1454 open to port this fix into the 4.13 release, so you should be able to take advantage of that. Note that you can use MicroShift 4.13 engineering candidate builds from https://mirror.openshift.com/pub/openshift-v4/x86_64/microshift/ocp-dev-preview/latest-4.13/ before we officially release this code in rhocp-4.13 channel

I'm closing this PR due to the above
/close

PR #1478, opened 2023-03-09
OCPBUGS-8277: Use a different internal IP for apiserver connectivity

### LINKED ISSUE #1460: [BUG] Cannot validate certificate for 10.43.0.1 because it doesn't contain any IP SANs
#### What happened?

I am using the [MetalLB] load balancer and found that the speaker is attempting to connect to my Kubernetes API endpoint at https://10.43.0.1:443 but cannot validate the certificate for 10.43.0.1 due to this error: `Get \"https://10.43.0.1:443/api?timeout=32s\": x509: cannot validate certificate for 10.43.0.1 because it doesn't contain any IP SANs`

I've encountered this twice after upgrading my MicroShift RPMs.  The first time was when upgrading from 4.12.3 to 4.12.4 and to resolve the issue, I reverted my MicroShift version back to 4.12.2.  During the upgrade, I also noticed my `/etc/microshift/config.yaml` was malformed, so I corrected it.  A couple of weeks and upgrades later, I've noticed that I'm running into the same issue again, and the `/etc/microshift/config.yaml` appears to be fine this time around.

Any suggestions on what this issue might be?

API call at https://10.43.0.1:443/api should not run into an IP SAN issue as I would expect the API certificate to assume 10.43.0.1 to be trusted.

1. Installed a MicroShift 4.12 cluster
2. Installed [MetalLB] version 0.13.9 through these instructions:

```
$ oc apply -f https://raw.githubusercontent.com/metallb/metallb/v0.13.9/config/manifests/metallb-native.yaml
$ oc adm policy add-scc-to-user privileged -z controller -n metallb-system
$ oc adm policy add-scc-to-user privileged -z speaker -n metallb-system
```
3. Upgrade MicroShift package version (`sudo dnf update`).  I encountered this issue when upgrading from 4.12.3 to 4.12.4 and again from 4.12.5 to 4.12.6.

#### Environment
- MicroShift version (use `microshift version`):
```
$ rpm -qa | grep microshift
microshift-selinux-4.12.6-202303012057.p0.g50997a2.assembly.4.12.6.el8.noarch
microshift-networking-4.12.6-202303012057.p0.g50997a2.assembly.4.12.6.el8.x86_64
microshift-4.12.6-202303012057.p0.g50997a2.assembly.4.12.6.el8.x86_64
```
- Hardware configuration:
```
# cat /etc/os-release
VERSION="8.7 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.7 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/8/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.7
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"
```
- Kernel (e.g. `uname -a`):
```
# uname -a
Linux <hostname obfuscated> 4.18.0-425.13.1.el8_7.x86_64 #1 SMP Thu Feb 2 13:01:45 EST 2023 x86_64 x86_64 x86_64 GNU/Linux
```
- Others:

#### Relevant logs 

```
$ cat /etc/microshift/config.yaml
dns:
  baseDomain: <obfuscated>
network:
  clusterNetwork:
    - cidr: 10.42.0.0/16
  serviceNetwork:
    - 10.43.0.0/16
node:
  hostnameOverride: '<obfuscated>'
  nodeIP: '<obfuscated>'
apiServer:
  subjectAltNames: []
debugging:
  logLevel: "Normal"
```

The speaker pod happens to be an infrastructure application that is calling the API endpoint:
```
$ oc logs speaker-7zs9f -n metallb-system
{"branch":"dev","caller":"main.go:93","commit":"dev","goversion":"gc / go1.19.5 / amd64","level":"info","msg":"MetalLB speaker starting version 0.13.9 (commit dev, branch dev)","ts":"2023-03-07T22:40:28Z","version":"0.13.9"}
{"caller":"announcer.go:115","event":"createARPResponder","interface":"wlp2s0","level":"info","msg":"created ARP responder for interface","ts":"2023-03-07T22:40:28Z"}
{"caller":"announcer.go:124","event":"createNDPResponder","interface":"wlp2s0","level":"info","msg":"created NDP responder for interface","ts":"2023-03-07T22:40:28Z"}
{"caller":"announcer.go:115","event":"createARPResponder","interface":"br-int","level":"info","msg":"created ARP responder for interface","ts":"2023-03-07T22:40:28Z"}
{"caller":"announcer.go:115","event":"createARPResponder","interface":"ovn-k8s-mp0","level":"info","msg":"created ARP responder for interface","ts":"2023-03-07T22:40:28Z"}
{"caller":"announcer.go:115","event":"createARPResponder","interface":"virbr0","level":"info","msg":"created ARP responder for interface","ts":"2023-03-07T22:40:28Z"}
{"caller":"announcer.go:115","event":"createARPResponder","interface":"br-ex","level":"info","msg":"created ARP responder for interface","ts":"2023-03-07T22:40:28Z"}
{"caller":"announcer.go:124","event":"createNDPResponder","interface":"br-ex","level":"info","msg":"created NDP responder for interface","ts":"2023-03-07T22:40:28Z"}
{"level":"error","ts":"2023-03-07T22:40:28Z","msg":"Failed to get API Group-Resources","error":"Get \"https://10.43.0.1:443/api?timeout=32s\": x509: cannot validate certificate for 10.43.0.1 because it doesn't contain any IP SANs","stacktrace":"sigs.k8s.io/controller-runtime/pkg/cluster.New\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.1/pkg/cluster/cluster.go:161\nsigs.k8s.io/controller-runtime/pkg/manager.New\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.1/pkg/manager/manager.go:351\ngo.universe.tf/metallb/internal/k8s.New\n\t/go/go.universe.tf/metallb/internal/k8s/k8s.go:129\nmain.main\n\t/go/go.universe.tf/metallb/speaker/main.go:156\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
{"level":"error","ts":"2023-03-07T22:40:28Z","logger":"setup","msg":"unable to start manager","error":"Get \"https://10.43.0.1:443/api?timeout=32s\": x509: cannot validate certificate for 10.43.0.1 because it doesn't contain any IP SANs","stacktrace":"go.universe.tf/metallb/internal/k8s.New\n\t/go/go.universe.tf/metallb/internal/k8s/k8s.go:149\nmain.main\n\t/go/go.universe.tf/metallb/speaker/main.go:156\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
```

[MetalLB]: https://github.com/openshift/microshift/blob/main/docs/howto_load_balancer.md

pacevedom:

Closes #1460

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
-	ctrl "k8s.io/kubernetes/pkg/controlplane"
 	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/microshift/pkg/util"
@@ -466,16 +465,24 @@ func (c *MicroshiftConfig) ReadAndValidate(configFile string) error {
 	}
 	c.Cluster.DNS = clusterDNS
 
-	// If KAS advertise address is not configured then grab it from the service
+	// If KAS advertise address is not configured then compute it from the service
 	// CIDR automatically.
 	if len(c.KASAdvertiseAddress) == 0 {
 		// unchecked error because this was done when getting cluster DNS
 		_, svcNet, _ := net.ParseCIDR(c.Cluster.ServiceCIDR)
-		_, apiServerServiceIP, err := ctrl.ServiceIPRange(*svcNet)
-		if err != nil {
-			return fmt.Errorf("error getting apiserver IP: %v", err)
+		// Since the KAS advertise address was not provided we will default to the
+		// next immediate subnet after the service CIDR. This is due to the fact
+		// that using the actual apiserver service IP as an endpoint slice breaks
+		// host network pods trying to reach apiserver, as the VIP 10.43.0.1:443 is
+		// not translated to 10.43.0.1:6443. It remains unchanged and therefore
+		// connects to the ingress router instead, triggering all sorts of errors.
+		nextSubnet, exceed := cidr.NextSubnet(svcNet, 32)
+		if exceed {
+			return fmt.Errorf("unable to compute next subnet from service CIDR")
 		}
-		c.KASAdvertiseAddress = apiServerServiceIP.String()
+		// First and last are the same because of the /32 netmask.
+		firstValidIP, _ := cidr.AddressRange(nextSubnet)
+		c.KASAdvertiseAddress = firstValidIP.String()
 		c.SkipKASInterface = false
 	} else {
 		c.SkipKASInterface = true
--- a/docs/howto_firewall.md
+++ b/docs/howto_firewall.md
@@ -8,6 +8,7 @@ It is mandatory to allow MicroShift pods the access to the internal CoreDNS and 
 |:-------------|:---------------------|:-----------|
 |10.42.0.0/16  |Yes                   |Pod network access to other pods |
 |10.43.0.0/16  |No                    |ClusterIP service network, used by pods to access services (like CoreDNS and MicroShift API) |
+|10.44.0.0/32  |No                    |Next available subnet from ClusterIP service network, used by pods to access apiserver |
 |169.254.169.1 |Yes                   |Special IP to access services backed by host endpoints, like MicroShift API Server |
 
 The following ports are optional and they should be considered for MicroShift if a firewall is enabled.
--- a/docs/network/host_networking.md
+++ b/docs/network/host_networking.md
@@ -34,6 +34,8 @@ Not all IP addresses are attached to specific physical interfaces, some of them 
 
 **NOTE:** These IPs shall be reserved for MicroShift.
 
+**NOTE:** There is another special IP reserved for MicroShift's apiserver. In order to allow external access using host IPs a new local IP must be allocated for the apiserver. It defaults to the first IP in the next available subnet from the service CIDR. If service CIDR is 10.43.0.0/16, then the new IP will be 10.44.0.0/32. This default IP is added to the loopback interface to allow connectivity without ovnk.
+
 ## Interfaces
 
 The following physical network interfaces are created or modified by ovn-kubernetes:
--- a/docs/network/ovn_kubernetes_traffic_flows.md
+++ b/docs/network/ovn_kubernetes_traffic_flows.md
@@ -242,12 +242,12 @@ What traffic uses this flow?
 
 Example: from `pod-1` to kubernetes APIServer service `10.43.0.1:443`.
 
-|step|where                    |srcIP:port      |dstIP:port          |comment                                                                                 |
-|:---|:------------------------|:---------------|:-------------------|:---------------------------------------------------------------------------------------|
-|1   |ovn node switch          |10.42.0.12      |10.43.0.1:443       |a loadbalancer rule DNAT’s the service IP 10.43.0.1:443 to endpoint 192.168.122.14:6443 |
-|2   |ovn cluster router       |10.42.0.12      |192.168.122.14:6443 |a router policy rule routes the packet to ovn-k8s-mp0 via ovn node switch               |
-|3   |ovn node switch          |10.42.0.12      |192.168.122.14:6443 |forward to ovn-k8s-mp0                                                                  |
-|4   |ovn-k8s-mp0              |10.42.0.12      |192.168.122.14:6443 |packet is received by the host                                                          |
+|step|where                    |srcIP:port      |dstIP:port     |comment                                                                            |
+|:---|:------------------------|:---------------|:--------------|:----------------------------------------------------------------------------------|
+|1   |ovn node switch          |10.42.0.12      |10.43.0.1:443  |a loadbalancer rule DNAT’s the service IP 10.43.0.1:443 to endpoint 10.44.0.0:6443 |
+|2   |ovn cluster router       |10.42.0.12      |10.44.0.0:6443 |a router policy rule routes the packet to ovn-k8s-mp0 via ovn node switch          |
+|3   |ovn node switch          |10.42.0.12      |10.44.0.0:6443 |forward to ovn-k8s-mp0                                                             |
+|4   |ovn-k8s-mp0              |10.42.0.12      |10.44.0.0:6443 |packet is received by the host                                                     |
 
 Step 1, load balancer in ovn node switch:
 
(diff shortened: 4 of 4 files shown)


pkg/config/config.go
```diff
@@ -466,19 +472,21 @@ func (c *MicroshiftConfig) ReadAndValidate(configFile string) error {
...
-	// If KAS advertise address is not configured then grab it from the service
+	// If KAS advertise address is not configured then compute it from the service
 	// CIDR automatically.
 	if len(c.KASAdvertiseAddress) == 0 {
 		// unchecked error because this was done when getting cluster DNS
 		_, svcNet, _ := net.ParseCIDR(c.Cluster.ServiceCIDR)
-		_, apiServerServiceIP, err := ctrl.ServiceIPRange(*svcNet)
-		if err != nil {
-			return fmt.Errorf("error getting apiserver IP: %v", err)
+		// Using first valid IP from next subnet as using the actual apiserver IP
```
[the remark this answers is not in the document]
pacevedom: What it does is to get the next immediate subnet from the service CIDR and use that IP.
Given that we need a non-service-CIDR IP to setup for the apiserver endpoint this is the most trivial approach I could think of.
This is also why both parameters (the actual address and the lo interface getting it) are now configurable, since this is an additional IP and there might be cases where we need a different one because of collisions.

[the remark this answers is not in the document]
dhellmann: Could you expand the comment in the code to explain all of that? The text that is there now might have been clear to someone who understood how we were already choosing an IP, but was not enough for me to understand what was going on.

pkg/config/config.go
```diff
@@ -161,6 +160,9 @@ type ApiServer struct {
 	// AdvertiseAddress for endpoint slices in kubernetes service. Developer
 	// only parameter, wont show in show-config commands or docs.
 	AdvertiseAddress string `json:"advertiseAddress,omitempty"`
+	// SkipInterfaceConfiguration determines whether the loopback interface
+	// should be assigned the AdvertiseAddress configured before.
+	SkipInterfaceConfiguration bool `json:"skipInterfaceConfiguration,omitempty"`
```
dhellmann, line 165: Why do we want to make the user make this choice? Is this something we can figure out on our own?

pkg/config/config.go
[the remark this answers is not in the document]
pacevedom: Done

pkg/config/config.go
pacevedom: Not really, this is a complex. You have all these possibilities:
- Custom AdvertiseAddress + SkipInterfaceConfiguration=true: This means the api server is either in a different node or is using an already configured interface.
- Custom AdvertiseAddress + SkipInterfaceConfiguration=false: This is deliberately ignoring the default next_subnet_after_service_cidr ip, as there might be collisions with that range or simply put it is required to have a different subnet. It does configure the lo interface with the ip.
- No AdvertiseAddress + SkipInterfaceConfiguration=true: Same as having a custom advertise address, there might be an interface already configured with the first ip from next_subnet_after_service_cidr.
- No AdvertiseAddress + SkipInterfaceConfiguration=false: Default everything, it will configure the first valid ip from next_subnet_after_service_cidr in lo interface. This will be the common case.

zshi-redhat: Should we make the `SkipInterfaceConfiguration` an implicit (devel-only) option and assume it is false for single node deployment? because it seems `SkipInterfaceConfiguration=true` is for multi-node consideration.

pacevedom: Multi node does require this option, however you could also have an IP range you can use in the node and there is no need to configure it for lo interface. I wonder if this is something that could happen.
Also, it defaults to false: https://github.com/openshift/microshift/pull/1478/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R251

dhellmann: I'm still unclear on why we need the boolean at all. If the user provides a custom address, we can say that they must pre-configure the interface to use with that address, even if they just use the loopback interface. If they do not provide an address, then we will configure the loopback interface with an address we choose.

pacevedom: So are we ok to say that if not using default then the configuration side of the interface lies on the admin? Having in mind that custom here means "anything different than "service-CIDR-next-subnet". If that is ok then I am totally fine to remove the exposed option!

dhellmann:
@pacevedom maybe we could take #1442 before this one? It should be easy to add a method to report when the user has provided a value for the IP and update the interface management logic based on that.

pkg/config/config.go
dhellmann: > So are we ok to say that if not using default then the configuration side of the interface lies on the admin? Having in mind that custom here means "anything different than service-CIDR-next-subnet". If that is ok then I am totally fine to remove the exposed option!

Yes, I think that's consistent with how we separate responsibilities for the OS setup in other cases.

pkg/config/config.go
```diff
@@ -466,19 +472,24 @@ func (c *MicroshiftConfig) ReadAndValidate(configFile string) error {
...
-		_, apiServerServiceIP, err := ctrl.ServiceIPRange(*svcNet)
-		if err != nil {
-			return fmt.Errorf("error getting apiserver IP: %v", err)
+		// Since the KAS advertise address was not provided we will default to the
+		// next immediate subnet after the service CIDR. This is due to the fact
+		// that using the actual apiserver service IP as an endpoint slice breaks
+		// host network pods trying to reach apiserver, as the VIP 10.43.0.1:443 is
+		// not translated to 10.43.0.1:6443. It remains unchanged and therefore
+		// connects to the ingress router instead, triggering all sorts of errors.
+		nextSubnet, exceed := cidr.NextSubnet(svcNet, 32)
```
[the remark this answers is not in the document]
pacevedom: Yes, that should be documented.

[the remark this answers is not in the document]
pacevedom: Added docs commit.

PR #1486, opened 2023-03-10
USHIFT-289 Changes to microshift-validate/ for ARM CI

pmtk:


validate-microshift/kuttl-test.sh
```diff
@@ -5,12 +5,19 @@ set -x
...
 
-KUTTL_VERSION="0.10.0"
+KUTTL_VERSION="0.15.0"
 KUTTL="$ROOT/bin/kuttl"
 
+if [ $(uname -p) == "x86_64" ]; then
+    ARCH=x86_64
+else
+    ARCH=arm64
+fi
```
ggiguash, line 15: A nit: to make this code future proof, we'd better have a case here for supported platforms with an error indication for the unsupported ones.

scripts/devenv-builder/configure-vm.sh
```diff
@@ -52,23 +52,24 @@ fi
...
+    GO_VER=1.19.4
+    GO_ARCH=$([ "$(uname -i)" == "x86_64" ] && echo "amd64" || echo "arm64")
+    curl -L -o "go${GO_VER}.linux-${GO_ARCH}.tar.gz" "https://go.dev/dl/go${GO_VER}.linux-${GO_ARCH}.tar.gz" &&
+        sudo rm -rf "/usr/local/go${GO_VER}" && \
+        sudo mkdir -p "/usr/local/go${GO_VER}" && \
+        sudo tar -C "/usr/local/go${GO_VER}" -xzf "go${GO_VER}.linux-${GO_ARCH}.tar.gz" --strip-components 1 && \
+        sudo rm -rfv /usr/local/bin/{go,gofmt} && \
+        sudo ln --symbolic /usr/local/go${GO_VER}/bin/{go,gofmt} /usr/local/bin/ && \
+        rm -rfv "go${GO_VER}.linux-${GO_ARCH}.tar.gz"
+
```
ggiguash, line 72: This is a good catch! Strictly speaking, we do not need these utilities unless we build the binaries.
@jogeo - can you verify that this change does not break the QE pipeline?

ggiguash: @pmtk , let's wait for QE confirmation before we merge this PR.
It's LGTM for me

validate-microshift/kuttl-test.sh
```diff
@@ -5,14 +5,30 @@ set -x
 
 ROOT="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/../")"
 
-KUTTL_VERSION="0.10.0"
+KUTTL_VERSION="0.15.0"
 KUTTL="$ROOT/bin/kuttl"
```
dhellmann, line 9: While we're updating this script, could we change this to put the file in the build output directory that is already ignored by git? That makes it nicer when running the script on a dev system.

```suggestion
KUTTL="$ROOT/_output/bin/kuttl"
```

scripts/devenv-builder/configure-vm.sh
ggiguash: @pmtk , as I suspected, this change breaks QE flow. Can we revert it to merge the other fixes for now?

PR #1487, opened 2023-03-10
USHIFT-949: cache getAllHostnames

dhellmann:
getAllHostnames is meant to run once on startup to help populate
configuration. In the test suite it may run many times, and it
occasionally blocks looking up the FQDN for hosts. This change caches
the first set of results to avoid blocking on future calls and to make
the tests run more quickly. This change is safe because MicroShift
should be restarted (automatically or by the user) if the network
settings of the host change in a significant way.

### DIFF
--- a/etcd/vendor/github.com/openshift/microshift/pkg/config/config.go
+++ b/etcd/vendor/github.com/openshift/microshift/pkg/config/config.go
@@ -17,6 +17,7 @@ import (
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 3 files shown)


ggiguash:
@dhellmann , this PR also contains two other changes that have already been merged. Is this OK, or we should rebase the commit against the latest main branch?

dhellmann:
> @dhellmann , this PR also contains two other changes that have already been merged. Is this OK, or we should rebase the commit against the latest main branch?

If you mean the changes in `etcd/vendor/`, those are the result of updating the config module in the etcd sub-module. We made those changes in the main config code, but did not re-vendor them into etcd.

PR #1488, opened 2023-03-11
USHIFT-335: fix mDNS server tests

dhellmann:
Limit the tests to IPv4 lookups only. Restructure the code so each address
lookup is its own subtest. Simplify the actual test using the assert
package to verify that the expected address appears in the list of
responses.


dhellmann:
I've updated the ticket for this to suggest creating an e2e test instead of trying to make the unit tests work. There are too many system dependencies for this to be a reliable unit test.

/close

PR #1489, opened 2023-03-11
OCPBUGS-9996: make lvms configuration more robust

dhellmann:
Instead of requiring the user to configure their system with a "rhel"
volume group, look at the volume groups that exist and pick one as the default. If there is only 1, use it. If there is a "microshift" VG, use it. Otherwise, disable the CSI driver.

/assign @copejon

### DIFF
--- a/docs/default_csi_plugin.md
+++ b/docs/default_csi_plugin.md
@@ -40,11 +40,17 @@ is run as.
 
 ## System Requirements
 
-### Volume Group Name
+### Default Volume Group
 
-The default integration of LVMS assumes a volume-group named `rhel`. LVMS's node-controller expects that volume
-group to exist prior to launching the service. If the volume group does not exist, the node-controller will fail to
-start and enter a CrashLoopBackoff state.
+If there is only one volume group on the system, LVMS uses it by
+default. If there are multiple volume groups, and no configuration
+file, LVMS looks for a volume group named `rhel`. If there is no
+volume group named `rhel`, LVMS is disabled.
+
+LVMS expects all volume groups to exist prior to launching the
+service. If LVMS is configured to use a volume group that does not
+exist, the node-controller Pod will fail and enter a CrashLoopBackoff
+state.
 
 ### Volume Size Increments
 
--- a/docs/default_csi_plugin.md
+++ b/docs/default_csi_plugin.md
@@ -3,7 +3,7 @@
 > **IMPORTANT!** The default LVMS configuration is intended to match the developer environment described in [MicroShift Development Environment](./devenv_setup.md). See section **[Configuring LVMS](#Configuring-LVMS)** for guidance on configuring LVMS for your environment.
 
 MicroShift enables dynamic storage provisioning out of the box with the LVMS CSI plugin. This plugin is a downstream
-Red Hat fork of TopoLVM. This provisioner will create a new LVM logical volume in the `rhel` volume group for each
+Red Hat build of TopoLVM. This provisioner will create a new LVM logical volume in the `rhel` volume group for each
 PersistenVolumeClaim(PVC), and make these volumes available to pods. For more information on LVMS, visit the repo's
 [README](https://github.com/red-hat-storage/topolvm).
 
--- a/docs/default_csi_plugin.md
+++ b/docs/default_csi_plugin.md
@@ -32,7 +32,7 @@ Full documentation of the config spec can be found at [github.com/red-hat-storag
 #### Path
 
 The user provided lvmd config should be written to the same directory as the MicroShift config.  If an lvmd configuration file
-does not exist in `/etc/microshift/lvmd.yaml`, MicroShift will use default values. 
+does not exist in `/etc/microshift/lvmd.yaml`, MicroShift will use default values.
 
 ## System Requirements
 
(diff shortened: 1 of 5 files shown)


copejon:
I'm pretty hesitant to make educated guesses about users' systems.  For one, it means user's can't expect uniform default behavior from the platform and will get different results based on their storage stack.  It'd be safer instead to select the VG that backs the user's homedir.  This would mesh well with current dev env we suggest with a microshift user, be predictable, and allow users to select special VGs if they want.  Making the 'rhel' name special also makes the storage configuration just a little bit more esoteric - one more reason to assume home VG if none are provided.

ggiguash:
Maybe, we should not attempt selecting any volumes if 'rhel' does not exist and nothing is specified in the config? 
Disabling CSI in case of an invalid config is great.

dhellmann:
> I'm pretty hesitant to make educated guesses about users' systems. For one, it means user's can't expect uniform default behavior from the platform and will get different results based on their storage stack. It'd be safer instead to select the VG that backs the user's homedir. This would mesh well with current dev env we suggest with a microshift user, be predictable, and allow users to select special VGs if they want. Making the 'rhel' name special also makes the storage configuration just a little bit more esoteric - one more reason to assume home VG if none are provided.

The point about predictability is fair. There is no user with a home directory, so I'm not sure how to use that to be more deterministic. Just look for `/home`? Maybe the volume group where `/` is would be a better rule?

How many hosts where someone is likely to run MicroShift actually have multiple volume groups?

> Maybe, we should not attempt selecting any volumes if 'rhel' does not exist and nothing is specified in the config?
> Disabling CSI in case of a non-existent config is great.

Disabling if there is no 'rhel' VG and the user has not provided a config file would be more predictable. It's not a minimal configuration for a nice out-of-the-box experience, though, so I was hoping to make things "just work."

dhellmann:
> My vote would be to make it simple and robust rather than "clever". There's no end to users' imagination, so we may inadvertedly end up in situations where the volume selection logic does not behave well.

It's not selecting a volume. It's selecting a _volume group_. The vast majority of systems running MicroShift are only going to have 1 of those.

dhellmann:
@ggiguash @copejon , I've changed the logic to prefer "rhel" but choose the single existing VG if it is named otherwise. If there are multiples, then it will require the user to configure the driver explicitly.

copejon:
> @ggiguash @copejon , I've changed the logic to prefer "rhel" but choose the single existing VG if it is named otherwise. If there are multiples, then it will require the user to configure the driver explicitly.

+1 I think that's a pretty reasonable fallback.

ggiguash:
@dhellmann, do we need to update GitHub docs with this change? I think we need to at least mention this new logic in the config doc, maybe a CSI design as well.

dhellmann:
> @dhellmann, do we need to update GitHub docs with this change? I think we need to at least mention this new logic in the config doc, maybe a CSI design as well.

Good point. See cb4e618d0. While I was working on that, I made a few other changes in that file in other commits.

dhellmann:
CI system timed out accessing images

/test e2e-openshift-conformance-reduced

docs/default_csi_plugin.md
```diff
@@ -31,20 +31,22 @@ Full documentation of the config spec can be found at [github.com/red-hat-storag
...
+file, LVMS looks for a volume group named `rhel`. If there is no
+volume group named `rhel`, LVMS is disabled.
 
-The default integration of LVMS assumes a volume-group named `rhel`. LVMS's node-controller expects that volume
-group to exist prior to launching the service. If the volume group does not exist, the node-controller will fail to
-start and enter a CrashLoopBackoff state.
+LVMS expects all volume groups to exist prior to launching the
+service. If LVMS is configured to use a volume group that does not
+exist, the node-controller Pod will fail and enter a CrashLoopBackoff
+state.
```
copejon, line 49:
```suggestion
state.


> Note: The VolumeGroup selected by MicroShift may be completely allocated to system and user LogicalVolumes. Be sure the VolumeGroup has sufficient unallocated capacity for workloads and extend the VolumeGroup as necessary.
```

copejon: Not sure if it's strictly necessary to remind users of this, but it would hopefully head off the inevitable "no space available" bug report.

pkg/config/lvmd/lvmd.go
```diff
@@ -16,21 +19,87 @@ const (
...
+func (l *Lvmd) IsEnabled() bool {
+	return len(l.DeviceClasses) > 0
+}
+
+// DefaultLvmdConfig returns a configuration struct for Lvmd with
+// default settings based on the current host. If the default volume
+// group name "rhel" is found, that volume group is used, otherwise
+// the largest volume group is used. If there are no volume groups,
+// the configuration returned will report that it is not enabled (see
+// IsEnabled()).
```
ggiguash, line 38: A nit: the comment needs to be fixed. The code no longer picks the largest volume group.

dhellmann: I fixed this and added some unit tests in the latest draft.

ggiguash:
@dhellmann , the `kutil`test is failing with the following errors in the CI:
```
   case.go:364: failed in step 30-
    case.go:366: deployments.apps "topolvm-controller" not found
    case.go:366: daemonsets.apps "topolvm-node" not found
```

Looks like the CI is falling back to the case with CSI disabled.

ggiguash:
@dhellmann, last error remaining:
```
# github.com/openshift/microshift/pkg/config/lvmd [github.com/openshift/microshift/pkg/config/lvmd.test]
pkg/config/lvmd/lvmd_test.go:22:5: unknown field 'DisabledReason' in struct literal of type Lvmd
pkg/config/lvmd/lvmd_test.go:90:5: unknown field 'DisabledReason' in struct literal of type Lvmd
```

dhellmann:
> @dhellmann, last error remaining:
> 
> ```
> # github.com/openshift/microshift/pkg/config/lvmd [github.com/openshift/microshift/pkg/config/lvmd.test]
> pkg/config/lvmd/lvmd_test.go:22:5: unknown field 'DisabledReason' in struct literal of type Lvmd
> pkg/config/lvmd/lvmd_test.go:90:5: unknown field 'DisabledReason' in struct literal of type Lvmd
> ```

That's what I get for not running the unit tests locally before pushing. :-/

PR #1491, opened 2023-03-13
Add flag to turn off built-in loadbalancer service controller

pliurh:
If a user wants to use an external loadbalancer service implementation like MetalLB, the built-in loadbalancer controller needs to be turned off to avoid interference.

The new flag 'disableLoadBalancerController' is false by default.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -103,8 +103,9 @@ type MicroshiftConfig struct {
 	BaseDomain       string        `json:"baseDomain"`
 	Cluster          ClusterConfig `json:"cluster"`
 
-	Ingress IngressConfig      `json:"-"`
-	Etcd    InternalEtcdConfig `json:"etcd"`
+	Ingress                       IngressConfig      `json:"-"`
+	Etcd                          InternalEtcdConfig `json:"etcd"`
+	DisableLoadBalancerController bool               `json:"disableLoadBalancerController"`
 }
 
 // Top level config file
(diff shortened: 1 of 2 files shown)


ggiguash:
@pliurh, if we're implementing this, the following also needs to be updated:
- https://github.com/openshift/microshift/blob/main/docs/howto_config.md
- https://github.com/openshift/microshift/blob/main/packaging/microshift/config.yaml

If you prefer, we can do this in the #1432 PR after the current one is merged. This should help us avoid unnecessary rebases.

pliurh:
> If you prefer, we can do this in the #1432 PR after the current one is merged. This should help us avoid unnecessary rebases.

Yeah, I will update the doc in  #1432

pkg/config/config.go
```diff
@@ -103,8 +103,9 @@ type MicroshiftConfig struct {
 	BaseDomain       string        `json:"baseDomain"`
 	Cluster          ClusterConfig `json:"cluster"`
 
-	Ingress IngressConfig      `json:"-"`
-	Etcd    InternalEtcdConfig `json:"etcd"`
+	Ingress                       IngressConfig      `json:"-"`
+	Etcd                          InternalEtcdConfig `json:"etcd"`
+	DisableLoadBalancerController bool               `json:"disableLoadBalancerController"`
```
dhellmann, line 108: We will want this to be a string field so we can use an enum.

dhellmann (commented):
I agree that we want the new flag. I would prefer to avoid changing the config data structure until #1442 has merged, so I'm placing a temporary hold. We don't need to backport this change to 4.13, so there's no rush, and we can talk about where in the config the flag should be.

/hold

pkg/config/config.go
pliurh: May I ask why we prefer an enum instead of a boolean here?

dhellmann: Of coruse, I should have given more detail.

Our general API design guidelines have good background: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#do-not-use-boolean-fields

The short version is that it's difficult to expand the meaning of the field in the future (Imagine what we would do if we start embedding MetalLB in MicroShift, for example.), it is difficult to change the default, and in our case it is difficult to tell if a user has provided a value or not because the zero value is a valid value. We could use a pointer field for that, but an enum is easier to work with.

copejon:
This is not a trivial API change and is going to set a precedent for how we eventually support disabling other platform components.  We may want to discuss this in more detail with the broader group before considering merging.

pliurh:
> This is not a trivial API change and is going to set a precedent for how we eventually support disabling other platform components. We may want to discuss this in more detail with the broader group before considering merging.

 @dhellmann @copejon I've created an EPIC https://issues.redhat.com/browse/USHIFT-963 to track this effort.

PR #1496, opened 2023-03-13
USHIFT-955: MicroShift Cockpit plugin PoC

ggiguash:
Use the https://github.com/cockpit-project/starter-kit repo as a starter and put it under microshift/cockpit-plugin directory

The PR contains two commits:
- Initial version of starter-kit (without its .git)
- An implementation of YamlFile component and Makefile adjustments

It is recommended to look at the second commit only.

Closes [USHIFT-955](https://issues.redhat.com//browse/USHIFT-955)


cockpit-plugin/src/YamlFile.jsx
```diff
@@ -0,0 +1,124 @@
...
+        promise.then((content, tag) => {
+            if (!content) {
+                fileName += ".default";
+                const promise = cockpit.file(fileName).read();
+                promise.then((content, tag) => {
+                    if (content) {
+                        this.updateState(content);
+                        this.validateContent();
+                    } else {
+                        this.updateState("", false, true, "Failed to open " + this.props.fileName);
```
pmtk, line 33: This notifies about failing to read `${filename}.default`, I think include of suffix would be more accurate and helpful in case of error

PR #1506, opened 2023-03-15
ETCD-403: Revert user configuration for etcd and change default max quota

dusk125:
To simplify the configuration, this PR reverts the changes made by #1435. Also, this sets the maximum on-disk quota for etcd to be the maximum allowed (and aligned with Openshift) as the default to reduce the likelihood that a customer will hit the quota during runtime.

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -21,11 +21,7 @@ apiServer:
 debugging:
   logLevel: ""
 etcd:
-  quotaBackendSize: ""
-  defragCheckFreq: ""
-  doStartupDefrag: false
-  minDefragSize: ""
-  maxFragmentedPercentage: 0
+  memoryLimitMB: 0
 ```
 
 ## Default Settings
@@ -49,11 +45,7 @@ apiServer:
 debugging:
   logLevel: "Normal"
 etcd:
-  quotaBackendSize: "2Gi"
-  defragCheckFreq: "5m"
-  doStartupDefrag: true
-  minDefragSize: "100Mi"
-  maxFragmentedPercentage: 45
+  memoryLimitMB: 0
 ```
 
 ## Service NodePort range
@@ -89,6 +81,14 @@ List of ports that you must avoid:
 | 10248/tcp     | kubelet healthz port
 | 10259/tcp     | kube scheduler
 |---------------|-----------------------------------------------------------------|
+
+## Etcd Memory Limit
+
+By default, etcd will be allowed to use as much memory as it needs to handle the load on the system; however, in memory constrained systems, it may be preferred or necessary to limit the amount of memory etcd is allowed to use at a given time.
+
+Setting the `memoryLimitMB` to a value greater than 0 will result in a soft memory limit being applied to etcd; etcd will be allowed to go over this value during operation, but memory will be more aggresively reclaimed from it if it does. A value of `128` megabytes is the recommended starting place for this limit; however, the configuration floor is `50` megabytes - attempting to set the limit below 50 megabytes will result in the configuration being 50 megabytes.
+
+Please note that values between 50 and 128 megabytes will heavily trade off memory footprint for etcd performance: the lower the memory limit, the more time etcd will spend on paging memory to disk and will take longer to respond to queries or even timing requests out if the limit is low and the etcd usage is high.
 
 # Auto-applying Manifests
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,7 +17,6 @@ import (
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/pflag"
 
-	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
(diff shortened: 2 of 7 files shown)


dhellmann:
We are pretty confident that we do want this change, so I'll approve the PR. We should wait for the measurements before merging it, as Allen says. 

/approve

dusk125:
So it looks like etcd has a memory overhead of 30-60% in addition to attempting to hold the database file(s) in memory. Doug and I agreed that keeping the quota at 8GB and allowing the user to opt-in to a memory limit was a good way forward. The most recent version of this PR introduces a memory limit for etcd that a user can add if they want; by default, there's no limit.

I ran a synthetic 'put' test against a fresh etcd server with varying memory limits and found that 50MB is a good minimum before etcd performance degrades significantly. Therefore, Microshift config will fail to parse if a user attempts to set a value lower than 50MB.

pkg/config/config.go
```diff
@@ -102,18 +91,27 @@ type MicroshiftConfig struct {
...
+	Network   Network   `json:"network"`
+	Node      Node      `json:"node"`
+	ApiServer ApiServer `json:"apiServer"`
+	Debugging Debugging `json:"debugging"`
+	Etcd      Etcd      `json:"etcd"`
+}
+
+const (
+	// Etcd performance degrades significantly if the memory available is less than 50MB, enfore this minimum.
+	EtcdMinimumMemoryLimit = "50M"
```
dhellmann, line 110: In the team meeting today we discussed making this 128M to start. 

```suggestion
	EtcdMinimumMemoryLimit = "128M"
```

pkg/config/config.go
```diff
@@ -422,35 +421,21 @@ func (c *MicroshiftConfig) ReadFromConfigFile(configFile string) error {
...
+
+		limit, err := resource.ParseQuantity(config.Etcd.MemoryLimit)
 		if err != nil {
-			return fmt.Errorf("failed to parse etcd quotaBackendSize: %v", err)
+			return fmt.Errorf("failed to parse etcd memory limit: %v", err)
 		}
-		if !q.IsZero() {
-			c.Etcd.QuotaBackendBytes = q.Value()
+		if limit.Value() < min.Value() {
+			return errors.New("attempted to set etcd memory limit beneath the minimum of 50MB, please select a higher value.")
```
dhellmann, line 435: This error message should use the default value from the variable instead of repeating it.

dhellmann (commented):
Could you update docs/howto_config.md as part of this PR, too?

pkg/config/config.go
dusk125: This variable is the minimum allowed value, not the default. If the user tries to configure etcd to limit it to less than 50MB, Microshift will error.

dhellmann: Got it.

dhellmann:
I had one nit, and there's a hold. We can address the nit separately, so if we're ready to move ahead otherwise please remove the hold.

pkg/controllers/etcd.go
```diff
@@ -59,24 +63,34 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
...
 			"--uid=root",
 			"--scope",
-			"--property", "BindsTo=microshift.service",
 			"--unit", "microshift-etcd",
+			"--property", "BindsTo=microshift.service",
+		)
+
+		if s.memoryLimit != "" {
+			args = append(args,
+				"--property", fmt.Sprintf("MemoryHigh=%v", s.memoryLimit),
```
dhellmann, line 82: We're going to need some additional protection for this parameter. I tried `128Mi` and `systemd-run` couldn't understand it.

Perhaps the field should just be an integer for the number of megabytes to allow, instead of the units value?

pkg/config/config.go
dusk125: So you mean to say that if limit is less than min, use the min value silently?

pkg/config/config.go
```diff
@@ -102,18 +90,28 @@ type MicroshiftConfig struct {
...
+}
+
+const (
+	// Etcd performance degrades significantly if the memory available is less than 50MB, enfore this minimum.
+	EtcdMinimumMemoryLimit = 50.0
+)
+
+type Etcd struct {
+	// Set a memory limit, in megabytes, on the etcd process; etcd will begin paging memory when it gets to this value. 0 means no limit.
+	MemoryLimit uint64 `json:"memoryLimit"`
```
dhellmann, line 114: We have units in some of the other names, even though they aren't exposed. What do you think about including units in this exposed name?

```suggestion
	MemoryLimitMB uint64 `json:"memoryLimitMB"`
```

docs/howto_config.md
```diff
@@ -21,11 +21,7 @@ apiServer:
 debugging:
   logLevel: ""
 etcd:
-  quotaBackendSize: ""
-  defragCheckFreq: ""
-  doStartupDefrag: false
-  minDefragSize: ""
-  maxFragmentedPercentage: 0
+  memoryLimit: 0
```
dhellmann, line 24:
```suggestion
  memoryLimitMB: 0
```

docs/howto_config.md
```diff
@@ -49,11 +45,7 @@ apiServer:
 debugging:
   logLevel: "Normal"
 etcd:
-  quotaBackendSize: "2Gi"
-  defragCheckFreq: "5m"
-  doStartupDefrag: true
-  minDefragSize: "100Mi"
-  maxFragmentedPercentage: 45
+  memoryLimit: 128
```
dhellmann, line 48:
```suggestion
  memoryLimitMB: 128
```

dhellmann (commented):
2 doc nits and then I think this is ready to merge

dhellmann:
This version looks good. Let's see if these are just test flakes.

dusk125:
I was able to see that etcd is running in the sos report in the CI failures, so I think these are just flakes as well.

dhellmann:
I think this strikes the right balance. Let's see what CI has to say.

/lgtm

PR #1509, opened 2023-03-15
PERFSCALE-2108: Enable logging for Microshift built-in kubelet metrics

sjug:
- Make external cadvisor process conditional
- Create role to create metrics service account
- Add firewall port for kubelet endpoint
- Create role to modify prometheus config
- Fix file location in microshift-start
- Append kubelet endpoint to prometheus config in playbook


dhellmann (commented):
I take it from the setup that Prometheus is running outside of MicroShift?

/approve

sjug:
> I take it from the setup that Prometheus is running outside of MicroShift?

The ansible logging tasks will set up Prometheus & Grafana on a specified host. I generally run the logging setup on a second host so I don't add resource consumption to the MicroShift node.

PR #1513, opened 2023-03-16
NOISSUE: Add jira auto clone tool to scripts

pacevedom:
Python tool to automate bug handling:
- Checks Fix versions and mismatches for original and cloned bugs.
- Checks Target version mismatches with Fix versions.
- Checks gaps in Target version against Fix versions and creates clones for those that are not present.
- Automatically creates blocks links between clones and originals.
- Automatically synchronize clones with originals for Fix versions.
- Alert about manual actions required, such as setting Target version/Fix versions.

Closes #<Issue Number>

### DIFF
--- a/scripts/jira/cloner.py
+++ b/scripts/jira/cloner.py
@@ -1,0 +1,216 @@
+import jira
+import os
+from tabulate import tabulate
+import argparse
+from tqdm import tqdm
+
+JQL_GLOBAL_QUERY = 'filter = "MicroShift - Bugs in Project" and resolution = Unresolved order by key asc'
+JQL_ISSUE_QUERY = 'filter = "MicroShift - Bugs in Project" and resolution = Unresolved and key in ({}) order by key asc'
+JIRA_SERVER = 'https://issues.redhat.com'
+JIRA_URL_PREFIX = JIRA_SERVER+'/browse/'
+
+def is_original_issue(issue):
+    for link in issue.fields.issuelinks:
+        if link.type.name != 'Cloners':
+            continue
+        if hasattr(link, 'outwardIssue'):
+            return False
+    return True
+
+def get_issue_fix_versions(issue):
+    l = []
+    if hasattr(issue.fields, 'fixVersions') and issue.fields.fixVersions is not None:
+        for x in issue.fields.fixVersions:
+            if hasattr(x, 'name'):
+                l.append(x.name)
+    if len(l) == 0:
+        raise RuntimeError(f"[{issue.key}] Fix versions empty")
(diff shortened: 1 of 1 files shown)


dhellmann:
So this is a tool 1 person can use to manage all of the bug tickets? Rather than something an individual needs to run with 1 bug ticket they are working on?

pacevedom:
> So this is a tool 1 person can use to manage all of the bug tickets? Rather than something an individual needs to run with 1 bug ticket they are working on?

Added some more stuff now if compared to the original commit.
The idea for the tool is to have something that:
- Watches and fixes any inconsistencies for bugs.
- Is able to run without supervision (mostly).
- Is able to run in a job.

You can have it fix/setup your own bugs with -i option, or have it run widespread on the whole MicroShift bugs filter.
Since it needs a JIRA token, it can't do anything you can't do with your user.

scripts/jira/cloner.py
```diff
@@ -0,0 +1,216 @@
...
+    if not is_original_issue(issue):
+        return actions
+
+    if target_version != fix_versions[0]:
+        actions.append([JIRA_URL_PREFIX+issue.key, "", f"Target version does not target first Fix version. {target_version} != {fix_versions[0]}. Please check."])
+        return actions
+
+    fix_versions_missing = set(fix_versions[1:])
+    clones = get_clone_by_issues(issue, connection)
+    for clone in clones:
```
dhellmann, line 112: The body of this loop has a lot of conditionals and exception handling. We can land this version, but you might be able to simplify this by creating a "command" abstraction so that this function can be split between building a list of commands to run and another function that actually runs them (or not, if dry-run is set).

scripts/jira/cloner.py
```diff
@@ -0,0 +1,216 @@
...
+            else:
+                try:
+                    set_fix_versions(clone, fix_versions, connection)
+                except RuntimeError as e:
+                    actions.append([JIRA_URL_PREFIX+clone.key, JIRA_URL_PREFIX+issue.key, f"Unable to update Fix versions: {e}"])
+                    continue
+                finally:
+                    actions.append([JIRA_URL_PREFIX+clone.key, JIRA_URL_PREFIX+issue.key, "Fix versions updated to match original"])
+
+        if clone_fix_versions != fix_versions:
```
dhellmann, line 128: Does the order of the versions matter here? I don't know if Jira sorts the values or if we might fail to recognize "4.13.0, 4.12.z" being the same as "4.12.z, 4.13.0", for example.

dhellmann (commented):
/lgtm
/retest-required

PR #1515, opened 2023-03-16
USHIFT-978 Building RPMs in container

pmtk:


hack/Containerfile.rpm-builder
```diff
@@ -0,0 +1,12 @@
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.13
+
+RUN rm -rfv /etc/yum.repos.d/ci-rpm-mirrors.repo /etc/yum.repos.d/localdev* && \
+    dnf install \
+        --setopt=tsflags=nodocs \
+        --setopt=install_weak_deps=False \
+        --repo=rhel-8-for-x86_64-baseos-rpms \
```
ggiguash, line 7: We're supporting RHEL 9 now in the main branch. Can we use RHEL9 containers / dependencies?

Makefile
```diff
@@ -214,6 +214,14 @@ image-build-iso: rpm
 iso: image-build-configure image-build-iso
 .PHONY: iso
 
+rpm-builder:
+	podman build -v /etc/pki/entitlement/:/etc/pki/entitlement -t rpm-builder - < ./hack/Containerfile.rpm-builder
+.PHONY: rpm-builder
+
+rpm-in-container:
```
ggiguash, line 221: Let's simplify the process by having one top-level target? 

We already have 'make rpm' & 'make srpm' - can we add 'make rpm-podman'?

ggiguash:
A general question: why would we want to build in a container if we have a devenv? Are you thinking to use this for building 4.12 s/w on a RHEL9 machine for example?

hack/Containerfile.rpm-builder
pmtk: This is because we use `registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.13`.
I see that `rhel-9-release-golang-1.19-openshift-4.13` exists but it fails due to some certificates for `rhel-9-for-x86_64-baseos-rpms` repo.
Also, we're still using `rhel-8-release-golang-1.19-openshift-4.13` in the CI jobs

pmtk:
> A general question: why would we want to build in a container if we have a devenv? Are you thinking to use this for building 4.12 branch on a RHEL9 machine for example?

We'll be able to use the same image CI uses, i.e. rhel8 (or 9) + go1.19 (which isn't available in the repos yet) and drop the workaround with installing two versions of Go, therefore fixing the microshift.spec to require right Go version (now we "require" 1.18, but it'll fail if 1.19 is not used).
That go1.1x workaround happened twice already so someday it'll happen again, so this would be one less thing to keep in mind.

I feel like QE's pipeline would benefit from being able to build in a container. From what I've seen recently, they only run `configure-vm.sh` to get that go1.19 side-installed, everything else is in the kickstart.
If we could get `make iso` containerized QE could drop dev-vm altogether most likely speeding up the procedure.
Same thing locally on laptop, build ISO without needing build vm (most likely with limited resources compared to the laptop).

ggiguash:
Do we need a pull secret for pulling the Go container image?

ggiguash:
Should we also add "go" invocation from the container and get rid of the "go" dependency dillema in the devenv all together? I mean, we could remove "go" installation from configure-vm.sh and have container builds executed by default for compilation and RPM creation.

pmtk:
> Should we also add "go" invocation from the container and get rid of the "go" dependency dillema in the devenv all together? I mean, we could remove "go" installation from configure-vm.sh and have container builds executed by default for compilation and RPM creation.

Maybe, we need to think about this.
I'm partial to leaving it as is because I build bins on laptop and copy on VM - I have Go on my laptop either way.
This would also be appreciated by devs with MacOS as Docker on Mac is actually docker on linux vm on Mac

Makefile
```diff
@@ -214,6 +214,20 @@ image-build-iso: rpm
 iso: image-build-configure image-build-iso
 .PHONY: iso
 
+rpm-podman:
+	@if [ -z "$(shell podman images -q microshift-rpm-builder)" ]; then \
```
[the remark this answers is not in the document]
ggiguash: +1 on letting podman figure out if there's a need to (re)build. It comes with a bonus of automatic rebuild when something changes in the Containerfile.

ggiguash:
@pmtk , let's pull the Containerfile out of the hack directory. 
We have a Dockerfile under packaging/images/openshift-ci directory. Since we're using a CI image, should we put the new file somewhere nearby?

Makefile
```diff
@@ -214,6 +214,21 @@ image-build-iso: rpm
...
+		podman build \
+			-v /etc/pki/entitlement/:/etc/pki/entitlement \
+			-t microshift-rpm-builder - < ./hack/Containerfile.rpm-builder ; \
```
ggiguash, line 221: Do we want to include a version in the image name, i.e. `microshift-4.13-rpm-builder-rhel8-go1.19`? I mean, we should be able to easily cross-compile from different branches on the same machine now, but images may be different.

Makefile
```diff
@@ -214,6 +214,25 @@ image-build-iso: rpm
 iso: image-build-configure image-build-iso
 .PHONY: iso
 
+RPM_BUILDER_IMAGE_TAG=$(shell \
+  gawk 'BEGIN { FS=":" } \
+	/FROM/ { \
+		match($$0, /rhel-([0-9]+).*golang-([0-9]\.[0-9]{2})-openshift-([0-9]+\.[0-9]+)/,m);\
+		printf "%s-rhel%s-go%s",m[3],m[1],m[2]; }' packaging/images/Containerfile.rpm-builder)
```
ggiguash, line 221: Since we're building the container from make, can we simplify this by defining a variable in the Makefile and passiing it to the Containerfile?

```
Makefile:
GOLANG_CONTAINER=release:rhel-8-release-golang-1.19-openshift-4.13

podman build -v GOLANG_CONTAINER=$GOLANG_CONTAINER ...

Containerfile:
FROM registry.ci.openshift.org/openshift/$GOLANG_CONTAINER
```

pmtk: Following your suggestion, does `microshift-builder:rhel-8-release-golang-1.19-openshift-4.13` sound good as an image `name:tag`?

ggiguash: I think so, because the CI image name/version is the main identifier for it.
Unless, of course, there's a limitation in docker not allowing to use that name as-is.

PR #1519, opened 2023-03-17
WIP DO NOT MERGE test less compressed RPMs

pmtk:
/hold
/cc

### DIFF
--- a/packaging/rpm/make-rpm.sh
+++ b/packaging/rpm/make-rpm.sh
@@ -62,7 +62,7 @@ EOF
   # it yields w19.zstdio, which means compression level 19 using zstd algorithm. To
   # speed this up it runs w19T8.zstdio, which is the 8 thread version of the same
   # algorithm.
-  rpmbuild --quiet ${RPMBUILD_OPT} --define "_topdir ${RPMBUILD_DIR}" --define "_binary_payload w19T8.zstdio" "${RPMBUILD_DIR}"SPECS/microshift.spec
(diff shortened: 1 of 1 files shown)


pmtk:
build log without the change:

```
2023-03-16T15:53:00.770027545Z [1/2] STEP 5/6: RUN make rpm
2023-03-16T15:58:15.842346006Z [1/2] STEP 6/6: RUN TARGET_ARCH=aarch64 make rpm
2023-03-16T16:03:39.962084758Z [2/2] STEP 1/16
```

packaging/rpm/make-rpm.sh
```diff
@@ -62,7 +62,7 @@ EOF
   # it yields w19.zstdio, which means compression level 19 using zstd algorithm. To
   # speed this up it runs w19T8.zstdio, which is the 8 thread version of the same
   # algorithm.
```
dhellmann, line 64: It looks like this comment will also need to be updated.

I wonder if we need any compression at all if this is just for CI?

pmtk:
Actually it looks like the difference is within an error, that minute saved might've been just lucky run on CI.
Not sure if worth changed the RHEL/Fedora building defaults
```
2023-03-17T14:44:48.108607559Z [1/2] STEP 5/6: RUN make rpm
2023-03-17T14:48:32.235223875Z [1/2] STEP 6/6: RUN TARGET_ARCH=aarch64 make rpm
2023-03-17T14:51:56.898985730Z [2/2] STEP 1/16:
```

pmtk:
From further investigation (connected to building RPMs in a container) it seems that bulk of the time is (obviously) spent on compiling go binaries. Given that CI is always starting with clean slate (empty go cache), there's not much to do

PR #1523, opened 2023-03-17
USHIFT-591: log when exiting due to certificate rotation

dhellmann:
When the certificates are about to expire MicroShift restarts to regenerate
them. This change introduces an extra goroutine to watch for that event
specifically and log a message explaining the restart. It then propagates
the shutdown by canceling the main context used by everything else to know
when to shutdown.


pacevedom (approved):
/lgtm

PR #1525, opened 2023-03-18
Override osbuild-composer to use rhel-92 nightly repositories.

jogeo:
Provides [rhel-92](https://issues.redhat.com//browse/rhel-92) rpm repo overrides for the composer image build and dev env VM.
**Which issue(s) this PR addresses**:[USHIFT-980](https://issues.redhat.com/browse/USHIFT-980

### DIFF
--- a/scripts/devenv-builder/configure-vm.sh
+++ b/scripts/devenv-builder/configure-vm.sh
@@ -47,6 +47,20 @@ if [ $RHEL_SUBSCRIPTION = true ] ; then
     fi
 fi
 
+# Temporarily use RHEL 9.2 nightly repositories
+sudo tee /etc/yum.repos.d/rhel-92.repo >/dev/null <<EOF
+[rhel-9-for-x86_64-appstream-nightly-rpms]
+name = Red Hat Enterprise Linux 9 for x86_64 - AppStream Nightly (RPMs)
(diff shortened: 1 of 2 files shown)


scripts/devenv-builder/configure-vm.sh
```diff
@@ -47,6 +47,20 @@ if [ $RHEL_SUBSCRIPTION = true ] ; then
     fi
 fi
 
+# Temporarily use RHEL 9.2 nightly repositories
+sudo tee /etc/yum.repos.d/rhel-92.repo >/dev/null <<EOF
+[rhel-9-for-x86_64-appstream-nightly-rpms]
+name = Red Hat Enterprise Linux 9 for x86_64 - AppStream Nightly (RPMs)
+baseurl = http://download.eng.tlv.redhat.com/rhel-9/nightly/RHEL-9/latest-RHEL-9.2.0/compose/AppStream/x86_64/os/
```
dhellmann, line 54: This is a region-specifc URL. Is there a location that redirects to something local to the user?

dhellmann (commented):
One nit that we can fix in another PR.

What does this do to a development environment? Is it safe for all hosts to be configured this way? Since we expect to encounter this issue again in the future, do we want a separate "mode" for the script so we can keep the logic and just update the URLs over time?

/lgtm

dhellmann:
/lgtm cancel

CI is pretty unhappy

ggiguash:
We need to discuss how to proceed with this fix
/hold

PR #1526, opened 2023-03-19
USHIFT-979: Sanitize the values of wait timeout and max boots from greenboot.conf

ggiguash:
In case of an invalid input, the following messages will be logged.
```
Could not parse MICROSHIFT_WAIT_TIMEOUT_SEC value 'Abc': using '300' instead
GREENBOOT_MAX_BOOTS value '35' is not in the [1..9] range: using '3' instead
```

Closes [USHIFT-979](https://issues.redhat.com//browse/USHIFT-979)

### DIFF
--- a/packaging/greenboot/functions.sh
+++ b/packaging/greenboot/functions.sh
@@ -15,7 +15,8 @@ OCGET_CMD="oc get ${OCCONFIG_OPT}"
 #
 # The base value for the timeout and the maximum boot attempts can be defined in
 # the /etc/greenboot/greenboot.conf file using the MICROSHIFT_WAIT_TIMEOUT_SEC
-# and GREENBOOT_MAX_BOOTS settings.
+# and GREENBOOT_MAX_BOOTS settings. These values can be in the [60..9999] range
+# for MICROSHIFT_WAIT_TIMEOUT_SEC and the [1..9] range for GREENBOOT_MAX_BOOTS.
 #
 # args: None
 # return: Print the recommended timeout value to stdout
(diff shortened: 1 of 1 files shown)


packaging/greenboot/functions.sh
```diff
@@ -15,19 +15,38 @@ OCGET_CMD="oc get ${OCCONFIG_OPT}"
...
+    # Read and verify the wait timeout value, allowing for the [60..9999] range
     local base_timeout=${MICROSHIFT_WAIT_TIMEOUT_SEC:-300}
+    local reSecs='^[1-9]{1}[0-9]{0,3}$'
+    if [[ ! ${base_timeout} =~ $reSecs ]] ; then
+        base_timeout=300
+        echo "MICROSHIFT_WAIT_TIMEOUT_SEC value '${MICROSHIFT_WAIT_TIMEOUT_SEC}' is invalid: using '${base_timeout}' instead"
+    fi
+    if [[ ${base_timeout} -lt 60 ]] ; then
+        base_timeout=60
+        echo "MICROSHIFT_WAIT_TIMEOUT_SEC value '${MICROSHIFT_WAIT_TIMEOUT_SEC}' is invalid: using '${base_timeout}' instead"
```
dhellmann, line 37: It would be more helpful to say "lower than the minimum value (60)" or something similar here, so an admin who sees the message understand what is wrong with the value.

packaging/greenboot/functions.sh
```diff
@@ -15,19 +15,38 @@ OCGET_CMD="oc get ${OCCONFIG_OPT}"
...
     # Source Greenboot configuration file if it exists
     local conf_file=/etc/greenboot/greenboot.conf
     [ -f "${conf_file}" ] && source ${conf_file}
+
+    # Read and verify the wait timeout value, allowing for the [60..9999] range
     local base_timeout=${MICROSHIFT_WAIT_TIMEOUT_SEC:-300}
+    local reSecs='^[1-9]{1}[0-9]{0,3}$'
+    if [[ ! ${base_timeout} =~ $reSecs ]] ; then
+        base_timeout=300
+        echo "MICROSHIFT_WAIT_TIMEOUT_SEC value '${MICROSHIFT_WAIT_TIMEOUT_SEC}' is invalid: using '${base_timeout}' instead"
```
dhellmann, line 33: It would be more helpful to say something like "could not parse" rather than just "invalid" here, so an admin who sees the message understands what is wrong.

packaging/greenboot/functions.sh
```diff
@@ -15,19 +15,38 @@ OCGET_CMD="oc get ${OCCONFIG_OPT}"
...
+        base_timeout=60
+        echo "MICROSHIFT_WAIT_TIMEOUT_SEC value '${MICROSHIFT_WAIT_TIMEOUT_SEC}' is invalid: using '${base_timeout}' instead"
+    fi
+
+    # Read and verify the max boots value, allowing for the [1..9] range
+    local max_boots=${GREENBOOT_MAX_BOOTS:-3}
+    local reBoots='^[1-9]{1}$'
+    if [[ ! ${max_boots} =~ $reBoots ]] ; then
+        max_boots=3
+        echo "GREENBOOT_MAX_BOOTS value '${GREENBOOT_MAX_BOOTS}' is invalid: using '${max_boots}' instead"
```
dhellmann, line 45: Same comment about minimum value.

dhellmann (commented):
/lgtm

PR #1548, opened 2023-03-22
NOISSUE: Add user filtering and sprint handling to jira script

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/scripts/jira/cloner.py
+++ b/scripts/jira/cloner.py
@@ -55,6 +55,16 @@ def get_parent_issue(issue, connection):
             return connection.issue(link.outwardIssue.key)
     return None
 
+def get_sprint(issue):
+    if not hasattr(issue.fields, 'customfield_12310940'):
+        return None
+    if issue.fields.customfield_12310940 is None or len(issue.fields.customfield_12310940) == 0:
+        return None
+    last_sprint = issue.fields.customfield_12310940[-1]
+    first_index = last_sprint.find('id=')+3
+    last_index = last_sprint.find(',', first_index)
+    return last_sprint[first_index:last_index]
+
 def is_issue_a_cve(issue):
     for label in issue.fields.labels:
         if label.startswith('CVE-'):
@@ -84,7 +94,7 @@ def remove_needs_fix_version_label(issue):
         if l == 'needs-fix-version':
             continue
         labels.append({'name': l})
-    issue.ipdate(fields={
+    issue.update(fields={
         'labels': labels
     })
 
@@ -116,6 +126,11 @@ def clone_issue(issue, target, connection):
     new_issue = connection.create_issue(data_dict)
     connection.create_issue_link("Cloners", inwardIssue=new_issue.key, outwardIssue=issue.key)
     connection.create_issue_link("Blocks", inwardIssue=issue.key, outwardIssue=new_issue.key)
+
+    sprint = get_sprint(issue)
+    if sprint is not None:
+        connection.add_issues_to_sprint(sprint, [new_issue])
+
     return new_issue
 
 def add_blocks_link(issue, parent, connection):
@@ -203,6 +218,9 @@ def scan_issue(issue, connection):
         if has_fix_versions_label(issue):
             actions.append(Action(issue.key, "Remove Fix versions label", remove_needs_fix_version_label, issue=issue))
 
+    if get_sprint(issue) is None:
+        actions.append(Action(issue.key, "Issue needs to be included in a sprint", None))
+
     if is_original_issue(issue):
         actions.extend(scan_original_issue(issue, connection))
     else:
@@ -223,7 +241,7 @@ if __name__ == '__main__':
     connection = jira.JIRA(
         server=JIRA_SERVER,
         token_auth=args.token)
-    
+
     jql_query = JQL_GLOBAL_QUERY
     if len(args.issue) > 0:
         jql_query = JQL_ISSUE_QUERY.format(args.issue)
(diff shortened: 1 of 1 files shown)


scripts/jira/cloner.py
```diff
@@ -203,30 +220,43 @@ def scan_issue(issue, connection):
...
+
     if is_original_issue(issue):
         actions.extend(scan_original_issue(issue, connection))
     else:
         actions.extend(scan_cloned_issue(issue, connection))
     return actions
 
+def query_build(issue, user):
+    query = JQL_FILTER_QUERY
+    if len(issue) > 0:
```
dhellmann, line 234: issue is a string? So this could just be `if issue`?

scripts/jira/cloner.py
```diff
@@ -203,30 +220,43 @@ def scan_issue(issue, connection):
...
+        query += f' and {JQL_FILTER_ISSUE.format(issue)}'
+    if len(user) > 0:
```
dhellmann, line 236: Same?

scripts/jira/cloner.py
```diff
@@ -203,30 +220,43 @@ def scan_issue(issue, connection):
         if has_fix_versions_label(issue):
             actions.append(Action(issue.key, "Remove Fix versions label", remove_needs_fix_version_label, issue=issue))
 
+    if get_sprint(issue) is None:
+        actions.append(Action(issue.key, "Issue needs to be included in a sprint", None))
```
dhellmann, line 224: What if we don't intend to backport a change in the current sprint?

ggiguash: If the original issue is in the sprint, the new one better be put in the same sprint. Otherwise, it is forgotten.
We can always manually reshuffle sprints when the issue is on the Agile Board.

scripts/jira/cloner.py
pacevedom: You are right, done.

scripts/jira/cloner.py
pacevedom: Done.

scripts/jira/cloner.py
pacevedom: +1 on what Giga said. Plus, you get the alert from the script about an issue without a sprint, and it will clone bugs using the same sprint than the parent. If needed, you can change it manually afterwards.

pacevedom:
/hold
Found some bugs, waiting for fixes.

scripts/jira/cloner.py
```diff
@@ -55,6 +57,16 @@ def get_parent_issue(issue, connection):
             return connection.issue(link.outwardIssue.key)
     return None
 
+def get_sprint(issue):
+    if not hasattr(issue.fields, 'customfield_12310940'):
+        return None
+    if issue.fields.customfield_12310940 is None or len(issue.fields.customfield_12310940) == 0:
+        return None
+    last_sprint = issue.fields.customfield_12310940[-1]
```
dhellmann, line 65:
```suggestion
    sprints = getattr(issue.fields, SPRINT_FIELD_ID, [])
    if not sprints:
        return None
    last_sprint = sprints[-1]
```

scripts/jira/cloner.py
```diff
@@ -75,7 +87,7 @@ def set_target_version(issue, version):
 def set_needs_fix_version_label(issue):
     labels = issue.fields.labels + ['needs-fix-version']
     issue.update(fields={
-        'labels': [{'name': x} for x in labels]
+        'labels': [x for x in labels]
```
dhellmann, line 90:
```suggestion
        'labels': labels,
```

Do we need a new list there?

scripts/jira/cloner.py
```diff
@@ -4,8 +4,10 @@
 import argparse
 from tqdm import tqdm
 
-JQL_GLOBAL_QUERY = 'filter = "MicroShift - Bugs in Project" and status in (New, Assigned, Post, "To Do", "In Progress", "Code Review") order by key asc'
-JQL_ISSUE_QUERY = 'key in ({}) order by key asc'
+JQL_FILTER_QUERY = 'filter = "MicroShift - Bugs in Project" and status in (New, Assigned, Post, "To Do", "In Progress", "Code Review")'
+JQL_FILTER_ISSUE = 'key in ({})'
+JQL_FILTER_USER = 'assignee in ({})'
+JQL_ORDER_BY = 'order by key asc'
```
dhellmann, line 10:
```suggestion
JQL_ORDER_BY = 'order by key asc'
SPRINT_FIELD_ID = 'customfield_12310940'
```

scripts/jira/cloner.py
```diff
@@ -93,30 +105,51 @@ def set_target_version(issue, version):
         'customfield_12319940': [{'name': version}]
```
dhellmann, line 105:
```suggestion
       SPRINT_FIELD_ID: [{'name': version}]
```

scripts/jira/cloner.py
```diff
@@ -93,30 +105,51 @@ def set_target_version(issue, version):
...
+    })
+
+def set_fix_versions(issue, fix_versions):
+    issue.update(fields={
+        'fixVersions': [{'name': x} for x in fix_versions]
+    })
+
+def set_qa_contact(issue, contact):
+    issue.update(fields={
+        'customfield_12315948': {'value': contact}
```
dhellmann, line 122: This custom field ID should be a constant, too.

dhellmann (commented):
A few suggestions for your next iteration, but nothing to hold this up.

/lgtm

PR #1549, opened 2023-03-22
[WIP] [OCPBUGS-7977] Missed 'etcdctl' in microshift-etcd

copejon:

OCPBUGS-7977

Add the ectdctl subcommands to microshift-etcd.  To accomplish this, I exported the etcdctl rootCmd variable with a patch so it can be imported directly.

This places the top level commands of etcdctl at the root of microshift-etcd, see below.  This method admittedly only really provides the syntactic sugar of reusing the microshift-etcd command.  An alternative method would be to include etcdctl as a binary dependency in the scripts/tools.go file and pin it's version in go.mod.  Meaning we'd simply package etcdctl alongside microshift.

Size Comparison:
|component|size Mb|
|---|---|
|etcdtl|18M|
|microshift-etcd|43M
|microshift-etcd(ctl)|67M|

With a size increase on microshift-etcd of 24M.

```
[microshift@ushift-dev ~]$ microshift-etcd --help
Usage:
  microshift-etcd [flags]
  microshift-etcd [command]

Available Commands:
  alarm       Alarm related commands
  auth        Enable or disable authentication
  check       commands for checking properties of the etcd cluster
  compaction  Compacts the event history in etcd
  completion  Generate the autocompletion script for the specified shell
  defrag      Defragments the storage of the etcd members with given endpoints
  del         Removes the specified key or range of keys [key, range_end)
  elect       Observes and participates in leader election
  endpoint    Endpoint related commands
  get         Gets the key or a range of keys
  help        Help about any command
  lease       Lease related commands
  lock        Acquires a named lock
  make-mirror Makes a mirror at the destination etcd cluster
  member      Membership related commands
  move-leader Transfers leadership to another etcd cluster member.
  put         Puts the given key into the store
  role        Role related commands
  run
  snapshot    Manages etcd node snapshots
  txn         Txn processes all the requests in one transaction
  user        User related commands
  version     Prints the version of etcdctl
  version     Print MicroShift-etcd version information
  watch       Watches events stream on keys or prefixes

Flags:
  -h, --help                           help for microshift-etcd
      --log-flush-frequency duration   Maximum number of seconds between log flushes (default 5s)
  -v, --v Level                        number for the log level verbosity
      --vmodule moduleSpec             comma-separated list of pattern=N settings for file-filtered logging (only works for the default text log format)

Use "microshift-etcd [command] --help" for more information about a command.
[microshift@ushift-dev ~]$
```


copejon:
I'm going to close this as there are some broader design considerations that I'd like to understand better before ham-fistedly importing all of etcdctl.  The fact alone that embedding it is more storage hungry that simply including the binary is reason enough I think.

PR #1557, opened 2023-03-23
WIP microshift specific tests

pmtk:
/cc


e2e/tests/0030-router-smoke-test.sh
```diff
@@ -0,0 +1,42 @@
...
+source "${SCRIPT_PATH}/../helpers.sh"
+
+cleanup() {
+    oc delete route hello-microshift
+    oc delete service hello-microshift
+    oc delete -f "${SCRIPT_PATH}/assets/hello-microshift.yaml"
+}
+trap cleanup EXIT
+
+gcloud compute firewall-rules update "${INSTANCE_PREFIX}" --allow tcp:22,icmp,tcp:80
```
dhellmann, line 17: Can we split these test scripts up so it is possible to run them against a local deployment, too?

e2e/tests/0050-greenboot.sh
```diff
@@ -0,0 +1,132 @@
...
+
+set -euo pipefail
+IFS=$'\n\t'
+set -x
+
+SCRIPT_PATH="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")"
+
+scp "$SCRIPT_PATH/assets/busybox_running_check.sh" "$USHIFT_USER@$USHIFT_IP":~/busybox_running_check.sh
+
+ssh -q "$USHIFT_USER@$USHIFT_IP" <<'ENDSSH'
```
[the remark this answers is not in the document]
pmtk: Definitely

e2e/helpers.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+gscp() {
```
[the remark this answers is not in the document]
dhellmann: I like that

e2e/tests/0030-router-smoke-test.sh
```diff
@@ -0,0 +1,37 @@
...
+# shellcheck disable=SC2317  # Don't warn about unreachable commands in this function
+cleanup() {
+    oc delete route hello-microshift || true
+    oc delete service hello-microshift || true
+    oc delete -f "${SCRIPT_PATH}/assets/hello-microshift.yaml" || true
+    firewall::close_port 80 tcp || true
+}
+trap cleanup EXIT
+
+firewall::open_port 80 tcp
```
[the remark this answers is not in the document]
dhellmann: TIL, nice

e2e/cluster-debug-info.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
```
dhellmann, line 1: At what point can this script just become a call to sos report?

e2e/main.sh
```diff
@@ -0,0 +1,134 @@
...
+  subjectAltNames:
+  - '"$USHIFT_IP"'
+EOF'
+    ssh "$USHIFT_USER@$USHIFT_IP" "sudo systemctl enable --now microshift"
+}
+
+microshift_debug_info() {
+    log "Gathering debug info"
+    scp "$SCRIPT_DIR/cluster-debug-info.sh" "$USHIFT_USER@$USHIFT_IP:/tmp/cluster-debug-info.sh"
+    ssh "$USHIFT_USER@$USHIFT_IP" "sudo KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig /tmp/cluster-debug-info.sh"
```
dhellmann, line 71: Is there any reason the KUBECONFIG variable can't be set from inside the debug script?

e2e/main.sh
```diff
@@ -0,0 +1,134 @@
...
+    log "${test} - RUNNING"
+
+    microshift_cleanup
+    microshift_setup
+    microshift_check_readiness
+    konfig=$(microshift_get_konfig)
+
+    start=$(date +%s)
+    set +e
+    KUBECONFIG="$konfig" "${SCRIPT_DIR}/tests/${test}"
```
dhellmann, line 91: If this function took the full path to the test script as argument then a developer could use tab completion when running a test from the command line.

e2e/main.sh
```diff
@@ -0,0 +1,134 @@
...
+    res=$?
+    set -e
+    rm -f "${konfig}"
```
dhellmann, line 94: Is it worth using a trap to do this cleanup?

e2e/cluster-debug-info.sh
pacevedom: As soon as the sos version we need is publicly available this script may be retired.

pmtk: I don't remember the details, but I think we still don't have everything in SOS? Does it have "feature parity" with the cluster-debug-info.sh?

pmtk: I take ^ back. Just looked at sos from `master` and it's pretty dope.
@dhellmann @pacevedom what do you think about leaving couple commands in the job log to have quickly glanceable info in case of error? Like `oc get {pods,nodes}` and maybe `oc get events -A --sort-by=.metadata.creationTimestamp | head -n 20`. Given our problems with topolvm image pulling I'd get quickly annoyed to go through downloading and extracting sos report only to find that issue there :P

e2e/main.sh
```diff
@@ -0,0 +1,151 @@
...
+    tmpfile=$(mktemp /tmp/microshift-e2e-konfig.XXXXXX)
+    ssh "$USHIFT_USER@$USHIFT_IP" 'sudo cat /var/lib/microshift/resources/kubeadmin/'"$USHIFT_IP"'/kubeconfig' >"$tmpfile"
+    echo "$tmpfile"
+}
+
+microshift_check_readiness() {
+    local test_output="${1}"
+    log "Waiting for MicroShift to become ready"
+    ssh "$USHIFT_USER@$USHIFT_IP" \
+        "sudo /etc/greenboot/check/required.d/40_microshift_running_check.sh | \
```
dhellmann, line 56: Passing shell commands to ssh to run with pipes and loops and things can make quoting pretty messy. Can we process the output of the script outside of ssh? Or can we make the script being run produce just the output we want (maybe via a command line switch)?

We also, at one point, talked about just copying all of the test code into the VM and then running everything there. I can see some issues with that, but the appeal is that is simplifies the test runner logic, at least in some cases. Does it make the implementation of the tests harder, though?

e2e/main.sh
```diff
@@ -0,0 +1,151 @@
...
+microshift_setup() {
+    local test_output="${1}"
+    log "Setting up and starting MicroShift"
+    ssh "$USHIFT_USER@$USHIFT_IP" 'cat << EOF | sudo tee /etc/microshift/config.yaml
+---
+apiServer:
+  subjectAltNames:
+  - '"$USHIFT_IP"'
+EOF' &>"${test_output}/0001-setup.log"
+    ssh "$USHIFT_USER@$USHIFT_IP" "sudo systemctl enable --now microshift" &>>"${test_output}/0001-setup.log"
```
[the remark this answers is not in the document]
pmtk: That was my initial approach, but then piping whole function output to file leads to messages like "setting up microshift" to be in that file instead of printed out to console. Unless there's some magic with streams (like "separate runner's std{err,out} from test's std{err,out})?

Slightly related: I consider this script as an intermediate step - move tests here, see what we need, what we want.
I hope that we'll be able to move to some other solution. I spent some time over the weekend [PoCing python script](https://github.com/pmtk/microshift/commit/d25d9f6ff4b81c9dcefe05471dc8b0ab8c29fb4a) and I see needs beside what we have here:
- junit xml is very nice addition
- tagging/labeling tests (e.g. to group "non-destructive" like smoke tests we have now together without tearing down and setting up microshift in between; as opposed to greenboot test that tears down microshift multiple times)

e2e/main.sh
pmtk: I feel that tests like router/lb smokes would not be as sane ran from within.
Also - reboot test (which I haven't ported(? - there's nothing to port there actually...) yet)

I know that current state of the PR is "in between" approaches, but I didn't wanted to make too many changes to existing tests (esp greenboot) hoping for iterative approach.

Re that monstrosity after pipe - it just adds timestamp to beginning of the line, somehow I couldn't find moreutils package on RHEL (which would provide [ts](https://linux.die.net/man/1/ts) and that would be much simpler).
Also, it's not needed at all - just put it there because I noticed that greenboot check takes way too long (pod restart check takes 50s for each namespace sequentially)

pmtk: I just remembered testing I've done simulating "power loss during initialization" - test similar to this (or just "do 100 reboots")  has no place running from the microshift itself. 

Having external tests runner feels cleaner to me, but there's nothing prohibiting the test suite from the microshift (like selected tests running on dev vm against dev vm), but that's probably not what you meant?
Are we talking about test runner on dev VM or just having test cases in scripts, copied to remote and executed there? Even though bash scripts are much simpler at times due to being native to shell (duh) or `set -x` providing full debug info, they feel like a mess.

dhellmann: I'm sorry, yes, you've pointed out that reboot case to me a few times before, and I keep forgetting it.

I'm trying to minimize the complexity of the commands embedded in the ssh arguments. Maybe we can isolate just those cases as temporary scripts? A guideline like "if you wouldn't type it on the command line, don't pass it to ssh as an argument" could give us a heuristic. Maybe that's stricter than needed, though?

e2e/main.sh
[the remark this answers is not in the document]
dhellmann: OK, let's keep things as you have them for now, then.

e2e/main.sh
pmtk: Can we skip this? This conflicts with the current way and I'm not sure if it's worth spending time marry what you proposed with current glob-like approach. Esp that I'd like to migrate from bash

dhellmann: Sure, it sounds like you have other plans so that's fine. Moving away from bash has my full support. :-)

dhellmann (commented):
I'm happy with this version. I'll leave the PR open to let some of the rest of the team look at it, since there are others who have been following the work more closely than I have.

e2e/main.sh
```diff
@@ -0,0 +1,197 @@
...
+function_should_be_exported() {
+    local fname=${1}
+    if ! declare -F "${fname}"; then
+        log "WARNING: Function '${fname}' is unexported. It is expected that function is provided for interacting with cloud provider"
+        return 1
+    fi
+}
+
+check_passwordless_ssh() {
+    ssh -o BatchMode=yes "${USHIFT_USER}@${USHIFT_IP}" "true" || {
```
ggiguash, line 58: There are quite a few repetitions for the `ssh` command invocation. If we create a `run_ssh_command` function, it should also allow us to transparently execute commands locally if the host/user are the same.

e2e/main.sh
```diff
@@ -0,0 +1,197 @@
...
+
+    all_successful=true
+    for t in ${to_run}; do
+        run_test "${t}" || all_successful=false
+    done
+    "${all_successful}"
+}
+
+[ $# -eq 0 ] && {
+    usage "Expected arguments"
```
ggiguash, line 192:
```suggestion
    usage "Missing expected arguments"
```

e2e/tests/0050-greenboot.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+IFS=$'\n\t'
+PS4='+ $(date "+%T.%N")\011 '
+set -x
+
+SCRIPT_PATH="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")"
+
+scp "${SCRIPT_PATH}/../../docs/config/busybox_running_check.sh" "${SCRIPT_PATH}/assets/greenboot-test.sh" "${USHIFT_USER}@${USHIFT_IP}":/tmp/
```
ggiguash, line 10: Let's copy those to the user home dir if possible to avoid security remarks about predictable file names in /tmp?

e2e/tests/0100-reboot.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Allow for `ssh` command errors (255 exit code) like "connection closed by remote host"
+# Fail on other errors (coming from the command executed remotely itself)
+if [ "${res}" -ne 0 ] && [ "${res}" -ne 255 ]; then
+    exit 1
+fi
+
+wait_until ssh "${USHIFT_USER}@${USHIFT_IP}" "true"
+# Just check if KAS is up and serving
+wait_until oc get node
```
ggiguash, line 42: This should not be necessary as the readiness check script has the code to deal with microshift not yet running.

ggiguash (changes requested):

e2e/main.sh
```diff
@@ -0,0 +1,219 @@
...
+
+microshift_setup() {
+    local output_dir="${1}"
+    log "Setting up and starting MicroShift"
+    ssh_cmd 'cat << EOF | sudo tee /etc/microshift/config.yaml
+---
+apiServer:
+  subjectAltNames:
+  - '"${USHIFT_IP}"'
+EOF' &>"${output_dir}/0001-setup.log"
```
pacevedom, line 109: Is this needed? Should it remain as part of the installation?

pmtk: But it's only done in CI, missing for local VMs (unless someone does it manually). I'd prefer using IPs over hostnames if we can

pacevedom: Tests only require a host for the execution, be that a hostname or an IP, right? I would put the requirement on configuring the system outside of the tests so that they are generic for any environment.
In CI you have that as part of the install steps, and in a dev environment you may get away with default configuration. For example, I use "microshift-rhel9" as hostname for my VM and never use its IP, therefore the default config is ok.

PR #1558, opened 2023-03-23
USHIFT-642: Static analysis for .go, .sh, and .py

pmtk:


dhellmann:
This looks OK. It would be nice if the tools could be run by a developer using a container image, but I suppose that wouldn't work well in CI where the job is already running in a container?

pmtk:
> This looks OK. It would be nice if the tools could be run by a developer using a container image, but I suppose that wouldn't work well in CI where the job is already running in a container?

Not sure about running container within a job - by glancing https://www.redhat.com/sysadmin/podman-inside-kubernetes it might need some extra privileges that CI probably don't support.
Other way would be to use that container with tools as a base and copy sources into it, e.g.:
```
    FROM linter-tools
    COPY microshift /tmp/
    WORKDIR /tmp/microshift/
```
then in a job: `cd /tmp/microshift && make lint-xyz` - we'd have to find a way to keep this straightforward for dev as well: `make lint` would start a container with sources mounted (just like rpm-podman).

scripts/fetch_tools.sh
```diff
@@ -0,0 +1,138 @@
...
+
+    _install "$url" "$checksum" "$filename" "yq_linux_${arch}"
+}
+
+[[ "$(uname -o)" == "GNU/Linux" ]] || { echo "Script only runs on Linux"; exit 1; }
+[[ "$(uname -p)" =~ x86_64|aarch64 ]] || { echo "Only x86_64 and aarch64 architectures are supported"; exit 1; }
+
+if [ $# -eq 1 ]; then
+    "get_$1"
+else
```
ggiguash, line 133: Can we add check for input and proper usage message in case it's wrong?

ggiguash: Also, let's generate a help message if run without arguments? 
We can download all the tools using a special "all" arg.

scripts/fetch_tools.sh
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+ARCH="$(uname -p)"
+DEST_DIR="${DEST_DIR:-_output/bin}"
```
ggiguash, line 7: This implies the script can only be run from the top-level microshift directory. Let's make is relative to the script location?

scripts/fetch_tools.sh
```diff
@@ -0,0 +1,138 @@
...
+    if ! sha256sum -c "${tmp}/checksum.txt" < "$tmp/$filename" &>/dev/null; then
+        echo "  Checksum for $filename doesn't match"
+        echo "    Expected: $checksum"
+        echo "         Got: $(sha256sum < "$tmp/$filename" | cut -d' ' -f1)"
+        return 1
+    fi
+
+    if [[ "$(file --brief --mime-type "$tmp/$filename")" != "application/x-executable" ]]; then
+        (cd "$tmp" && tar xvf "$filename" --transform 's,.*\/,,g' --wildcards "*/$initial_filename" >/dev/null)
+    fi
```
ggiguash, line 48: What are we doing here? Please, also add a comment in the code before the "if" block

pmtk: added comment

dhellmann:
2b0632593c0a368c12e852adc3e5fa28ea4dfec9 also resolves USHIFT-987

PR #1559, opened 2023-03-23
add storage config documentation

copejon:
Add storage configuration, including how to configure multiple device classes, to the config doc.

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -157,3 +157,7 @@ Restart the MicroShift service to apply the manifests and verify that the `busyb
 sudo systemctl restart microshift
 oc get pods -n busybox
 ```
+
+## Storage Configuration
+
+MicroShift's included CSI plugin manages LVM LogicialVolumes to provide persistent workload storage. For specific storage configuration, refer to the dedicated [documentation](./storage/configuration.md).
--- a/docs/storage/configuration.md
+++ b/docs/storage/configuration.md
@@ -1,0 +1,98 @@
+# Storage Configuration
+
+LVM backed storage is indirectly configured through `/etc/microshift/lvmd.yaml`. LVMD is a sub-component of the CSI plugin responsible for managing LVM host operations.  An example with default values is provided by MicroShift at `/etc/microshift/lvmd.yaml.default`.  Users may copy this file and rename it to `/etc/microshift/lvmd.yaml` to customize their configuration.
+
+### Default Config
+
+Below is an example of what users will find at `/etc/microshift/lvmd.yaml.default`.  
+
+```yaml
+# Unix domain socket endpoint of gRPC
+#socket-name: /run/lvmd/lvmd.socket
+
+device-classes:
+  # The name of a device-class
+  #- name: default
+
+    # The group where this device-class creates the logical volumes
+    #volume-group: microshift
+
+    # Storage capacity in GiB to be spared
+    #spare-gb: 0
+
+    # A flag to indicate that this device-class is used by default
+    #default: true
+
+    # The number of stripes in the logical volume
+    #stripe: ""
+
+    # The amount of data that is written to one device before moving to the next device
+    #stripe-size: ""
+
+    # Extra arguments to pass to lvcreate, e.g. ["--type=raid1"]
+    #lvcreate-options:
+      #- ""
+```
+
+### Device Classes
+
+The storage configuration must specify at least one device class.  One and only one device class must be set as default.
+
+Multiple device classes may be defined in the `device-classes` array, and they can be a mix of thick and thin volume configurations. E.g.
+
+```yaml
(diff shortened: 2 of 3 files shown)


docs/howto_config.md
```diff
@@ -157,3 +157,104 @@ Restart the MicroShift service to apply the manifests and verify that the `busyb
 sudo systemctl restart microshift
 oc get pods -n busybox
 ```
+
```
dhellmann, line 160: It seems like this new content should be part of `docs/default_csi_plugin.md`. What do you think?

dhellmann: Now it's in both files.

ggiguash: I agree that we need to "split" the text:
* The design document should not contain the specifics of the configuration file internals. 
* The configuration file should list the options and explain them to the end-user

dhellmann: For CNI, which also has a separate configuration file, we put the docs in docs/network/default_cni_plugin.md. Can we put all of the CSI docs in the CSI file, for consistency?

ggiguash: Yes, we need to be consistent. 

I would prefer to have all of the generic/CSI/CNI configuration options documented at the same place in the howto_config.md file. I'm thinking about design as more of a stable text, while config options/details change more often.

Either way, if we choose to keep this text in the design, I think it's fine as long as we create links to the appropriate sections from the howto_config.md file. It would be nice to have all the configuration options reachable from a single doc.

docs/storage/configuration.md
```diff
@@ -0,0 +1,98 @@
...
+```yaml
+# Unix domain socket endpoint of gRPC
+#socket-name: /run/lvmd/lvmd.socket
+
+device-classes:
+  # The name of a device-class
+  #- name: default
+
+    # The group where this device-class creates the logical volumes
+    #volume-group: rhel
```
dhellmann, line 18:
```suggestion
    #volume-group: microshift
```

docs/storage/configuration.md
```diff
@@ -0,0 +1,98 @@
...
+- Setting spare-gb to anything other than 0 is not recommended because it does not behave predictably and usually results in more space being allocated that expected.
+
+### Storage Class
+
+Storage Classes provide the workload layer interface for selecting a given device class.  The following storage class parameters are supported.
+
+- `csi.storage.k8s.io/fstype` selects filesystem type: `xfs` or `ext4` are supported
+- `topolvm.io/device-class` maps the backend device class. If not provided, the default device class is assumed.
+
+The relationship between storage classes and device classes is many-to-one.  This allows admins to provide varying sets of parameters for the same backing device class, e.g. to provide xfs and ext4 variants.
```
dhellmann, line 74:
```suggestion
Multiple storage classes can refer to the same device class.  This allows admins to provide varying sets of parameters for the same backing device class, e.g. to provide xfs and ext4 variants.
```

docs/storage/configuration.md
```diff
@@ -0,0 +1,98 @@
...
+parameters:
+  "csi.storage.k8s.io/fstype": "xfs" [2]
+  "topolvm.io/device-class": "ssd" [3]
+provisioner: topolvm.io [4]
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer [5]
+allowVolumeExpansion: false [6]
+```
+
+1. Denotes this storage class is the cluster's default.  If a PVC does not specify a storage class, this class is assumed. Exclude the annotation if not a default storage class.
```
dhellmann, line 93: Maybe number these items based on the order above for users who are reading the `.md` file offline?

dhellmann (commented):
A couple of wording suggestions, but this is organized well.

docs/storage/configuration.md
```diff
@@ -0,0 +1,98 @@
+# Storage Configuration
+
+LVM backed storage is indirectly configured through `/etc/microshift/lvmd.yaml`. LVMD is a sub-compnent of the CSI plugin responsible for managing LVM host operations.  An example with default values is provided by MicroShift at `/etc/microshift/lvmd.yaml.default`.  Users may copy this file and rename it to `/etc/microshift/lvmd.yaml` to customize their configuration.
```
ggiguash, line 3:
```suggestion
LVM backed storage is indirectly configured through `/etc/microshift/lvmd.yaml`. LVMD is a sub-component of the CSI plugin responsible for managing LVM host operations.  An example with default values is provided by MicroShift at `/etc/microshift/lvmd.yaml.default`.  Users may copy this file and rename it to `/etc/microshift/lvmd.yaml` to customize their configuration.
```

docs/storage/configuration.md
copejon: done

dhellmann (commented):
/lgtm

PR #1563, opened 2023-03-24
USHIFT-1002: add script to disable microshift

dhellmann:
This script disables MicroShift and removes the OVS configuration
needed for OVN-K, restoring the system to a state where the admin can
use Cockpit to modify the host network settings.

/assign @zshi-redhat

### DIFF
--- a/scripts/microshift-disable.sh
+++ b/scripts/microshift-disable.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+SCRIPT_NAME=$(basename $0)
+
+# Exit if the current user is not 'root'
+if [ $(id -u) -ne 0 ] ; then
+    echo "The '${SCRIPT_NAME}' script must be run with the 'root' user privileges"
+    exit 1
+fi
+
+echo "Stopping MicroShift..."
+systemctl stop microshift
+
+echo "Stopping pods managed by kubelet..."
+systemctl stop kubepods.slice
+
+echo "Removing OVS configuration..."
+/usr/bin/configure-ovs.sh OpenShiftSDN
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -143,6 +143,7 @@ install -d %{buildroot}%{_bindir}
 install -p -m755 ./_output/microshift %{buildroot}%{_bindir}/microshift
 install -p -m755 ./_output/microshift-etcd %{buildroot}%{_bindir}/microshift-etcd
 install -p -m755 scripts/microshift-cleanup-data.sh %{buildroot}%{_bindir}/microshift-cleanup-data
+install -p -m755 scripts/microshift-disable.sh %{buildroot}%{_bindir}/microshift-disable
 
 restorecon -v %{buildroot}%{_bindir}/microshift
 restorecon -v %{buildroot}%{_bindir}/microshift-etcd
@@ -258,6 +259,7 @@ systemctl enable --now --quiet openvswitch || true
 %{_bindir}/microshift
 %{_bindir}/microshift-etcd
 %{_bindir}/microshift-cleanup-data
+%{_bindir}/microshift-disable
 %{_unitdir}/microshift.service
 %{_sysconfdir}/crio/crio.conf.d/microshift.conf
 %dir %{_sysconfdir}/microshift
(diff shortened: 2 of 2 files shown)


dhellmann:
I need to add some documentation, too.

/hold

scripts/microshift-disable.sh
```diff
@@ -0,0 +1,20 @@
...
+fi
+
+echo "Stopping MicroShift..."
+systemctl stop microshift
+
+echo "Stopping pods managed by kubelet..."
+systemctl stop kubelet.slice
+
+echo "Removing OVS configuration..."
+/usr/bin/configure-ovs.sh OpenShiftSDN
```
zshi-redhat, line 20: There is a partial cleanup issue when `br-int` is removed in this script, details captured in this ticket: https://issues.redhat.com/browse/OCPBUGS-7778. This could result in pod fails to start after restart microshift service.
To fix the partial cleanup issue, we could either:
1. require a node reboot after running this script or
2. add an parameter in configure-ovs.sh to skip `br-int` bridge removal. for example: `/usr/bin/configure-ovs.sh disable`

zshi-redhat (commented):
@dhellmann  Thanks for working on the fix!

ggiguash:
@dhellmann / @zshi-redhat - a lot of this logic is already integrated into the microshift-cleanup-data.sh script. 
Do you think we'd better add --stop option to the existing script instead of creating a new one?

scripts/microshift-disable.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ $(id -u) -ne 0 ] ; then
+    echo "The '${SCRIPT_NAME}' script must be run with the 'root' user privileges"
+    exit 1
+fi
+
+echo "Stopping MicroShift..."
+systemctl stop microshift
+
+echo "Stopping pods managed by kubelet..."
+systemctl stop kubelet.slice
```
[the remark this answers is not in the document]
dhellmann: I updated both, thanks.

scripts/microshift-disable.sh
dhellmann: The reboot option is probably the safest. That would bring everything up freshly with the new IP settings. I'll add that to the document.

dhellmann:
> @dhellmann / @zshi-redhat - a lot of this logic is already integrated into the microshift-cleanup-data.sh script.
> Do you think we'd better add --stop option to the existing script instead of creating a new one?

We could also do that. To start, I want to provide this customer with something very simple that they could potentially integrate with a Cockpit UI they're building. Having the steps in a standalone script will make it easier for them to remix it if they want to do that.

scripts/microshift-disable.sh
ggiguash: The problem is that deleting the `br-int` interface may cut off the network connectivity. This is what happens when we're uninstalling MicroShift RPMs.

scripts/microshift-disable.sh
```diff
@@ -0,0 +1,20 @@
...
+    exit 1
+fi
+
+echo "Stopping MicroShift..."
+systemctl stop microshift
+
+echo "Stopping pods managed by kubelet..."
+systemctl stop kubepods.slice
+
+echo "Removing OVS configuration..."
```
ggiguash, line 19: This script generates a lot of output. Do we want to redirect it to /dev/null?

scripts/microshift-disable.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
ggiguash, line 1: The name of the script may be a bit misleading as it does not disable microshift, only stops services/pods and reverts OVN configuration. MicroShift will be started after boot.

scripts/microshift-disable.sh
zshi-redhat: @ggiguash Yes, deleting br-int interface will disconnect pod network, good capture on that!
We could either reboot the node to fix it or not deleting br-int in this script.

ggiguash: The reboot can only be performed from the current script to make sure the operation is atomic and network is back. 
However, since we're trying to temporarily "undo" OVN configuration for updating the IP address from cockpit, this creates a race with the reboot. 
Looks like we need to add the new IP setting + reboot to the same script to have everything consistent. Any other ideas?

dhellmann: My goal isn't to combine everything into 1 script, it's to have a process they can follow.

What network connectivity is cut off when `br-int` is removed? I didn't see a problem using a browser outside of the VM to access cockpit on the VM. In fact, the cockpit UI updated live as the script ran.

ggiguash: External network connectivity may be disrupted when switching between SDN and OVN modes if there is no reboot. Until now, this happened when uninstalling and reinstalling MicroShift. With this script, we're adding one more similar flow.

dhellmann:
We no longer expect to need this.

/close

PR #1565, opened 2023-03-25
configure-vm: add option to not install build deps

pmtk:


scripts/devenv-builder/configure-vm.sh
```diff
@@ -5,87 +5,118 @@
...
 
+start=$(date +%s)
+
 function usage() {
-    echo "Usage: $(basename $0) [--no-build] <openshift-pull-secret-file>"
-    echo "  --no-build   Do not build MicroShift code and install MicroShift RPMs"
+    echo "Usage: $(basename $0) [--no-build] [--no-build-deps] [--force-firewall] <openshift-pull-secret-file>"
+    echo ""
+    echo "  --no-build         Do not build, install and start MicroShift"
+    echo "  --no-build-deps    Do install dependencies for building binaries and RPMs (implies --no-build)"
```
ggiguash, line 19:
```suggestion
    echo "  --no-build-deps    Do not install dependencies for building binaries and RPMs (implies --no-build)"
```

scripts/devenv-builder/configure-vm.sh
```diff
@@ -5,87 +5,118 @@
...
-# Only RHEL requires a subscription
-if grep -q 'Red Hat Enterprise Linux' /etc/redhat-release ; then
-    RHEL_SUBSCRIPTION=true
-fi
+title() {
+    echo -e "\E[34m$1\E[00m"
+}
 
-if [ $# -ne 1 ] && [ $# -ne 2 ]; then
+while true; do
```
ggiguash, line 30:
```suggestion
while [ $# -gt 1 ] ; do
```

scripts/devenv-builder/configure-vm.sh
```diff
@@ -5,87 +5,118 @@
...
+    --no-build-deps)
+        INSTALL_BUILD_DEPS=false
+        BUILD_AND_RUN=false
+        shift
+        ;;
+    --force-firewall)
+        FORCE_FIREWALL=true
+        shift
+        ;;
+    *) break ;;
```
ggiguash, line 45: Need to call usage here to avoid invalid options.
```suggestion
    *)  usage ;;
```

scripts/devenv-builder/configure-vm.sh
```diff
@@ -117,17 +149,22 @@ gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-NFV
 EOF
 fi
 
-if [ $BUILD_AND_INSTALL = true ] ; then
-    sudo dnf localinstall -y ~/microshift/_output/rpmbuild/RPMS/*/*.rpm
+# Set up pull secret for CRI-O regardless if it'll be installed in next step or later from another source
+title "Setting up pull secret for CRI-O"
+sudo mkdir -p /etc/crio/
+sudo cp -f ${OCP_PULL_SECRET} /etc/crio/openshift-pull-secret
+sudo chmod 600 /etc/crio/openshift-pull-secret
```
ggiguash, line 156: If we want to make the pull secret setup unconditional, let's do it without changing the order: first install RPMs and then the pull secret. This is to adhere to the current docs.

scripts/devenv-builder/configure-vm.sh
```diff
@@ -137,35 +174,59 @@ else
...
+    sudo systemctl enable crio >/dev/null
+
+    title "Enabling and starting MicroShift"
+    sudo systemctl enable microshift --now >/dev/null
+
+    sudo bash -c "while [ ! -e /var/lib/microshift/resources/kubeadmin/kubeconfig ]; do sleep 5s; done"
+
+    title "Copying kubeconfig to ~/.kube/config"
+    mkdir -p ~/.kube
+    sudo cat /var/lib/microshift/resources/kubeadmin/kubeconfig | tee ~/.kube/config >/dev/null
```
ggiguash, line 208: I'm not sure this is necessary here. The idea is to verify MicroShift can be started and clean it up, so that it's possible to run standalone exe in the devenv. 
So, I would remove this and fix the `watch oc get pod -A` command to use the original kubeconfig file.

ggiguash:
As a general command, I saw that you added `/dev/null` redirects on some commands in the script. I understand this is because we now have "titles". However, this script is a mirror of the devenv_setup.md doc and the redirects are not in it. 
So, I think we should try be as similar to the documentation as possible.

scripts/devenv-builder/configure-vm.sh
ggiguash: I noticed the unconditional call to this code fails in CI. Why do we need it unconditionally?
```
+ echo -e '\E[34mSetting up pull secret for CRI-O\E[00m'
+ sudo mkdir -p /etc/crio/
+ sudo cp -f /etc/crio/openshift-pull-secret /etc/crio/openshift-pull-secret
cp: '/etc/crio/openshift-pull-secret' and '/etc/crio/openshift-pull-secret' are the same file
```

pmtk: I'll swap the order so it's matching what's in the doc (although I wish there was "literal bash" - readable script with embedded docs, because that's a duplication right now).
I want this unconditionally to set up the pull secret without installing RPMs for the CI - cri-o will be pulled in when we install microshift RPMs but that happens after this script

scripts/devenv-builder/configure-vm.sh
pmtk: Done

scripts/devenv-builder/configure-vm.sh
pmtk: Done

scripts/devenv-builder/configure-vm.sh
pmtk: Done

scripts/devenv-builder/configure-vm.sh
pmtk: removed

PR #1575, opened 2023-03-27
OCPBUGS-10766: Change apiserver preferred addresses

pacevedom:
In order to avoid DNS configuration unless there is no other way, try using external and internal IPs to reach kubelet from apiserver. This allows having whatever hostname+domain name combinations without trying to resolve the name externally.
To allow this we need to include the node IP in the kubelet certificate as this is what the apiserver will target.

Closes #<Issue Number>

### DIFF
--- a/pkg/controllers/kube-apiserver.go
+++ b/pkg/controllers/kube-apiserver.go
@@ -133,7 +133,7 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 			"kubelet-certificate-authority":   {cryptomaterial.CABundlePath(kubeCSRSignerDir)},
 			"kubelet-client-certificate":      {cryptomaterial.ClientCertPath(kubeletClientDir)},
 			"kubelet-client-key":              {cryptomaterial.ClientKeyPath(kubeletClientDir)},
-			"kubelet-preferred-address-types": {"Hostname"},
+			"kubelet-preferred-address-types": {"InternalIP", "Hostname"},
 
 			"proxy-client-cert-file":           {cryptomaterial.ClientCertPath(aggregatorClientCertDir)},
 			"proxy-client-key-file":            {cryptomaterial.ClientKeyPath(aggregatorClientCertDir)},
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -180,7 +180,7 @@ func certSetup(cfg *config.MicroshiftConfig) (*certchains.CertificateChains, err
 						Name:         "kubelet-server",
 						ValidityDays: cryptomaterial.ShortLivedCertificateValidityDays,
 					},
-					Hostnames: []string{cfg.NodeName},
+					Hostnames: []string{cfg.NodeName, cfg.NodeIP},
 				},
 			),
 		),


dhellmann:
I'm not sure about the implications of some of these changes.

> In order to avoid DNS configuration unless there is no other way, try using external and internal IPs to reach kubelet from apiserver. 

What are the circumstances where a host won't know its own name? What doesn't work when that's the situation?

> This allows having whatever hostname+domain name combinations without trying to resolve the name externally. To allow this we need to include the node IP in the kubelet certificate as this is what the apiserver will target.

Why do we want multiple combinations? Should we just use the node IP?

pliurh:
> What are the circumstances where a host won't know its own name? What doesn't work when that's the situation?

We build microshift with CGO disabled. Thus golang doesn't use glibc to resolve a hostname, which means the `myhostname` NSS plugin is unavailable. Microshift can only use `hosts` and `dns`. So when we don't have microshift's hostname configured in the DNS server or the hosts file, microshift cannot know its own IP by its hostname.

pkg/controllers/kube-apiserver.go
```diff
@@ -133,7 +133,7 @@ func (s *KubeAPIServer) configure(cfg *config.MicroshiftConfig) error {
 			"kubelet-certificate-authority":   {cryptomaterial.CABundlePath(kubeCSRSignerDir)},
 			"kubelet-client-certificate":      {cryptomaterial.ClientCertPath(kubeletClientDir)},
 			"kubelet-client-key":              {cryptomaterial.ClientKeyPath(kubeletClientDir)},
-			"kubelet-preferred-address-types": {"Hostname"},
+			"kubelet-preferred-address-types": {"ExternalIP", "InternalIP", "ExternalDNS", "InternalDNS", "Hostname"},
```
pliurh, line 136: In microshift, only 'InternalIP' and 'Hostname' are provided by kubelet. Shall we remove the address types we don't use?

pacevedom: In order to allow the CNCF conformance tests requiring multinode I dropped there the external counterparts, just in case.

pacevedom: Wait, let me simplify it because you are right.

pacevedom: Done. This also addresses your comment @dhellmann.
```bash
$ KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig oc get node -o jsonpath='{.items[*].status.addresses}' | jq
[
  {
    "address": "192.168.122.215",
    "type": "InternalIP"
  },
  {
    "address": "microshift-dev.local",
    "type": "Hostname"
  }
]
```

dhellmann:
> > What are the circumstances where a host won't know its own name? What doesn't work when that's the situation?
> 
> We build microshift with CGO disabled. Thus golang doesn't use glibc to resolve a hostname, which means the `myhostname` NSS plugin is unavailable. Microshift can only use `hosts` and `dns`. So when we don't have microshift's hostname configured in the DNS server or the hosts file, microshift cannot know its own IP by its hostname.

OK, so the host _does_ know its name, but the library we're using to look it up doesn't look in the places we need it to.

/jira-refresh

PR #1576, opened 2023-03-27
USHIFT-1008: Justify why healthz endpoints are not TLS verified

pacevedom:
Includes http client configuration instead of changing global http package variables.

Closes #<Issue Number>


dhellmann (commented):
/lgtm

PR #1577, opened 2023-03-27
USHIFT-1016: microshift crashes if LVM is not supported by host

copejon:

Closes USHIFT-1016

Check for the `vg` utility in $PATH and if not found, do not deploy topolvm.  Searching `vg` is sufficient since topolvm-node uses command line utilities found in $PATH to interact with LVM, so if `vg` isn't present, we know topolvm won't be able to function.

TODO: update documentation


pkg/config/lvmd/lvmd.go
```diff
@@ -140,3 +141,10 @@ func NewLvmdConfigFromFile(p string) (*Lvmd, error) {
 	l.Message = fmt.Sprintf("Read from %s", p)
 	return l, nil
 }
+
+func LvmSupported() bool {
+	if _, err := exec.LookPath("vg"); err == nil {
```
dhellmann, line 146: The command we run above is `vgs`.

pkg/components/storage.go
```diff
@@ -71,6 +71,12 @@ func startCSIPlugin(cfg *config.MicroshiftConfig, kubeconfigPath string) error {
 		}
 	)
 
+	// check for lvm utility 'vg', if not present, log and do not start CSI plugin
+	if ! lvmd.LvmSupported() {
+		klog.Warning("LVM utility 'vg' not found in $PATH, skipping CSI deployment")
```
dhellmann, line 76: This warning message assumes knowledge here about how the utility function in another package works. You could avoid leaking that knowledge by having LvmSupported() return the error and logging that directly.

pkg/config/lvmd/lvmd.go
copejon: fixed

PR #1584, opened 2023-03-28
Start using RHEL 9.2 in the devenv and edge

ggiguash:
* Switch the development environment to use RHEL 9.2 beta
* Enable osbuild configuration to use RHEL 9.2 beta channels
* Use golang that comes with the OS
* Fix the golang dependency in the MicroShift RPM spec file

Closes:
* [USHIFT-1023](https://issues.redhat.com//browse/USHIFT-1023)
* [OCPBUGS-10472](https://issues.redhat.com//browse/OCPBUGS-10472)

### DIFF
--- a/docs/devenv_setup.md
+++ b/docs/devenv_setup.md
@@ -4,7 +4,7 @@ It is recommended to review the current document and use the automation instruct
 
 ## Create Development Virtual Machine
 Start by downloading one of the supported boot images for the `x86_64` or `aarch64` architecture:
-* RHEL 9.1 from https://developers.redhat.com/products/rhel/download
+* RHEL 9.2 from https://developers.redhat.com/products/rhel/download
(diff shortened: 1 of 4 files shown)


ggiguash:
OCPBUGS-10472

dhellmann:
Are we confident about committing to a move to a beta version of the OS? I'd feel more comfortable making this optional so we can still develop easily on 9.1 if we have to.

ggiguash:
> Are we confident about committing to a move to a beta version of the OS? I'd feel more comfortable making this optional so we can still develop easily on 9.1 if we have to.

I think we may want to try it and roll back in case there is a stopper issue. It's the OS used for OCP 4.13, so it should also be stable enough for MicroShift.

dhellmann:
The test just passed and then failed again.

/override ci/prow/e2e-reboot

PR #1598, opened 2023-03-30
Fix LVM-not-detected log rendering

copejon:
Oneliner to make the warning log line render the error message.

### DIFF
--- a/pkg/components/storage.go
+++ b/pkg/components/storage.go
@@ -72,7 +72,7 @@ func startCSIPlugin(cfg *config.MicroshiftConfig, kubeconfigPath string) error {
 	)
 
 	if err := lvmd.LvmSupported(); err != nil {
-		klog.Warning("skipping CSI deployment: %v", err)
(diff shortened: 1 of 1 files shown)


pkg/components/storage.go
```diff
@@ -72,7 +72,7 @@ func startCSIPlugin(cfg *config.MicroshiftConfig, kubeconfigPath string) error {
 	)
 
 	if err := lvmd.LvmSupported(); err != nil {
-		klog.Warning("skipping CSI deployment: %v", err)
+		klog.Warningf("skipping CSI deployment: %v", err)
```
eggfoobar, line 75: I'm not 100% on how klog handles error types, but I think we should change the `%v` to `%w`. The go team has a nice little FAQ on some of these conventions. [see](https://github.com/golang/go/wiki/ErrorValueFAQ#i-am-already-using-fmterrorf-with-v-or-s-to-provide-context-for-an-error-when-should-i-switch-to-w)

pacevedom (commented):
/lgtm

PR #1600, opened 2023-03-30
USHIFT-950: add verification step to ensure etcd/vendor is kept up to date

dhellmann:
As we make changes to code in microshift that is vendored into etcd, we
need to ensure that we also update the vendored copy at the same time.

This change adds a Makefile target `vendor-etcd` that updates the
`etcd/vendor` directory and another target `verify-vendor-etcd` that fails
if there are any files added or modified after running `make vendor-etcd`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -287,3 +287,15 @@ vendor:
 		git mailinfo /dev/null /dev/stderr 2<&1- < $$p | git apply --reject || exit 1; \
 	done
 .PHONY: vendor
+
+# Update the etcd dependencies, including especially MicroShift itself.
+vendor-etcd:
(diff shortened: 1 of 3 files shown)


dhellmann:
I expect the verify job to fail with the first version of the PR. I will update it with the updated etcd/vendor files and remove the WIP flag after the first set of tests run.

dhellmann:
Without updating etcd/vendor the verify job fails with

```
ERROR:

You need to run 'make vendor-etcd' and commit the results to include
these files in the PR:

etcd/vendor/github.com/openshift/microshift/pkg/config/config.go
etcd/vendor/github.com/openshift/microshift/pkg/util/cert.go
```

PR #1613, opened 2023-04-03
USHIFT-1004: delete no longer used Dockerfile

pmtk:
Looking at PR that removed usage of the Dockerfile in CI, I think it was used for running MicroShift in container: https://github.com/openshift/release/pull/29745/files


pacevedom (approved):
/lgtm

PR #1614, opened 2023-04-03
USHIFT-1030: fix pylint finding issues

chiragkyal:

https://issues.redhat.com/browse/USHIFT-1030

This PR will

- Add `hack/verify-py.sh` script to create a virtual python environment in `_output/venv` and install all python dependency packages in that environment and run pylint. 
   - If pylint is already installed, it will run against all `.py` files. 
- Add `scripts/requirements.txt`, which lists all the dependency packages 
- Fix pylint finding issues and disabled a few rules
- Add `make verify-py` in the `verify` CI job


scripts/auto-rebase/handle-assets.py
```diff
@@ -19,18 +26,23 @@
 
 
 def merge_paths(pathl, pathr):
+    """Merge two paths into one"""
```
dhellmann, line 29: The first time I looked at this function I thought we should replace it with os.path.join, so adding a little detail about how it is different from that function would be useful.

```suggestion
    """Merge two paths into one by removing any '/' prefix from pathr and then appending it to pathl"""
```

scripts/auto-rebase/rebase.py
```diff
@@ -92,32 +110,40 @@ def make_sure_rebase_script_created_new_commits_or_exit(git_repo, base_branch):
...
+    considered functional, False otherwise.
+    """
     logging.info(f"Deciding if PR should be created by diffing against {base_branch} branch")
     diffs = git_repo.active_branch.commit.diff(base_branch)
     logging.info(f"Following files changed: {[ d.a_path for d in diffs ]}")
 
-    for d in diffs:
-        if 'scripts/auto-rebase/' in d.a_path:
-            logging.info(f" - {d.a_path} - ignoring")
+    for d_diff in diffs:
```
dhellmann, line 121: What rule led to this change?

dhellmann (commented):
Most of these changes look good. I had a couple of questions but nothing jumps out at me as problematic.

Some of these scripts are dev tools and are not tested in CI, so I wonder if we should apply the same approach we said we would for the bash linter? That would give us a chance to test some of them individually in case we miss an issue with visual inspection.

scripts/auto-rebase/rebase.py
chiragkyal: `C0103: Variable name "d" doesn't conform to snake_case naming style (invalid-name)` 

`diff_itr` or `d_diff` were two substitutes that came to my mind, and chose to use `d_diff`.

scripts/auto-rebase/handle-assets.py
chiragkyal: Thanks for the suggestion, addressed.

scripts/auto-rebase/rebase.py
```diff
@@ -138,66 +168,82 @@ def try_get_rebase_branch_ref_from_remote(remote, branch_name):
...
+            _extra_msgs.append(f"PR #{pull_req.number} for '{branch_name}' was already merged but rebase.sh produced results")
         else:
-            _extra_msgs.append(f"PR #{pr.number} for '{branch_name}' exists already but was closed")
+            _extra_msgs.append(f"PR #{pull_req.number} for '{branch_name}' exists already but was closed")
         return None
-    return pr
+    return pull_req
 
 
 def generate_pr_description(branch_name, amd_tag, arm_tag, prow_job_url, rebase_script_succeded):
```
chiragkyal, line 237: This function is complaining with : `W0613: Unused argument 'branch_name' (unused-argument)`, I believe `branch_name` is not required here and can be removed, unless we missed anything.

scripts/jira/cloner.py
```diff
@@ -277,48 +317,48 @@ def query_build(issue, user):
...
-        exit(0)
+        sys.exit(0)
 
     answer = ''
     if args.auto_accept:
         answer = 'y'
     while answer not in ['y', 'n']:
-        answer = input(f'Perform {len(actions)} non manual actions? [Y/N]').lower()
+        answer = input(f'Perform {len(actions_list)} non manual actions? [Y/N]').lower()
     if answer == 'n':
```
chiragkyal, line 356: If  I'm not wrong, `answer` is not a constant here, it's a variable. Still, pylint is thinking it as a constant and suggesting to use UPPER_CASE.
`C0103: Constant name "answer" doesn't conform to UPPER_CASE naming style (invalid-name)`

scripts/auto-rebase/rebase.py
```diff
@@ -350,6 +432,10 @@ def get_token(org, repo):
 
 
 def main():
```
chiragkyal, line 434: At Line#447, the suggestion is to change 
```python
REMOTE_DRY_RUN = False if try_get_env(DRY_RUN_ENV, die=False) == "" else True
```
line due to `R1719: The if expression can be replaced with 'not test' (simplifiable-if-expression)` issue.

I think this line can be changed to 

```python
REMOTE_DRY_RUN = bool(try_get_env(DRY_RUN_ENV, die=False))
```

is there any better approach to re-write this ?

scripts/auto-rebase/handle-assets.py
chiragkyal, line 1: We should rename the file to `handle_assets.py` to conform to snake_case naming style. Before doing that, I want to double-check the dependencies on this file. 

I see it will impact [here](https://github.com/openshift/microshift/blob/41bf2ec949e23e3942850ea86b376c238ccb8478/scripts/auto-rebase/rebase.sh#L659). Apart from this, is there any other place from where this file can get invoked?

scripts/release-notes/gen-ec-release-notes.py
chiragkyal, line 1: This file also needs to be renamed: `gen_ec_release_notes.py`. Is it safe to rename it?

chiragkyal:
Apart from there, there are a lot of 

- `C0301: Line too long (135/100) (line-too-long)`
- `W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)`

I think we can snooze them, because it may impact code readability with not so valuable code changes.

chiragkyal:
>Some of these scripts are dev tools and are not tested in CI, so I wonder if we should apply the same approach we said we would for the bash linter? That would give us a chance to test some of them individually in case we miss an issue with visual inspection.

For sure, so if I'm not wrong, the plan would be to merge one file at a time.

scripts/jira/cloner.py
```diff
@@ -277,48 +317,48 @@ def query_build(issue, user):
...
         print(f"Unable to retrieve issues: {e}")
-        exit(1)
+        sys.exit(1)
 
-    actions = []
+    actions_list = []
     for i in tqdm(range(len(query))):
         try:
-            issue = query[i]
+            issuee = query[i]
```
pmtk, line 337: What is reason for this change?

pmtk:
Before merging this needs be tested in rebase scenario - either manually running `rebase.py` or creating a dummy PR in `o/release` to rehearse it

scripts/jira/cloner.py
```diff
@@ -1,7 +1,23 @@
-import jira
+# pylint: disable=too-few-public-methods, broad-except
```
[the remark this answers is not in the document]
dhellmann: +1

scripts/jira/cloner.py
```diff
@@ -20,36 +37,41 @@ def is_original_issue(issue):
...
-    l = []
+    """Returns a list of clone issues for the given issue."""
+    lst = []
     for link in issue.fields.issuelinks:
         if link.type.name != 'Cloners':
             continue
         if hasattr(link, 'inwardIssue'):
-            l.append(connection.issue(link.inwardIssue.key))
-    return l
+            lst.append(connection.issue(link.inwardIssue.key))
```
dhellmann, line 46: I wonder if we actually want that naming rule applied. Maybe we should have a conversation about it on slack?

scripts/jira/cloner.py
dhellmann: I wonder if line 311 is making it think it's a constant? Maybe that's another rule we want to turn off?

scripts/release-notes/gen-ec-release-notes.py
dhellmann: Sure, that's fine.

scripts/jira/cloner.py
chiragkyal: It's complaining for both 311 and 313 lines.  And [C0103](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/invalid-name.html) is a very broad check for all types of naming convention, so if we want to turn it off, we can do that for only these two lines. Like this:


```python
    answer = '' #pylint: disable=invalid-name
    if args.auto_accept:
        answer = 'y' #pylint: disable=invalid-name
    while answer not in ['y', 'n']:
        answer = input(f'Perform {len(actions_list)} non manual actions? [Y/N]').lower()
    if answer == 'n':
        sys.exit(0)

```

scripts/auto-rebase/rebase.py
chiragkyal: Yeah, that's also better. Addressed.

scripts/auto-rebase/handle-assets.py
```diff
@@ -19,18 +26,23 @@
 
 
 def merge_paths(pathl, pathr):
+    """Merge two paths into one by removing any '/' prefix from pathr and then appending it to pathl"""
```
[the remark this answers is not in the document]
chiragkyal: Changed this to :

```
    """
    Merge two paths depending upon following conditions:
    - If `pathr` is absolute (starts with `/`), then discard the leading `/` and return rest `pathr`.
    - If `pathr` is relative, then return `pathl/pathr`.
    """
```

scripts/auto-rebase/handle-assets.py
```diff
@@ -89,26 +106,28 @@ def handle_file(file, dst_dir="", src_prefix=""):
     copy(src, dst)
 
 
-def handle_dir(dir, dst_dir="", src_prefix=""):
-    dst = merge_paths(dst_dir, dir['dir'])
-    new_src_prefix = merge_paths(src_prefix, dir['src'] if "src" in dir else "")
+def handle_dir(dir_, dst_dir="", src_prefix=""):
```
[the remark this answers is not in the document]
chiragkyal: That's right, `dir` is a Python keyword and conflicts with the variable name.

scripts/auto-rebase/handle-assets.py
```diff
@@ -89,26 +106,28 @@ def handle_file(file, dst_dir="", src_prefix=""):
...
+        handle_file(file, dst, new_src_prefix)
 
-    for d in dir.get('dirs', []):
-        handle_dir(d, dst, new_src_prefix)
+    for sub_dir in dir_.get('dirs', []):
+        handle_dir(sub_dir, dst, new_src_prefix)
 
 
 def main():
+    """Main function for handling assets based on the recipe file."""
```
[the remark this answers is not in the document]
chiragkyal: Checked a few examples in [PEP 257 – Docstring Conventions](https://peps.python.org/pep-0257/#one-line-docstrings), and most of them end with `.`  So, I've added `.` for all other docstring to be consistent

scripts/auto-rebase/handle-assets.py
```diff
@@ -118,7 +137,9 @@ def main():
         sys.exit(1)
 
     recipe_filepath = "./scripts/auto-rebase/assets.yaml"
-    recipe = yaml.load(open(recipe_filepath).read(), Loader=Loader)
+    with open(recipe_filepath, encoding='utf-8') as recipe_file:
+        recipe = yaml.load(recipe_file.read(), Loader=Loader)
```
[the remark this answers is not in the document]
chiragkyal: May be https://peps.python.org/pep-0343/ could be a good read

scripts/auto-rebase/presubmit.py
```diff
@@ -16,29 +24,33 @@
 RECIPE_FILEPATH = "./scripts/auto-rebase/assets.yaml"
 
 
-def build_assets_filelist_from_asset_dir(dir, prefix=""):
-    dir_path = os.path.join(prefix, dir['dir'])
-    return ([os.path.join(dir_path, f['file']) for f in dir.get('files', [])] +
+def build_assets_filelist_from_asset_dir(asset_dir, prefix=""):
+    """Recursively builds a list of assests filepaths from an asset directory."""
```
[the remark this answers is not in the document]
chiragkyal: Good catch, addressed

scripts/auto-rebase/presubmit.py
```diff
@@ -16,29 +24,33 @@
...
+    return ([os.path.join(dir_path, f['file']) for f in asset_dir.get('files', [])] +
             reduce(lambda x, y: x+y,
-                   [build_assets_filelist_from_asset_dir(subdir, dir_path) for subdir in dir.get('dirs', [])],
+                   [build_assets_filelist_from_asset_dir(subdir, dir_path) for subdir in asset_dir.get('dirs', [])],
                    []))
 
 
 def build_assets_filelist_from_recipe(recipe):
-    return reduce(lambda x, y: x+[y] if type(y) == str else x+y,
+    """Builds a list of assests filepaths from a recipe file."""
```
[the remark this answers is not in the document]
chiragkyal: lol, typo

scripts/auto-rebase/rebase.py
```diff
@@ -92,32 +110,40 @@ def make_sure_rebase_script_created_new_commits_or_exit(git_repo, base_branch):
...
+        logging.info(f" - File {d_diff.a_path} is considered functional")
         return True
 
     return False
 
 
 def get_remote_with_token(git_repo, token, org, repo):
+    """
+    Returns the Git remote for the given repository
+    using the provided personal access token.
```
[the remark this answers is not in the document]
chiragkyal: Changed to 
```
    """
    Returns the Git remote for the given repository using
    the provided installation (or personal) access token.
    """
```

scripts/auto-rebase/rebase.py
```diff
@@ -138,66 +168,82 @@ def try_get_rebase_branch_ref_from_remote(remote, branch_name):
...
         logging.info(f"Branch '{branch_name}' already exists on remote")
         return matching_remote_refs[0]
 
+    return None
 
 def is_local_branch_based_on_newer_base_branch_commit(git_repo, base_branch_name, remote_branch_name, local_branch_name):
     """
     Compares local and remote rebase branches by looking at their start on base branch.
-    Returns True if local branch is starts on newer commit and needs to be pushed to remote, otherwise False.
+    Returns True if local branch is starts on newer commit and needs to be pushed to remote,
```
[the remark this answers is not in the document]
chiragkyal: Done

scripts/auto-rebase/rebase.py
```diff
@@ -214,32 +260,43 @@ def generate_pr_description(branch_name, amd_tag, arm_tag, prow_job_url, rebase_
...
+    pull_req.update()  # arm64 release or prow job url might've changed
+    logging.info(f"Updated PR #{pull_req.number}: {title}\n{desc}")
 
 
-def post_comment(pr, comment=""):
+def post_comment(pull_req, comment=""):
+    """
+    Posts a comment on a GitHub pull request.
+    If there are any extra messages in the global `_extra_msgs` list,
+    they will be appended to the comment.
```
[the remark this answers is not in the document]
chiragkyal: Changed to 
```
    """
    Posts a comment on a GitHub pull request with
    the contents of the global `_extra_msgs` list.
    """
```

scripts/auto-rebase/rebase.py
```diff
@@ -56,33 +67,40 @@ def try_get_env(var_name, die=True):
...
-    result = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True)
+    result = subprocess.run(
+        args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True, check=False)
     logging.info(f"Return code: {result.returncode}. Output:\n" +
                  "==================================================\n" +
                  f"{result.stdout}" +
                  "==================================================\n")
     end = timer() - start
-    logging.info(f"Script returned code: {result.returncode}. It ran for {end/60:.0f}m{end%60:.0f}s.")
+    logging.info(f"Script returned code: {result.returncode}. " +
```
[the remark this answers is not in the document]
chiragkyal: There are still a lot of `line-too-long` issues, and I think we should turn this rule off. Because it's reducing code readability with not so impactful code changes. WDYT?

scripts/jira/cloner.py
```diff
@@ -277,48 +317,48 @@ def query_build(issue, user):
...
-        query = connection.search_issues(jql_str=jql_query, maxResults=9999)
+        query = conn.search_issues(jql_str=jql_query, maxResults=9999)
         print(f"Scanning {len(query)} issues")
     except Exception as e:
         print(f"Unable to retrieve issues: {e}")
-        exit(1)
+        sys.exit(1)
 
-    actions = []
+    actions_list = []
```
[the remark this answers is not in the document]
chiragkyal: [C0103](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/invalid-name.html) is a very broad check for all types of naming conventions; I think we should keep it.

scripts/jira/cloner.py
chiragkyal: `W0621: Redefining name 'issue' from outer scope (line 338) (redefined-outer-name)`

Because `issue` is getting used in most of the function definitions. So, either change it in `main` function or in all the functions where `issue` is used. I chose to change in `main` function for simplicity and less line of change.

scripts/jira/cloner.py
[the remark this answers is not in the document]
chiragkyal: The main reason for this change is because `actions` and `query` are conflicting with the local variables defined in other functions. We need to either change in `main` function or in all other functions where it has been used.

scripts/auto-rebase/rebase.py
chiragkyal: Reverted the changes and added `--variable-naming-style=any` to mute snake_case style.

scripts/jira/cloner.py
chiragkyal: Added `#pylint: disable=invalid-name` for these two line to turn off the rule. It seems like a false positive.

scripts/auto-rebase/rebase.py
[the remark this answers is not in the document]
chiragkyal: As discussed, disabled `line-too-long` check

scripts/jira/cloner.py
chiragkyal: Reverted the changes and turned off the rule.

hack/verify-py.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+ROOTDIR=$(git rev-parse --show-toplevel)
+REQ_FILE=${ROOTDIR}/requirements.txt
+VENV="/tmp/venv"
+PYLINT="pylint"
+
+if ! command -v ${PYLINT} &>/dev/null; then
+    
+    check="$(python3 -m pip install -r ${REQ_FILE} 2>&1 | { grep 'Permission denied' || true; })"
```
dhellmann, line 10: Why not just use the virtualenv all the time?

requirements.txt
```diff
@@ -0,0 +1,7 @@
+PyGithub==1.56
```
dhellmann, line 1: Over time, especially as we set up e2e tests, we are likely to have different requirements files for different purposes. This one should probably move to hack as something like `hack/requirements.txt` so we can differentiate it from `validate-microshift/requirements.txt` in the future.

dhellmann (commented):
I had 2 comments, but nothing to hold this up. They can be addressed in another PR, unless you want to fix them as part of addressing other comments.

I want to leave this open for @pmtk to approve, since he is more familiar with the code being changed and had previous review comments.

hack/verify-py.sh
chiragkyal: The pylint installation worked fine directly in RHEL dev VM, without virtualenv. And as virtualenv will put a little overhead in the VM, I think we should do that only for CI.

requirements.txt
pmtk: I think that `script/auto-rebase/rebase.py` and (if we decide so) `e2e/tests.py` should have separate `requirements.txt` in their own dirs. Does it make sense from how python project operate?

pmtk (commented):
Let's get "venv in CI and locally" and "/requirements.txt" topics sorted out and I think we're good to go

hack/verify-py.sh
chiragkyal: Yes, we need all the packages listed in `requirements.txt`  for pylint to run successfully, otherwise it will throw `(import-error)`. 

For having venv locally, we could setup that in `_output/` folder and install all the dependencies there. In that case, we should run all python screipts from that venv, and install further dependencies there only. I should not have 2 copies of dependencies.

requirements.txt
chiragkyal: I am unsure what is the recommended way, however we need all the dependencies installed before running pylint. We may have to loop over all the `requirements.txt` files if we choose to have multiple.

hack/verify-py.sh
chiragkyal: As discussed in Slack, we may later on, create a separate PR to run all .py scripts from venv rather than requiring to have deps installed on system (usually /usr/, and not $HOME...)

dhellmann: It's not really recommended practice to use pip to install things directly into the system python space, so I would prefer we use virtualenv all the time.

chiragkyal: Okay, I've updated the script to create a virtual environment all the time.

PR #1622, opened 2023-04-05
USHIFT-1052: Speed up greenboot check by checking Pod restarts in parallel

pmtk:


packaging/greenboot/microshift-running-check.sh
```diff
@@ -119,9 +119,22 @@ for i in ${!PODS_NS_LIST[@]}; do
...
 
 # Verify that MicroShift core pods are not restarting
+declare -A pidNamespace
 for i in ${!PODS_NS_LIST[@]}; do
     CHECK_PODS_NS=${PODS_NS_LIST[$i]}
 
     echo "Checking pod restart count in the '${CHECK_PODS_NS}' namespace"
-    namespace_pods_not_restarting ${CHECK_PODS_NS}
+    namespace_pods_not_restarting ${CHECK_PODS_NS} &
+    pid=$!
```
ggiguash, line 128: Is there no race-condition here for the PID assignment? The function called in the background spawns many sub-shells, so how do we know that we're getting the "right" PID?

packaging/greenboot/microshift-running-check.sh
```diff
@@ -119,9 +119,22 @@ for i in ${!PODS_NS_LIST[@]}; do
...
+    pidNamespace["${pid}"]="${CHECK_PODS_NS}"
+done
+
+set +e
+for pid in "${!pidNamespace[@]}"; do
+    wait "${pid}"
```
ggiguash, line 134: Can we just wait for all the sub-shells to exit? We should not need to disable exit-on-error - let's just fail the script if the background shell returns an error?

ggiguash: We should get an output about the failed background process, right?

dhellmann (commented):
Since this is production code, it would be good to link the PR to a jira ticket

packaging/greenboot/microshift-running-check.sh
pmtk: AFAIK `$!` is for most recent background process and we're not spawning any of those in `namespace_pods_not_restarting`. I did a small test which would suggest that we would be safe anyway:
```
+ pids=()
+ pids+=($!)
+ sleep 2
```

packaging/greenboot/microshift-running-check.sh
pmtk: > We should get an output about the failed background process from bash, right?

You mean something like this?
```
[1]+  Exit 1 
```
I think it's only in interactive sessions

PR #1624, opened 2023-04-05
USHIFT-662: remove user directory configuration file

dhellmann:
Simplify the configuration approach by removing the search in multiple
places for a configuration file. Use only the file in
/etc/microshift/config.yaml.

### DIFF
--- a/docs/howto_config.md
+++ b/docs/howto_config.md
@@ -1,5 +1,5 @@
 # Configuration
-The MicroShift configuration file must be located at `~/.microshift/config.yaml` (user-specific) and `/etc/microshift/config.yaml` (system-wide), while the former takes precedence if it exists. A sample `/etc/microshift/config.yaml.default` configuration file is installed by the MicroShift RPM and it can be used as a template when customizing MicroShift.
+The MicroShift configuration file must be located at `/etc/microshift/config.yaml`. A sample `/etc/microshift/config.yaml.default` configuration file is installed by the MicroShift RPM and it can be used as a template when customizing MicroShift.
 
 The format of the `config.yaml` configuration file is as follows.
 
--- a/etcd/vendor/github.com/openshift/microshift/pkg/config/files.go
+++ b/etcd/vendor/github.com/openshift/microshift/pkg/config/files.go
@@ -10,10 +10,9 @@ import (
 )
 
 const (
-	DefaultUserConfigFile   = "~/.microshift/config.yaml"
-	defaultUserDataDir      = "~/.microshift/data"
-	DefaultGlobalConfigFile = "/etc/microshift/config.yaml"
-	defaultGlobalDataDir    = "/var/lib/microshift"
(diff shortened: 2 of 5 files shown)


eggfoobar:
> /cc @eggfoobar
> 
> I used a simple const in this version, but when I went to look at removing the logic for finding the data dir some of the tests tried to change the value, so I could be talked into exposing a function as the interface for this value, too. Let me know what you think.

After looking at the code, it doesn't seem like the data dir is designed to be configured by the user, is that correct? If that is the case, then I would be on board with keeping it simple and making it a constant like you planned. Since it seems the only thing that test is checking for is that `validateNodeName` is working, maybe doing a little dependency injection and passing the dir there would be the best approach? Then the test would just need a small change but remain the same for the most part. What do you think?

```go
// function change
fun (c *Config) validateNodeName(isDefaultNodeName bool, dataDir string)
``` 

```go
// test change
func setupSuiteDataDir(t *testing.T) (string, func()) {
	tmpdir, err := os.MkdirTemp("", "microshift")
	if err != nil {
		t.Errorf("failed to create temp dir: %v", err)
	}
	return tmpdir, func() {
		os.RemoveAll(tmpdir)
	}
}
```

dhellmann:
> > /cc @eggfoobar
> > I used a simple const in this version, but when I went to look at removing the logic for finding the data dir some of the tests tried to change the value, so I could be talked into exposing a function as the interface for this value, too. Let me know what you think.
> 
> After looking at the code, it doesn't seem like the data dir is designed to be configured by the user, is that correct? If that is the case, then I would be on board with keeping it simple and making it a constant like you planned. Since it seems the only thing that test is checking for is that `validateNodeName` is working, maybe doing a little dependency injection and passing the dir there would be the best approach? Then the test would just need a small change but remain the same for the most part. What do you think?
> 
> ```go
> // function change
> fun (c *Config) validateNodeName(isDefaultNodeName bool, dataDir string)
> ```
> 
>  ```go
> // test change
> func setupSuiteDataDir(t *testing.T) (string, func()) {
> 	tmpdir, err := os.MkdirTemp("", "microshift")
> 	if err != nil {
> 		t.Errorf("failed to create temp dir: %v", err)
> 	}
> 	return tmpdir, func() {
> 		os.RemoveAll(tmpdir)
> 	}
> }
> ```

Yes, something like that might work. I opened a separate ticket for that, so we can proceed with this PR and I'll do the refactoring in the next one.

ggiguash:
@dhellmann , can you also update the docs [here](https://github.com/openshift/microshift/blob/1fc4f98f49d869425483605a1cc7f0588ba3278b/docs/howto_config.md?plain=1#L2)?

dhellmann:
> @dhellmann , can you also update the docs [here](https://github.com/openshift/microshift/blob/1fc4f98f49d869425483605a1cc7f0588ba3278b/docs/howto_config.md?plain=1#L2)?

Updated in the most recent draft.

PR #1631, opened 2023-04-06
OCPBUGS-11512: avoid requiring jq during build

dhellmann:
Use jq during rebase to extract version information from the release
files so it can be read directly by make during the build process.


dhellmann:
On second thought, this isn't going to be one we can cherry-pick as-is. The version numbers will be wrong. I'll prepare another PR instead.

PR #1635, opened 2023-04-07
WIP POC MicroShift e2e using robot framework

pmtk:
See most recent commit


e2e-robot/Makefile
```diff
@@ -0,0 +1,43 @@
+
+.PHONY: setup dry-run with-filter with-tag all
+
+setup:
+	python3 -m venv .venv && \
+	source .venv/bin/activate && \
+	python3 -m pip install -r requirements.txt
```
dhellmann, line 7: You can run `.venv/bin/python3 -m pip` without activating the environment.

e2e-robot/Makefile
```diff
@@ -0,0 +1,43 @@
...
+
+dry-run:
+	source .venv/bin/activate && \
+	robot \
```
dhellmann, line 11: Same here, if you run `.venv/bin/robot` it should work without activating the environment.

e2e-robot/Makefile
```diff
@@ -0,0 +1,43 @@
...
+		--dryrun \
+		--outputdir ../_output/e2e-$$(date +'%Y%m%d-%H%M%S') \
+		./tests/microshift.robot
+
+with-filter:
+	source .venv/bin/activate && \
+	robot \
+		-v USHIFT_IP:192.168.122.166 \
```
dhellmann, line 19: The IP could come from running `./scripts/devenv-builder/manage-vm.sh ip`

e2e-robot/tests/microshift.robot
```diff
@@ -0,0 +1,138 @@
...
+Check HTTP Response
+    [Arguments]    ${result}
+    Log    ${result.stdout}
+    Log    ${result.stderr}
+    Should Be Equal As Integers    ${result.rc}    0
+    Should Match Regexp    ${result.stdout}    HTTP.*200
+    Should Match    ${result.stdout}    *Hello MicroShift*
+
+Create Pod With PVC
+    Run With Kubeconfig    oc create -f ../e2e/tests/assets/pod-with-pvc.yaml
```
dhellmann, line 109: Are these paths relative to the directory where someone runs robot, or to the location of the `.robot` file?

pmtk: I think it's relative to current working dir which, when running `make` inside `e2e-robot`, is `e2e-robot/`

e2e-robot/Makefile
pmtk: Yeah, it could also be passed via ENV vars or other means. I just hardcoded it to reduce typing when PoCing this :)
Not tying it to other scripts can result in scalability - run whatever test suite you want against any VM in parallel.

In regards to ostree upgrade/rollback I have this idea (not sure how it'll pan out):
if we go with "nested virt" approach, then we could setup a job (sanity periodic most likely) that would spawn multiple nested VMs and do some testing in parallel.
Like this mad idea: build 4.14.0 **base image** (as in qemu backing image, so new VMs can be spawned quickly) and do upgrades (rather by installing new RPMs than preparing whole new image)
```
 .0 ----
 |  \    \
.1   .2   .3
 |    |
.2   .3
 |
.3
```

PR #1636, opened 2023-04-07
USHIFT-1069: add script for automating ticket status based on branch

dhellmann:


scripts/jira/start_ticket.py
```diff
@@ -0,0 +1,217 @@
...
+        field_details = fields_by_name[name]
+        return ticket.update(fields={
+            field_details['id']: value,
+        })
+    return get_field_value, set_field_value
+
+
+def get_active_sprint(server):
+    """Return the active sprint for the USHIFT project."""
+    valid_sprints = server.sprints(SCRUM_BOARD)
```
pacevedom, line 98: You can include `state = active` and that should save the filtering below. There is a maxResults that defaults to 50, which for us means 150 weeks, so it will stop working next year?

scripts/jira/start_ticket.py
```diff
@@ -0,0 +1,217 @@
...
+        action='store_const',
+        const='Code Review',
+        help='mark the ticket as ready for code review',
+    )
+    parser.add_argument(
+        '--in-progress',
+        dest='status',
+        action='store_const',
+        const='In Progress',
+        help='mark the ticket as ready for code review',
```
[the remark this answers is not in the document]
dhellmann: Updated

scripts/jira/start_ticket.py
dhellmann: Does the state filter fix that or are you saying I need to ask for more results?

dhellmann: The USHIFT board includes sprints from multiple projects, so we still need the loop. I guess I can remove the date checking portion, so I've done that.

pacevedom (approved):
/lgtm

PR #1637, opened 2023-04-08
Multinode: ovn-kubernetes

zshi-redhat:
This commit is the first step to introduce multinode capability in microshift, focusing on networking part.

- add a hidden flag --master
- split ovnk manifests to common, single-node and multi-node
- apply multi-node manifests for ovnk
- change cluster mtu based on multinode flag

To run microshift in multinode mode:
$ microshift run --master


pkg/config/config.go
```diff
@@ -33,6 +33,8 @@ type Config struct {
 	// Internal-only fields
 	Ingress      IngressConfig `json:"-"`
 	userSettings *Config       `json:"-"` // the values read from the config file
+
+	MultiNode MultiNode `json:"multiNode"` // the value read from commond line
```
dhellmann, line 37:
```suggestion
	MultiNode MultiNode `json:"-"` // the value read from commond line
```

I think we will want to hide this structure from the json marshaller, so that the show-config command does not expose it.

pkg/cmd/run.go
```diff
@@ -33,19 +33,31 @@ func NewRunMicroshiftCommand() *cobra.Command {
...
-				return err
-			}
-			return RunMicroshift(cfg)
-		},
+	}
+
+	var master bool
+
+	flags := cmd.Flags()
+	flags.BoolVar(&master, "master", false, "act as control plane node in multinode mode")
```
dhellmann, line 41: Using a command line flag means the instructions for running in the configuration for certification won't be able to use the default systemd unit. That's not necessarily a bad thing. For example, it means it will be easier to not support multi-node mode. But it's something we will need to be aware of and handle in our automation for the certification test job.

pkg/config/node.go
```diff
@@ -19,6 +19,13 @@ type Node struct {
 	NodeIP string `json:"nodeIP"`
 }
 
+type MultiNode struct {
+	Enabled bool `json:"enabled"`
+	// only one master node is supported, no control plane HA
+	Master  string   `json:"master"`
+	Workers []string `json:"workers"`
```
dhellmann, line 26: What populates the list of workers? Does the user need to know them in advance?

fzdarsky:
Can we call this “controlplane” instead of “master”, a language that Red Hat (and the K8s community) try to deprecate?

Also, how about we revert to the original `—roles=controlplane,node` approach, which is more flexible than a boolean flag?

pkg/config/node.go
zshi-redhat: I should probably remove it in this commit. I was thinking to use it for generating kubelet kubeconfig for each extra worker node. It could be handled in a follow up PR if needed when adding multi worker nodes support in microshift main binary.

pkg/cmd/run.go
zshi-redhat: @dhellmann understood, good point! 
I could change it to a hidden config file option instead if that sounds better.
I'll keep it for now, and change it anytime later if you prefer.

zshi-redhat:
> Can we call this “controlplane” instead of “master”, a language that Red Hat (and the K8s community) try to deprecate?
> 
I will use control plane.

> Also, how about we revert to the original `—roles=controlplane,node` approach, which is more flexible than a boolean flag?

sounds good, will update to use `--roles`

pkg/cmd/run.go
dhellmann: Yes, let's leave it as a command line option for now. It will be easy to change it later to make the automation more convenient, if we need to.

dhellmann:
> > Can we call this “controlplane” instead of “master”, a language that Red Hat (and the K8s community) try to deprecate?
> 
> I will use control plane.
> 
> > Also, how about we revert to the original `—roles=controlplane,node` approach, which is more flexible than a boolean flag?
> 
> sounds good, will update to use `--roles`

We talked about using the standard kubelet binary for the worker. If we do that, do we need MicroShift to have separate roles?

zshi-redhat:
> > > Can we call this “controlplane” instead of “master”, a language that Red Hat (and the K8s community) try to deprecate?
> > 
> > 
> > I will use control plane.
> > > Also, how about we revert to the original `—roles=controlplane,node` approach, which is more flexible than a boolean flag?
> > 
> > 
> > sounds good, will update to use `--roles`
> 
> We talked about using the standard kubelet binary for the worker. If we do that, do we need MicroShift to have separate roles?

Looking at the [previous implementation](https://github.com/openshift/microshift/commit/b6a32cf41e0722246b78cdb11f366d03c3e497ac), I think `roles` is to distinguish the `control plane` and `node` in microshift main binary so that:
1. `node` role only runs kubelet service
2. `controlplane` role runs service other than kubelet
3. `controlplane` + `node` role runs all microshift services

If we use standard kubelet binary for extra worker nodes, `roles` may not be needed since microshift just needs to support `controlplane` + `node` role. `controlplane` role is not what we wanted to support.

zshi-redhat:
Still needs to fix the assets structure in scripts/auto-rebase/assets.yaml which causes ci/prow/verify job fail.
@pmtk Do we have script for generating scripts/auto-rebase/assets.yaml or is it manual?

pkg/config/node.go
zshi-redhat: removed `Workers` and move the `Multinode` structure to a separate file pkg/config/multinode.go

pkg/config/multinode.go
```diff
@@ -0,0 +1,22 @@
...
+// MultinodeConfig populates multinode configurations to Config.MultiNode
+func (c *Config) MultinodeConfig(enabled bool) {
+	if !enabled {
+		return
+	}
+	c.MultiNode.Enabled = enabled
+	c.MultiNode.Controlplane = c.Node.NodeIP
+
+	// Use controlplane node IP as APIServer backend (instead of 10.44.0.0)
+	c.ApiServer.AdvertiseAddress = c.Node.NodeIP
```
zshi-redhat, line 19: @pacevedom  This overrides the advertiseAddress IP of APIServer for multi-node since the service cdir cannot be accessed by extra worker nodes. Do you see any issue with API certificate by doing so?

pacevedom: Yes, we do need some changes for this to work. Let me create the PR.

pacevedom: I checked and everything is already there, so no additional changes required.

pkg/cmd/run.go
```diff
@@ -33,19 +33,29 @@ func NewRunMicroshiftCommand() *cobra.Command {
...
-				return err
-			}
-			return RunMicroshift(cfg)
-		},
+	}
+
+	var controlplane bool
+
+	flags := cmd.Flags()
+	flags.BoolVar(&controlplane, "controlplane", false, "act as both control plane and node in multinode mode")
```
dhellmann, line 41: If we're not doing "modes" this can be "multinode" instead of "controlplane".

etcd/vendor/github.com/openshift/microshift/pkg/config/multinode.go
```diff
@@ -0,0 +1,22 @@
...
+
+type MultiNode struct {
+	Enabled bool `json:"enabled"`
+	// only one controlplane node is supported
+	// IP address of control plane node
+	Controlplane string `json:"controlplane"`
+}
+
+// MultinodeConfig populates multinode configurations to Config.MultiNode
+func (c *Config) MultinodeConfig(enabled bool) {
```
dhellmann, line 11:
```suggestion
func (c *Config) MultiNode(enabled bool) {
```

It might be good to have this function return a new Config instance, so it's clear that any transformation is one way. We can do that in another PR, though.

dhellmann (commented):
A couple of small suggestions, but there's nothing to hold this up.

I'll leave this for @pacevedom to approve.

/assign @pacevedom

pacevedom (approved):
/lgtm

zshi-redhat:
seeing permission denied failures in the CI job:
```
oc logs -n openshift-ingress pod/router-default-d55c4fb97-9sfnc router
exec container process `/usr/bin/openshift-router`: Permission denied
```

rebased against master.

ggiguash:
> seeing permission denied failures in the CI job:
> 
> ```
> oc logs -n openshift-ingress pod/router-default-d55c4fb97-9sfnc router
> exec container process `/usr/bin/openshift-router`: Permission denied
> ```
> 
> rebased against master.

That's a problem in our CI configuration, which manifests itself in all the PRs.

PR #1639, opened 2023-04-09
[WIP]: USHIFT-1061: auto config generator

eggfoobar:

This PR adds automatic generation of our config from the struct, we then are able to use go generate to update both our default yaml as well as our document yaml to keep things in sync with minimal effort and the source of truth being our source struct.


pkg/config/etcd.go
```diff
@@ -11,7 +11,7 @@ const (
 type EtcdConfig struct {
 	// Set a memory limit on the etcd process; etcd will begin paging
 	// memory when it gets to this value. 0 means no limit.
-	MemoryLimitMB uint64 `json:"memoryLimitMB"`
+	MemoryLimitMB uint64 `json:"memoryLimitMB" config:"comment,default=0"`
```
dhellmann, line 14: I see kubebuilder directives to set defaults elsewhere, but this uses a different approach. What's the difference?

vendor/github.com/getkin/kin-openapi/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
dhellmann, line 1: If we run the generate step as part of verifying changes in CI, but not as part of building the RPM, could we avoid vendoring the dependencies for the tool entirely? We might want to skip vendoring these dependencies even if we make the separate go module mentioned in an earlier comment.

hack/config_crd.go
```diff
@@ -0,0 +1,19 @@
+// +kubebuilder:object:generate=true
+// +groupName=config.microshift.openshift.io
+// +versionName=v1
+
+// This file only exists to help translate our config struct
+// to an OpenAPIV3 spec via controller-gen, it should not be exported
```
dhellmann, line 6: What brings controller-gen into the build environment?

packaging/microshift/config.yaml
```diff
@@ -1,35 +1,37 @@
+#!! Do Not Edit
+#!! This is a generated file
+apiServer:
+    # Kube apiserver advertise address to work around the certificates issue when requiring external access using the node IP. This will turn into the IP configured in the endpoint slice for kubernetes service. Must be a reachable IP from pods. Defaults to service network CIDR first address.
```
dhellmann, line 4: We're going to want to review the comments we are exposing. We need a way to have a comment like this, which is developer facing, separate from the help text we give to admins. The "...work around the certificates issue..." part shouldn't be exposed.

dhellmann (commented):
I like the direction this is going better than earlier versions. I have a few questions/comments inline about details. The actual YAML generator is pretty small, but could use some comments as an aid to future maintainers.

go.mod
```diff
@@ -229,6 +229,13 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
 
+require github.com/getkin/kin-openapi v0.115.0
```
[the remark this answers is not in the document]
eggfoobar: Sorry Doug, was still making changes, had investigated an approach that would remove this dependency and it proved to be fruitful. Side note, is `yaml-gen` good or should we change it to `config-gen`?

pkg/config/etcd.go
eggfoobar: Cleaned those up, we should be good to rely on just kubebuilder comments now.

hack/config_crd.go
eggfoobar: Currently, that would be just doing a simple `go install` similar to the [existing lichen](https://github.com/openshift/microshift/blob/main/scripts/fetch_tools.sh#L153), I'll add that in this PR

packaging/microshift/config.yaml
eggfoobar: Agreed

go.mod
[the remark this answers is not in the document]
dhellmann: I like config-gen, but I'm not hung up on it.

hack/config_crd.go
eggfoobar: Added controller-gen to the fetch script, and it should be pulled in during the config-verify now

hack/yaml-gen/util.go
```diff
@@ -0,0 +1,165 @@
...
+}
+func parseScalarJSONExample(rawJson *v1ext.JSON) string {
+	results := parseScalarJSONValue(rawJson)
+	if len(results) == 0 {
+		return ""
+	}
+	return toYaml(results)
+}
+
+// Best effort to parse raw json examples, controller-gen will break if the examples
```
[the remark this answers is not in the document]
eggfoobar: Still working on my inner author.

hack/yaml-gen/util.go
```diff
@@ -0,0 +1,165 @@
...
+func parseArrayJSONValue(rawJson *v1ext.JSON) []*yaml.Node {
+	if rawJson == nil {
+		return nil
+	}
+	var arrayContainer []interface{}
+	if err := json.Unmarshal(rawJson.Raw, &arrayContainer); err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
eggfoobar: Yeah that's what I was thinking, if there is bad input the controller-gen will fail to parse it before we do. In the case of mismatch, I updated the panic messages to include type information for easier debugging and a comment on maybe why we want to just panic in these functions. It does keep the code a little cleaner. If we aren't comfortable with it, we can bubble up the error with out a problem, the end result is the same since these functions will execute primarily in the template engine.

hack/yaml-gen/opts.go
```diff
@@ -0,0 +1,108 @@
...
+	case opt.fileInput != "":
+		f, err := os.Open(opt.fileInput)
+		if err != nil {
+			return err
+		}
+		dataReader = f
+	default:
+		dataReader = os.Stdin
+	}
+	defer dataReader.Close()
```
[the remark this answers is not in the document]
eggfoobar: Good call out, I don't see why we need to close it, changed it

hack/config-gen/README.md
```diff
@@ -0,0 +1,77 @@
...
+  -a, --api-output string              output path for openapi spec if desired
+  -f, --file string                    default is stdin
+  -h, --help                           help for yaml-gen
+  -o, --output string                  output path, default is stdout
+  -t, --template string                template file to use
+```
+
+Use as a go generate command example
+```go
+//go:generate sh -c "controller-gen crd paths=../../hack/yaml-gen/configcrd output:stdout | go run -mod vendor ../../hack/yaml-gen -a ../../cockpit-plugin/packaging/config-openapi-spec.json -o ../../packaging/microshift/config.yaml"
```
dhellmann, line 29: These commands use the old name, too

hack/config-gen/README.md
```diff
@@ -0,0 +1,77 @@
+# YAML-Gen
```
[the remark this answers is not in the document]
eggfoobar: Oops, updated, thanks!

packaging/microshift/config.yaml
```diff
@@ -1,35 +1,37 @@
+#!! Do Not Edit
```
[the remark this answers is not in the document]
eggfoobar: Sounds good, removed it

hack/yaml-gen/util.go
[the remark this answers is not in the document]
eggfoobar: Sounds like a plan, easy enough change if it comes to it, I did some quick testing and with the updated error messages, it should be clear where something went wrong.

dhellmann (commented):
I like this version. We mentioned a few improvements in comment threads, but those can come later. Remove the hold when you're ready.

/lgtm

dhellmann:
/hold

give @pmtk a chance to look at this

PR #1642, opened 2023-04-10
USHIFT-1050:  Add hadolint : Linter for Dockerfiles and Containerfiles

chiragkyal:

Closes https://issues.redhat.com/browse/USHIFT-1050

This PR adds [hadolint](https://github.com/hadolint/hadolint) as a linter for Dockerfiles and Containerfiles

### DIFF
--- a/packaging/images/openshift-ci/Dockerfile.test-runtime
+++ b/packaging/images/openshift-ci/Dockerfile.test-runtime
@@ -1,14 +1,15 @@
-# hadolint global ignore=DL3002,DL4006,SC3037,SC2086
 FROM registry.access.redhat.com/ubi8/ubi:latest
 USER root
-RUN echo -e '[google-cloud-sdk]\n\
-name=Google Cloud SDK\n\
-baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64\n\
-enabled=1\n\
-gpgcheck=1\n\
-repo_gpgcheck=1\n\
-gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg\n\
-      https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg' > /etc/yum.repos.d/google-cloud-sdk.repo
+RUN printf '%s\n' \
+    '[google-cloud-sdk]' \
+    'name=Google Cloud SDK' \
+    'baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64' \
+    'enabled=1' \
+    'gpgcheck=1' \
+    'repo_gpgcheck=1' \
+    'gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg' \
+    '      https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg' > /etc/yum.repos.d/google-cloud-sdk.repo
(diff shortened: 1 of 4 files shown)


chiragkyal:
Enabled `DL3007` rule to cross-check the linter in CI. Expecting the `verify` job to fail due to DL3007

chiragkyal:
As expected, the [CI Job](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/1642/pull-ci-openshift-microshift-main-verify/1645312039356207104)  failed with error
```
./docs/config/Containerfile.tinyproxy:1 DL3007 warning: Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag
./packaging/images/openshift-ci/Dockerfile.test-runtime:2 DL3007 warning: Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag
```

Disabling the rule back.

packaging/images/openshift-ci/Dockerfile.test-runtime
```diff
@@ -1,3 +1,4 @@
+# hadolint global ignore=DL3002,DL4006,SC3037,SC2086
```
chiragkyal, line 1:
```
DL3002 warning: Last USER should not be root
SC3037 warning: In POSIX sh, echo flags are undefined.
SC2086 info: Double quote to prevent globbing and word splitting.
DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it.
```

ggiguash: I think we may want to move the first warning to the global flags and fix the other 3 in the file.

ggiguash: Let's try to avoid having local suppressions.

chiragkyal: - DL3002 : Moved to common config
- SC3037 : replaced `echo -e` with `printf` with some formatting changes
- SC2086 : Added Double quote
- DL4006: Added `SHELL ["/bin/bash", "-o", "pipefail", "-c"]`   Ref [here](https://github.com/hadolint/hadolint/wiki/DL4006)

packaging/images/openshift-ci/Dockerfile.test-runtime
```diff
@@ -1,13 +1,15 @@
...
 USER root
-RUN echo -e '[google-cloud-sdk]\n\
-name=Google Cloud SDK\n\
-baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64\n\
-enabled=1\n\
-gpgcheck=1\n\
-repo_gpgcheck=1\n\
-gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg\n\
-      https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg' > /etc/yum.repos.d/google-cloud-sdk.repo
+RUN printf '%s\n' \
```
dhellmann, line 3: Does this produce the same output? It seems like we would lose the line breaks between each of the separate values?

Could we put this config file in git as a file and then copy it into the right place in the container at this step?

dhellmann: TIL

```
$ printf "%s\n" foo bar
foo
bar
```

PR #1643, opened 2023-04-10
OCPBUGS-11395: Remove stale kubeconfigs

pacevedom:
Remove any kubeconfig that is not part of the hostname or SAN values in configuration.

Closes #<Issue Number>

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -33,6 +33,8 @@ import (
 	"github.com/openshift/microshift/pkg/util"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial/certchains"
+
+	"k8s.io/klog/v2"
 )
 
 func initCerts(cfg *config.Config) (*certchains.CertificateChains, error) {
@@ -401,6 +403,10 @@ func initKubeconfigs(
 		}
 	}
 
+	if err := cleanupStaleKubeconfigs(cfg); err != nil {
+		klog.Warningf("Unable to remove stale kubeconfigs: %v", err)
+	}
+
 	if err := util.KubeConfigWithClientCerts(
 		cfg.KubeConfigPath(config.KubeAdmin),
 		cfg.ApiServer.URL,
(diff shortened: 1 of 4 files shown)


pkg/cmd/init.go
```diff
@@ -512,3 +518,30 @@ func certsToRegenerate(cs *certchains.CertificateChains) ([][]string, error) {
...
+	deleteDirs := make([]string, 0)
+	for _, file := range files {
+		if !file.IsDir() {
+			continue
+		}
+		if _, ok := currentKubeconfigs[file.Name()]; !ok {
+			deleteDirs = append(deleteDirs, filepath.Join(cfg.KubeConfigRootAdminPath(), file.Name()))
+		}
+	}
+	klog.Infof("Stale kubeconfig directories: %v", deleteDirs)
```
dhellmann, line 540: Maybe instead of logging a list here, the loop below could log right before it tries to remove each item? That would make the logs easier to read and would fit with the "unit of work" style logging. https://docs.openstack.org/oslo.log/latest/user/guidelines.html#overall-logging-principles

pacevedom: Didnt know about this guide, it is done now.

dhellmann:
This is a good candidate for an automated test. Do we have the parts in place to add a test like that in this PR, or is the work to make that possible still in progress?

pacevedom:
> This is a good candidate for an automated test. Do we have the parts in place to add a test like that in this PR, or is the work to make that possible still in progress?

We could make it but not on this PR until https://github.com/openshift/microshift/pull/1557 merges. That way we can include it in a new e2e test that does not need another VM in CI.

dhellmann:
> > This is a good candidate for an automated test. Do we have the parts in place to add a test like that in this PR, or is the work to make that possible still in progress?
> 
> We could make it but not on this PR until #1557 merges. That way we can include it in a new e2e test that does not need another VM in CI.

If the cleanup function took a directory name as argument, we could have some tests for it in the unit test framework.

pacevedom:
> > > This is a good candidate for an automated test. Do we have the parts in place to add a test like that in this PR, or is the work to make that possible still in progress?
> > 
> > 
> > We could make it but not on this PR until #1557 merges. That way we can include it in a new e2e test that does not need another VM in CI.
> 
> If the cleanup function took a directory name as argument, we could have some tests for it in the unit test framework.

Done

dhellmann (commented):
/lgtm

PR #1645, opened 2023-04-10
USHIFT-1053: remove user data dir

dhellmann:
Remove the user data directory and always use the global path.


pkg/config/node.go
```diff
@@ -36,13 +36,13 @@ func (c *Config) CanonicalNodeName() string {
...
-func (c *Config) establishNodeName() (string, error) {
+func (c *Config) establishNodeName(dataDir string) (string, error) {
 	name := c.CanonicalNodeName()
-	filePath := filepath.Join(GetDataDir(), ".nodename")
+	filePath := filepath.Join(dataDir, ".nodename")
 	contents, err := os.ReadFile(filePath)
 	if os.IsNotExist(err) {
 		// ensure that dataDir exists
-		os.MkdirAll(GetDataDir(), 0700)
+		os.MkdirAll(DataDir, 0700)
```
eggfoobar, line 45: This is looking great, small nit, I think this needs to be the lowercase `dataDir`.

dhellmann: good eye, yes, you're right

PR #1646, opened 2023-04-10
USHIFT-1071 licensecheck verify

eggfoobar:


Makefile
```diff
@@ -275,13 +275,13 @@ clean: clean-cross-build
 .PHONY: clean
 
 licensecheck: microshift bin/lichen
-	bin/lichen -c .lichen.yaml microshift
+	$(GO_BUILD_BINDIR)/lichen -c .lichen.yaml $(GO_BUILD_BINDIR)/microshift
```
ggiguash, line 278: We have a different pattern for installing the tools now. It goes via the scrtips/fetch_tools.sh script.
Can we update this code to use this script, which will also make the lichen exe location more standard under _output/bin, etc.?

Makefile
```diff
@@ -275,13 +275,13 @@ clean: clean-cross-build
...
 
 bin:
 	mkdir -p $@
 
 bin/lichen: bin vendor/modules.txt
-	GOBIN=$(realpath ./bin) go install github.com/uw-labs/lichen@latest
+	GOBIN=$(realpath $(GO_BUILD_BINDIR)) go install github.com/uw-labs/lichen@latest
```
ggiguash, line 284: Let's not use ./bin path. The tool should be installed under _output instead.

Makefile
eggfoobar: Sure, moved over the installation to the script

Makefile
eggfoobar: Yeah, that was the issue, the ./bin dir was at the root, that `GO_BUILD_BINDIR` var points to `_output/bin`, moved it over to the script for this installation step.

PR #1651, opened 2023-04-10
USHIFT-1050: Use container image for running hadolint

ggiguash:
The `hadolint` executable from GitHub crashes on x86_64 platform. 
Using container image as a w/a until the problem is fixed.

Closes [USHIFT-1050](https://issues.redhat.com//browse/USHIFT-1050)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,8 +154,8 @@ verify-py:
 
 .PHONY: verify-container
 verify-container:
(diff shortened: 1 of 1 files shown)


ggiguash:
This does not work in CI because the verify job is already run inside a container.
/close

PR #1652, opened 2023-04-10
USHIFT-1050: Fix fetch_tools to use _output directory for temporary files

ggiguash:
Using /tmp/ directory caused `hadolint` SEGV
Closes [USHIFT-1050](https://issues.redhat.com//browse/USHIFT-1050)


ggiguash:
Did not help solving the problem in all environments.
/close

PR #1659, opened 2023-04-12
OCPBUGS-11538: upgrade openvswitch package version for 4.14

zshi-redhat:
Minimal openvswitch version required by ovn-kubernetes in rhel9.2(4.14) is openvswitch3.1.

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -89,7 +89,7 @@ The microshift-selinux package provides the SELinux policy modules required by M
 
 %package networking
 Summary: Networking components for MicroShift
-Requires: openvswitch2.17
+Requires: openvswitch3.1
 Requires: NetworkManager
 Requires: NetworkManager-ovs
 Requires: jq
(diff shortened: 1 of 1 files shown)


ggiguash:
@zshi-redhat, before merging this change, we need to update the environment configuration procedure because openvswitch3.1 package is not available for 4.13/4.14 builds from https://mirror.openshift.com/ at present.

Where do we expect to get this package from?
/hold

scripts/devenv-builder/configure-vm.sh
```diff
@@ -102,6 +103,15 @@ baseurl=https://mirror.openshift.com/pub/openshift-v4/\$basearch/dependencies/rp
 enabled=1
 gpgcheck=0
 skip_if_unavailable=0
+EOF
+
+    sudo tee "/etc/yum.repos.d/${FAST_DATAPATH_REPO_NAME}.repo" >/dev/null <<EOF
+[${FAST_DATAPATH_REPO_NAME}]
+name=fast datapath for RHEL ${OSVERSION}
+baseurl=https://rhsm-pulp.corp.redhat.com/content/dist/layered/rhel9/$(uname -i)/fast-datapath/os/
```
ggiguash, line 111: This is an internal URL, so we cannot use it.

ggiguash: We may want to ask the ART team to include the v3 package alongside the v2.
However, it's still not clear to me where we expect getting the v3 package in production.

ggiguash:
The packages are now published at mirror.openshift.com
/unhold

PR #1663, opened 2023-04-12
NO-ISSUE: Add JIRA bug trend generator

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/scripts/jira/bugtrend.py
+++ b/scripts/jira/bugtrend.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python3
+# pylint: disable=too-few-public-methods, broad-except, line-too-long
+"""
+This script generates trend charts for bugs in MicroShift. It groups
+created and resolved bugs per week over the last 6 months from the
+moment the script is executed, producing a chart with results for both
+metrics and the trend.
+
+Args:
+    To use this script, provide the following arguments:
+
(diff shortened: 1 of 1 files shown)


dhellmann:
Doesn't Jira have built-in tools for producing these sorts of graphs?

pacevedom:
> Doesn't Jira have built-in tools for producing these sorts of graphs?

Yes, but we have an issue with that. JIRA produces these charts using the resolution field, which in our case does not transition to Done until errata takes over and includes the bug in an advisory. In other words, charts lag until automation kicks in.
This happens quite late, so JIRA charts don't show the current situation for the team. What this tool does instead is take the transition of the status to verified as another measure for classifying the issue as resolved. Once an issue is in that status, the team does not need any further involvement.

dhellmann:
> > Doesn't Jira have built-in tools for producing these sorts of graphs?
> 
> Yes, but we have an issue with that. JIRA produces these charts using the resolution field, which in our case does not transition to Done until errata takes over and includes the bug in an advisory. In other words, charts lag until automation kicks in. This happens quite late, so JIRA charts don't show the current situation for the team. What this tool does instead is take the transition of the status to verified as another measure for classifying the issue as resolved. Once an issue is in that status, the team does not need any further involvement.

We can edit those filters in Jira.

pacevedom:
Close for now. If we need something not supported in JIRA we can bring it up again.

PR #1670, opened 2023-04-13
USHIFT-1081: update release note tool to look more places for the latest release

dhellmann:
This change works around a problem we have seen where updates to the
mirror do not update the "latest-4.x" version link by trying both of
our supported architectures instead of only looking at the ARM builds.


dhellmann:
https://github.com/openshift-eng/aos-cd-jobs/pull/3632 may make this less necessary, but I think it's still an improvement.

PR #1673, opened 2023-04-14
OCPBUGS-11538: Fix rpm spec history for ovs version upgrade

zshi-redhat:
openvswitch3.1 is required by rhel9.2 by both
4.13 & 4.14 releases


ggiguash:
/retitle OCPBUGS-11538: Fix rpm spec history for ovs version upgrade

PR #1686, opened 2023-04-19
USHIFT-1114: Optimize static analysis scripts and checks in Makefile

ggiguash:
* Rearrange verify scripts into ./scripts/verify directory
* Rename license check make file target to be consistent with others
* Add verify-licenses.sh script to fail when a new license is discovered
* Switch to checking vulnerabilities in binary, rather than source parsing mode. This is faster and not memory consuming.

Closes [USHIFT-1114](https://issues.redhat.com//browse/USHIFT-1114)


Makefile
```diff
@@ -145,22 +145,20 @@ verify-golangci:
...
+	./scripts/fetch_tools.sh govulncheck && \
+	./_output/bin/govulncheck ./...
 
 # We use the IGNORE var instead of `shellcheck disable=all` for files that are taken from
 # upstream and we wish to keep them with out added shellcheck declaratives.
 .PHONY: verify-sh
 verify-sh:
-	IGNORE="configure-ovs*.sh" ; \
+	IGNORE="configure-ovs.sh" ; \
 	./scripts/fetch_tools.sh shellcheck && \
```
dhellmann, line 156: Maybe we can eventually have a wrapper script for this step like we do for verify-py.sh

dhellmann (commented):
I like these changes. They're marked WIP and there's a merge conflict, so I'll wait for the update to lgtm.

PR #1687, opened 2023-04-19
USHIFT-1116: Use connect-to instead of resolve in router smoke test

pacevedom:
Use --connect-to instead of --resolve option in curl to support both IPs and hostnames in the e2e tests.

Closes #<Issue Number>


dhellmann:
If this is WIP, that's OK, but I don't think we want to land changes to test or production code without a Jira ticket for tracking.

PR #1668, opened 2023-04-13
USHIFT-1064: lightweight, unprivileged container image "MicroShift Test"

chiragkyal:
**Which issue(s) this PR addresses**:

Closes [USHIFT-1064](https://issues.redhat.com/browse/USHIFT-1064)

PR #1691, opened 2023-04-20
USHIFT-1064: use microshift-test image for e2e test

chiragkyal:

Closes  [USHIFT-1064](https://issues.redhat.com/browse/USHIFT-1064)

### DIFF
--- a/e2e/tests/0030-router-smoke-test.sh
+++ b/e2e/tests/0030-router-smoke-test.sh
@@ -36,7 +36,7 @@ for _ in $(seq "${RETRIES}"); do
 
     [ ${result} -eq 0 ] &&
         echo "${response}" | grep -q -E "HTTP.*200" &&
-        echo "${response}" | grep -q "Microshift Test" &&
+        echo "${response}" | grep -q "MicroShift Test" &&
         exit 0
 
     sleep "${BACKOFF}"
--- a/e2e/tests/0030-router-smoke-test.sh
+++ b/e2e/tests/0030-router-smoke-test.sh
@@ -14,23 +14,23 @@ BACKOFF=3s
 
 # shellcheck disable=SC2317  # Don't warn about unreachable commands in this function
 cleanup() {
-    oc delete route hello-microshift || true
-    oc delete service hello-microshift || true
-    oc delete -f "${SCRIPT_PATH}/assets/hello-microshift.yaml" || true
+    oc delete route microshift-test || true
+    oc delete service microshift-test || true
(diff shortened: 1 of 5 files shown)


chiragkyal:
> @chiragkyal , do we have microshift-test-workload Dockerfile somewhere in the repo? If not, can we include it in this review?

Yes, I have merged that Dockerfile [here](https://github.com/openshift/microshift/pull/1668/files)

e2e/tests/0030-router-smoke-test.sh
```diff
@@ -36,7 +36,7 @@ for _ in $(seq "${RETRIES}"); do
 
     [ ${result} -eq 0 ] &&
         echo "${response}" | grep -q -E "HTTP.*200" &&
-        echo "${response}" | grep -q "Hello MicroShift" &&
+        echo "${response}" | grep -q "Microshift Test" &&
```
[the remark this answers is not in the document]
chiragkyal: Sure, we can do that. Before committing that changes, I'll need to build  & push the Dockerfile manually to quay, otherwise, the e2e will fail.

e2e/tests/assets/hello-microshift.yaml
```diff
@@ -8,9 +8,7 @@ spec:
   terminationGracePeriodSeconds: 0
   containers:
   - name: hello-microshift
```
chiragkyal, line 10: Should we change all `hello-microshift` to `microshift-test` ? (pod, svc, route etc.)

e2e/tests/0030-router-smoke-test.sh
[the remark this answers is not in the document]
chiragkyal: Done

e2e/tests/assets/hello-microshift.yaml
ggiguash: Probably, yes, but that also requires updates on the CI side because we should also update the file name to be consistent. We could do this in 2 steps not to break the CI:
- Create a fixed copy of hello-microshift.yaml called microshift-test.yaml
- Start using it in the CI
- Delete the hello-microshift.yaml in a separate PR

pmtk: I don't think there's a need for that. e2e runs directly from this repository, so if you rename the file and change references in tests that use it, then we're good

chiragkyal: Okay, that sounds good. I've renamed the file and changed the references in the test.

ggiguash:
This is LGTM, but we need to wait for the arm jobs to be fixed to make sure the arm image is properly working.
/lgtm

PR #1692, opened 2023-04-20
NP-764: Enable no uplink OVN-K gateway bridge mode

pliurh:
Use no uplink OVN-K gateway bridge to decouple br-ex from host physical interface.
Use ovs-vsctl to create br-ex bridge instead of NM, as we don't need to run DHCP client for br-ex interface anymore.


dhellmann:
@pliurh I think that if you rebase this PR on the HEAD of the main branch we will get better results from the CI jobs.

anuragthehatter:
/hold

This is failing on QE Ushift clusters causing node to go into NOTREADY state with error
```container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: No CNI configuration file in /etc/cni/net.d/. ```

```10-ovn-kubernetes.conf``` is being absent under ```/etc/cni/net.d/```

**Comment by pmtk** [comment]
[Today's rebase](https://github.com/openshift/microshift/pull/1937) brought in changes for amd64 (it seems that we already had them for arm - introduced in [2023-06-15-173508](https://arm64.ocp.releases.ci.openshift.org/releasestream/4.14.0-0.nightly-arm64/release/4.14.0-0.nightly-arm64-2023-06-15-173508), we had `2023-06-16-015443`)
```
- ovn-kubernetes image-amd64 23c67c62ed1d912387791a1d8eb422b367c422dc to 5217351470aaa77bcfca04a1b6ee1ebf2bfbe7d9
  - eacdaa7 2023-06-13T09:44:38+02:00 Validate port before deleting conntrack flow
  - d829d7d 2023-06-13T09:44:21+02:00 Use no-uplink gateway bridge in compact-mode e2e test
  - f764a79 2023-06-13T09:44:16+02:00 Allow external gateway bridge without uplink port In local gateway mode
```

**Comment by pmtk** [comment]
some infra problems
/retest

**Comment by pliurh** [comment]
@anuragthehatter We have the up-to-date ovn-k d/s image. Could you test this PR again?

**Comment by anuragthehatter** [comment]
> @anuragthehatter We have the up-to-date ovn-k d/s image. Could you test this PR again?

Thanks @pliurh Cluster installed fine. Testing i sbeing conducted. So we expect br-ex to be unmanaged now as a result of this PR and of course no uplink to br-ex?

**Comment by pliurh** [comment]
> So we expect br-ex to be unmanaged now as a result of this PR and of course no uplink to br-ex?

Yes, we don't need NM to manage br-ex, as we don't need to run the DHCP client on it.

**Comment by dhellmann** [comment]
/override ci/prow-test-rebase

The asset file modified here is not managed by the rebase script.

**Comment by dhellmann** [comment]
@pliurh who is the best person to review this change?

File: pkg/node/netconfig.go [hunk 1]
```diff
@@ -31,6 +31,7 @@ const (
 	componentNetworkConfiguration = "network-configuration"
 	// Interface name where to add service IP
 	loopbackInterface = "lo"
+	ovnExternalBridge = "br-ex"
```
zshi-redhat, line 34: There is another const in pkg/config/ovn/ovn.go for the "br-ex", can we reuse it or define it in the ovn.go file if the ovnGatewayInterface is going to be removed?
https://github.com/openshift/microshift/blob/main/pkg/config/ovn/ovn.go#L17

packaging/systemd/microshift-ovs-init.service
```diff
@@ -6,6 +6,6 @@ Requires=openvswitch.service NetworkManager.service
 
 [Service]
 Type=oneshot
-ExecStart=/bin/bash /usr/bin/configure-ovs.sh OVNKubernetes
+ExecStart=/bin/bash /usr/bin/configure-ovs.sh
```
zshi-redhat, line 9: Where do we revert the previous changes applied by `configure-ovs.sh OVNKubernetes`?

pkg/node/netconfig.go
```diff
@@ -76,9 +77,15 @@ func (n *NetworkConfiguration) Run(ctx context.Context, ready chan<- struct{}, s
...
 
 func (n *NetworkConfiguration) addServiceIPLoopback() error {
-	link, err := netlink.LinkByName(loopbackInterface)
-	if err != nil {
-		return err
+	var link netlink.Link
+	var err error
+
+	link, err = netlink.LinkByName(ovnExternalBridge)
+	if _, ok := err.(netlink.LinkNotFoundError); ok {
```
zshi-redhat, line 84: Who will create `br-ex`in the no-uplink mode in a new microshift deployment?
I wonder if this service address would always be assigned to loopback device.

pkg/util/net.go
```diff
@@ -36,23 +35,17 @@ import (
 var previousGatewayIP string = ""
 
 func GetHostIP() (string, error) {
-	// Prefer OVN-K gateway IP if it is the CNI
-	gatewayIP, err := ovn.GetOVNGatewayIP()
```
zshi-redhat, line 40: shall we remove the `GetOVNGatewayIP` in ovn.go if it is no longer used?

zshi-redhat (commented):
Several other comments besides the inline ones:
1. shall we deprecate the "gatewayInterface" in this PR?
2. we should no longer need to listen on the "br-ex" interface in the mdns controller.
3. do we "need to" consider upgrade from default ovnk to no-uplink ovnk?

packaging/systemd/microshift-ovs-init.service
pliurh: In /usr/bin/configure-ovs.sh, the NM key files will be removed.

pkg/util/net.go
pliurh: Yes, I think we can remove it too.

pkg/node/netconfig.go
pliurh: Yes

pliurh:
> Several other comments besides the inline ones:
> 
> 1. shall we deprecate the "gatewayInterface" in this PR?

Yes.

> 2. we should no longer need to listen on the "br-ex" interface in the mdns controller.

Agree.

> 3. do we "need to" consider upgrade from default ovnk to no-uplink ovnk?

The `configure-ovs.sh` will handle the upgrade scenario. It will remove the NM config files and the OVS bridges and ports, then create br-ex with ovs-vsctl command.

pkg/node/netconfig.go
pliurh: configure-ovs.sh will create `br-ex` with `ovs-vsctl` command instead of NM.

pkg/config/ovn/ovn.go
```diff
@@ -30,20 +30,10 @@ const (
...
-	// disable microshift-ovs-init.service.
-	// OVS bridge "br-ex" needs to be configured manually when disableOVSInit is true.
-	DisableOVSInit bool `json:"disableOVSInit,omitempty"`
-	// Uplink interface for OVS bridge "br-ex"
-	GatewayInterface string `json:"gatewayInterface,omitempty"`
-}
-
 func (o *OVNKubernetesConfig) Validate() error {
 	// br-ex is required to run ovn-kubernetes
 	err := o.validateOVSBridge()
```
zshi-redhat, line 39: This validation would fail if br-ex doesn't exist, maybe the validation shall be removed given br-ex is created by ovnk.

pkg/node/netconfig.go
zshi-redhat: In a fresh deployment, the service address will always be assigned to loopback device since this node network configuration runs before the InfrastructureServices controller (where CNI is started). Is it expected?

pkg/config/ovn/ovn.go
pliurh: I think that is what we expect. `br-ex` shall be created by `microshift-ovs-init.service` before microshft starts.

pkg/node/netconfig.go
pliurh: The service address will be assigned to `lo` only if `br-ex` doesn't exist. It means the cluster is started without microshift-ovs-init.service. This is for the case that we use a non-OVN-K CNI.

pkg/config/ovn/ovn.go
zshi-redhat: ack+. I mis-read the commit message as the br-ex will be created by ovnkube-node.

zshi-redhat:
@jcaamano  Could you also take a look? 
This is about integrating no-uplink ovnk gateway change in microshift, specifically the configure-ovs.sh script is updated to handle upgrade from uplink NM settings to no-uplink NM settings.

packaging/systemd/configure-ovs.sh
```diff
@@ -477,284 +145,25 @@ get_bridge_physical_interface() {
...
+
+if [ "${nm_need_rollback}" = "true" ]; then
+  # Remove NM config files create by previous release when doing upgrade
   rollback_nm
-  
-  # Remove bridges created by ovn-kubernetes
-  ovs-vsctl --timeout=30 --if-exists del-br br-int -- --if-exists del-br br-local
 fi
+
+ovs-vsctl --timeout=15 --may-exist add-br br-ex
```
jcaamano, line 165: I am curious to know if there is a specific reason this command is duped here and 2 lines below

jcaamano:
> @jcaamano Could you also take a look? This is about integrating no-uplink ovnk gateway change in microshift, specifically the configure-ovs.sh script is updated to handle upgrade from uplink NM settings to no-uplink NM settings.

That part LGTM, just that small comment

pmtk (commented):
LGTM

I also verified upgrade from 4.13 to (this+https://github.com/openshift/microshift/pull/1957) and worked perfectly

packaging/systemd/configure-ovs.sh
pliurh: Good catch. It shall be removed.

assets/components/ovn/single-node/master/daemonset.yaml
```diff
@@ -417,7 +418,7 @@ spec:
             memory: 60Mi
         env:
         - name: OVN_KUBE_LOG_LEVEL
-          value: "4"
+          value: "5"
```
zshi-redhat, line 421: nit: why changing this to 5?

pliurh: This was for debugging. Let me change it back.

pkg/node/netconfig.go
dhellmann: The details in this comment thread are an important part of the design. Are they written down somewhere? If not, could you add them in a comment in this function, please? You can use a follow-up PR, since this one looks like it is almost ready to be merged.

dhellmann (commented):
This version looks good and seems to be passing the tests now that the topolvm image issue is resolved. I have one comment about some documentation, but that can (and probably should) be addressed in a simple follow-up PR, rather than requiring another iteration on this one.

The ARM tests are still a bit flakey for reasons outside of our control, so if those fail we can still merge this, unless you have reason to think the failure is related to the code.

@zshi-redhat do you want to apply the LGTM when you're happy?

zshi-redhat:
> This version looks good and seems to be passing the tests now that the topolvm image issue is resolved. I have one comment about some documentation, but that can (and probably should) be addressed in a simple follow-up PR, rather than requiring another iteration on this one.

@pliurh please file a followup PR for the documentation change.

> 
> The ARM tests are still a bit flakey for reasons outside of our control, so if those fail we can still merge this, unless you have reason to think the failure is related to the code.

agree, looking at the job history, it has 4 pass in May and zero pass in June. And the failed tests are different in each run. I will go ahead and lgtm the PR.

> 
> @zshi-redhat do you want to apply the LGTM when you're happy?

/lgtm

PR #1693, opened 2023-04-21
OCPBUGS-11967: Set OVN-k gateway next hop to an internal IP address

ggiguash:
Closes OCPBUGS-11967


ggiguash:
We decided to implement this conditionally on MicroShift startup when the default gateway is not set. 
Closing this PR now
/close

PR #1700, opened 2023-04-23
USHIFT-1061: Move hack/config-gen into cmd/generate-config directory

ggiguash:
Automatic configuration generation and verification is part of the development and CI procedures now. 
Moving the sources and scripts out of `hack` directory.


generate-config/cmd.go
```diff
@@ -17,7 +17,7 @@ func newCommand() *cobra.Command {
 	opt := configGenOpts{}
 
 	cmd := &cobra.Command{
-		Use:   "config-gen",
+		Use:   "generate-config",
```
pmtk, line 20: Could we use `cmd/generate-config/`? It more consistent with go dir pattern and doesn't pollute root dir

ggiguash: Done

ggiguash:
/retitle USHIFT-1061: Move hack/config-gen into cmd/generate-config directory

PR #1707, opened 2023-04-25
OCPBUGS-11967: Use dummy next-hop masquerade address as OVN-K default gateway

pliurh:
When there is no default gateway available for the host, use '169.254.169.4' as the OVN-K default gateway.

We will revisit this in 4.14, if we decide to use this setting for all scenarios.

Closes #<Issue Number>

### DIFF
--- a/pkg/components/networking.go
+++ b/pkg/components/networking.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/microshift/pkg/assets"
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/config/ovn"
+	"github.com/vishvananda/netlink"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/components/networking.go
```diff
@@ -81,16 +82,24 @@ func startCNIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
...
 		return err
 	}
-
+	defaultGWExist, err := ovn.IsDefaultGatewayExist(netlink.FAMILY_V4)
+	if err != nil {
+		klog.Warningf("Failed to check if default gateway exist on br-ex %v %v", crb, err)
+		return err
+	}
+	if !defaultGWExist {
+		klog.Info("cannot find default gateway way on br-ex, use dummy next-hop masquerade address as OVN-K default gateway")
```
ggiguash, line 91:
```suggestion
		klog.Info("Cannot find default gateway on br-ex, using dummy next-hop masquerade address as OVN-K default gateway")
```

zshi-redhat (commented):
The change works for me.
One question: do we need to monitor the default route in sysconfwatch and switch between online and offline modes?

pliurh:
> One question: do we need to monitor the default route in sysconfwatch and switch between online and offline modes?

I prefer to ask users to restart microshift.service in such a case. Since ovn-k doesn't support changing '--gateway-nexthop' on-the-fly. And the configMap change cannot trigger pod restart, as we run ovn-k in a DaemonSet.

ggiguash:
> The change works for me. One question: do we need to monitor the default route in sysconfwatch and switch between online and offline modes?

@zshi-redhat We have automatic restart in case of an IP change. Do you think there are cases when default route would change without IP change occurring?

zshi-redhat:
> > The change works for me. One question: do we need to monitor the default route in sysconfwatch and switch between online and offline modes?
> 
> @zshi-redhat We have automatic restart in case of an IP change. Do you think there are cases when default route would change without IP change occurring?

@ggiguash  I remember you mentioned to unplug the cable to test isolated network, when you do that, I assume the IP won't change, right?

pkg/config/ovn/ovn.go
```diff
@@ -196,3 +197,73 @@ func IsOVNKubernetesInternalInterface(name string) bool {
 
 	return excludedInterfacesRegexp.MatchString(name)
 }
+
+func HasDefaultGateway(family int) (bool, error) {
```
dhellmann, line 201: The logic here looks OK. I think the logging needs some work.

We expect to need to operate in environments with no gateway, with multiple NICs, etc. Therefore, I'm not sure it makes sense to warn as we search for the default gateway.

Can we log at each step using info level messages like "Did not find gateway for route..." and then when the gateway is found say something like "Found gateway for route..." so that someone debugging a system can see how MicroShift is interpreting the system configuration?

pliurh: Added logs.

pliurh:
@jcaamano PTAL

pliurh:
/hold

There is OVN-K fix on the way https://github.com/ovn-org/ovn-kubernetes/pull/3570

PR #1710, opened 2023-04-25
NO-ISSUE: Make authorized keys content replacement more robust in build.sh

ggiguash:
Authorised keys generated by `yubikeys` used `;` character in the contents. 
Changing to the `^` separator in the `sed` command to make the command more robust.


dhellmann:
@ausil would a `|` work?

ausil:
@dhellmann I tried and it did not work either

PR #1714, opened 2023-04-25
USHIFT-1122: use container-selinux version based on OS version

dhellmann:


packaging/rpm/microshift.spec
```diff
@@ -18,8 +18,9 @@
 # SELinux specifics
 %global selinuxtype targeted
 %define selinux_policyver 3.14.3-67
-%define container_policyver 2.189.0
-%define container_policy_epoch 3
+# FIXME: Specified by make-rpm.sh
```
pmtk, line 21: I don't think ART's build system likes comments in .spec. @vfreex  PTAL

dhellmann: CI didn't seem to like it anyway.

PR #1717, opened 2023-04-26
USHIFT-1202: Use hostnames over IPs in apiserver when connecting to kubelet

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/controllers/kube-apiserver.go
+++ b/pkg/controllers/kube-apiserver.go
@@ -125,7 +125,14 @@ func (s *KubeAPIServer) configure(cfg *config.Config) error {
 			"kubelet-certificate-authority":   {cryptomaterial.CABundlePath(kubeCSRSignerDir)},
 			"kubelet-client-certificate":      {cryptomaterial.ClientCertPath(kubeletClientDir)},
 			"kubelet-client-key":              {cryptomaterial.ClientKeyPath(kubeletClientDir)},
-			"kubelet-preferred-address-types": {"InternalIP", "Hostname"},
+			// MicroShift nodes expose these two types of addresses. In order to support having more than one
+			// node with the current approach (which is running a stand alone kubelet and share certificates
+			// with the master node) we need to use names only because of the way certificates are generated.
+			// The names are fed through the subjectAltNames configuration parameter, and using master node
+			// IP in the certificates when having more than one node is not possible due to apiserver SNI
+			// limitations. For this, we prefer using names and IPs as a fallback, supporting both single
+			// and multi node.
+			"kubelet-preferred-address-types": {"Hostname", "InternalIP"},
 
 			"proxy-client-cert-file":           {cryptomaterial.ClientCertPath(aggregatorClientCertDir)},
 			"proxy-client-key-file":            {cryptomaterial.ClientKeyPath(aggregatorClientCertDir)},


zshi-redhat:
This was tested to work in my local multi-node env.  LGTM

pkg/cmd/init.go
```diff
@@ -181,7 +181,12 @@ func certSetup(cfg *config.Config) (*certchains.CertificateChains, error) {
 						Name:         "kubelet-server",
 						ValidityDays: cryptomaterial.ShortLivedCertificateValidityDays,
 					},
-					Hostnames: []string{cfg.Node.HostnameOverride, cfg.Node.NodeIP},
+					// Kubelet hostnames must contain those addresses advertised by a node. In MicroShift these
+					// correspond to the node's IP address and the hostname. In order to accommodate having more
+					// than one node for CNCF certification tests, the subject alternative names from the
```
dhellmann, line 186: Let's not mention the tests here. We may have other reasons to deal with multi-node in the future.

```suggestion
					// than one node, the subject alternative names from the
```

pkg/controllers/kube-apiserver.go
```diff
@@ -125,7 +125,7 @@ func (s *KubeAPIServer) configure(cfg *config.Config) error {
 			"kubelet-certificate-authority":   {cryptomaterial.CABundlePath(kubeCSRSignerDir)},
 			"kubelet-client-certificate":      {cryptomaterial.ClientCertPath(kubeletClientDir)},
 			"kubelet-client-key":              {cryptomaterial.ClientKeyPath(kubeletClientDir)},
-			"kubelet-preferred-address-types": {"InternalIP", "Hostname"},
+			"kubelet-preferred-address-types": {"Hostname", "InternalIP"},
```
dhellmann, line 128: A comment here (or in the commit message) explaining how this change fixes the problem would help future maintainers.

dhellmann (commented):
The commit messages still say WIP.

pkg/cmd/init.go
pacevedom: Done

pkg/controllers/kube-apiserver.go
pacevedom: Done

dhellmann (commented):
The implementation and new comments look OK, but it wasn't clear from the slack thread whether we still need this. Zenghui and Jeremy should take a look.

/assign @zshi-redhat @jerpeter1

pkg/controllers/kube-apiserver.go
```diff
@@ -125,7 +125,14 @@ func (s *KubeAPIServer) configure(cfg *config.Config) error {
 			"kubelet-certificate-authority":   {cryptomaterial.CABundlePath(kubeCSRSignerDir)},
 			"kubelet-client-certificate":      {cryptomaterial.ClientCertPath(kubeletClientDir)},
 			"kubelet-client-key":              {cryptomaterial.ClientKeyPath(kubeletClientDir)},
-			"kubelet-preferred-address-types": {"InternalIP", "Hostname"},
+			// MicroShift nodes expose these two types of addresses. In order to support having more than one
+			// node with the current approach (which is running a stand alone kubelet and share certificates
+			// with the master node) we need to use names only because of the way certificates are generated.
+			// The names are fed through the subjectAltNames configuration parameter, and using master node
```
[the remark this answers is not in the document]
pacevedom: Yes, it is part of the current approach. If we are to generate the certificates and then manually signing them with the CA (for which you need root access to microshift) then it should not be needed. But that is a bit more complex.

pkg/controllers/kube-apiserver.go
```diff
@@ -125,7 +125,14 @@ func (s *KubeAPIServer) configure(cfg *config.Config) error {
 			"kubelet-certificate-authority":   {cryptomaterial.CABundlePath(kubeCSRSignerDir)},
 			"kubelet-client-certificate":      {cryptomaterial.ClientCertPath(kubeletClientDir)},
 			"kubelet-client-key":              {cryptomaterial.ClientKeyPath(kubeletClientDir)},
-			"kubelet-preferred-address-types": {"InternalIP", "Hostname"},
+			// MicroShift nodes expose these two types of addresses. In order to support having more than one
```
jerpeter1, line 128: The current approach (using standalone kubelet) does not share certs with the master node, it uses a cert with cluster-admin privileges to bootstrap: https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/

zshi-redhat:
> The implementation and new comments look OK, but it wasn't clear from the slack thread whether we still need this. Zenghui and Jeremy should take a look.
> 
> /assign @zshi-redhat @jerpeter1

I think we still want the first commit, which is a bug fix according to @pacevedom, not specific to multi-node.
Whether second commit is required depending on how kubelet runs on the extra worker node, will have a follow-up discussion.

pkg/controllers/kube-apiserver.go
pacevedom: In light of this I will remove the second commit, as its handled correctly by bootstrapping.

jerpeter1:
I had to make a minor adjustment to my script, but I tested this and it works as intended.

/lgtm

pacevedom:
/override ci/prow/e2e-openshift-conformance-reduced-arm
/override ci/prow/microshift-e2e-arm
arm jobs are flaky due to a kernel bug.
/override ci/prow/e2e-openshift-conformance-reduced
Overriding conformance reduced as the failure relates to 1.27 rebase. Should this PR fail then any test trying to retrieve anything from the kubelet would fail, which is not the case.

PR #1732, opened 2023-04-28
USHIFT-1218: Implement ostree updates from the local repository and edge commit

ggiguash:
* Add a special mode to the image-builder/build.sh script to generate edge commit archives
* Add an HTTP server to the ISO to allow local ostree updates
* Update documentation of Edge ISO with the new scenario

Closes [USHIFT-1218](https://issues.redhat.com//browse/USHIFT-1218)

### DIFF
--- a/docs/rhel4edge_iso.md
+++ b/docs/rhel4edge_iso.md
@@ -56,9 +56,11 @@ Optional arguments:
           included in the image (default: none)
   -embed_containers
           Embed the MicroShift container dependencies in the image
-  -ostree_server_name name_or_ip
-          Name or IP address and optionally port of the ostree
-          server (default: 127.0.0.1:8080)
+  -ostree_server_url URL
+          URL of the ostree server (default: http://127.0.0.1:8085/repo)
+  -build_edge_commit
+          Build edge commit archive instead of an ISO image. The
+          archive contents can be used for serving ostree updates.
   -lvm_sysroot_size num_in_MB
           Size of the system root LVM partition. The remaining
           disk space will be allocated for data (default: 10240)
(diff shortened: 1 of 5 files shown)


scripts/image-builder/config/kickstart.ks.template
```diff
@@ -81,4 +81,33 @@ for uuid in $(nmcli -f uuid,type,autoconnect connection | awk '$2 == "ethernet"
...
+
+    # Create an HTTP server for the repository
+    cat > /etc/systemd/system/${OSTREE_SERV} <<EOF
+[Unit]
+Description=Python HTTP server for the local ostree repository
+
+[Service]
+User=root
+WorkingDirectory=${OSTREE_REPO}
+ExecStart=/usr/bin/python -m http.server 8080
```
pmtk, line 102: https://www.osbuild.org/guides/user-guide/building-ostree-images.html#setting-up-an-http-server

> Note: You might be wondering why to bother with a container when you can just use "python -m http.server". The problem is that OSTree produces way too many requests and the Python HTTP server simply fails to keep up with OSTree

ggiguash: Is this documented somewhere that Python server is overwhelmed by the ostree requests?
I would like to avoid complicating the current solution if possible.

ggiguash: I switched to using Caddy as an HTTP server. Using Apache HTTPD is too complex for the configuration and I wanted to avoid running containers for this.

docs/rhel4edge_iso.md
```diff
@@ -56,9 +56,11 @@ Optional arguments:
           included in the image (default: none)
   -embed_containers
           Embed the MicroShift container dependencies in the image
-  -ostree_server_name name_or_ip
-          Name or IP address and optionally port of the ostree
-          server (default: 127.0.0.1:8080)
+  -ostree_server_url URL
+          URL of the ostree server (default: http://127.0.0.1:8080/repo)
```
dhellmann, line 60: We might want to pick a different port by default to leave that common port for users' apps. That would also let us pick a port that is likely to have firewall rules blocking access from outside the host, too.

docs/rhel4edge_iso.md
```diff
@@ -202,6 +204,123 @@ oc get cs
...
+
+# Done
+```
+
+> Optionally specify additional `-microshift_rpms`, `-custom_rpms`, or
+> `-embed_containers` arguments to customize the edge commit archive contents.
+
+Copy the produced `microshift-container-0.0.1-commit.tar` archive to the
+MicroShift server and unpack it at the `/var/lib/ostree-server-local` directory,
+deleting the previous contents.
```
dhellmann, line 298: How does a user stage multiple deployments this way? Does having rpm-ostree download the deployment leave it staged for rollback, or manually switching to/from it later?

dhellmann (commented):
I have a couple of questions about details inline, but I like the approach.

docs/rhel4edge_iso.md
ggiguash: Good point. Updated to 8085, which does not seem to be allocated to any "known" service.

docs/rhel4edge_iso.md
ggiguash: Currently, there is no option to have more than one update in the same commit. 
The rollback works unrelated to the ostree server because it's stored locally.

ggiguash: Having mutliple commits is possible to arrange by adding another ostree source. Currently, we have the default `edge`, so we can add more and have all those available. 
Let me see if it's easy to document it.

ggiguash: Actually, I did not find an option to merge more than one commit in the same repo.

ggiguash: I think we may want to merge this in the current form and get feedback from users about possible other scenarios to support in the devenv. WDYT?

dhellmann (commented):
This looks good. I wonder if @pacevedom has given any thought about whether it can be used in our CI?

/lgtm

PR #1708, opened 2023-04-25
USHIFT-1119: Add load-balancer Robot Framework test

pacevedom:
**Which issue(s) this PR addresses**:

Closes #<Issue Number>

PR #1734, opened 2023-04-29
USHIFT-1198: add tests for version API

dhellmann:
- update the way commands are run to return stdout
- log the remote host used for tests when connecting
- add shell runner for robot framework tests
- add tests for version config map and command line output
- move variable file up one level and make the default an example file
- update ssh connection so if no key is provided the active agent is used instead

### DIFF
--- a/test/resources/microshift-process.resource
+++ b/test/resources/microshift-process.resource
@@ -1,0 +1,22 @@
+*** Settings ***
+Library    Process
+Library    SSHLibrary
+Library    String
+Library    OperatingSystem
+Resource    ../resources/systemd.resource
+Resource    ../resources/microshift-host.resource
+Library    ../resources/YAML.py
+
+
+*** Keywords ***
+MicroShift Version
+    [Documentation]    Run the version command
+    Login MicroShift host
+    ${version_text}    ${rc}=    Execute Command
+    ...    microshift version -o yaml
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    ${rc}    0
+    Should Not Be Empty    ${version_text}
+    Logout MicroShift Host
+    ${version}=  Yaml Parse  ${version_text}
+    RETURN  ${version}
--- a/test/resources/microshift-process.resource
+++ b/test/resources/microshift-process.resource
@@ -3,7 +3,6 @@ Library    Process
 Library    SSHLibrary
 Library    String
 Library    OperatingSystem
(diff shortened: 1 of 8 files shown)


dhellmann:
/hold

This builds on #1708 and should wait for that PR.

test/resources/microshift-process.resource
```diff
@@ -0,0 +1,34 @@
...
+Parse Version Line
+    [Arguments]  ${line}
+    ${version_raw}=  Fetch From Right  ${line}  :
+    ${version}=  Strip String  ${version_raw}
+    RETURN  ${version}
+
+
+Restart MicroShift
+    [Documentation]  Restart the MicroShift service
+    Systemctl  restart  microshift
```
jogeo, line 34: This should be utilizing "Login MicroShift host" to ensure this restart is performed on the target MicroShift host.

test/resources/oc.resource
```diff
@@ -0,0 +1,13 @@
...
+Library     YAML.py
+
+*** Keywords ***
+
+Oc Get
+    [Arguments]  ${type}  ${namespace}  ${resource}
+
+    ${yaml_text}=  Run With Kubeconfig  oc get -n ${namespace} -o yaml ${type} ${resource}
+    ${yaml_data}=  Yaml Parse  ${yaml_text}
+
```
jogeo, line 12: Can we add the "Set Log Level    TRACE" and "Log     ${yaml_text}" keywords be added here?

test/resources/systemd.resource
```diff
@@ -0,0 +1,66 @@
...
+    END
+
+    Wait Until Keyword Succeeds
+    ...    10x
+    ...    10s
+    ...    Execute Command
+    ...    [ $(systemctl show -p SubState --value ${unit_name}) = ${state} ]
+    ...    timeout=10s    return_stdout=True    return_stderr=True
+
+    Logout MicroShift Host
```
jogeo, line 66: It would be better to factor out the Login/Logout steps, so they are not needed within each Keyword defined here.

dhellmann: I went back and forth on that. I'll see what it looks like if the suite setup opens the connection.

dhellmann: The tests run quite a bit faster if we login at the suite level, so that's what the latest version does.

jogeo:
Thanks for the updated @dhellmann
/lgtm

PR #1735, opened 2023-04-29
USHIFT-1199: add automated tests for etcd memory limit management

dhellmann:


dhellmann:
/hold

These tests depend on #1734 

These tests introduce instability to the overall test run by restarting microshift and not waiting long enough for it to restart.

dhellmann:
/hold

The e2e tests are not running on this PR because only test code is being changed. See https://github.com/openshift/release/pull/39797

test/resources/microshift-config.resource
```diff
@@ -0,0 +1,82 @@
...
+    [Documentation]  Remove any configuration file
+    ${stdout}  ${rc}=  Execute Command
+    ...    rm -f /etc/microshift/config.yaml
+    ...    sudo=True  return_rc=True
+
+Upload MicroShift Config  # robocop: disable=too-many-calls-in-keyword
+    [Documentation]  Upload a new configuration file to the MicroShift host
+    [Arguments]  ${config_content}
+
+    Log  ${config_content}
```
pacevedom, line 62: When using trace log level it shows the arguments contents. We do use trace level when not a dry run: https://github.com/openshift/microshift/blob/main/test/run.sh#L91

test/resources/microshift-config.resource
```diff
@@ -0,0 +1,82 @@
...
+
+    Remove File  ${local_tmp}
+
+    ${stdout}  ${rc}=  Execute Command
+    ...    mv ${remote_tmp} /etc/microshift/config.yaml
+    ...    sudo=True  return_rc=True
+    Should Be Equal As Integers  0  ${rc}
+
+    ${stdout}  ${rc}=  Execute Command
+    ...    chown root:root /etc/microshift/config.yaml
```
pacevedom, line 80: MicroShift runs as root so it will be able to read anything, isnt it?

test/resources/microshift-process.resource
```diff
@@ -20,3 +22,42 @@ MicroShift Version
...
+    [Documentation]  Wait for various checks to ensure MicroShift is online.
+    Wait Until Keyword Succeeds
+    ...  30x
+    ...  10s
+    ...  MicroShift Is Ready
+    Wait Until Keyword Succeeds
+    ...  30x
+    ...  10s
+    ...  MicroShift Is Live
+    # We could also wait for relevant pods. Can we restructure the
```
pacevedom, line 46: Actually it does work even if not in ostree. You will see an error about not being in an ostree host but thats only cosmetic. Checks run and return codes reflect situation. Needs some refactoring though, as it always expects topolvm to be up, even if there is no lvmd.

test/resources/microshift-process.resource
```diff
@@ -20,3 +22,42 @@ MicroShift Version
...
+    # Use separate stop and start steps to avoid a race condition with
+    # restart where we can't tell when it is down and so we can't tell
+    # whether we should count it as back up yet. Forcing the service
+    # down, then back up, ensures that when we do things like check
+    # systemd settings for the running process we get the values we
+    # expect without worrying about the race. Any test that modifies
+    # the MicroShift configuration file will be more reliable as a
+    # result.
+    Systemctl With Retry  stop  microshift.service
+    Sleep  10 seconds
```
pacevedom, line 61: Do we need to sleep? Stop only returns after MicroShift is completely down.

test/resources/microshift-process.resource
```diff
@@ -20,3 +22,42 @@ MicroShift Version
...
+    Systemctl With Retry  start  microshift.service
+    Wait For MicroShift
```
pacevedom, line 63: Greenboot tests do this in a different way, will ping you when its up for review. Actually it would benefit from some of the work in this PR.

test/resources/systemd.resource
```diff
@@ -0,0 +1,81 @@
...
+    END
+
+    ${stdout}  ${stderr}  ${rc}=  Execute Command
+    ...    systemctl ${verb} ${unit_name}
+    ...    sudo=True
+    ...    return_stdout=True
+    ...    return_stderr=True
+    ...    return_rc=True
+    Log  ${stdout}
+    Log  ${stderr}
```
pacevedom, line 48: Regular info log level already logs these.

test/resources/systemd.resource
```diff
@@ -0,0 +1,81 @@
...
+    IF  ${rc} != 0
+        ${status_text}=  Execute Command
+        ...    systemctl status ${unit_name}
+        ...    sudo=True
+        ...    return_stdout=True
+        Log  ${status_text}
+        ${log_text}=  Execute Command
```
pacevedom, line 55: This is already covered by sos report, isnt it? It doesnt even matter if the test succeeds or not (it fails here, though). You are going to get a huge output from this, mixed with the other tests/keywords output.

test/suites/etcd.robot
```diff
@@ -0,0 +1,80 @@
...
+    Restart MicroShift
+    Sleep  10 seconds  # Wait for systemd to catch up
+
+Setup With Custom Config
+    [Documentation]  Install a custom config and restart MicroShift
+    [Arguments]  ${config_content}
+    ${merged}=  Extend MicroShift Config  ${config_content}
+    Upload MicroShift Config  ${merged}
+    Restart MicroShift
+    Sleep  10 seconds  # Wait for systemd to catch up
```
pacevedom, line 72: The Restart MicroShift keyword is already waiting for all pods to be running, does systemd require more than that?

test/resources/microshift-process.resource
dhellmann: We may not need to sleep this long, but calling systemctl too quickly was not stable.

test/resources/microshift-process.resource
dhellmann: @pmtk has some "wait" logic in one of his PRs, too. It would be ideal if we could come up with 1 good strong implementation of that.

test/resources/systemd.resource
dhellmann: Removed.

test/resources/systemd.resource
dhellmann: We may want to cut this down, but I found it useful in debugging to see the logs from the process at the time of the command inline. I could make it 20 lines?

test/suites/etcd.robot
dhellmann: It doesn't currently wait for pods, though it could in the future. I'll see if removing this makes the test less stable.

dhellmann: Seems OK, I've removed it.

pacevedom (approved):
Will address comments in further PRs, as the main target of the PR looks good. Merging for progress and lets keep iterating the common functions.
/lgtm

PR #1736, opened 2023-04-29
USHIFT-1222: make pylint configuration more static

dhellmann:
Pylint incorporates new rules based on the version of pylint and
python. This PR adds a configuration file that should help us keep the
rule set we apply more consistent over time, avoiding confusing results
from different errors appearing in different settings.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -90,7 +90,13 @@ disable=raw-checker-failed,
         suppressed-message,
         useless-suppression,
         deprecated-pragma,
-        use-symbolic-message-instead
+        use-symbolic-message-instead,
+        # The names of modules for robot framework libraries do not
+        #  conform to the default name rule.
+        invalid-name,
+        # The linter runs in a virtualenv without robot framework, so
+        # cannot import modules.
+        import-error,
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
@@ -121,7 +127,7 @@ output-format=text
 reports=no
 
 # Activate the evaluation score.
-score=yes
+score=no
 
 
 [REFACTORING]
@@ -461,7 +467,7 @@ property-classes=abc.abstractproperty
 #typevar-rgx=
 
 # Naming style matching correct variable names.
-variable-naming-style=snake_case
+variable-naming-style=any
 
 # Regular expression matching correct variable names. Overrides variable-
 # naming-style. If left empty, variable names will be checked with the set
--- a/.pylintrc
+++ b/.pylintrc
@@ -97,6 +97,8 @@ disable=raw-checker-failed,
         # The linter runs in a virtualenv without robot framework, so
         # cannot import modules.
         import-error,
+        # Humbug
+        no-self-use,
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/.pylintrc
+++ b/.pylintrc
@@ -99,6 +99,8 @@ disable=raw-checker-failed,
         import-error,
         # Humbug
         no-self-use,
+        # Too many false-positives
+        no-member,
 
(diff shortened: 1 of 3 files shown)


scripts/verify/verify-py.sh
```diff
@@ -21,19 +21,18 @@ create_venv() {
...
+
+    local pyfiles
     pyfiles=$(find . -type d \( -path ./_output -o -path ./vendor -o -path ./assets -o -path ./etcd/vendor \) -prune -o -name '*.py' -print)
 
-    for f in ${pyfiles} ; do
-        echo "pylint: ${f}"
-        ${pylint} --variable-naming-style=any --score=n "${f}"
-    done
+    # shellcheck disable=SC2086
+    ${pylint} --rcfile="${ROOTDIR}/.pylintrc" ${pyfiles}
```
ggiguash, line 34: What's the advantage of running the lint on all files at once, rather than processing file-by-file?

dhellmann: If we process one file at a time, the job stops at the first file that has an error. If we process them all at once, all of the files are processed without stopping the job, so we get all of the errors at one time. We can also potentially add parallel execution, but I didn't bother with that here.

ggiguash: Should we still print the list of files that are checked? I think it might be interesting to see that in the CI to make sure nothing is missed by accident. We could leave the loop printing the names without running the command, and then just run the command all at once.

dhellmann: Added

.pylintrc
```diff
@@ -0,0 +1,595 @@
...
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once). You can also use "--disable=all" to
+# disable everything first and then re-enable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use "--disable=all --enable=classes
+# --disable=W".
+disable=raw-checker-failed,
```
chiragkyal, line 86: we can disable a few other common rules like `logging-fstring-interpolation,logging-not-lazy,line-too-long`, and remove them from file specific configuration.

.pylintrc
```diff
@@ -0,0 +1,595 @@
...
+        suppressed-message,
+        useless-suppression,
+        deprecated-pragma,
+        use-symbolic-message-instead,
+        # The names of modules for robot framework libraries do not
+        #  conform to the default name rule.
+        invalid-name,
+        # The linter runs in a virtualenv without robot framework, so
+        # cannot import modules.
+        import-error,
```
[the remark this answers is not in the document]
dhellmann: I'm not sure the effort of keeping all of our tools co-installable is worth the benefit of this rule. If a module tries to import something that doesn't exist, it will fail when we run it in some other CI job, right?

.pylintrc
dhellmann: Let's make those tweaks in follow-up PRs.

.pylintrc
```diff
@@ -0,0 +1,595 @@
...
+        bad-inline-option,
+        locally-disabled,
+        file-ignored,
+        suppressed-message,
+        useless-suppression,
+        deprecated-pragma,
+        use-symbolic-message-instead,
+        # The names of modules for robot framework libraries do not
+        #  conform to the default name rule.
+        invalid-name,
```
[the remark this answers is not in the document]
dhellmann: Done

.pylintrc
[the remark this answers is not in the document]
chiragkyal: yeah, right. agreed on that point.

.pylintrc
dhellmann: Those updates are in #1751

PR #1743, opened 2023-05-01
OCPBUGS-11421: remove version specifier for container-selinux

dhellmann:
We depend on cri-o and can rely on it to pull in any specific version
needed.  Currently neither cri-o nor podman specify a minimum version.


dhellmann:
We think this needs https://github.com/openshift/release/pull/38842

PR #1751, opened 2023-05-02
USHIFT-1234: globally disable a few pylint rules that are commonly an issue

dhellmann:
Disable logging-fstring-interpolation, logging-not-lazy, line-too-long
because we have valid code that violates those rules in a few places
where we disable the rules locally.


dhellmann:
/assign @chiragkyal 

This is the follow-up to the previous PR where we talked about disabling these rules globally.

PR #1765, opened 2023-05-06
USHIFT-1240: add runner for robot framework tests

dhellmann:
I could not figure out how to make the Makefile let me run one test at
a time with valid arguments, so I moved most of the content to a wrapper
script to use instead.


test/variables.yaml.example
```diff
@@ -3,4 +3,5 @@ USHIFT_HOST: microshift
 # USHIFT_USER: User to log into MicroShift's host
 USHIFT_USER: microshift
 # SSH Private key to use when logging into MicroShift's host.
+# Unset this variable to use ssh agent.
 SSH_PRIV_KEY: /home/microshift/.ssh/id_rsa
```
ggiguash, line 7: Can we use env variable like `$HOME` or `~/.ssh` here instead of `/home/microshift`?

test/run.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
```
ggiguash, line 2: We usually add at least `set -e` in the scripts to prevent uncontrolled command failure.

test/run.sh
```diff
@@ -0,0 +1,84 @@
...
+            ;;
+    esac
+done
+shift $((OPTIND-1))
+
+if [ ! -f "${RF_VARIABLES}" ]; then
+    echo "Please create a variables file at ${RF_VARIABLES}" 1>&2
+    echo "See ${RF_VARIABLES}.example for the expected content." 1>&2
+    exit 1
+fi
```
ggiguash, line 56: We renamed the default `yaml`, so one needs to make changes to the project for running scripts. 
Can we avoid that by having the defaults in place for running in `devenv`?

test/run.sh
```diff
@@ -0,0 +1,84 @@
...
+
+if [ ! -d "${RF_VENV}" ]; then
+    python3 -m venv "${RF_VENV}" && \
+        "${RF_VENV}/bin/python3" -m pip install -r "${SCRIPTDIR}/requirements.txt"
```
ggiguash, line 60: `&&` would not be necessary if `set -e` is there.

test/run.sh
```diff
@@ -0,0 +1,84 @@
...
+fi
+
+cd "${SCRIPTDIR}" || (echo "Did not find ${SCRIPTDIR}" 1>&2; exit 1)
```
ggiguash, line 63: `set -e` would not require the error message.

ggiguash (changes requested):

test/run.sh
pmtk: I always use opts from http://redsymbol.net/articles/unofficial-bash-strict-mode/ (although IFS might interfere with some things at times)
```
#!/bin/bash
set -euo pipefail
IFS=$'\n\t'
```

dhellmann (commented):
Updated based on comments and to add some changes that enable ssh-agent logins for VMs for cases where the ssh key requires a password.

test/run.sh
dhellmann: done

test/run.sh
```diff
@@ -0,0 +1,87 @@
...
+    "${RF_BINARY}" \
+        --dryrun \
+        --outputdir "${OUTDIR}" \
+        ${TESTS}
+else
+    "${RF_BINARY}" \
+        --randomize all \
+        --loglevel DEBUG \
+        -V "${RF_VARIABLES}" \
+        -x xunit.xml \
```
pacevedom, line 84: CI looks for this file as `junit.xml`, `xunit.xml` wont render in the prow summary.

test/run.sh
dhellmann: I added some documentation explaining how to manage the setup

dhellmann:
@pacevedom I added the top-level Makefile target, as we discussed. I removed the Makefile in the test directory as part of that.

@ggiguash I added some setup instructions to explain the configuration file. Let me know if that addresses your concerns about the defaults.

ggiguash:
This is LGTM for me. I'm delegating the final approval to @pacevedom

pacevedom:
/override ci/prow/microshift-e2e-arm
/override 
ci/prow/e2e-openshift-conformance-reduced-arm
Overriding both of these jobs since arm is flaky in AWS until we figure it out. Failures include unexpected restarts which render connection timeouts and refusals.

pacevedom:
/override
ci/prow/e2e-openshift-conformance-reduced-arm

pacevedom (commented):
/lgtm

PR #1775, opened 2023-05-09
OCPBUGS-11967: Use ovn-k masquerade nexthop address as gateway route's default gw

pliurh:
With this change, the gateway router of ovn-k can always forward egress traffic to host kernel.

It will fix the following scenarios:
1. no default gateway host
2. the connection to host default gateway is down

Closes #<Issue Number>


dhellmann:
This change looks good to me.

I see that @jogeo asked @anuragthehatter to look at it, is that for pre-merge testing?

zshi-redhat:
LGTM

PR #1777, opened 2023-05-09
USHIFT-986: Use UBI image and EPEL repo for running tinyproxy container for CRIO proxy doc

ggiguash:
The usage of this container is documented [here](https://github.com/openshift/microshift/blob/main/docs/howto_http_proxy.md)
Closes [USHIFT-986](https://issues.redhat.com//browse/USHIFT-986)

### DIFF
--- a/docs/config/Containerfile.tinyproxy
+++ b/docs/config/Containerfile.tinyproxy
@@ -1,13 +1,17 @@
-FROM docker.io/alpine:latest
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 
-RUN apk add --no-cache tinyproxy \
- && apk add --no-cache nano
+# The tinyproxy package is only available in EPEL7
+RUN microdnf install -y dnf && \
+    microdnf clean -y all && \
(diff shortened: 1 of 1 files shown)


docs/config/Containerfile.tinyproxy
```diff
@@ -1,13 +1,16 @@
-FROM docker.io/alpine:latest
+FROM registry.access.redhat.com/ubi9/ubi-minimal
 
-RUN apk add --no-cache tinyproxy \
- && apk add --no-cache nano
+# The tinyproxy package is only available in EPEL7
```
dhellmann, line 3: Is there some more modern replacement?

ggiguash: There is, but not in "official" repositories. At least, not something I could find.
The alternative is to use more complex proxies like nginx, etc., or build tinyproxy from sources.

dhellmann: I don't think that's really necessary. I don't think we would add EPEL instructions to product docs anyway, but it's OK in the project docs.

dhellmann (commented):
/lgtm

PR #1778, opened 2023-05-09
OCPBUGS-12977: Fix log message when CSI is disabled

ggiguash:
Minor fix in the message formatting.
Closes OCPBUGS-12977


dhellmann:
/lgtm

Do we want to backport this?

PR #1792, opened 2023-05-11
USHIFT-1250: Added Frequently used ports to be used by default in R4E…

shpaz:
As most Microshift implementations will include a central management tool (such as ACM/OCM/ArgoCD) that manages a fleet of Microshift clusters, we need to expose the `kube-api` port to the outside world, so that other tools can access it remotely. 

As this is the bread and butter of every Kubernetes cluster and the `kube-api` server port is permanent to every Microshift installation, this port should be opened automatically in the host's OS Firewall (`firewalld`). 

There are two options: 

Opening this port within the blueprint `.toml` file, thus makes it an integral part of every image 
Opening this port within the kickstart `.ks` file, will provide the same functionality 
As port `6443` is permanent, I think it'll be more reasonable to have it as part of the blueprint `.toml` configuration file.

 

As for `SSH` (port 22), it's reasonable that a user deploying Microshift will want to remotely access it (for configuration management) considering the fact Microshift doesn't have the machine-config API.

 

A suggested blueprint file can be used as the following (taking an example from the blueprint [file](https://github.com/openshift/microshift/blob/f4d833eb9775047189da5e2093eb8f5ea19db0f2/scripts/image-builder/config/blueprint_v0.0.1.toml) Microshift's repo): 

```bash
[customizations.firewall]
ports = ["6443:tcp"]

[customizations.firewall.services]
enabled = ["ssh"] 
```

In this example, we add to the blueprint those two ports, so that when the build.sh script is being run, those ports are automatically added to the created blueprint (which means to the installer artifact as well).


dhellmann:
/hold

This isn't consistent with how we have discussed managing edge devices in the past. Most will be deployed in environments where incoming connections need to be blocked, and agents on the host need to reach out to find instructions. In settings where that is not the case, users would be free to open the necessary ports and we should document that, but I don't think we want this as the default.

Let's talk about this before proceeding.

fzdarsky:
> This isn't consistent with how we have discussed managing edge devices in the past. Most will be deployed in environments where incoming connections need to be blocked, and agents on the host need to reach out to find instructions. In settings where that is not the case, users would be free to open the necessary ports and we should document that, but I don't think we want this as the default.

I completely concur with @dhellmann. We are targeting a model with RHDE (with or without MicroShift) in which all network connections for device and workload management are _device-initiated_, such that by default we can operate with zero open server ports on the host.

There are multiple reasons for this model, including better security posture and scalability, but most importantly it's the only model that works reliably under all networking conditions, e.g. when the device's IP address changes regularly (e.g. it's connected via cellular network or public WiFi), when it's behind a NAT or firewall, when it's without network for extended periods of time.

We therefore consider ssh'ing or kubectl'ing into the device an anti-pattern for production deployments. Of course, customers are free to open ports as needed, but we don't want to encourage this in the product / product engineering.

Instead, please consider an agent-based approach, e.g. deploying the ACM Managed Cluster Agent and/or ArgoCD to pull from the management hub or GitHub. If SSH access is needed for troubleshooting, consider using a reverse SSH tunnel.

DanielFroehlich:
The decision to open this port has to be made by the customer.  There are other ways to manage, which do not require to open this port. So doing that per default  could create unnecessary risks to customers.

ggiguash:
@shpaz, Do you think that adding an `-open_firewall_ports port1,...,portN` option to  the `scripts/image-builder/build.sh` script would provide the required flexibility without compromising the default settings?

shpaz:
@ggiguash As long as we provide users the ability to easily configure it and document it as needed. 

If you guys take a look, you'll see that the `kickstart.ks.template` template file doesn't include port `6443` in `%post`. As users may use the automated procedure, port `6443` may be forgotten and a whole new build/deployment will have to be executed (or the firewall port will be opened by hand, which IMO is less preferred). 

We should allow users to pick and choose those ports using the automation, plus document it as needed so that it won't cause confusion.

ggiguash:
@shpaz, thank you for confirming. I will open a PR for the team review adding a more flexible port selection.
Let's close this one for now.

/close

PR #1819, opened 2023-05-18
USHIFT-1083 `microshift admin data backup` command

pmtk:


pkg/admin/backup.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := os.RemoveAll(dest); err != nil {
+		klog.Errorf("Failed to remove %s directory: %v", tmp_dest, err)
+		return err
+	}
+
+	if err := os.Rename(tmp_dest, dest); err != nil {
+		klog.Errorf("Failed to rename %s to %s: %v", tmp_dest, dest, err)
+		return err
+	}
+	klog.Infof("Backed %s to %s", config.DataDir, dest)
```
dhellmann, line 48:
```suggestion
	klog.Infof("Backed up %s to %s", config.DataDir, dest)
```

pkg/admin/backup.go
```diff
@@ -0,0 +1,50 @@
...
+		klog.Errorf("Failed to make backup - copy failed: %v", err)
+		return err
+	}
+
+	if err := os.RemoveAll(dest); err != nil {
+		klog.Errorf("Failed to remove %s directory: %v", tmp_dest, err)
+		return err
+	}
+
+	if err := os.Rename(tmp_dest, dest); err != nil {
```
dhellmann, line 44: Should the temporary directory be removed at this point?

pkg/admin/backup.go
```diff
@@ -0,0 +1,50 @@
...
+func MakeBackup(target, name string) error {
+	if err := config.CreateDir(target); err != nil {
+		klog.Errorf("Failed to create dir %s: %v", target, err)
+		return err
+	}
+
+	dest := filepath.Join(target, name)
+	tmp_dest := dest + "-tmp"
+
+	args := append(cpArgs, config.DataDir, tmp_dest)
```
dhellmann, line 31: What happens here if the destination folder already exists?

dhellmann (commented):
This looks like a good start.

How can we make sure MicroShift is not running before taking the backup? Is that actually a requirement or does cp work ok in that case?

pmtk:
> How can we make sure MicroShift is not running before taking the backup? Is that actually a requirement or does cp work ok in that case?

That would be the thing with `ps aux` I had in previous PR, thanks for the reminder

pkg/admin/backup.go
pmtk: It was renamed  to final destination/name (without suffix), so it's gone, right?

pkg/admin/backup.go
pmtk: That's a good catch. Do you think that error in that case is good for now? I'd like to not add another param (`--force`) right now (maybe later for non-ostree systems)

pkg/admin/backup.go
dhellmann: Sorry, I meant if there's an error in the rename call.

pkg/admin/backup.go
dhellmann: If we get an error, that would be OK. I suspect we will end up with another copy of the directory inside the existing one.

pmtk: I didn't explain it properly... I want to add an error if the path already exists (so there's neither overwriting nor copying into)

pkg/admin/backup.go
pmtk: Probably? Not really sure. But it would definitely help in "new deployment, failed backup, reboot, tmp_dest already exists, problem). I'll add that

pkg/admin/backup.go
dhellmann: That sounds like a good plan, yes.

pkg/admin/run_check.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"k8s.io/klog/v2"
+)
+
+func microshiftShouldNotRun() error {
+	cmd := exec.Command("sh", "-c", "ps aux | grep  -v grep | grep 'microshift run'")
```
ggiguash, line 12: Is this better done by running `pidof` command? Note that we should also check microshift-etcd.
```
pidof microshift microshift-etcd && echo Running
```

pmtk: Unfortunately `pidof` also includes pid of `microshift admin backup`

pacevedom: Should we instead use systemctl? `systemctl show --no-pager microshift` produces machine readable format.

ggiguash: For pidof, we could also use `--omit-pid` option for the current PID of the caller. In any case, we should call something more robust than `ps` command, possibly `systemctl` that Pablo suggested.

pmtk: I wanted to check for actual process, not systemd unit - ensure on a more general "linux process level" than just systemd

pmtk: But if we don't want  such strong guarantee, then we can go with what Pablo suggested

pkg/admin/data/backup.go
```diff
@@ -0,0 +1,136 @@
...
+		"--preserve",
+		"--reflink=auto",
+	}
+)
+
+// MakeBackup backs up MicroShift data (/var/lib/microshift) to
+// target/name/ (e.g. /var/lib/microshift-backups/backup-00001).
+func makeBackup(cfg BackupConfig) error {
+	if cfg.Storage == "" {
+		return fmt.Errorf("backup storage must not be empty")
```
dhellmann, line 28: Can these validation steps be performed at a place where we could provide an error message that includes the command line option they need to use? We wouldn't want the CLI interface to leak into this low level module, but if I saw this error text I'm not sure I would know immediately what I should have done differently.

pkg/admin/data/backup.go
```diff
@@ -0,0 +1,136 @@
...
+	if found {
+		klog.Infof("Directory %s already exists", path)
+		return nil
+	}
+
+	err = util.MakeDir(path)
+	if err != nil {
+		return fmt.Errorf("failed creating %s: %w", path, err)
+	}
+	return err
```
dhellmann, line 81: It would be good to log here that the directory was created.

I think also explicitly returning nil, instead of err, would make it clearer that this is the success outcome.

pkg/admin/data/backup.go
```diff
@@ -0,0 +1,136 @@
...
+	return nil
+}
+
+func removeDirBeforeRenaming(path string) error {
+	found, err := util.FileExists(path)
+	if err != nil {
+		return fmt.Errorf("failed to check if %s exists: %w", path, err)
+	}
+	if found {
+		klog.Infof("Removing %s before renaming tmp directory", path)
```
dhellmann, line 103: Does this function know why the directory is being removed?

pkg/admin/data/backup.go
```diff
@@ -0,0 +1,136 @@
...
+		if err := os.RemoveAll(path); err != nil {
+			return fmt.Errorf("failed to remove %s: %w", path, err)
+		}
+	}
```
dhellmann, line 107: Log success

pkg/admin/data/backup.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func renameDir(from, to string) error {
+	klog.InfoS("Renaming directory", "from", from, "to", to)
+	err := os.Rename(from, to)
+	if err == nil {
+		return nil
+	}
+
+	klog.ErrorS(err, "Renaming directory failed - deleting 'from' dir", "from", from, "to", to)
```
dhellmann, line 130: Cleaning up if the rename fails seems like something the caller should do.

pkg/admin/data/run_check.go
```diff
@@ -0,0 +1,36 @@
...
+
+const (
+	expectedState = "inactive"
+)
+
+var (
+	services = []string{"microshift.service", "microshift-etcd.scope"}
+)
+
+func microshiftShouldNotRun() error {
```
dhellmann, line 19:
```suggestion
func microshiftIsNotRunning() error {
```

dhellmann (commented):
The overall logic looks good. I have a couple of specific points inline, and I think it would be worth reviewing the logging advice at https://docs.openstack.org/oslo.log/latest/user/guidelines.html#overall-logging-principles and thinking about how to apply that here, both for the use of different log levels and for the "unit of work" approach described for info log messages.

pkg/admin/data/manager.go
```diff
@@ -0,0 +1,30 @@
...
+	// Name is backup's directory name
+	Name string
+}
+
+type Manager interface {
+	Backup(BackupConfig) error
+	Restore(BackupConfig) error
+}
+
+func NewManager() *manager {
```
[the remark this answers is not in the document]
pmtk: I'm planning to use it like this ([src](https://github.com/openshift/microshift/pull/1826/files#diff-e78e17834c76663b756393d04ebe7aceb8ea4059d44b7d2ae3ee4e9999aead6eR64)):
```
a.dataManager.Backup(data.BackupConfig{
			Storage: config.BackupsDir,
			Name:    prevBootHealth.DeploymentID,
		})
```
When I'll need to restore backup for another deployment, I don't want to create another manager.
Should manager take `backupsDir` and Backup/Restore functions take name of the backup?

pkg/admin/data/backup.go
pmtk: Yes, I'll add that. But I'd like to leave these here as well (speaking from experience, I already managed to pass empty deployID when working on https://github.com/openshift/microshift/pull/1826/ resulting in backup being in `cfg.Storage` rather than `cfg.Storage/cfg.Name`)

pkg/admin/data/backup.go
pmtk: You mean that detail like "before renaming" should be logged in caller, rather than here? And make it just "if dir exists, remove it"?

pkg/admin/data/manager.go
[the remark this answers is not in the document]
dhellmann: I think you're right that the backup directory could be held in the manager, and then the functions could take arguments that are more dynamic, like the name to restore from. A method like `List()` would also use the backup directory, but not the name value in `BackupConfig`.

pkg/cmd/admin.go
```diff
@@ -0,0 +1,70 @@
...
+
+			if err := data.MicroShiftIsNotRunning(); err != nil {
+				return fmt.Errorf("microshift must not be running: %w", err)
+			}
+
+			return nil
+		},
+	}
+	v := version.Get()
+	data.PersistentFlags().String(
```
dhellmann, line 48: What's the use case for making the storage location configurable from the command line?

pkg/admin/data/backup.go
```diff
@@ -0,0 +1,152 @@
...
+		return err
+	} else if backupAlreadyExists {
+		klog.V(2).InfoS("Backup already exists - renaming temporarily", "path", dest)
+		if err := renamePath(dest, dest_old); err != nil {
+			return err
+		}
+	}
+
+	if err := renamePath(dest_tmp, dest); err != nil {
+		klog.Errorf("Renaming path failed - renaming %s back and deleting %s: %v", dest_old, dest_tmp, err)
```
dhellmann, line 61: `dest_old` will only exist if the original location existed when this function started, so if the rename on line 60 fails for some other reason then the logic in this section is going to produce more errors.

What are the situations that could result in the location already existing? We can definitely write code to produce a unique name, so I think we only have to worry about the collision if the name comes from the user via the command line? What if we just refuse to take the backup in that case? Or don't let the user specify the full backup location (give us a directory and we pick a name in that directory)?

pkg/cmd/admin.go
pmtk: I'd say nice to have for RPM systems where backups are manual. But i don't have concrete use case other than NFS (or different partition) mount for backups, but then CoW goes out of window

pkg/admin/data/backup.go
pmtk: Same code will be used in automated ostree backing up so 3rd boot into the same healthy deployment will overwrite backup taken during 2nd boot.

pkg/cmd/admin.go
dhellmann: OK, that makes sense. In that case I think we're going to not want to overwrite an existing backup, so we should think about how to handle that. Maybe just a check in the CLI code, before we get into the low-level code, where we do want old backups overwritten.

pkg/admin/data/backup.go
dhellmann: We might want to move some of that overwrite logic "up" the call stack to a place where the code knows whether it is being invoked by the user or by automation. That will make the edge cases easier to reason about.

pmtk: Added function to interface `BackupExists(BackupName)` and it's called in the CLI. It will refuse to overwrite a backup unless `--force` is provided

pkg/admin/data/manager.go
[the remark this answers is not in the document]
pmtk: Changed it so the manager has "storage" as field and Backup/Restore only take `BackupName`

pkg/cmd/admin.go
pmtk: Added `BackupExists()` and `--force`

pkg/admin/data/data_manager.go
```diff
@@ -0,0 +1,151 @@
...
+		return err
+	} else if backupExists {
+		if err := renamePath(dest, old); err != nil {
+			return err
+		}
+		klog.InfoS("Temporarily renamed existing backup", "backup", dest, "renamed", old)
+	}
+
+	if err := renamePath(tmp, dest); err != nil {
+		klog.Errorf("Renaming path failed - renaming %s back and deleting %s: %v", old, tmp, err)
```
dhellmann, line 88: It's going to be interesting to set up a test for this error case.

pkg/admin/data/data_manager.go
```diff
@@ -0,0 +1,151 @@
...
+		klog.InfoS("Removed path", "path", path)
+	}
+	return nil
+}
+
+func copyDataDir(dest string) error {
+	cmd := exec.Command("cp", append(cpArgs, config.DataDir, dest)...) //nolint:gosec
+	klog.InfoS("Executing command", "cmd", cmd)
+
+	out, err := cmd.CombinedOutput()
```
dhellmann, line 127: We don't have to change it in this PR, but we may want to split handling of stdout and stderr. If there's an error after copying 50 files, we may not want all of the names of the files that were successfully copied in the error message.

dhellmann (commented):
/lgtm

/hold for QE verification

pkg/admin/data/data_manager.go
dhellmann: We talked on slack about a way to generate unique names to avoid having to rename anything. Let's land this version and make that change in another PR.

dhellmann (commented):
/lgtm
/hold

@jogeo, please remove the hold when you've had a chance to verify the change.

pkg/admin/data/data_manager.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if found, err := pathExists(string(dm.storage)); err != nil {
+		return err
+	} else if !found {
+		if makeDirErr := util.MakeDir(string(dm.storage)); makeDirErr != nil {
+			return fmt.Errorf("making %s directory failed: %w", dm.storage, makeDirErr)
+		}
+		klog.InfoS("Backup storage directory created", "path", dm.storage)
+	} else {
+		klog.InfoS("Backup storage directory already existed", "path", dm.storage)
```
dhellmann, line 61: This indicates an attempt to write over an existing backup, right? I would expect that to be an error.

pmtk: Not really, that's the "parent directory" where backups are stored. Maybe you have how can we name it better ("backup storage") or maybe just simply delete this log (and keep only one about creating /var/lib/microshift-backups/)?

dhellmann: Ah, I see what's going on now.

dhellmann (commented):
/lgtm
/hold

Holding for @jogeo to verify again

jogeo:
I've preformed some early testing on this PR and it behaves as expected.
/unhold

PR #1832, opened 2023-05-23
USHIFT-1246: Enable custom variables file and venv dir in RF

pacevedom:
Avoid having fixed locations for some of the practicalities of running RF both locally and on CI. Allow changing both the dir for venv for python dependencies and also the variables file. This also allows QE have multiple files locally and select when running the tests.

Closes #<Issue Number>


dhellmann:
/hold cancel
/override ci/prow/e2e-openshift-conformance-reduced
/override ci/prow/microshift-e2e

There's no reason to think the tests are failing because of this change.

PR #1837, opened 2023-05-25
NO-ISSUE Update python scripts to use python3

pmtk:
Recently we updated `build_root` image to one based on RHEL9 which ships python3 instead of python2. Change is needed to fix rebase procedure

### DIFF
--- a/scripts/auto-rebase/rebase.py
+++ b/scripts/auto-rebase/rebase.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 """
 This Python script automates the process of rebasing a Git branch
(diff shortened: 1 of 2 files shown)


scripts/auto-rebase/rebase.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
ggiguash, line 1: Can you in explain the description why?

ggiguash: Do we need to make the same change in all the Python scripts?

pmtk:
Tested with rehearsal https://github.com/openshift/release/pull/39713
/unhold

PR #1843, opened 2023-05-26
USHIFT-1284: make test login to microshift host more reliable

dhellmann:
Use 'Get Varible Value' to see if an ssh key has been provided instead
of looking for the name in all of the variables ourselves.


dhellmann:
/hold

The e2e tests are not running on this PR because only test code is being changed. See https://github.com/openshift/release/pull/39797

PR #1854, opened 2023-05-30
USHIFT-1286: Use exit code 1 in greenboot failures

pacevedom:
All MicroShift failures return 1 except for the service not being able to start. This happens because it shortcuts error checking by sending a SIGTERM to itself so that it exits immediately. The default handler for this signal returns exit code 143, whereas all the other known failure conditions return 1.
When testing this behavior the greenboot script is not consistent when something fails. This change creates a specific handler for TERM so that the exit code is also 1.

Closes #<Issue Number>

### DIFF
--- a/packaging/greenboot/microshift-running-check.sh
+++ b/packaging/greenboot/microshift-running-check.sh
@@ -10,8 +10,17 @@ PODS_CT_LIST=(2                        1                    1                 2 
 # shellcheck source=packaging/greenboot/functions.sh
 source /usr/share/microshift/functions/greenboot.sh
 
+# Set the term handler to convert exit code to 1
+trap 'return_failure' TERM
+
 # Set the exit handler to log the exit status
 trap 'script_exit' EXIT
+
+# The term handler to override the default behavior and have a uniform and
+# homogeneous exit code in all controlled situations.
+function return_failure() {
+    exit 1
+}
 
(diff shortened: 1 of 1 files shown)


packaging/greenboot/microshift-running-check.sh
```diff
@@ -10,9 +10,18 @@ PODS_CT_LIST=(2                        1                    1                 2
 # shellcheck source=packaging/greenboot/functions.sh
 source /usr/share/microshift/functions/greenboot.sh
 
+# Set the term handler to convert exit code to 1
+trap 'return_failure' TERM
+
 # Set the exit handler to log the exit status
 trap 'script_exit' EXIT
```
ggiguash, line 17: Is `script_exit` called when TERM arrives?

pacevedom: Yes, here's an example:
```
[root@microshift-dev microshift]# /etc/greenboot/check/required.d/40_microshift_running_check.sh
STARTED
GRUB boot variables:
boot_success=1
boot_indeterminate=0
Greenboot variables:
None
error: loading sysroot: opendir(ostree/repo): No such file or directory
The ostree status:

Waiting 300s for MicroShift service to be active and not failed
The microshift.service systemd unit is failed. Terminating...
FAILURE
[root@microshift-dev microshift]# echo $?
1
```

PR #1862, opened 2023-06-01
USHIFT-1292: Install GCC in dev VM

jerpeter1:
Installing microshift via configure-vm.sh was working before, but today it was failing because gcc wasn't installed:

```
# Building RPM packages
error: Failed build dependencies:
	gcc is needed by microshift-4.14.0_0.nightly_2023_05_29_174116_20230531233827_e4dff3d6-1.el9.x86_64
```


ggiguash:
@jerpeter1 , could you add in the description a few words on why this is necessary?

PR #1866, opened 2023-06-01
USHIFT-1312 Test: Rebooting healthy system should result in backing up the data

pmtk:
Tests functionality of https://github.com/openshift/microshift/pull/1874

### DIFF
--- a/test/resources/microshift-host.resource
+++ b/test/resources/microshift-host.resource
@@ -25,3 +25,15 @@ Login MicroShift Host
 Logout MicroShift Host
     [Documentation]  Close the open ssh connection to the MicroShift host
     SSHLibrary.Close Connection
+
+Reboot MicroShift Host
+    [Documentation]    Reboot the MicroShift host
+    ...
+    ...    Expects that SSH connection to MicroShift host is active.
+
+    SSHLibrary.Start Command    reboot    sudo=True
+    Sleep    30s
+    Wait Until Keyword Succeeds    5m    15s    Login MicroShift Host
+
+    ${rc}=    SSHLibrary.Execute Command    true    return_stdout=False    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
--- a/test/resources/ostree.resource
+++ b/test/resources/ostree.resource
@@ -1,7 +1,8 @@
 *** Settings ***
-Documentation    Keywords for OSTree-based systems
+Documentation       Keywords for OSTree-based systems
 
-Library     SSHLibrary
+Library             SSHLibrary
+Library             DataFormats.py
 
 
 *** Variables ***
@@ -17,14 +18,14 @@ Get Booted Deployment Status
     ...    rpm-ostree status --booted --json
     ...    return_rc=True
     Should Be Equal As Integers    0    ${rc}
-    ${json}=    Evaluate    json.loads('''${ostree_status}''')    json
-    RETURN    ${json}
+    ${status}=    Json Parse    ${ostree_status}
+    RETURN    ${status}
 
 Get Booted Deployment ID
     [Documentation]    Get ID of currently booted deployment
 
     ${status}=    Get Booted Deployment Status
-    RETURN    ${status["deployments"][0]["id"]}
+    RETURN    ${status.deployments[0].id}
 
 Get Booted Deployment Backup Prefix Path
     [Documentation]    Get backup path prefix for current deployment
--- a/test/resources/ostree.resource
+++ b/test/resources/ostree.resource
@@ -1,0 +1,62 @@
+*** Settings ***
+Documentation    Keywords for OSTree-based systems
+
+Library     SSHLibrary
+
+
+*** Variables ***
+${BACKUP_STORAGE}       /var/lib/microshift-backups
+${HEALTH_FILE}          /var/lib/microshift-backups/health.json
+
+
+*** Keywords ***
+Get Booted Deployment Status
+    [Documentation]    Get rpm-ostree status of booted deployment as a json
+
+    ${ostree_status}    ${rc}=    Execute Command
+    ...    rpm-ostree status --booted --json
+    ...    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+    ${json}=    Evaluate    json.loads('''${ostree_status}''')    json
+    RETURN    ${json}
+
+Get Booted Deployment ID
+    [Documentation]    Get ID of currently booted deployment
+
+    ${status}=    Get Booted Deployment Status
+    RETURN    ${status["deployments"][0]["id"]}
+
+Get Booted Deployment Backup Prefix Path
+    [Documentation]    Get backup path prefix for current deployment
+    ...
+    ...    Prefix path is BACKUP_STORAGE/deployment-id.
+    ...    Globbing directories starting with the prefix will yield
+    ...    list of backups for the deployment.
+
+    ${current_deployment_id}=    Get Booted Deployment ID
+    Should Not Be Empty    ${current_deployment_id}
+    Should Not Be Empty    ${BACKUP_STORAGE}
+    ${backup_path}=    Catenate    SEPARATOR=/    ${BACKUP_STORAGE}    ${current_deployment_id}
+    RETURN    ${backup_path}
+
+Get System Health
+    [Documentation]    Get system health information from health.json file
+
+    ${health}    ${stderr}    ${rc}=    Execute Command
+    ...    jq -r '.health' ${BACKUP_STORAGE}/health.json
+    ...    sudo=True    return_rc=True    return_stderr=True
+    Should Be Equal As Integers    0    ${rc}
+    RETURN    ${health}
+
+Does Backup For Booted Deployment Exist
+    [Documentation]    Checks if any backup for current deployment exists
+    ...    by finding paths starting with "backup prefix path"
+
+    ${deploy_id}=    Get Booted Deployment ID
+    ${output}    ${rc}=    Execute Command
+    ...    find ${BACKUP_STORAGE} -name '${deploy_id}*' | wc -l
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+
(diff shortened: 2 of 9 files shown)


pmtk:
This test will probably require setting up current `microshift-e2e` to ignore `ostree` tags or provide mechanism to not run tests on non-ostree-based systems

test/suites/upgrade_ostree.robot
```diff
@@ -0,0 +1,99 @@
...
+    [Documentation]    Checks if property of a service has expected value
+    [Arguments]    ${service}    ${property}    ${expected_value}
+
+    ${value}    ${rc}=    Execute Command
+    ...    systemctl show ${service} -P ${property}
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+    Should Be Equal As Strings    ${expected_value}    ${value}
+
+Reboot
```
pmtk, line 67: From @dhellmann ([src](https://github.com/openshift/microshift/pull/1826#discussion_r1212161152))
> This could go into the microshift-host resource file.

test/suites/upgrade_ostree.robot
```diff
@@ -0,0 +1,99 @@
...
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+
+Wait Until Greenboot Health Checks Are Done
+    [Documentation]    Wait until greenboot healthchecks are done
+
+    Wait Until Keyword Succeeds    5m    15s
+    ...    Check Service Property    greenboot-healthcheck.service    SubState    exited
+
+Wait Until MicroShift Is Ready
```
pmtk, line 51: From @dhellmann ([src](https://github.com/openshift/microshift/pull/1826#discussion_r1212164372)) 
> There's a similar keyword in my PR with etcd tests that does a bit more. https://github.com/openshift/microshift/pull/1735/files#diff-4f6b7fe814ca7246b4cc856e49319e6e2f33ba6ac2abe86995e7a42cafab7afcR36
>
> I'd like to make that reusable keyword as robust as we can.

test/suites/upgrade_ostree.robot
```diff
@@ -0,0 +1,99 @@
...
+    [Documentation]    Wait until MicroShift service is ready
+
+    Wait Until Keyword Succeeds    2m    15s
+    ...    Check Service Property    microshift.service    ActiveState    active
+
+Check Service Property
```
pmtk, line 57: From @dhellmann ([src](https://github.com/openshift/microshift/pull/1826#discussion_r1212165319)):
> See https://github.com/openshift/microshift/pull/1735/files#diff-9f059e74d79eab4cda8c994b93013638f88857b968777109b9bc462d89fa5f7fR10 as well

pmtk:
/cc @dhellmann @pacevedom @jogeo @dhensel-rh 

This PR needs some sort mechanism to not run tests intended for R4E/ostree-based systems on RPM.
Approaches that come to mind:
- doing a runtime check and deciding if test should run or be skipped
- using tags to filter out irrelevant test suites
any other ideas?

test/resources/microshift-host.resource
```diff
@@ -25,3 +25,15 @@ Login MicroShift Host
 Logout MicroShift Host
     [Documentation]  Close the open ssh connection to the MicroShift host
     SSHLibrary.Close Connection
+
+Reboot MicroShift Host
```
dhellmann, line 29: This would be more robust if we looked for a different boot id after the reboot, to avoid any race condition with issuing the reboot command and sshd stopping. It doesn't seem like the most likely race, so let's make that change in another PR.

test/suites/upgrade-ostree.robot
```diff
@@ -0,0 +1,85 @@
...
+Remove Existing Backup For Current Deployment
+    [Documentation]    If exists, remove backup for current deployment
+
+    ${exists}=    Does Backup For Booted Deployment Exist
+    IF    ${exists} == ${False}    RETURN
+
+    ${deployID}=    Get Booted Deployment ID
+
+    ${rm_output}    ${rm_stderr}    ${rc}=    Execute Command
+    ...    find ${BACKUP_STORAGE} -name '${deployID}*' -exec rm -rf {} +
```
dhellmann, line 71: We know the exact path, right? Can we just remove it, instead of using `find`?

dhellmann (commented):
This looks OK, but I see the e2e test is failing with an error in this new test suite

```
Keyword 'Greenboot Health Check Exited' failed after retrying for 5 minutes. The last error was: dead != exited
```

test/suites/upgrade-ostree.robot
dhellmann: OK, we probably want to do something to limit the depth of the find call, then. Is it enough to say `rm -rf ${deployID}*` instead of using find at all?

And we probably want to check for deployID being empty, to ensure we don't delete everything.

pmtk: `find` was only looking into `/var/lib/microshift-backups/` but I changed it to `rm -rf` because it's easier to read

test/suites-ostree/upgrade-ostree.robot
```diff
@@ -0,0 +1,82 @@
+*** Settings ***
```
pmtk, line 1: This file is in separate dir so e2e CI job doesn't try to run it against non-ostree system.
When we have e2e for ostree systems we can think if we want to move it back to `suites/` and use tags (or other method to filter tests)

pmtk:
/remove-label tide/merge-method-squash

Remembered that tide's squash results in ugly commit messages. I'll squash the commits myself

dhellmann (commented):
This looks good to me. I'll leave it for @jogeo to approve since he wanted a chance to review, too.

test/resources/ostree.resource
```diff
@@ -0,0 +1,62 @@
...
+    Should Not Be Empty    ${current_deployment_id}
+    Should Not Be Empty    ${BACKUP_STORAGE}
+    ${backup_path}=    Catenate    SEPARATOR=/    ${BACKUP_STORAGE}    ${current_deployment_id}
+    RETURN    ${backup_path}
+
+Get System Health
+    [Documentation]    Get system health information from health.json file
+
+    ${health}    ${stderr}    ${rc}=    Execute Command
+    ...    jq -r '.health' ${BACKUP_STORAGE}/health.json
```
jogeo, line 46: 'Evaluate    json.loads' was used above in keyword "Get Booted Deployment Status". Should we be consistent in the pattern we use to ingest and parse json?

test/suites-ostree/upgrade-ostree.robot
```diff
@@ -0,0 +1,77 @@
...
+    [Documentation]    Checks if greenboot-healthcheck finished running
+
+    ${value}=    Get Systemd Setting    greenboot-healthcheck.service    SubState
+    Should Be Equal As Strings    ${value}    exited
+
+Remove Existing Backup For Current Deployment
+    [Documentation]    Remove backups for current deployment
+
+    ${path}=    Get Booted Deployment Backup Prefix Path
+
```
jogeo, line 61: Should we test path here with 'Should Not Be Empty'? I see that 'Get Booted Deployment Backup Prefix Path' has a check for empty values. However, if the code is changed to use a different key word at some point to get path we would not want to accidentally call 'rm -rf *'

jogeo (commented):
Thank you for adding this test, it looks good over all. I've left a few questions and comments.

test/resources/ostree.resource
jogeo: As a side note, 'jq' is used by scripts such as microshift_set_healthy.sh, so it's not a bad thing that we are also using it in our test scripts. My initial comment was just a reaction to seeing json parsed in the same PR using two different approaches.

test/resources/ostree.resource
```diff
@@ -0,0 +1,62 @@
...
+
+*** Keywords ***
+Get Booted Deployment Status
+    [Documentation]    Get rpm-ostree status of booted deployment as a json
+
+    ${ostree_status}    ${rc}=    Execute Command
+    ...    rpm-ostree status --booted --json
+    ...    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+    ${json}=    Evaluate    json.loads('''${ostree_status}''')    json
```
dhellmann, line 20: For YAML, I found it handy to parse the text and then return a RF DottedDict type to make accessing the data from RF keywords easier. See the YAML library in the resources directory.

test/resources/ostree.resource
```diff
@@ -0,0 +1,62 @@
...
+    RETURN    ${json}
+
+Get Booted Deployment ID
+    [Documentation]    Get ID of currently booted deployment
+
+    ${status}=    Get Booted Deployment Status
+    RETURN    ${status["deployments"][0]["id"]}
```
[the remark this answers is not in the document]
pmtk: This uses kw `Get Booted Deployment Status` which runs `rpm-ostree status --booted --json`
Thanks to `--booted` it will only output one deployment, but it's still using the same data structure, i.e. `deployments` in an array. With `--booted` it will only have 1 element.

Without that switch, `0` could be either booted or staged deployment

test/resources/ostree.resource
pmtk: That's a good point. `jq` here felt simpler, otherwise I'd have either `cat` and use `json.loads` or... I don't know what, I don't think `SSHLibrary.Get File` supports privilege escalation and `health.json` is only readable by root

dhellmann: I like the idea of keywords for grabbing files only readable by root. [Save Default MicroShift Config](https://github.com/openshift/microshift/blob/main/test/resources/microshift-config.resource#L13) uses `cat` to grab the contents of the config file, and it could use the same reusable keyword for that step. Then something like [Upload String To File](https://github.com/openshift/microshift/blob/main/test/resources/microshift-config.resource#L13) could be modified to take arguments for the owner and mode of the output file.

ggiguash:
Let's merge this in the current form. The tests are not going to run because they are in a separate folder.
We now have a metal environment to run this tests, so let's try those out.
/lgtm

PR #1868, opened 2023-06-01
NOISSUE: [release-4.13] Fix configure-vm.sh script to use release-4.13 branch by default

ggiguash:
Necessary for QE procedures to build the right source code in the release-4.13 branch.

### DIFF
--- a/scripts/devenv-builder/configure-vm.sh
+++ b/scripts/devenv-builder/configure-vm.sh
@@ -92,7 +92,7 @@ if ${BUILD_AND_RUN}; then
     # Build MicroShift
     # https://github.com/openshift/microshift/blob/main/docs/devenv_setup.md#build-microshift
     if [ ! -e ~/microshift ]; then
-        git clone https://github.com/openshift/microshift.git ~/microshift
+        git clone https://github.com/openshift/microshift.git -b release-4.13 ~/microshift
     fi
     cd ~/microshift
 


dhellmann:
How does the caller get the script to run? Could the caller clone the repo and the script require that it is present? That way we don't have to keep updating the content of the script every time we create a branch.

ggiguash:
> How does the caller get the script to run? Could the caller clone the repo and the script require that it is present? That way we don't have to keep updating the content of the script every time we create a branch.

At present, this script is copied to a brand new machine and it is the one that's cloning the repo. Since we're talking about the system bootstrap, it's a chicken-and-egg problem.

jogeo:
QE checks out the MicroShift repository, so we can use the configure script to setup the build machine. In one scenario  we download the RPMs from ART, so we don't build from source. In a second scenario we checkout microshift but then also builds from source. This second scenario is where we get into the situation that a release-4.13 branch has been checked out but then when the configure-vm.sh script is called it again checks out microshift but uses the main branch.

jogeo:
> How does the caller get the script to run? Could the caller clone the repo and the script require that it is present? That way we don't have to keep updating the content of the script every time we create a branch.

QE is going to update our pipeline to make sure that ~/microshift is created the first time microshift is cloned, so configure-vm.sh will skip it's git clone call.

dhellmann:
> > How does the caller get the script to run? Could the caller clone the repo and the script require that it is present? That way we don't have to keep updating the content of the script every time we create a branch.
> 
> QE is going to update our pipeline to make sure that ~/microshift is created the first time microshift is cloned, so configure-vm.sh will skip it's git clone call.

That feels like a good compromise for long-term maintenance.

PR #1869, opened 2023-06-01
USHIFT-1298: enable cgo to support loading local libraries

dhellmann:
We had original disabled cgo, because it increases the binary size. We will need to accept that footprint increase.

```
# With CGO_ENABLED=0

$ du -sh _output/bin/microshift*
110M	_output/bin/microshift
42M	_output/bin/microshift-etcd

# With CGO_ENABLED=1

$ du -sh _output/bin/microshift*
141M	_output/bin/microshift
55M	_output/bin/microshift-etcd
```


pacevedom (approved):
/lgtm

pacevedom:
Need a rebase for the e2e to work.

PR #1870, opened 2023-06-01
OCPBUGS-12736: Fix microshift-etcd signal handling by calling wait on the process after creating it

ggiguash:
Call cmd.Wait on the microshift-etcd process asynchronously to make sure SIGCHLD is handled in case the etcd process exits before the microshift process. In this scenario, microshift should exit right after to avoid flooding the system log with error messages about not being able to connect to the etcd service.

Closes OCPBUGS-12736

### DIFF
--- a/pkg/controllers/etcd.go
+++ b/pkg/controllers/etcd.go
@@ -103,7 +103,19 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
 		return fmt.Errorf("%s failed to start: %v", s.Name(), err)
 	}
 
-	// Make sure microshift-etcd is terminated whether it failed to properly start or MicroShift stops
+	// Handle microshift-etcd termination before microshift process exits
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			klog.Warningf("%v failed waiting on process to finish: %+v", s.Name(), err)
+		}
+		klog.Infof("%v process quit: %v", s.Name(), cmd.ProcessState.String())
+
+		// Exit microshift to trigger microshift-etcd restart
+		klog.Warning("microshift-etcd process terminated prematurely, restarting MicroShift")
+		os.Exit(0)
+	}()
+
+	// Handle microshift-etcd termination after microshift process exits
 	defer func() {
 		if err := cmd.Process.Signal(os.Interrupt); err != nil {
 			klog.Warningf("%v failed interrupting the process: %+v", s.Name(), err)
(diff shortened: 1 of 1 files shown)


pkg/controllers/etcd.go
```diff
@@ -103,18 +103,26 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
...
+
+	// Handle microshift-etcd termination after microshift process exits
 	defer func() {
 		if err := cmd.Process.Signal(os.Interrupt); err != nil {
 			klog.Warningf("%v failed interrupting the process: %+v", s.Name(), err)
 			if err := cmd.Process.Kill(); err != nil {
 				klog.Warningf("%v failed killing the process: %+v", s.Name(), err)
 			}
 		}
-		if err := cmd.Wait(); err != nil {
```
pmtk, line 114: Shouldn't we wait until microshift-etcd finishes? I hope that etcd shutdown sequence won't be interrupted

ggiguash: On Linux, the process management is using fork->exec->wait pattern, which we need to mimic in Go.
Wait is a blocking function, which only exits after the process terminates. 
It's the way to detect and properly "release" an orphan process in case `microshift-etcd` exits prematurely.

pmtk: But now you have no guarantee that cmd.Wait from line 108 will return, go doesn't wait for goroutines by itself. Is there a risk that we signal etcd to finish, return from this function and almost immediately whole `microshift` exits not giving chance that `goroutine` to wait for etcd to finish?

ggiguash: If microshift exits first without running Wait on microshift-etcd, the orphaned process is cleaned up by the `init` (pid 1) OS process. 
You can simulate this situation by running `kill -9` on microshift - the OS would take care of the clean up in this case.

pmtk: So problem we're solving here is detecting microshift-etcd exit and restarting to get it up again? If there are no more caveats to this solution, then it looks good enough to me but I'd like others to take a look

dusk125:
lgtm, thanks for fixing this Giga!

pkg/controllers/etcd.go
```diff
@@ -103,18 +103,26 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
...
+	// Handle microshift-etcd termination before microshift process exits
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			klog.Warningf("%v failed waiting on process to finish: %+v", s.Name(), err)
+		}
+		klog.Infof("%v process quit: %v", s.Name(), cmd.ProcessState.String())
+
+		// Exit microshift to trigger microshift-etcd restart
+		klog.Warning("microshift-etcd process terminated prematurely, restarting MicroShift")
+		os.Exit(0)
```
dhellmann, line 115: Instead of abruptly exiting we should cancel the run context (https://github.com/openshift/microshift/blob/main/pkg/cmd/run.go#L118) to make sure that other parts of the process have a chance to clean up after themselves before the main() function exits. I don't know if the ctx argument to this function can do that or if we need to pass that cancel function in separately.

ggiguash: We have a few other places where we call exit. And, we have a task to replace those.
Should we do it in another PR so that all exits are consistent?

PR #1873, opened 2023-06-02
USHIFT-1299 make SSH re-connect after reboot more robust

pmtk:
require several successful connection attempts before deciding that it's really working

tries to mitigate situation when one SSH command might be successful, but next one might get "connection refused"

### DIFF
--- a/e2e/tests/0040-reboot.sh
+++ b/e2e/tests/0040-reboot.sh
@@ -8,19 +8,27 @@ set -x
 
 SCRIPT_PATH="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")"
 
-RETRIES=25
-INTERVAL=10
+RETRIES=60
+INTERVAL=5
 SUCCESSFUL_TRIES=3
 
-wait_until() {
-    successes=0
-    local cmd=$*
+get_remote_boot_timestamp() {
+    ssh "${USHIFT_USER}@${USHIFT_IP}" "date -d \"\$(uptime -s)\" +%s"
+}
+
+wait_for_system_up() {
+    local prev_boot_start="${1}"
+    local successes=0
+
     for _ in $(seq "${RETRIES}"); do
-        if ${cmd}; then
+        curr_boot_start=$(get_remote_boot_timestamp || true)
+
+        if [[ "${curr_boot_start}" -gt "${prev_boot_start}" ]]; then
           successes=$((successes+=1))
         else 
           successes=0
         fi
+
         if [[ "${successes}" -ge "${SUCCESSFUL_TRIES}" ]]; then
             return 0
         fi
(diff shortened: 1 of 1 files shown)


e2e/tests/0040-reboot.sh
```diff
@@ -8,14 +8,23 @@ set -x
...
     local cmd=$*
     for _ in $(seq "${RETRIES}"); do
-        ${cmd} && return 0
-        sleep "${BACKOFF}"
+        if ${cmd}; then
+          successes=$((successes+=1))
+        else 
+          successes=0
+        fi
+        if [[ "${successes}" -ge "${SUCCESSFUL_TRIES}" ]]; then
```
ggiguash, line 24: Successful retries do not guarantee that the host was booted as ssh might work before the boot when it takes time.

A way to check this may be the following:
1. Get the current date: date +"%Y-%m-%d %H:%M:%S"
2. reboot
3. Run "ssh hostname uptime -s"
4. The reboot occurred if the uptime timestamp is greater than the saved timestamp

ggiguash: Another, probably simpler way is to run "who -b" before the boot and after the boot and just make sure they are different.

pmtk: I think that 3 attempts with 10s pause between them should be good enough.
To be sure, I checked journal on shutdown; sshd was off shortly after issuing reboot
```
Jun 02 08:38:13.176469 localhost.localdomain systemd-logind[808]: The system will reboot now!
Jun 02 08:38:13.308323 localhost.localdomain systemd[1]: sshd.service: Deactivated successfully.
```
This test will be reimplemented in RF pretty soon, so maybe it's good enough for now?

ggiguash: 30s is not enough when microshift is running because it takes time to shut down crio and microshift (at least in my VM environment). 
The "who -b" is really simple. Run one before and then after reboot in loop and compare results.

pmtk: sshd is shut down first, you cannot log in when microshift and crio are still shutting down:
```
Jun 02 08:38:13 localhost.localdomain systemd-logind[808]: The system will reboot now!
Jun 02 08:38:13 localhost.localdomain systemd[1]: sshd.service: Deactivated successfully.
Jun 02 08:39:17 localhost.localdomain systemd[1]: microshift.service: Deactivated successfully.
Jun 02 08:39:17 localhost.localdomain systemd[1]: microshift.service: Consumed 55.740s CPU time.
Jun 02 08:39:17 localhost.localdomain systemd[1]: crio.service: Deactivated successfully.
Jun 02 08:39:17 localhost.localdomain systemd[1]: crio.service: Consumed 13.671s CPU time.
```
So it seems that we don't need to track boot time in this case

PR #1881, opened 2023-06-05
WIP USHIFT-1084 Restore backup if previous was unhealthy

pmtk:


pkg/admin/data/data_manager.go
```diff
@@ -91,20 +92,50 @@ func (dm *manager) Backup(name BackupName) error {
...
+
+	if exists, err := dm.BackupExists(name); err != nil {
+		return fmt.Errorf("checking if backup %s exists failed: %w", name, err)
+	} else if !exists {
+		klog.ErrorS(nil, "Backup to restore does not exist", "name", name)
+		return fmt.Errorf("backup %s does not exist", name)
+	}
+
+	tmp := fmt.Sprintf("%s.%s", config.DataDir, rand.String(8))
+	klog.InfoS("Temporarily renaming data dir", "data", config.DataDir, "renamedTo", tmp)
```
pmtk, line 118: This is asymmetric to `Backup()` (there it's caller's responsibility to ensure the backup dir does not exist).
Should we make caller's responsibility ensure `/var/lib/microshift` does not exist before restoring from a backup?

pkg/admin/prerun/prerun.go
```diff
@@ -31,36 +32,49 @@ func (hi *HealthInfo) IsHealthy() bool {
...
-		klog.InfoS("System was not healthy - skipping backup")
-		return nil
-	}
+	klog.InfoS("Previous boot", "health", health.Health, "deploymentID", health.DeploymentID, "bootID", health.BootID)
 
-	dataManager, err := data.NewManager(config.BackupsDir)
-	if err != nil {
-		return err
+	if health.IsHealthy() {
+		return pr.backup(health)
```
pmtk, line 68: As implementation grows, I expect this to not be end of the workflow.
For example, when rolling back from healthy system, we want to take backup, but then we want to restore another (compatible with rollback's deployID). Storage migration is another example.
We'll get there

pkg/admin/prerun/prerun.go
```diff
@@ -76,20 +90,84 @@ func Perform() error {
...
 	}
 
-	removeOldBackups(dataManager, backupsForDeployment)
+	pr.removeOldBackups(backupsForDeployment)
 
 	return nil
 }
 
+func (pr *PreRun) restore() error {
+	// TODO: Check if containers are already running (i.e. microshift.service was restarted)?
```
pmtk, line 103: Situation when system was unhealthy, so it restored the backup, but then crashed (or exited due to IP change or smth else)  and was restarted before red/green script had chance to update `health.json`: it would restore again.

At this point some Pods might already run, so should we protect against restoring in such case? Or maybe it's not a problem.

pkg/admin/data/data_manager.go
copejon: IIUC, `/var/lib/microshift` should never(?) exist before executing the Restore business logic, so it'd be an invariant of Restore(). If that's the case, IMO it should be checked by Restore(), not it's caller. Ditto for Backup().

pkg/admin/data/data_manager.go
```diff
@@ -91,20 +92,50 @@ func (dm *manager) Backup(name BackupName) error {
...
+
+	if name == "" {
+		return &EmptyArgErr{"name"}
+	}
+
+	if exists, err := dm.BackupExists(name); err != nil {
+		return fmt.Errorf("checking if backup %s exists failed: %w", name, err)
+	} else if !exists {
+		klog.ErrorS(nil, "Backup to restore does not exist", "name", name)
+		return fmt.Errorf("backup %s does not exist", name)
```
dhellmann, line 114:
```suggestion
		return fmt.Errorf("failed to restore backup, %s does not exist", name)
```

pkg/admin/data/data_manager.go
```diff
@@ -91,20 +92,50 @@ func (dm *manager) Backup(name BackupName) error {
...
+	klog.InfoS("Restoring the data", "storage", dm.storage, "name", name, "data", config.DataDir)
+
+	if name == "" {
+		return &EmptyArgErr{"name"}
+	}
+
+	if exists, err := dm.BackupExists(name); err != nil {
+		return fmt.Errorf("checking if backup %s exists failed: %w", name, err)
+	} else if !exists {
+		klog.ErrorS(nil, "Backup to restore does not exist", "name", name)
```
dhellmann, line 113: We can let the caller log the error we return on the next line

pkg/admin/data/data_manager.go
dhellmann: Restore should essentially be

```
mv /var/lib/microshift /var/lib/microshift.save
cp /var/lib/microshift-backups/whatever /var/lib/microshift (using the same flags used when doing the backup)
rm -rf /var/lib/microshift.save
```

If the `cp` fails, then recovery is

```
rm -rf /var/lib/microshift
mv /var/lib/microshift.save /var/lib/microshift
```

I think you want all of that logic managed from 1 function to ensure all of the error handling paths that involve recovery are covered.

Are there files that only exist in `/var/lib/microshift` (and not in the backup) that we need to keep, like history metadata?

pkg/admin/prerun/prerun.go
dhellmann: As a first iteration, let's assume it's safe to restore from the same backup multiple times.

jerpeter1:
re: https://github.com/openshift/microshift/pull/1881#discussion_r1218449482

> If the cp fails, then recovery is ... `rm -rf /var/lib/microshift`

Removing the files as an automated response to an error could potentially remove some of the information needed to figure out what the error happened in the first place - another move operation might be safer (assuming that running out of disk space isn't the error which caused the problem)

PR #1882, opened 2023-06-05
USHIFT-1267: Allow for keeping container images when cleaning MicroShift data

ggiguash:
* Add the `--keep-images` option to `microshift-cleanup-data` script
* Use this option in the greenboot tests to speed it up
* Fixed bug when deleting pods

```
$ ./scripts/microshift-cleanup-data.sh 
Stop all MicroShift services, also cleaning their data

Usage: microshift-cleanup-data.sh <--all [--keep-images] | --ovn>
   --all         Clean all MicroShift and OVN data
   --keep-images Keep container images when cleaning all data
   --ovn         Clean OVN data only
```

Closes [USHIFT-1267](https://issues.redhat.com//browse/USHIFT-1267)

### DIFF
--- a/scripts/microshift-cleanup-data.sh
+++ b/scripts/microshift-cleanup-data.sh
@@ -5,13 +5,15 @@ SCRIPT_NAME=$(basename "$0")
 # to allow for pod deletion when MicroShift is stopped
 PODS_NS_LIST=(openshift-service-ca openshift-ingress openshift-dns openshift-storage openshift-ovn-kubernetes)
 FULL_CLEAN=true
+KEEP_IMAGE=false
 
 function usage() {
     echo "Stop all MicroShift services, also cleaning their data"
     echo ""
-    echo "Usage: ${SCRIPT_NAME} <--all | --ovn>"
-    echo "   --all   Clean all MicroShift and OVN data"
-    echo "   --ovn   Clean OVN data only"
+    echo "Usage: ${SCRIPT_NAME} <--all [--keep-images] | --ovn>"
+    echo "   --all         Clean all MicroShift and OVN data"
+    echo "   --keep-images Keep container images when cleaning all data"
+    echo "   --ovn         Clean OVN data only"
     exit 1
 }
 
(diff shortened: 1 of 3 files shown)


scripts/microshift-cleanup-data.sh
```diff
@@ -56,7 +57,8 @@ function stop_clean_pods() {
             if [ "$(echo "${ocp_pods}" | wc -w)" -eq 0 ] ; then
                 break
             fi
-            crictl rmp -f "${ocp_pods}" &>/dev/null
+            # shellcheck disable=SC2086
+            crictl rmp -f ${ocp_pods} &>/dev/null
```
ggiguash, line 61: This was a bug introduced when fixing shellcheck errors. The command expects multiple arguments, so quotes are not allowed.

scripts/microshift-cleanup-data.sh
```diff
@@ -116,17 +119,25 @@ if ${FULL_CLEAN} ; then
...
     done
 fi
 
+# Get the list of pod namespaces excluding 'openshift-ovn-kubernetes'
+ns_list=$(crictl pods -o json | jq -r '.items[].metadata.namespace' | sort | uniq | grep -v '^openshift-ovn-kubernetes$')
+# The 'openshift-ovn-kubernetes' namespace must be in the end of the list
+# to allow for pod deletion when MicroShift is stopped
+for ns in ${ns_list} openshift-ovn-kubernetes ; do
+    PODS_NS_LIST+=("${ns}")
+done
```
ggiguash, line 139: Namespace list is computed dynamically now, including whatever namespaces user created.

dhellmann: Why delete user workloads?

ggiguash: We're deleting OVN pods, so networking will break. I thought it might be good to clean up user pods as well.
@dhellmann , if you think it's best not to touch those, I can revert to a hard-coded list like previously.

pmtk:
There's also function `microshift_cleanup()` in `e2e/main.sh`, shall we change the invocation there as well?

ggiguash:
@pacevedom , please also update this doc section.
https://github.com/openshift/microshift/blob/main/docs/devenv_cloud.md?plain=1#L120

PR #1884, opened 2023-06-05
update asset verification script to look for redundant instructions

dhellmann:
Avoid issues like #1883 in the future

/assign @copejon @pmtk


copejon:
Tested locally.  I saw it catch duplicated values.  After fixing them, it succeeded. 

/lgtm

PR #1888, opened 2023-06-06
OCPBUGS-13078: Remove externalGatewayInterface from configure-ovs.sh

pliurh:
Ignore externalGatewayInterface flag in ovn.yaml.

### DIFF
--- a/packaging/systemd/configure-ovs.sh
+++ b/packaging/systemd/configure-ovs.sh
@@ -666,12 +666,6 @@ if [ "$1" == "OVNKubernetes" ]; then
     else
       rm -f "${iface_default_hint_file}"
     fi
-    externalGatewayInterface=$(get_cni_config "externalGatewayInterface")
-    if [ "$externalGatewayInterface" != "" ]; then
-      write_iface_external_hint "${extra_bridge_file}" "${externalGatewayInterface}"
-    else
-      rm -f "${extra_bridge_file}"
-    fi
   fi
   # Configures NICs onto OVS bridge "br-ex"
   # Configuration is either auto-detected or provided through a config file written already in Network Manager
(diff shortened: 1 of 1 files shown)


zshi-redhat:
@pliurh  do we need the logic in configure-ovs.sh to handle upgrade (e.g. remove the br-ex1)?

packaging/systemd/configure-ovs.sh
```diff
@@ -737,13 +731,8 @@ if [ "$1" == "OVNKubernetes" ]; then
...
   convert_to_bridge "$iface" "br-ex" "phys0" "${BRIDGE_METRIC}"
-  # Check if we need to configure the second bridge
-  if [ -f "$extra_bridge_file" ] && (! nmcli connection show br-ex1 &> /dev/null || ! nmcli connection show ovs-if-phys1 &> /dev/null); then
-    interface=$(head -n 1 $extra_bridge_file)
-    convert_to_bridge "$interface" "br-ex1" "phys1" "${BRIDGE1_METRIC}"
-  fi
-  # Check if we need to remove the second bridge
-  if [ ! -f "$extra_bridge_file" ] && (nmcli connection show br-ex1 &> /dev/null || nmcli connection show ovs-if-phys1 &> /dev/null); then
+  # Remove the second bridge
+  if [ -f "$extra_bridge_file" ] && (nmcli connection show br-ex1 &> /dev/null || nmcli connection show ovs-if-phys1 &> /dev/null); then
```
zshi-redhat, line 735: Shall we remove the check `[ -f "$extra_bridge_file" ]`?
In line 760, the second bridge and interfaces are not going to be activated even if $extra_bridge_file exist.

packaging/systemd/configure-ovs.sh
```diff
@@ -757,9 +746,6 @@ if [ "$1" == "OVNKubernetes" ]; then
     fi
   done < <(nmcli -g NAME c)
   connections+=(ovs-if-phys0 ovs-if-br-ex)
-  if [ -f "$extra_bridge_file" ]; then
```
zshi-redhat, line 760: the $extra_bridge_file is used in other places in the script, do we want to remove them all at once in this PR or leave it until the detach ovnk gateway change lands in microshift?

packaging/systemd/configure-ovs.sh
pliurh: This is for removing the existing extra bridge NM config files if they exist.

packaging/systemd/configure-ovs.sh
pliurh: We will remove most of the content from configure-ovs.sh in 4.14 when the detach ovnk gateway change lands. All the NM configurations will be removed when users upgrade to 4.14.

When we backport this PR to 4.13, for clusters that have br-ex1 configured, we need to keep `extra_bridge_file` in the script to ensure the behavior of `get_default_interface` is unchanged.

zshi-redhat: ack+.

PR #1892, opened 2023-06-06
USHIFT-1287, USHIFT-1315: prepare EC 1 build

dhellmann:
* restore build dependency on golang
* update image references to 4.14 EC 1 build

/assign @ggiguash @pacevedom

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -46,6 +46,17 @@ BuildRequires: make
 BuildRequires: policycoreutils
 BuildRequires: systemd
 
+# DO NOT REMOVE
+#
+# Without the dependency on the compiler via this file, the production
+# build pipeline does not work. Sometimes that pipeline requires
+# compiling with a version of go that is not released into public RHEL
+# repos, yet, so we do not specify a version here. Instead, the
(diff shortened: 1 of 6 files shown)


packaging/rpm/microshift.spec
```diff
@@ -46,6 +46,17 @@ BuildRequires: make
...
+# DO NOT REMOVE
+#
+# Without the dependency on the compiler via this file, the production
+# build pipeline does not work. Sometimes that pipeline requires
+# compiling with a version of go that is not released into public RHEL
+# repos, yet, so we do not specify a version here. Instead, the
+# version is checked in the prep step below.
+#
+BuildRequires: golang
+# DO NOT REMOVE
```
ggiguash, line 58: I think we can remove `DO NO REMOVE` statements. The comment seems to be enough on explaining why.

ggiguash:
@dhellmann , please also update [this line](https://github.com/openshift/microshift/blob/main/scripts/devenv-builder/configure-vm.sh#L85) in `configure-vm.sh` script to install the `golang` RPM. Otherwise, when installing the devenv, we won't be able to build an RPM.

PR #1914, opened 2023-06-10
USHIFT-1328: document how to pass options to the test runner

dhellmann:
Since we have a wrapper script for running the robot runner, it is not
necessarily obvious how to pass arguments through to the runner to
control which tests are run. Add a couple of examples to the
documentation.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -76,5 +76,36 @@ Options:
   -o DIR   The output directory.
 ```
 
+### Running a Single Suite
+
 By default, all of the test suites will be run. To run a subset,
 specify the filenames as arguments on the command line.
+
+```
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -76,5 +76,36 @@ Options:
...
+To run tests with specific tags, use the `-i` option:
+
+```
+$ ./run.sh -- -i etcd suites/*.robot
+```
+
+For more options, see the help output for the `robot` command.
+
+```
+$ ../_output/robotenv/bin/robot -h
```
ggiguash, line 110: In the commands, can we make sure they all run from the same location? The first command runs from the microshift root dir: `./test/run.sh` - I think we might want to fix the rest to be the same.

PR #1941, opened 2023-06-21
Increase retries and backoff of loadbalancer service smoke test

pliurh:

Closes #<Issue Number>

### DIFF
--- a/e2e/tests/0031-loadbalancer-smoke-test.sh
+++ b/e2e/tests/0031-loadbalancer-smoke-test.sh
@@ -9,8 +9,8 @@ set -x
 
 SCRIPT_PATH="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")"
 
-RETRIES=3
-BACKOFF=3s
+RETRIES=10
+BACKOFF=6s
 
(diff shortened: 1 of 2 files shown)


pmtk:
FYI There's also load balancer test in Robot Framework: https://github.com/openshift/microshift/blob/main/test/suites/load-balancer.robot
I think that is the one that is flaky

e2e/tests/0031-loadbalancer-smoke-test.sh
```diff
@@ -25,6 +25,8 @@ oc create -f "${SCRIPT_PATH}/assets/microshift-test.yaml"
 oc create service loadbalancer microshift-test --tcp=5678:8080
 oc wait pods -l app=microshift-test --for condition=Ready --timeout=60s
 
+timeout 10s bash -c "until kubectl get service microshift-test --output=jsonpath='{.status.loadBalancer}' | grep ingress; do sleep ${BACKOFF} ; done"
+
 for _ in $(seq "${RETRIES}"); do
```
ggiguash, line 30: Should we just increase the RETRIES=10 and BACKOFF=6, so that we test the service for 1m.

PR #1944, opened 2023-06-21
USHIFT-1088 Refuse to start if existing data is older than 1 y-stream

pmtk:


pkg/admin/prerun/version.go
```diff
@@ -118,5 +120,13 @@ func checkVersionDiff(execVer, dataVer versionMetadata) (bool, error) {
 		return false, fmt.Errorf("executable (%s) is older than existing data (%s): migrating data to older version is not supported", execVer.String(), dataVer.String())
 	}
 
+	if execVer.Minor > dataVer.Minor {
+		if execVer.Minor-1 == dataVer.Minor {
+			return true, nil
+		} else {
```
dhellmann, line 126: nit: this else isn't needed, the code will fall through to the alternative case.

dhellmann (commented):
/lgtm

PR #1948, opened 2023-06-22
USHIFT-1303: Separate libvirt installation and startup into manage-vm.sh script

ggiguash:
To speed up VM creation in the `metal-test` CI job, it is necessary to run the `create-vm.sh` scripts in parallel.
The problem is that `dnf` command executed from within the script corrupts its database when run on parallel.

This change is separation `libvirt` initialization so that is can be called before the VM creation is triggerred.
P.S. Also removing the obsolte `os-type` argument of `virt-install`

Prerequisite for https://github.com/openshift/release/pull/40471


scripts/devenv-builder/create-vm.sh
```diff
@@ -52,6 +49,12 @@ if [ "${SWAPSIZE}" -eq 0 ] ; then
     sed -i "s;^part swap;#part swap;" "${KICKSTART_FILE}"
 fi
 
+# Allow the VM creation to run in parallel using multiple instances of this script
+# Note: If 'dnf' command is run in parallel, its database is corrupted
+# === Start critical section ===
+exec {LOCK_FD}<"$0"
```
pmtk, line 55: Can you explain it or link documentation? I'm not familiar with this pattern

scripts/devenv-builder/create-vm.sh
```diff
@@ -52,6 +49,12 @@ if [ "${SWAPSIZE}" -eq 0 ] ; then
...
+flock --exclusive ${LOCK_FD}
+
 sudo dnf install -y libvirt virt-manager virt-install virt-viewer libvirt-client qemu-kvm qemu-img sshpass
```
dhellmann, line 58: Could we move the dnf commands to another script that is only run once? We have a couple of different configure scripts that could be candidates, or we could make a new one.

PR #1949, opened 2023-06-22
USHIFT-1087 Simple list of blocked upgrades

pmtk:


pkg/admin/prerun/upgrade_blocking.go
```diff
@@ -0,0 +1,59 @@
...
+
+	m, err := unmarshalBlockedUpgrades(buf)
+	if err != nil {
+		return false, err
+	}
+
+	return isBlocked(m, execVersion.String(), dataVersion.String()), nil
+}
+
+func getBlockedUpgradesAsset() ([]byte, error) {
```
dhellmann, line 30: Is using an asset easier to manage than just embedding the list in this go module? Are you thinking of automation managing that file (in which case, yes, I agree YAML is going to be easier)?

dhellmann (commented):
/lgtm

pkg/admin/prerun/upgrade_blocking.go
pmtk: At this point I don't have any idea how we'll use this feature and manage the list. Asset just felt natural since we already embed some files that way but it's not a strong feeling and we can change it

PR #1951, opened 2023-06-22
USHIFT-1363: tools to build multiple images for ci

dhellmann:
This PR adds a framework for managing "test scenarios". It is possible to independently define images, VMs, and tests, then mix and match them to produce scenarios like "test upgrade and rollback from MicroShift 4.13 to 4.14 on RHEL 9.2". The PR includes a few scenarios as examples. The scripts have been tested on a RHEL hypervisor using remote access similar to what we will do in CI. The next phase of implementation will include some wrappers and integration scripts for linking in to CI, but these scripts are usable as-is for developers to run locally.


test/bin/download_images.sh
```diff
@@ -0,0 +1,42 @@
...
+    blueprint=$(basename ${blueprint_build} .edge-commit)
+    buildid=$(cat ${blueprint_build})
+    download_image "${buildid}"
+    commit_file="${buildid}-commit.tar"
+    echo "Unpacking ${commit_file} ${blueprint}"
+    tar -C "${IMAGEDIR}" -xf "${commit_file}"
+done
+
+echo "Updating references"
+(cd "${IMAGEDIR}" && ostree --repo=repo refs)
```
pmtk, line 32: `ostree refs` only lists the references, `ostree summary -u` is required instead
are you sure we can just extract all commits into one directory without using `ostree pull-local`?

test/image-blueprints/rhel92-microshift413.toml
```diff
@@ -0,0 +1,25 @@
+name = "rhel-9.2-microshift-4.13"
```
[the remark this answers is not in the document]
dhellmann: 7711195ec6e94074080c419802288c66dc39b6b4

scripts/verify/verify-shell.sh
```diff
@@ -20,5 +20,9 @@ CHECK_FILE_LIST=$(find . \( -type d \( ${IGNORE_PATHS} \) -o -name "${IGNORE_FIL
 
 for f in ${CHECK_FILE_LIST} ; do
     echo "shellcheck: ${f}"
-    "${SHELL_CHECK}" -x "${f}"
+    "${SHELL_CHECK}" \
+        --format=gcc \
+        --exclude=SC1091 \
+        --external-sources \
+        "${f}"
```
ggiguash, line 27: Can you add a comment in the code on why these options are necessary?

test/bin/wait_images.py
```diff
@@ -0,0 +1,110 @@
...
+        if build_ids:
+            time.sleep(30)
+
+
+if __name__ == '__main__':
+    cli_parser = argparse.ArgumentParser(add_help=True)
+    cli_parser.add_argument(
+        'build_id',
+        nargs='+',
+        help="a build id (UUID) from composer",
```
ggiguash, line 107: This should probably be fixed to denote multiple build_id(s)?

test/bin/wait_images.py
```diff
@@ -0,0 +1,110 @@
...
+import logging
+import subprocess
+import time
+
+logging.basicConfig(
+    level=logging.INFO,
+    format='%(asctime)s: %(message)s',
+)
+
+STATUS_COMMAND = ['sudo', 'composer-cli', 'compose', 'status', '--json']
```
ggiguash, line 18: I think the non-json output is easier to parse.

test/bin/start_osbuild_workers.sh
```diff
@@ -0,0 +1,20 @@
...
+
+if [ $# -eq 0 ]; then
+    error "Usage: ${0} number-of-workers"
+    exit 1
+fi
+num_workers="${1}"
+
+# Loop from 2 because we should already have at least 1 worker.
+for i in $(seq 2 "${num_workers}"); do
+    sudo systemctl start "osbuild-worker@${i}.service"
```
ggiguash, line 19: What stops the workers after they are finished?

test/bin/start_webserver.sh
```diff
@@ -0,0 +1,15 @@
...
+set -euo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${SCRIPTDIR}/common.sh"
+
+echo "Starting web server in ${IMAGEDIR}"
+cd "${IMAGEDIR}"
+
+pkill caddy || true
+nohup caddy file-server --browse --listen "0.0.0.0:8080" --root "${IMAGEDIR}" --access-log &
```
ggiguash, line 15: Why do we need access logs? It slows down the server.

ggiguash: Since we're running this on the main hypervisor host, I think we should use another port. 
The chances of 8080 being taken are very high.

test/bin/force_vm_settings.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -x
+#
+# This script is used by scenario.sh when creating a new VM to force
+# specific settings on the host, such as opening firewall ports.
+
+set -euo pipefail
+
```
ggiguash, line 7: If there's an assumption that we run with SUID==0, should we check it in the script?

test/bin/configure_hypervisor_firewall.sh
```diff
@@ -0,0 +1,15 @@
...
+# firewall for incoming connections to the VM network.
+
+set -euo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${SCRIPTDIR}/common.sh"
+
+VM_BRIDGE=$(sudo virsh net-info default | grep '^Bridge:' | awk '{print $2}')
+VM_BRIDGE_CIDR=$(ip -f inet addr show "${VM_BRIDGE}" | grep inet | awk '{print $2}')
+sudo firewall-cmd --permanent --zone=trusted --add-source="${VM_BRIDGE_CIDR}"
```
ggiguash, line 13: Why do we need this if we're explicitly opening port 8080 in the public zone? What other connections over the trusted network do we expect?

test/bin/composer_cleanup.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+#
+# This script cancels any composer jobs and deletes failed and completed jobs.
```
ggiguash, line 3: Should we also spin down the worker services here?

pmtk (commented):
Some minor things. Did you start working on incorporating the changes in `release` repo?

test/README.md
```diff
@@ -109,3 +109,187 @@ For more options, see the help output for the `robot` command.
...
+Composer. If you build new RPMs, you need to re-run
+`create_local_repo.sh` *and* build new images.
+
+### Creating the images
+
+Use `./bin/build_images.sh` to build all of the images for all of the
+blueprints available.
+
+Use `./bin/start_osbuild_workers.sh` to create multiple workers for
+building images in parallel.
```
pmtk, line 175: Maybe we could change the order? :)

scripts/image-builder/configure.sh
```diff
@@ -6,15 +6,15 @@ OSVERSION=$(awk -F: '{print $5}' /etc/system-release-cpe)
 sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
     cockpit-composer cockpit-machines bash-completion podman runc genisoimage \
     createrepo yum-utils selinux-policy-devel jq wget lorax rpm-build \
-    containernetworking-plugins
+    containernetworking-plugins caddy
```
pmtk, line 9: `caddy` is in the EPEL, it should be move to L17

test/bin/scenario.sh
```diff
@@ -0,0 +1,378 @@
...
+    local variable_file="${ARTIFACT_DIR}/e2e/${SCENARIO}/variables.yaml"
+    echo "Writing variables to ${variable_file}"
+    mkdir -p "$(dirname "${variable_file}")"
+    cat - <<EOF | tee "${variable_file}"
+VM_IP: ${vm_ip}
+API_PORT: ${api_port}
+LB_PORT: ${lb_port}
+USHIFT_HOST: ${public_ip}
+USHIFT_USER: redhat
+SSH_PRIV_KEY: ""
```
pmtk, line 280: I guess this need to be configurable?

test/bin/scenario.sh
```diff
@@ -0,0 +1,378 @@
...
+        --randomize all \
+        --loglevel TRACE \
+        --outputdir "${ARTIFACT_DIR}/e2e/${SCENARIO}" \
+        -x junit.xml \
+        -V "${variable_file}" \
+        "$@"
+}
+
+load_global_settings() {
+    local filename="${TESTDIR}/scenario_settings.sh"
```
pmtk, line 295: A template for this file would be nice

test/README.md
```diff
@@ -109,3 +109,187 @@ For more options, see the help output for the `robot` command.
...
+
+Add blueprints as TOML files in the `image-blueprints` directory, then
+add a short description of the image here for reference.
+
+Name | Purpose
+---- | -------
+rhel-9.2 | A simple RHEL image without MicroShift.
+rhel-9.2-microshift-4.13 | A RHEL 9.2 image with the latest MicroShift 4.13 z-stream installed and enabled.
+rhel-92-microshift-source | A RHEL 9.2 image with the RPMs built from source.
+rhel-90-microshift-source | A RHEL 9.0 image with the RPMS built from source.
```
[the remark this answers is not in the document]
dhellmann: I think the 9.2 vs 92 was a typo, I'll fix that.

Regarding the `/`, I was trying to be conservative about which characters are in the name because the same name is used as the blueprint name and ostree ref ID to make it easier for someone writing a test scenario to make sure they have the right value in place. I'm not sure if a blueprint can have a `/` in the name. If it can, and the scripting that uses the names won't be confused by the `/`, we could make that change. Can we do that in a follow-up? After we have all of this wired in and can automatically test a change it will be easier to ensure we have all of the changes we need.

test/README.md
dhellmann: Done

test/README.md
```diff
@@ -109,3 +109,187 @@ For more options, see the help output for the `robot` command.
...
+images needed for the test scenarios.
+
+Use `./bin/configure_hypervisor_firewall.sh` to set up the firewall
+rules that allow VMs to access the web server on the hypervisor.
+
+Use `./bin/scenario.sh` to create test infrastructure for a scenario
+with the `create` argument and a scenario directory name as input.
+
+```
+$ ./bin/scenario.sh create ./scenarios/rhel-9.2-microshift-source-standard-suite/
```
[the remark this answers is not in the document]
dhellmann: There's a section about global settings, but it looks like I left out the filename. Fixed.

test/bin/build_images.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo "Package source: ${name}"
+    sudo composer-cli sources info "${name}" | sed -e 's/gpgkeys.*/gpgkeys = .../g'
+done
+
+# Given a blueprint filename, extract the name value. It does not have
+# to match the filename, but some commands take the file and others
+# take the name, so we need to be able to have both.
+get_blueprint_name() {
+    local filename="${1}"
+    grep name "${filename}" | head -n 1 | cut -f2 -d= | sed -e 's/ //g' -e 's/"//g'
```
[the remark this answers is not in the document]
dhellmann: Oh, that is much nicer, yes. I'll make a note to update to use that. Can we do that in another PR?

test/bin/build_images.sh
```diff
@@ -0,0 +1,117 @@
...
+}
+
+# Track some dynamically created values
+BLUEPRINTS=""
```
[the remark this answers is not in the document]
dhellmann: It looks like it is not. Removed.

test/bin/configure_hypervisor_firewall.sh
dhellmann: I'm not entirely sure. I was not able to get the VMs to be able to see the server without this change. Maybe because by default the VMs cannot talk to the hypervisor?

test/bin/scenario.sh
```diff
@@ -0,0 +1,378 @@
...
+    wait_for_ssh "${ip}"
+}
+
+# Wait for greenboot health check to complete, without checking the results
+wait_for_greenboot() {
+    local vmname="${1}"
+    local ip="${2}"
+
+    echo "Waiting 5m for greenboot on ${vmname} to complete"
+    timeout 5m bash -c "until ssh redhat@${ip} \"sudo systemctl status greenboot-healthcheck | grep 'active (exited)'\"; do date; sleep 10; done"
```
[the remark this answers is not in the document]
dhellmann: Do you think the timeout needs to be configurable as part of the scenario? Or should we just make it larger? FWIW, this value is working today, but it's a bit tight.

test/bin/scenario.sh
dhellmann: Yes, I have a note to do that as part of the next step of integrating these scripts with the CI system, where we have a key but no agent.

test/bin/scenario.sh
dhellmann: Added

test/bin/start_osbuild_workers.sh
dhellmann: The command does not seem to report an error, no.

Nothing stops the workers. In CI, we're going to kill the entire host. On a local system, I expect the user to manage the processes and this is just a convenience script for starting several at once.

test/bin/wait_images.py
dhellmann: The help output shows that the option can be repeated, but the script isn't meant for anyone to use directly. It is called by build_images.sh.

scripts/image-builder/configure.sh
dhellmann: Thanks, fixed

dhellmann (commented):
@pmtk @ggiguash I think I've addressed all of your feedback. There were a couple of items that I think will be easier to deal with in future PRs, so I left those comment threads unresolved. I think this is ready for another review.

test/bin/composer_cleanup.sh
dhellmann: They don't take up many resources when there are no images to build.

test/package-sources/fast-datapath-rhel9.toml
```diff
@@ -0,0 +1,8 @@
+id = "fast-datapath"
+name = "Fast Datapath for RHEL 9"
+type = "yum-baseurl"
+url = "https://cdn.redhat.com/content/dist/layered/rhel9/${UNAME_M}/fast-datapath/os"
+check_gpg = true
+check_ssl = false
```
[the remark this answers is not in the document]
dhellmann: I copied this from somewhere. I'm not sure.

test/bin/force_vm_settings.sh
dhellmann: This script is invoked from `scenario.sh` using sudo. It's not meant to be used directly.

test/package-sources/fast-datapath-rhel9.toml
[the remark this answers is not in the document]
dhellmann: It seems to work if I enable ssl verification, so I've done that in both package source files.

test/bin/composer_cleanup.sh
ggiguash: Still, it might be a logical thing to do here, right?

test/bin/force_vm_settings.sh
ggiguash: We might want to either add sudo in the script or check for SUID==0 for consistency.
I'm sure we'd want to run this script directly because it's conveniently setting up the environment.

test/bin/start_osbuild_workers.sh
ggiguash: Since we have the "cancel" procedure, it seems like the right place to spin down the workers.
This is mostly required for the devenv, of course. But, also for consistency.

test/bin/start_webserver.sh
dhellmann: I added a variable to adjust this if we need to.

test/bin/force_vm_settings.sh
dhellmann: Done

test/bin/build_images.sh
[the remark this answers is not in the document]
dhellmann: done

test/bin/scenario.sh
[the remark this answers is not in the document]
dhellmann: I extended the timeout to 8 minutes and set up a variable to make it easy to adjust. Scenarios should even be able to override that value, if they need to.

test/README.md
```diff
@@ -109,3 +109,189 @@ For more options, see the help output for the `robot` command.
...
+Package source definition files are templates using `envsubst`,
+which means the files can have shell variables embedded.
+
+```
+id = "fast-datapath"
+name = "Fast Datapath for RHEL 9"
+type = "yum-baseurl"
+url = "https://cdn.redhat.com/content/dist/layered/rhel9/${UNAME_M}/fast-datapath/os"
+check_gpg = true
+check_ssl = false
```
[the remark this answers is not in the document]
dhellmann: Done

test/bin/scenario.sh
dhellmann: Done

pmtk (commented):
overall looks good to me, waiting for final version :)

test/bin/ci_phase_iso_boot.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
dhellmann: fixed

test/image-blueprints/rhel92.toml
```diff
@@ -0,0 +1,7 @@
+name = "rhel-9.2"
+description = ""
+version = "0.0.1"
+distro = "rhel-92"
+modules = []
+groups = []
+packages = []
```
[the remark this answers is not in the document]
dhellmann: fixed

test/README.md
```diff
@@ -109,3 +109,226 @@ For more options, see the help output for the `robot` command.
...
+configured before the tool is used in `./scenario_settings.sh`. You
+can copy `./scenario_settings.sh.example` as a starting point.
+
+`PUBLIC_IP` -- The public IP of the hypervisor, when accessing VMs
+remotely through port-forwarded connections.
+
+`SSH_PUBLIC_KEY` -- The name of the public key file to use for
+providing password-less access to the VMs.
+
+`SSH_PUBLIC_KEY` -- The name of the private key file to use for
```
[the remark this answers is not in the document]
dhellmann: fixed

test/bin/scenario.sh
```diff
@@ -0,0 +1,352 @@
...
+    ssh "redhat@${ip}" "chmod +x /home/redhat/force_vm_settings.sh && sudo /home/redhat/force_vm_settings.sh"
+}
+
+# Wait for greenboot health check to complete, without checking the results
+wait_for_greenboot() {
+    local vmname="${1}"
+    local ip="${2}"
+
+    echo "Waiting ${VM_BOOT_TIMEOUT} for greenboot on ${vmname} to complete"
+    timeout "${VM_BOOT_TIMEOUT}" bash -c "until ssh redhat@${ip} \"sudo journalctl -n 5 -u greenboot-healthcheck; sudo systemctl status greenboot-healthcheck | grep 'active (exited)'\"; do date; sleep 10; done"
```
pmtk, line 102: Are we sure `-n 5` is enough?

pmtk (commented):
just some nits we can address in separate PR
/lgtm
/hold

test/bin/scenario.sh
pmtk: I missed the loop :)

PR #1955, opened 2023-06-23
USHIFT-1361: Test happy ostree upgrade

pmtk:


test/suites-ostree/upgrade.robot
```diff
@@ -0,0 +1,88 @@
...
+    Login MicroShift Host
+
+Teardown
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+
+Perform And Verify Upgrade
+    [Documentation]    Template for performing a clean upgrade from one ref to another
+    ...    and verifying it was successful
+    [Arguments]    ${initial_ref}    ${target_ref}    ${cleanup}=${TRUE}
```
pmtk, line 41: This keyword doesn't care what initial state of the machine is - it'll clean up microshift, deploy 1st commit, then another.
It might not be what we want to some scenarios, but it might be useful for periodic and is definitely useful for development (no need to manually go back to initial commit)

dhellmann: Where does it get those values? Is it possible to make the test suite take inputs for the origin and destination image references so we can reuse it for multiple scenarios?

test/suites-ostree/upgrade.robot
```diff
@@ -0,0 +1,85 @@
+*** Settings ***
+Documentation       Tests related to MicroShift upgradeability on OSTree-based systems
```
dhellmann, line 2: We may want to name this file "healthy-upgrade" because I expect the flow for an unhealthy upgrade to be a bit different.

test/resources/systemd.resource
```diff
@@ -56,6 +56,8 @@ Systemctl    # robocop: disable=too-long-keyword
     END
     Should Be Equal As Integers    0    ${rc}
 
+    IF    "${verb}" == "enable"    RETURN
```
dhellmann, line 59: The docstring for the keyword should be updated if it supports "enable" as a verb.

dhellmann (commented):
One question and a couple of nits. I'll approve this with a hold and you can decide if you want to update this PR or submit another with the updates.

/lgtm
/hold

test/suites-ostree/upgrade.robot
pmtk: This keyword is used as a "test template" by test case `Upgrade`. Each line in the test case `Upgrade` contains args passed to this keyword.
https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#test-templates

I'm not sure if we'll use it like this. If we can spawn N VMs with desired initial ref, then we can cut the part about deploying initial ref but it's very very useful for local development

PR #1956, opened 2023-06-23
USHIFT-1085: feat: initial implementation of storage migration

eggfoobar:
Initial implementation for storage migrator

This currently only contains the code for the migrator itself. I tested a few ways of running this for performance, we're looking at around 2.8 seconds for a migration to run.

The calling code isn't present, I'll open up a new PR for that logic.

/assign @pmtk

### DIFF
--- a/pkg/admin/migration/migration.go
+++ b/pkg/admin/migration/migration.go
@@ -26,7 +26,7 @@ import (
 	"k8s.io/klog/v2"
 )
 
-var blackListResources = sets.NewString(
+var excludeResources = sets.NewString(
 	"events",
 )
 
@@ -42,23 +42,23 @@ type migrator struct {
 func NewMigrator(kubeConfigPath string) (*migrator, error) {
 	restConfig, err := clientcmd.BuildConfigFromFlags("", kubeConfigPath)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to build rest config: %w", err)
 	}
 	clientset, err := kubernetes.NewForConfig(restConfig)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to build kubernetes clientset config: %w", err)
 	}
 	crd, err := crdclient.NewForConfig(restConfig)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to build crd clientset config: %w", err)
 	}
 	apiservice, err := apiserviceclient.NewForConfig(restConfig)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to build apiservice client config: %w", err)
 	}
 	dynamic, err := dynamic.NewForConfig(restConfig)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to build dynamic client config: %w", err)
 	}
 
 	return &migrator{
@@ -92,12 +92,12 @@ func (d *migrator) start(ctx context.Context) (*MigrationResultList, error) {
 		go func(wg *sync.WaitGroup, sch schema.GroupVersionResource) {
 			defer wg.Done()
 			objectList := &unstructured.UnstructuredList{}
-			var migrationErr error
 
-			migrationErr = retry.OnError(retry.DefaultBackoff, canRetry, func() error {
-				objectList, migrationErr = d.list(ctx, sch, metav1.ListOptions{})
-				if migrationErr != nil {
-					return migrationErr
+			migrationErr := retry.OnError(retry.DefaultBackoff, canRetry, func() error {
+				var err error
+				objectList, err = d.client.Resource(sch).Namespace(metav1.NamespaceAll).List(ctx, metav1.ListOptions{})
+				if err != nil {
+					return err
 				}
 				return nil
 			})
@@ -106,35 +106,49 @@ func (d *migrator) start(ctx context.Context) (*MigrationResultList, error) {
 				errorOccured = true
 				migrationErr = fmt.Errorf("could not list resources: %+v", migrationErr)
 				lock.Lock()
+				defer lock.Unlock()
 				results.Items = append(results.Items, MigrationResult{Error: migrationErr, ResourceVersion: sch, Timestamp: time.Now()})
-				lock.Unlock()
 				return
 			}
 
+			localResults := &MigrationResultList{Status: MigrationSuccess}
 			for _, object := range objectList.Items {
 				ref := object
 				migrationErr := d.migrateOneItem(ctx, sch, &ref)
 				if migrationErr != nil {
 					errorOccured = true
 				}
-				lock.Lock()
-				results.Items = append(results.Items, MigrationResult{
+				localResults.Items = append(localResults.Items, MigrationResult{
 					Error:           migrationErr,
 					ResourceVersion: sch,
-					ObjectMeta:      getObjectMeta(&ref),
+					NamespacedName:  getNamespacedName(&ref),
 					Timestamp:       time.Now()})
-				lock.Unlock()
 			}
+
+			lock.Lock()
+			defer lock.Unlock()
+			results.Items = append(results.Items, localResults.Items...)
 		}(&wg, sch)
 	}
+	wg.Wait()
 	if errorOccured {
 		results.Status = MigrationFailure
 	}
-	wg.Wait()
-	klog.Infof("schema migration finished, it took %s to complete", time.Since(start).String())
+	klog.InfoS("schema migration finished", "duration", time.Since(start).String())
 	return results, nil
 }
 
+// findMigratableResources finds all the resources that potentially need
+// migration. Although all migratable resources are accessible via multiple
+// versions, the returned list only include one version.
+//
+// It builds the list in these steps:
+// 1. build a map from resource name to the groupVersions, excluding subresources, custom resources, or aggregated resources.
+// 2. exclude all the resource that is only available from one groupVersions.
+// 3. exclude the resource that does not support "list" and "update" (thus not migratable).
+//
+// More information can be found here:
+// https://github.com/kubernetes-sigs/kube-storage-version-migrator/blob/acdee30ced218b79e39c6a701985e8cd8bd33824/pkg/initializer/discover.go#L55-L125
 func (d *migrator) findMigratableResources(ctx context.Context) ([]schema.GroupVersionResource, error) {
 	customGroups, err := d.findCustomGroups(ctx)
 	if err != nil {
@@ -152,15 +166,15 @@ func (d *migrator) findMigratableResources(ctx context.Context) ([]schema.GroupV
 	for _, resourceList := range resourceLists {
 		gv, err := schema.ParseGroupVersion(resourceList.GroupVersion)
 		if err != nil {
-			klog.Errorf("cannot parse group version %s, ignored", resourceList.GroupVersion)
+			klog.ErrorS(err, "cannot parse group version, ignored", "groupVersion", resourceList.GroupVersion)
 			continue
 		}
 		if customGroups.Has(gv.Group) {
-			klog.V(4).Infof("ignored group %v because it's a custom group", gv.Group)
+			klog.InfoS("ignored because it's a custom group", "group", gv.Group)
 			continue
 		}
 		if aggregatedGroups.Has(gv.Group) {
-			klog.V(4).Infof("ignored group %v because it's an aggregated group", gv.Group)
+			klog.InfoS("ignored because it's an aggregated group", "group", gv.Group)
 			continue
 		}
 		for _, r := range resourceList.APIResources {
@@ -168,7 +182,8 @@ func (d *migrator) findMigratableResources(ctx context.Context) ([]schema.GroupV
 			if strings.Contains(r.Name, "/") {
 				continue
 			}
-			if blackListResources.Has(r.Name) {
+			// ignore excluded resources
+			if excludeResources.Has(r.Name) {
 				continue
 			}
 			// ignore resources that cannot be listed and updated
@@ -183,6 +198,8 @@ func (d *migrator) findMigratableResources(ctx context.Context) ([]schema.GroupV
 
 	ret := []schema.GroupVersionResource{}
 	for resource, groupVersions := range resourceToGroupVersions {
+		// if a resource only has one version, no migration is required
+		// resources that have more than one version are eligible for migration.
 		if len(groupVersions) == 1 {
 			continue
 		}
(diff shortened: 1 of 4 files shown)


pmtk:
What do you think about changing the order: first merging "partial start" (etcd+KAS) in places where migration would happen, and then adding code from this PR so migrator could already be executed?

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+					Timestamp:       time.Now()})
+				lock.Unlock()
+			}
+		}(&wg, sch)
+	}
+	if errorOccured {
+		results.Status = MigrationFailure
+	}
+	wg.Wait()
+	klog.Infof("schema migration finished, it took %s to complete", time.Since(start).String())
```
pmtk, line 134: BTW. Working on upgrade epic we (or I) moved to using `klog.InfoS`, trying to follow [K8s' developers logging guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md)

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+	client           dynamic.Interface
+	discoveryClient  discovery.ServerResourcesInterface
+	crdClient        v1.CustomResourceDefinitionInterface
+	apiserviceClient apiregistrationv1.APIServiceInterface
+}
+
+func NewMigrator(kubeConfigPath string) (*migrator, error) {
+	restConfig, err := clientcmd.BuildConfigFromFlags("", kubeConfigPath)
+	if err != nil {
+		return nil, err
```
pmtk, line 45: Not sure what kind of errors are produced here, but we might want to wrap them and provide more context, e.g.: 
```suggestion
		return nil, fmt.Errorf("failed to build rest config: %w", err)
```

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	for _, resourceList := range resourceLists {
+		gv, err := schema.ParseGroupVersion(resourceList.GroupVersion)
+		if err != nil {
+			klog.Errorf("cannot parse group version %s, ignored", resourceList.GroupVersion)
+			continue
+		}
+		if customGroups.Has(gv.Group) {
+			klog.V(4).Infof("ignored group %v because it's a custom group", gv.Group)
```
pmtk, line 159: I'd be careful with `V(4)`. Currently we only show `0` level. We might (at least for some time) have them on level 0 and see how many logs this produces

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/retry"
+	apiserviceclient "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
+
+	"k8s.io/klog/v2"
+)
+
+var blackListResources = sets.NewString(
```
pmtk, line 29: How about `excludeResources`?

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+			resourceToGroupVersions[r.Name] = gvs
+		}
+	}
+
+	ret := []schema.GroupVersionResource{}
+	for resource, groupVersions := range resourceToGroupVersions {
+		if len(groupVersions) == 1 {
+			continue
+		}
+		ret = append(ret, groupVersions[0].WithResource(resource))
```
pmtk, line 189: So what happens here: we are building a list of GVRs that at least two `groupVersion`s?
We don't care about `len == 0` because if there's `resource : groupVersions` in this map (added in L178-180), then `groupVersions` has at least 1 item?

Some explanation in the code would be nice

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+			continue
+		}
+		// error is not retriable
+		return fmt.Errorf("can not retry: %+v", err)
+	}
+}
+
+func (m *migrator) try(ctx context.Context, resource schema.GroupVersionResource, namespace, name string, item *unstructured.Unstructured, get bool) (bool, error) {
+	var err error
+	if get {
```
pmtk, line 232: Do we really need that? Could we always "get before put"?

I can image it's an extra load but at this point nothing is running on the cluster so MicroShift gets all the resources for itself.
On the other hand, what's the chance of the conflict if nothing is running... (but I'd rather have a "get before put" than end up with conflict for some weird reason that retries would remediate)

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+				time.Sleep(time.Duration(seconds) * time.Second)
+			case delay:
+				klog.Warningf("migration of %s will be retried after a %ds delay: %v", name, seconds, err)
+				time.Sleep(time.Duration(seconds) * time.Second)
+			case !delay && len(namespace) > 0:
+				klog.Warningf("migration of %s, in the %s namespace, will be retried: %v", name, namespace, err)
+			default:
+				klog.Warningf("migration of %s will be retried: %v", name, err)
+			}
+			continue
```
pmtk, line 223:
```suggestion
			seconds, delay := errors.SuggestsClientDelay(err)
			klog.ErrorS(err, "migration of an object will be retried", "name", name, "namespace", namespace, "delay", seconds)
			if delay {
			    time.Sleep(time.Duration(seconds) * time.Second)
			}
			continue
```

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+		Resource(resource).
+		Namespace(namespace).
+		Update(ctx, item, metav1.UpdateOptions{})
+	if err == nil {
+		return false, nil
+	}
+	return errors.IsConflict(err), err
+}
+
+func (m *migrator) list(ctx context.Context, resource schema.GroupVersionResource, options metav1.ListOptions) (*unstructured.UnstructuredList, error) {
```
pmtk, line 251: This function appears to be used only once, do we need it?

pkg/admin/migration/types.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Container for individual migration attempts
+type MigrationResult struct {
+	Error           error                       `json:"error,omitempty"`
+	ResourceVersion schema.GroupVersionResource `json:"resourceVersion"`
+	Timestamp       time.Time                   `json:"timestamp"`
+	ObjectMeta      *metav1.ObjectMeta          `json:"object,omitempty"`
+}
+
+type MigrationResultList struct {
```
pmtk, line 27: Not sure if we need that. Is there a way to log that some resource was migration from version A to B? We could log that in "real time", not packing all results into a collection and then iterating to log the items

pkg/admin/migration/types.go
```diff
@@ -0,0 +1,70 @@
...
+	MigrationSuccess MigratonStatus = "success"
+	MigrationFailure MigratonStatus = "failure"
+)
+
+// Container for individual migration attempts
+type MigrationResult struct {
+	Error           error                       `json:"error,omitempty"`
+	ResourceVersion schema.GroupVersionResource `json:"resourceVersion"`
+	Timestamp       time.Time                   `json:"timestamp"`
+	ObjectMeta      *metav1.ObjectMeta          `json:"object,omitempty"`
```
pmtk, line 24: We could think about using [NamespacedName](https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go#L27) if this is what we need from this

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+			objectList := &unstructured.UnstructuredList{}
+			var migrationErr error
+
+			migrationErr = retry.OnError(retry.DefaultBackoff, canRetry, func() error {
+				objectList, migrationErr = d.list(ctx, sch, metav1.ListOptions{})
+				if migrationErr != nil {
+					return migrationErr
+				}
+				return nil
+			})
```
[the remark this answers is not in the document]
eggfoobar: oops, good catch, yeah this was caught in a renaming refactor

pkg/admin/migration/migration.go
eggfoobar: It's more there for a safety concern, it won't execute unless there's a conflict, so if we do decide to run this during normal operation or what may happen, this would cover us.

pkg/admin/migration/migration.go
eggfoobar: Nah, we can get rid of it

pkg/admin/migration/types.go
eggfoobar: Oh right, perfect, thanks!

pkg/admin/migration/migration.go
eggfoobar: Sounds good, updating them

pkg/admin/migration/migration.go
eggfoobar: Great idea

pkg/admin/migration/migration.go
eggfoobar: Sure thing, updated

pkg/admin/migration/migration.go
eggfoobar: I like it, changing it

pkg/admin/migration/types.go
eggfoobar: Sure thing, we can do that. The primary reason it's here is just to give us raw access to write the information in any format we want if we need to store the results somewhere easily.

pkg/admin/migration/migration.go
eggfoobar: Good point, let me add some comments there and link back to their original documentation as well. One note, the original docs call out a TODO for improvements, but I can't find any indication that, that KEP has been implemented.

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,282 @@
...
+
+			if migrationErr != nil {
+				errorOccured = true
```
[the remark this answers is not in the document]
dhellmann: Data errors are recoverable, and it may make sense to collect all of them unless it makes the logic significantly more complicated.

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	errorOccured := false
+	start := time.Now()
+	klog.Info("schema migration started")
+
+	wg := sync.WaitGroup{}
+	lock := sync.Mutex{}
+	wg.Add(len(schemas))
+	for _, sch := range schemas {
+		go func(wg *sync.WaitGroup, sch schema.GroupVersionResource) {
```
dhellmann, line 92: How much of a performance benefit do we get by parallelizing this work? How many different workers are there running in parallel?

pkg/admin/migration/migration.go
```diff
@@ -0,0 +1,291 @@
...
+				if err != nil {
+					return err
+				}
+				return nil
+			})
+
+			if migrationErr != nil {
+				errorOccured = true
+				migrationErr = fmt.Errorf("could not list resources: %+v", migrationErr)
+				lock.Lock()
```
dhellmann, line 108: This seems like a good opportunity to use a channel to communicate between the goroutines, instead managing a lock.

pkg/admin/migration/migration.go
dhellmann: Do you see a benefit in running this version of storage migration code under "normal" operations, instead of running the standard controller?

pkg/admin/migration/types.go
dhellmann: Is the pre-run phase logging to the standard MicroShift log, or do we see these errors in the greenboot health check log? It would be nice if the greenboot log could at least report "there was a data migration error, check the MicroShift logs for details".

pkg/admin/migration/migration.go
eggfoobar: Honestly I kept going back and forth on doing things concurrently here, figured this might help incase there are a lot of schemas, but at our size there is no appreciable performance gain, there would be a worker per schema group here. I wouldn't mind getting rid of it to reduce complexity

dhellmann: Let's see how we can do without the complexity, then. If it turns into a problem, we'll have some understanding of _why_ and we can apply the right level of parallelism (a fixed number of workers, a worker per group, a worker per resource, etc.).

pkg/admin/migration/migration.go
eggfoobar: The one benefit I can think of is having an easier user experience path for re-running a migration instead of having to install the storage migration controller. In the event you would want to execute a migration after a resource conflict has been resolved manually, having a cockpit plugin button or running a `microshift admin migrate` command I think would be for a better flow. This assumes it would be the admin who would install it and not us.

pkg/admin/migration/types.go
eggfoobar: It's currently just using klog, I'm not sure how we send logs to greenboot but I love the idea. We should make it do that

pkg/admin/migration/migration.go
dhellmann: If the service won't start, there's not going to be a way to run any code that modifies the data, right? The user is going to have to roll back and do something to the system using the old code versions.

What sorts of "conflicts" are we talking about?

pkg/admin/migration/types.go
dhellmann: The klog output is going to go to stdout/stderr. I guess that's going to the systemd unit where pre-run is running, rather than a greenboot-specific log. Can we put these messages (and any others) somewhere for our health-check script to collect and report?

pkg/admin/migration/migration.go
eggfoobar: Sounds good, removed for now

pkg/admin/migration/migration.go
eggfoobar: Yeah you're right, this is over preparing for a future where a user might want to re-run a migration that takes a long time during normal operation. Removed it.

pkg/admin/migration/types.go
eggfoobar: Sure, I added some files here to use in greenboot, https://github.com/openshift/microshift/pull/1956/files#diff-e78e17834c76663b756393d04ebe7aceb8ea4059d44b7d2ae3ee4e9999aead6eR23-R24

eggfoobar:
/close

Closing PR in favor of other approaches. In this PR we were trying to implement this feature with out the cluster being fully up, upon further investigation, we will need the cluster up in order to fully support CRDs and their migration webhooks.

PR #1957, opened 2023-06-23
USHIFT-1091 USHIFT-1095: Handle upgrade from 4.13

pmtk:


pkg/admin/prerun/prerun.go
```diff
@@ -45,12 +47,14 @@ func New(dataManager data.Manager) *PreRun {
 func (pr *PreRun) Perform() error {
 	klog.InfoS("Starting pre-run")
 
+	if continuePrerun, err := pr.checkPreconditions(); err != nil {
```
dhellmann, line 50: It would be good to log before and after this call, since it's a distinct unit of work.

pkg/admin/prerun/prerun.go
```diff
@@ -109,6 +115,109 @@ func (pr *PreRun) Perform() error {
...
+
+		// 2
+		// TODO: Health needs to be extended into a history of boots and their health
+		// so prerun can look into the past and decide if a backup should be restored
+		// for currently running deployment
+		return false, fmt.Errorf("TODO: data is missing, but health file exists")
+	}
+
+	// 5
+	if !versionExists && !healthExists {
```
dhellmann, line 174: It would be good to be consistent about either having each case be 1 `if` statement with combined conditions (like here) or to use nested `if` statements (like above). I don't have a strong preference for which approach is used, but if nested conditions make it easier to ensure we are testing all of the relevant cases maybe that's the thing to do?

pkg/admin/prerun/prerun.go
```diff
@@ -109,6 +115,109 @@ func (pr *PreRun) Perform() error {
...
+		klog.InfoS("Data dir exists, but health and version files are missing: assuming upgrade from 4.13")
+		return false, pr.upgradeFrom413()
```
dhellmann, line 176: Why not perform the pre-run process when moving from 4.13?

pmtk: I tried that in first version (https://github.com/openshift/microshift/commit/333216d4bab887a7f4aacb2ff2b866f89b1c2acd#diff-e78e17834c76663b756393d04ebe7aceb8ea4059d44b7d2ae3ee4e9999aead6eR65) but I didn't like the workaround to fit in that process (manually constructing "4.13" + "healthy" and generating backup name from deployId without boot id)

pkg/admin/prerun/prerun.go
pmtk: What I like about current version is clear distinction of cases where data does and does not exist (seems to me to be kind of a big deal that deserves special treatment).
Lines from 174 to the end of the function could be inside an `else` but that's not idiomatic...
TBH I'm most happy with the current version - it's explicit that `if !dataExists` exhausts all the cases when data doesn't exist, and the rest are cases when it does. Turning that into `if !dataExists && !healthExists` and `if !dataExists && healthExists` makes me spend some brain cycles to think if there are other, implicit cases with `!dataExists`

pkg/admin/prerun/prerun.go
pmtk: I also expect that when we get to solving TODOs in this function, it might get renamed to resolve these special cases so all of this will be part of pre-run and what we have today will be "regular flow"

dhellmann:
/lgtm

That logic is easier to follow now, nice cleanup.

PR #1933, opened 2023-06-15
USHIFT-1304 [Feature]: CSI Snapshotting support

copejon:
This PR is an implementation of the ongoing work on PR openshift/enhancements#1406. Refer to the document there for details.

Merge Reqs:

- [x] code implementation
- [x] automated rebase enablement
~- [ ] documentation~ to be done in a follow-on PR
- [x] ~smoke-test~ robotframework test

PR #1966, opened 2023-06-27
[NO-ISSUE] use kustomize to generate smoke-test resources

copejon:
This was originally part of #1933.  This changes the way smoke-test assets are managed.  As a standalone change it's not that impactful (only the resize test is affected).

The snapshotting feature will add a handful more smoke-test assets and it will be more manageable to use a base pod definition and patch in varying PVC definitions.  The alternative is to use manifests that are almost completely identical, except for the claim.

This PR also fixes a script bug in the e2e/main.sh file by removing the indentation in the here-doc.


copejon:
Strike that, will leave these manifests as is and implement the changes in concert with #1933

PR #1969, opened 2023-06-28
USHIFT-1102: add static manifest details for running openshift-tests

chiragkyal:
Add documentation for applying static manifests before running openshift-tests. 
The implementation details are [here](https://github.com/openshift/origin/pull/27714).


docs/openshift_ci.md
```diff
@@ -12,7 +12,50 @@ Prior to opening a PR it is useful and efficient to be able to run tests locally
...
+Once we have the kubeconfig working, we need to set the static manifests pointed through `STATIC_CONFIG_MANIFEST_DIR` environment variable.
+
+```sh
+mkdir ~/manifests
+export STATIC_CONFIG_MANIFEST_DIR=~/manifests
+```
+
+Currently only manifests from `[infrastructure|network].config.openshift.io/v1` are supported as static manifests.
+
+Exemplary manifests:
```
ggiguash, line 24:
```suggestion
Example manifests:
```

ggiguash, line 24: Since these manifests are required for the normal test runtime, can we make the yaml snapshots into commands to create the files? Something like the following:
```
MICROSHIFT_URL=<microshift-server-url>
cat > ${STATIC_CONFIG_MANIFEST_DIR}/single-replica.yaml <<EOF
...
...
status:
  apiServerURL: ${MICROSHIFT_URL}
...
EOF

ggiguash (changes requested):

docs/openshift_ci.md
chiragkyal: Thanks for the suggestion. Added.

PR #1971, opened 2023-06-28
USHIFT-1371: Test agent for testing ostree upgrade failures

pmtk:


test/failure-agent/microshift-failure-agent.sh
```diff
@@ -0,0 +1,77 @@
...
+
+for scenario_request in "${scenarios[@]}"; do
+    scenario=${scenario_request%:*}
+    boot=${scenario_request#*:}
+    if [[ "${boot}" == "every" ]] || [[ "${boot}" -eq "${CURRENT_BOOT}" ]]; then
+        ${scenario}
+    fi
+done
+
+sleep infinity
```
dhellmann, line 77: Is this because it is a service? Is there not a way to specify that something runs once and exits?

pmtk: I don't want it to exit.
For example, I want to simulate that making a backup on first boot of new deployment fails, but I want it to succeed on second boot (greenboot's reboot) - for this reason, I want this service to last up until the shutdown and revert `chattr` - otherwise it'll be present in rolled back deployment as well

dhellmann: OK, I see what you're going for. I'm a little worried about relying on a clean exit handler, but let's see how it works in practice. If we find issues we can look at options like persisting the instructions for each boot separately.

PR #1978, opened 2023-06-30
USHIFT-1085: feat: support prerun with minimal services

eggfoobar:
Added function to launch minimal MicroShift services when needed. For now the idea is to keep it as a local function for prerun until we need to expose it. I tried to keep it simple at the call site by not sending the channels back to the caller to handle. 

This PR supports #1956 

/assign @pmtk

### DIFF
--- a/pkg/admin/prerun/prerun.go
+++ b/pkg/admin/prerun/prerun.go
@@ -36,11 +36,13 @@ func (hi *HealthInfo) IsHealthy() bool {
 
 type PreRun struct {
 	dataManager data.Manager
+	config      *config.Config
 }
 
-func New(dataManager data.Manager) *PreRun {
+func New(dataManager data.Manager, config *config.Config) *PreRun {
 	return &PreRun{
 		dataManager: dataManager,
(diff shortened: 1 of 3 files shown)


pkg/admin/prerun/prerun.go
```diff
@@ -187,6 +189,11 @@ func (pr *PreRun) regularPrerun() error {
 
 		if migrationNeeded {
 			_ = migrationNeeded
+			stop, err := runMinimalMicroshift(pr.config)
+			if err != nil {
+				return fmt.Errorf("minimal MicroShift run failed to be ready: %w", err)
+			}
+			defer stop()
```
pmtk, line 196: We can put these lines and L220-L224 to a separate function, like a `migrateStorage` that would call `runMinimalMicroshift()` and run the migrator

eggfoobar: Sounds like a plan, do you want me to put that in a separate function in this PR, or do it in https://github.com/openshift/microshift/pull/1956 ?

PR #1982, opened 2023-06-30
USHIFT-1380: Consider filepath as not existing even if exists but is empty

pmtk:


dhellmann:
The code looks ok, but the commit could use more of a description so I can understand the intent.

PR #1986, opened 2023-07-03
USHIFT-1383: Upgradeability: rename (un)healthy to "backup / restore" action

pmtk:
Reusing the same `health.json` across many boots with "boot health" persisted might be confusing.
Changing healthy to backup and unhealthy to restore might be beneficial for clearly describing the procedures.

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -240,10 +240,10 @@ install -p -m755 packaging/greenboot/microshift-running-check.sh %{buildroot}%{_
 
 install -d -m755 %{buildroot}%{_sysconfdir}/greenboot/red.d
 install -p -m755 packaging/greenboot/microshift-pre-rollback.sh %{buildroot}%{_sysconfdir}/greenboot/red.d/40_microshift_pre_rollback.sh
-install -p -m755 packaging/greenboot/microshift_set_unhealthy.sh %{buildroot}%{_sysconfdir}/greenboot/red.d/40_microshift_set_unhealthy.sh
+install -p -m755 packaging/greenboot/microshift_set_restore.sh %{buildroot}%{_sysconfdir}/greenboot/red.d/40_microshift_set_restore.sh
 
 install -d -m755 %{buildroot}%{_sysconfdir}/greenboot/green.d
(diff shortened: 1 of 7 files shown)


packaging/rpm/microshift.spec
```diff
@@ -240,10 +240,10 @@ install -p -m755 packaging/greenboot/microshift-running-check.sh %{buildroot}%{_
 
 install -d -m755 %{buildroot}%{_sysconfdir}/greenboot/red.d
 install -p -m755 packaging/greenboot/microshift-pre-rollback.sh %{buildroot}%{_sysconfdir}/greenboot/red.d/40_microshift_pre_rollback.sh
-install -p -m755 packaging/greenboot/microshift_set_unhealthy.sh %{buildroot}%{_sysconfdir}/greenboot/red.d/40_microshift_set_unhealthy.sh
+install -p -m755 packaging/greenboot/microshift_set_restore.sh %{buildroot}%{_sysconfdir}/greenboot/red.d/40_microshift_set_restore.sh
```
dhellmann, line 243: Renaming the scripts may have implications for https://github.com/openshift/openshift-docs/pull/61653

I like the idea of changing the action inside MicroShift. In this external interface script, though, staying more generic and using terms like "handle unhealthy" or "handle healthy" will let us be more flexible. There may eventually be multiple actions to take in those cases, of which one is recording whether to take a backup or restore from backup, and it would be odd to have the "set restore" script doing something other than that. 

Using the more generic term also exposes fewer implementation details, which aren't important from the perspective of the caller of the script.

test/suites-ostree/backup-restore.robot
```diff
@@ -17,8 +17,9 @@ ${USHIFT_USER}      ${EMPTY}
 
 
 *** Test Cases ***
-Rebooting Healthy System Should Result In Data Backup
-    [Documentation]    Check if rebooting healthy system will result in backing up of MicroShift data
+MicroShift Backs Up Data On Boot
```
dhellmann, line 20: Is "On Startup" more accurate?

pmtk:
/hold
Let's think about this

pmtk:
/close

I'm not sure about this

PR #1989, opened 2023-07-03
USHIFT-1098: add minor-version upgrade scenario

dhellmann:
Depends on https://github.com/openshift/release/pull/40720

/assign @pmtk

### DIFF
--- a/test/bin/get_latest_rpm_version.sh
+++ b/test/bin/get_latest_rpm_version.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+#
+# This script reports the version of the latest RPM in the local
+# repository.
+
+set -euo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${SCRIPTDIR}/common.sh"
+
+release_info_rpm=$(find "${LOCAL_REPO}" -name 'microshift-release-info-*.rpm' | sort | tail -n 1)
+if [ -z "${release_info_rpm}" ]; then
+    error "Failed to find microshift-release-info RPM in ${LOCAL_REPO}"
+    exit 1
+fi
+rpm -q --queryformat '%{version}' "${release_info_rpm}" 2>/dev/null
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -238,6 +238,12 @@ previous_minor_version() {
     echo $(( $(current_minor_version) - 1 ))
 }
 
+# Function to report the *next* minor version. If the current
+# version is "4.14.5", reports "15".
+next_minor_version() {
+    echo $(( $(current_minor_version) + 1 ))
+}
+
 # Run the tests for the current scenario
 run_tests() {
     local vmname="${1}"
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -15,7 +15,6 @@ PULL_SECRET="${PULL_SECRET:-${HOME}/.pull-secret.json}"
 PULL_SECRET_CONTENT="$(jq -c . "${PULL_SECRET}")"
 PUBLIC_IP=${PUBLIC_IP:-""}  # may be overridden in global settings file
 VM_BOOT_TIMEOUT=8m
-MAKE_VAR_FILE="${TESTDIR}/../Makefile.version.$(uname -m).var"
 
 full_vm_name() {
     local base="${1}"
@@ -224,8 +223,7 @@ remove_vm() {
 
 # Function to report the full current version, e.g. "4.13.5"
 current_version() {
-    # OCP_VERSION := 4.14.0-0.nightly-2023-06-30-131338
-    grep OCP_VERSION "${MAKE_VAR_FILE}" | awk '{print $3}' | cut -f1 -d-
+    "${SCRIPTDIR}/get_latest_rpm_version.sh"
 }
 
 # Function to report only the minor portion of the current version,
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -15,6 +15,7 @@ PULL_SECRET="${PULL_SECRET:-${HOME}/.pull-secret.json}"
 PULL_SECRET_CONTENT="$(jq -c . "${PULL_SECRET}")"
 PUBLIC_IP=${PUBLIC_IP:-""}  # may be overridden in global settings file
 VM_BOOT_TIMEOUT=8m
+MAKE_VAR_FILE="${TESTDIR}/../Makefile.version.$(uname -m).var"
 
 full_vm_name() {
     local base="${1}"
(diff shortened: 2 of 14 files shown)


pmtk:
I don't see RF artifacts for the scenarios. Are we missing `bin/ci_phase_test.sh`?

test/bin/scenario.sh
```diff
@@ -211,6 +212,25 @@ remove_vm() {
     rm -rf "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}"
 }
 
+
+# Function to report the full current version, e.g. "4.13.5"
+current_version() {
+    # OCP_VERSION := 4.14.0-0.nightly-2023-06-30-131338
+    grep OCP_VERSION "${MAKE_VAR_FILE}" | awk '{print $3}' | cut -f1 -d-
```
dhellmann, line 219: We parse the version within the Makefile, too. Maybe we should add targets there to print out the values we need so we're only doing the parsing in 1 place?

dhellmann:
/hold

Let's let the metal tests run. I see some changes that snuck in as part of a botched rebase, and I'll clean those up, too.

dhellmann:
I think the metal test job will be fixed in part by https://github.com/openshift/release/pull/41143/

test/bin/scenario.sh
pmtk: fyi: we already have https://github.com/openshift/microshift/blob/main/Makefile#L87-L93 so we might extend that and grep specific variable?

dhellmann: The image build script looks at the RPM in the local repo. I think that's going to be safer long term, since that's the actual package we're trying to install. I'll rework things so that script and scenario.sh use the same logic.

test/bin/get_latest_rpm_version.sh
```diff
@@ -0,0 +1,16 @@
...
+#
+# This script reports the version of the latest RPM in the local
+# repository.
+
+set -euo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${SCRIPTDIR}/common.sh"
+
+release_info_rpm=$(find "${LOCAL_REPO}" -name 'microshift-release-info-*.rpm')
```
ggiguash, line 11: Is there a chance we'll get more than one RPM? If so, we should use `sort | tail -1` on the result of `find`

test/bin/get_latest_rpm_version.sh
```diff
@@ -0,0 +1,16 @@
...
+if [ -z "${release_info_rpm}" ]; then
+    error "Failed to find microshift-release-info RPM in ${LOCAL_REPO}"
+    exit 1
+fi
+rpm -q --queryformat '%{version}' "${release_info_rpm}"
```
ggiguash, line 16: If we assume x.y.z version format, should we verify it here to avoid surprises? 
Note also that this command may produce warnings.

```
$ rpm -q  --queryformat '%{version}' microshift-release-info-4.14.0~ec.3-202307062134.p0.g0ba2d5c.assembly.ec.3.el9.noarch.rpm
warning: microshift-release-info-4.14.0~ec.3-202307062134.p0.g0ba2d5c.assembly.ec.3.el9.noarch.rpm: Header V3 RSA/SHA256 Signature, key ID 5cd7bf6a: NOKEY
4.14.0~ec.3
```

test/image-blueprints/rhel92-microshift-previous-minor.toml
```diff
@@ -0,0 +1,30 @@
...
+description = ""
+version = "0.0.1"
+modules = []
+groups = []
+distro = "rhel-92"
+
+# FIXME: Temporarily pin the openvswitch version.
+[[packages]]
+name = "openvswitch3.1"
+version = "3.1.0-14.el9fdp"
```
ggiguash, line 11: This is no longer necessary due to an explicit dependency inside microshift RPM

dhellmann: Do we have a published 4.13 build with that dependency in place, yet?

test/bin/get_latest_rpm_version.sh
dhellmann: Those warnings go to stderr, right?

The full string is the version. We need the actual full value in some cases and only the minor version in others. It's up to the caller to parse it.

test/bin/get_latest_rpm_version.sh
dhellmann: The framework automation prevents it, but I'll push a fix anyway after the tests finish running with this version.

test/image-blueprints/rhel92-microshift-previous-minor.toml
ggiguash: It's merged, but I do not know if it's published.

test/bin/get_latest_rpm_version.sh
ggiguash: I would at least add `2>/dev/null` redirection to filter out unnecessary information.

test/image-blueprints/rhel92-microshift-previous-minor.toml
dhellmann: OK, I'd rather keep this in place for now then to ensure that the upgrade job has a chance of passing. When we have a real fix, we can remove this. I'll open a ticket to make sure we do it when we branch for 4.15.

test/bin/get_latest_rpm_version.sh
dhellmann: OK

ggiguash:
/lgtm 
/hold

Please, unhold to merge.

dhellmann:
The new test failed, investigating.

dhellmann:
```
While pulling rhel-9.2-microshift-source: While fetching http://192.168.122.1:8080/repo/summary.sig: [7] Couldn't connect to server
```

That looks like caddy was not running, or at least not responding.

dhellmann:
I added a commit to ensure the web server is started. We may end up with merge conflicts, because I see that same change in some of @pmtk's PRs.

dhellmann:
@pmtk I ended up moving the "fake" RPMs to their own repositories. We will need to add another repo for the "upgrade from main to this PR" tests that @copejon is working on, and we may have others in the future. Let's keep the RPMs for different types of builds separate so it's easy to reason about what we should find in any given repository.

dhellmann:
The new test is passing: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_microshift/1989/pull-ci-openshift-microshift-main-microshift-metal-tests/1680266579004100608/artifacts/microshift-metal-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/upgrade-minor-version-healthy/log.html

PR #1993, opened 2023-07-04
USHIFT-1372: Backing up the data is top priority

pmtk:
- TestAgent.py for managing test agent config
- Split `ostree.resource`
- General cleanup of keywords related to upgrades
- New test `failed-upgrade.robot`

### DIFF
--- a/test/resources/ostree.resource
+++ b/test/resources/ostree.resource
@@ -3,86 +3,69 @@ Documentation       Keywords for OSTree-based systems
 
 Resource            systemd.resource
 Resource            microshift-process.resource
+Resource            ostree-data.resource
+Resource            ostree-health.resource
 Library             libostree.py
-
-
-*** Variables ***
-${DATA_DIR}             "/var/lib/microshift"
-${BACKUP_STORAGE}       "/var/lib/microshift-backups"
+Library             ../resources/TestAgent.py
 
 
 *** Keywords ***
-Wait Until Greenboot Health Check Exited
-    [Documentation]    Wait until greenboot healthchecks are done
+Make New SSH Connection
+    [Documentation]    Closes all SSH connections and makes a new one.
+    # Staging deployments potentially introduces multiple reboots
+    # which could break existing SSH connection
 
-    Wait Until Keyword Succeeds    10m    15s
-    ...    Greenboot Health Check Exited
+    SSHLibrary.Close All Connections
+    Login MicroShift Host
 
-Greenboot Health Check Exited
-    [Documentation]    Checks if greenboot-healthcheck finished running successfully (exited)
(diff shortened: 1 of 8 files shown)


test/resources/ostree.resource
```diff
@@ -3,86 +3,69 @@ Documentation       Keywords for OSTree-based systems
...
-MicroShift 413 Should Not Have Upgrade Artifacts
-    [Documentation]    Verifies that host running MicroShift 4.13
-    ...    does not have upgrade related artifacts.
+    Make New SSH Connection
 
-    ${version}=    MicroShift Version
-    IF    ${version.minor} == 13
-        Path Should Not Exist    ${BACKUP_STORAGE}
-        Path Should Not Exist    /var/lib/microshift/version
+    ${current_deploy}=    libostree.Get Booted Deployment Id
```
dhellmann, line 64: The logic here is looking at current and initial. Is it relevant what the "initial" deployment was, if that does not match the expected deployment? Is that check just for this fatal error handling?

pmtk: It's a shortcut to not wait whole 10/20 minutes just for it to fail because it rolled back

dhellmann: That makes sense.

test/suites-ostree/failed-upgrade.robot
```diff
@@ -0,0 +1,44 @@
...
+Suite Teardown      Teardown
+
+Test Tags           ostree
+
+
+*** Variables ***
+${USHIFT_HOST}      ${EMPTY}
+${USHIFT_USER}      ${EMPTY}
+
+${FAILING_REF}      ${EMPTY}
```
dhellmann, line 18: Does the image referred to here need to already include something that will cause a failure, or does the TestAgent call on line 30 inject that?

pmtk: Image needs [microshift-test-agent RPM](https://github.com/openshift/microshift/tree/main/test/agent). We can create an alias: `ostree refs --create microshift/failing microshift/source`

dhellmann: Since it is part of the test harness, we should add it to all of the blueprints. Can you do that in this PR?

If we need the alias, too, you could add that command to the image download script so it is always run, too. Do we want an alias for every image we build, or should we add them as we need them in a scenario?

pmtk: I added the alias to download. I only aliased microshift_source -> microshift_failing. I think we only need that one, so we can test that code under test behaves as expected during failure conditions

dhellmann (commented):
/lgtm

PR #1998, opened 2023-07-05
USHIFT-1390: Handle FDO scenario

pmtk:


pkg/admin/data/data_manager.go
```diff
@@ -171,6 +171,26 @@ func (dm *manager) Restore(name BackupName) error {
 	return nil
 }
 
+func (dm *manager) RemoveData() error {
+	cmd := exec.Command("rm", "-rf", config.DataDir) //nolint:gosec
```
dhellmann, line 175: Is this better than `os.RemoveAll()`? https://pkg.go.dev/os#RemoveAll

pkg/admin/prerun/prerun.go
```diff
@@ -288,22 +281,67 @@ func (pr *PreRun) restore() error {
...
+		return err
 	}
 
-	klog.Info("Finished restore")
-	return nil
+	if rollbackDeployID == "" {
+		// No backup for current deployment and there is no rollback deployment.
+		// This could be a unhealthy system that was manually rebooted to
+		// remediate the situation - let's not interfere: no backup, no restore, just proceed.
+		klog.InfoS("System has no rollback but health.json suggests system was rebooted - exiting prerun and proceeding with start")
```
dhellmann, line 308:
```suggestion
		klog.InfoS("System has no rollback but health.json suggests system was rebooted - skipping prerun")
```

This code should not assume what might come next, so let's just have the log messages comment on the local actions.

pkg/admin/prerun/prerun.go
```diff
@@ -288,22 +281,67 @@ func (pr *PreRun) restore() error {
...
+		"current-deployment-id", currentDeploymentID,
+		"health-deployment-id", health.DeploymentID)
+
+	if health.DeploymentID == rollbackDeployID {
+		return fmt.Errorf("deployment ID stored in health.json is the same as rollback's" +
+			" - MicroShift should not be updated from unhealthy system")
+	}
+
+	if health.DeploymentID == currentDeploymentID {
+		// There is no backup for current deployment, but there is a backup for the rollback.
```
dhellmann, line 323: I'm having trouble linking this comment and the condition being tested on line 322. Why do we know there is a backup for the rollback if the deployment ID matches the current deployment?

pkg/admin/prerun/prerun.go
```diff
@@ -288,22 +281,67 @@ func (pr *PreRun) restore() error {
...
+		// Let's restore it and act like it's first boot of newly staged deployment
+		backupsForRollback := getExistingBackupsForTheDeployment(existingBackups, rollbackDeployID)
+		if len(backupsForRollback) == 0 {
+			// This could happen if current deployment is unhealthy and rollback didn't run MicroShift
+			klog.InfoS("There is no backup for rollback deployment as well - removing existing data for clean start")
+			return pr.dataManager.RemoveData()
+		}
+
+		klog.InfoS("Restoring backup for a rollback deployment to perform migration and try starting again",
```
dhellmann, line 332: We don't seem to _know_ that there is a backup for the rollback until this point, so maybe the comment on line 323 is anticipating reaching this point?

pkg/admin/prerun/prerun.go
```diff
@@ -169,54 +169,38 @@ func (pr *PreRun) regularPrerun() error {
...
+	migrationNeeded, err := pr.checkVersions()
+	if err != nil {
+		return fmt.Errorf("failed version checks: %w", err)
+	}
 
-	if err := pr.dataManager.Backup(backupName); err != nil {
-		return fmt.Errorf("failed to create new backup %q: %w", backupName, err)
+	klog.InfoS("Completed version checks", "is-migration-needed?", migrationNeeded)
+
+	if !migrationNeeded {
```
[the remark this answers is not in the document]
pmtk: Don't think so, it might be a good idea so we actually test migration even though there's nothing to migrate

pkg/admin/prerun/prerun.go
pmtk: Wait, looks misplaced during rewrite

test/suites-ostree/fdo.robot
```diff
@@ -28,6 +28,7 @@ FIDO Onboarding Device
     ...    MicroShift data, unhealthy stored in health file,
     ...    and a deployment gap (no-microshift rollback ).
 
+    Mask Grub Boot Success Timer
```
dhellmann, line 31: Should we build this into the image blueprints or kickstart file?

dhellmann (commented):
One question that we can address in a follow-up.

/lgtm

@jogeo @dhensel-rh do you want to test this before we merge it?

/hold

test/suites-ostree/fdo.robot
pmtk: Not entirely sure, but I doubt blueprints have this capabilities. [They can enable/disable services](https://www.osbuild.org/guides/image-builder-on-premises/blueprint-reference.html#systemd-services), but it's not a service and it cannot be disabled, only masked. Furthermore, it's user-specific (--user) - notice no `sudo`. [0]

If we want to include it, my bet would on be kickstart but my internal paranoia would at least keep a check if this is not masked, so we don't run into this problem by accident wondering what's happening like I did.

----
[0]

```
$ systemctl --user disable grub-boot-success.timer
The unit files have no installation config (WantedBy=, RequiredBy=, Also=,
Alias= settings in the [Install] section, and DefaultInstance= for template
units). This means they are not meant to be enabled or disabled using systemctl.
 
Possible reasons for having this kind of units are:
• A unit may be statically enabled by being symlinked from another unit's
  .wants/ or .requires/ directory.
• A unit's purpose may be to act as a helper for some other unit which has
  a requirement dependency on it.
• A unit may be started when needed via activation (socket, path, timer,
  D-Bus, udev, scripted systemctl call, ...).
• In case of template units, the unit is meant to be enabled with some
  instance name specified.
```

on the other hand, masking seems to be just symlinking to /dev/null:
```
$ systemctl --user mask grub-boot-success.timer
Created symlink /home/redhat/.config/systemd/user/grub-boot-success.timer → /dev/null.
```

dhellmann (commented):
/lgtm

test/suites-ostree/fdo.robot
dhellmann: OK, maybe leaving it in the test code is the best way to be explicit that we need to take the step.

pmtk:
/unhold

Unholding to allow to make progress with other dev tasks

pmtk:
/hold

Depends on https://github.com/openshift/microshift/pull/2037

PR #1999, opened 2023-07-05
WIP. USHIFT-1387: add scenario test run script

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/test/bin/ci_phase_test.sh
+++ b/test/bin/ci_phase_test.sh
@@ -9,6 +9,10 @@ source "${SCRIPTDIR}/common.sh"
 
 cd "${TESTDIR}"
 
+if [ ! -d "${RF_VENV}" ]; then
+    "${ROOTDIR}/scripts/fetch_tools.sh" robotframework
+fi
+
 for scenario in ./scenarios/*.sh; do
     scenario_name="$(basename "${scenario}" .sh)"
     logfile="${SCENARIO_INFO_DIR}/${scenario_name}/run.log"
--- a/test/bin/ci_phase_test.sh
+++ b/test/bin/ci_phase_test.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+#
+# This script runs on the CI cluster, from the metal-tests step.
+
+set -xeuo pipefail
+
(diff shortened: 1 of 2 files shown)


test/bin/ci_phase_test.sh
```diff
@@ -0,0 +1,29 @@
...
+set -xeuo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${SCRIPTDIR}/common.sh"
+
+cd "${TESTDIR}"
+
+for scenario in ./scenarios/*.sh; do
+    scenario_name="$(basename "${scenario}" .sh)"
+    logfile="${SCENARIO_INFO_DIR}/${scenario_name}/run.log"
```
dhellmann, line 14: I think we need to override the SCENARIO_INFO_DIR to point to the location in the artifact dir where the scenario data has been downloaded in the previous step. We might need a new variable in common.sh to make overriding all of the other directory variables easier.

pacevedom: Yes, was just creating this commit to check whats needed from openshift/release and work it out from that starting point.
Base path for everything will also change once the microshift-test-payload image disappears, so we need to make it configurable in an easy way. Thanks for the heads up!

dhellmann: Is the artifact dir shared between pods in the CI cluster? Or should we use the shared directory and then copy the content into the artifact directory separately?

test/bin/scenario.sh
```diff
@@ -214,8 +215,7 @@ run_tests() {
     echo "Running tests with $# args" "$@"
 
     if [ ! -d "${RF_VENV}" ]; then
-        error "RF_VENV (${RF_VENV}) does not exist, create it with: ${ROOTDIR}/scripts/fetch_tools.sh robotframework"
-        exit 1
+        "${ROOTDIR}/scripts/fetch_tools.sh" robotframework
```
dhellmann, line 218: We can't do this in the scenario script because in CI several copies of this will eventually run in parallel. We could do it explicitly in the phase script instead.

PR #2003, opened 2023-07-06
manual 4.14 rebase to catch up

dhellmann:
./scripts/auto-rebase/rebase.sh to "registry.ci.openshift.org/ocp/release:4.14.0-0.nightly-2023-07-05-191022" "registry.ci.openshift.org/ocp-arm64/release-arm64:4.14.0-0.nightly-arm64-2023-07-05-221234" "registry.access.redhat.com/lvms4/lvms-operator-bundle:v4.12"

PR #2006, opened 2023-07-06
manual rebase 2023-07-06

dhellmann:


dhellmann:
The first commit in this series is part of #2005 and we should merge that separately, then squash the commits in this PR when it merges.

assets/components/csi-snapshot-controller/role.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Role
```
copejon, line 1: Leader election is disabled on the csi snapshotter. This being a manual rebase, let's go ahead and drop these (role / binding). I'm looking at the script now to see what's causing them to populate, but they shouldn't be included.

copejon:
> @copejon I think your comment [#2003 (comment)](https://github.com/openshift/microshift/pull/2003#issuecomment-1623881019) applies here, too.
> 
> Can you look at whatever we need to change in the rebase script to make that happen? Perhaps we should not merge this PR at all?

Looking at it now.  Shouldn't be a huge issue. I'll have a PR up in a bit to fix it, then we can retry the rebase.

dhellmann:
/close

obsoleted by #2008

PR #2013, opened 2023-07-07
OCPBUGS-15740: Fix show-config command output

pacevedom:
Command `microshift show-config` must show the running configuration. This command loaded the configuration from /etc/microshift/config.yaml if it existed, possibly showing wrong information (editing the file without restarting microshift yields wrong data).
Having no configuration file fills the config with defaults. This might also be wrong in the event of a hostname change, when this is automatically fixed to use the previous name. This command would show the new name when MicroShift is using the previous one.

New approach stores the running configuration upon start and returns it when asked for it.
If MicroShift is not running this command returns the last active config.

Closes #<Issue Number>

### DIFF
--- a/pkg/cmd/showConfig.go
+++ b/pkg/cmd/showConfig.go
@@ -33,6 +33,10 @@ func NewShowConfigCommand(ioStreams genericclioptions.IOStreams) *cobra.Command 
 				if err != nil {
 					cmdutil.CheckErr(err)
 				}
+				err = cfg.EnsureNodeNameHasNotChanged()
+				if err != nil {
+					cmdutil.CheckErr(err)
+				}
 			case "default":
 				cfg = config.NewDefault()
 			default:


pkg/config/files.go
```diff
@@ -47,18 +48,37 @@ func getActiveConfigFromYAML(contents []byte) (*Config, error) {
 // file exists, read it and require it to be valid. Otherwise return
 // the default settings.
 func ActiveConfig() (*Config, error) {
-	_, err := os.Stat(ConfigFile)
+	configFileName := ActiveConfigFile
```
dhellmann, line 51: This logic doesn't look right.

If we're going to cache the entire set of config somewhere, then we always need to update those settings in case the user has modified the input file.

Maybe we can focus this change on just reading the nodename file and using that value for the host name override?

pacevedom: This is a bit open to interpretation I think. The `microshift show-config` command should return the active config when microshift is running, right? Which means the configuration that is currently loaded. If you modify the config file then that wont take effect until next restart, which will persist the file again. But before restarting it any changes a user may do are not visible to microshift and displaying them in show-config might be misleading?

dhellmann: Yes, you're interpreting the command's purpose differently than it was originally intended. 

It was meant to show the combination of what is in `/etc/microshift/config.yaml` and any defaults that are not explicitly in the file. You are correct that the results may not be the same as what the current process is using, of course. Those settings are logged, so it's easy to get them from journalctl. It's harder to look at a partial configuration file and understand what other settings will be added to make up the full configuration.

pacevedom: Then you are right. I changed the logic to copy the run behavior and update it accordingly.
What I find misleading is the name of the command and the comment, but we can take that in a different PR.

dhellmann (commented):
/lgtm

PR #2015, opened 2023-07-07
[USHIFT-1374] specify an LVM thinpool in demo kickstart file

copejon:
USHIFT-1374

Add LVM thinpool provisioning to MicroShift's kickstarter files to support lvms thin volume provisioning, volume cloning, and volume snapshotting.


docs/config/microshift-starter.ks
```diff
@@ -16,6 +16,7 @@ part /boot --fstype=xfs --asprimary --size=800
 part pv.01 --grow
 volgroup rhel pv.01
 logvol / --vgname=rhel --fstype=xfs --size=10240 --name=root
+logvol none --thinpool --vgname=rhel --fstype="None" --size=100 --name=thin
```
ggiguash, line 19: Can you explain why the 100MB size of the pool?

copejon: Actually it's much too small. 

My feeling is that it'd be less complicated for users to switch the getting-started workflow to using only thin-vols. This would keep the number of storage classes device classes to 1 each, while also enabling snapshotting/cloning workflows.  

From the users' perspective, the existing storage workflows wouldn't change.
From MicroShift's perspective, the default storage class and lvmd would have to be reconfigured for thin vol provisioning (a relatively trivial change). 

I'll get some feedback on what a preferable thin/thick storage allocation would be from the team.

dhellmann: The size is up to the user, right? This is just a demo/starter file?

ggiguash: Let's specify a 1024 size in all cases. It should be big enough for the dev usage. 
Users will have their own layouts anyway.

test/kickstart-templates/kickstart.ks.template
```diff
@@ -28,6 +28,7 @@ part /boot --fstype=xfs --asprimary --size=800
 part pv.01 --grow
 volgroup rhel pv.01
 logvol / --vgname=rhel --fstype=xfs --size=REPLACE_LVM_SYSROOT_SIZE --name=root
+logvol none --thinpool --vgname=rhel --fstype="None" --size=100 --name=thin
```
ggiguash, line 31: Do we need to update the partition table comment above to reflect the thin volume?

ggiguash (changes requested):

docs/config/microshift-starter.ks
copejon: Right. I guess my concern is really how this would jive with MicroShift's config-defaulting behavior.  As is, MicroShift's out of the box experience wouldn't provide the snapshot/cloning features without the user setting up a thin-pool, and specifying an lvmd.yaml and storage class. Meaning the demo wouldn't be reflective of an user's unconfigured env.

That also brings up whether the defaulting logic should consider thin-pools in addition to vgs, and which it should prioritize.  Maybe we need to consider a VG/thin-pool naming or tagging scheme instead?

scripts/image-builder/config/kickstart.ks.template
```diff
@@ -28,6 +28,7 @@ part /boot --fstype=xfs --asprimary --size=800
 part pv.01 --grow
 volgroup rhel pv.01
 logvol / --vgname=rhel --fstype=xfs --size=REPLACE_LVM_SYSROOT_SIZE --name=root
+logvol none --thinpool --vgname=rhel --fstype="None" --size=100 --name=thin
```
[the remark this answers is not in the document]
copejon: Ah, yes.

copejon:
Per team discussion, thin-vol provisioning, snapshotting, and cloning lean more towards advanced storage features and aren't a good fit for a demo deployment.  User docs should be added/enhanced to describe host and cluster configuration for thin-pools.

PR #1987, opened 2023-07-03
USHIFT-1385: boot CI VMs in parallel

dhellmann:
/hold until https://github.com/openshift/release/pull/40720 merges
/assign @pacevedom @pmtk

PR #2016, opened 2023-07-07
USHIFT-1411: pin ovs dependency to 3.1.1

dhellmann:
This is a temporary work-around for OCPBUGS-15948

/assign @pmtk @pliurh
/cc @zshi-redhat @pacevedom


dhellmann:
/hold

It is not clear this is actually helpful.

Makefile
```diff
@@ -136,7 +136,7 @@ verify-fast: verify-go verify-assets verify-sh verify-py verify-config verify-rf
 
 # Full verification checks that should run in CI
 .PHONY: verify-ci
-verify-ci: verify-fast verify-images verify-licenses verify-containers
+verify-ci: verify-fast verify-images verify-containers
```
ggiguash, line 139: Why was the license verification removed?

dhellmann: The commit message for that commit shows the error message that is breaking the job.

```
2023/07/08 18:18:45 failed to evaluate licenses: invalid character 'G' looking for beginning of value
```

It affects all PRs, so I suspect an updated version of the tool is behaving differently. We don't add new dependencies often, so I thought rather than spending time figuring out why that tool is reporting an error I would just turn it off.

dhellmann: See https://github.com/openshift/microshift/pull/1987#issuecomment-1627712945 for example

ggiguash: Interesting. This problem does not occur on my development machine.

dhellmann: I was not able to reproduce it locally either. Perhaps it has something to do with the compiler version? Either way, I wanted to unblock us without spending time on a low-value tool.

ggiguash: I implemented a w/a in the https://github.com/openshift/microshift/pull/2020 PR

pmtk:
/test microshift-metal-tests

`[Errno 2] No such file or directory: '/var/cache/dnf/rhel-9-appstream-rhui-rpms-4266d693e48ae594/packages/git-core-2.39.3-1.el9_2.x86_64.rpm'`

pmtk:
Maybe we need something like `dnf clean cache/packages` or even all

pacevedom (approved):
/lgtm
/hold

PR #2019, opened 2023-07-08
USHIFT-1085: Storage Migration Controller Implementation

eggfoobar:
This PR builds upon the work done in the previous PR #1956 but with the cluster fully running instead of pre-run.

This was done since we need the cluster to be fully running with the kubelet in order to guarantee migrations for CRD Migration Webhooks.

/cc @dhellmann


pkg/controllers/kube-storage-version-trigger.go
```diff
@@ -0,0 +1,111 @@
...
+	kubeStorageVersionTrigger = "storage-version-migration-trigger"
+)
+
+type KubeStorageVersionTrigger struct {
+	kubeconfig string
+	healthPath string
+	healthPort string
+}
+
+func NewKubeStorageVersionTrigger(cfg *config.Config) *KubeStorageVersionTrigger {
```
dhellmann, line 41: What's the difference between the trigger and the migrator?

pmtk: There's [user guide](https://github.com/openshift/kubernetes-kube-storage-version-migrator/blob/master/USER_GUIDE.md#storage-version-migrator-in-a-nutshell) for it'd be nice to include that information in our repository (even as a comment in one of the files).

Looks like we'll need to include [checking if migration is complete and successful](https://github.com/openshift/kubernetes-kube-storage-version-migrator/blob/master/USER_GUIDE.md#check-if-migration-has-completed) in healthcheck script

pmtk: And we'll need to rethink `$DATA_DIR/version`  - current impl was based on assumption that migration would happen before "true" start, but with this we'll just start and healthcheck will assess if all is good.
Maybe it's a good thing and will simplify and increase flexibility of the upgradability. :thinking:

pkg/controllers/kube-storage-version-migrator.go
```diff
@@ -0,0 +1,118 @@
...
+	go func() {
+		errorChannel <- s.runMigrator(ctx)
+	}()
+
+	return <-errorChannel
+}
+
+func (s *KubeStorageVersionMigrator) runMigrator(ctx context.Context) error {
+	go func() {
+		err := util.StartHealthCheck(s.healthPath, s.healthPort)
```
pmtk, line 85: Can you explain what's the plan for this endpoint?
Now it's static `ok`, will it contain some info about migration itself later? Does Trigger/Migrator need health endpoints of each other?

eggfoobar: Yeah, I couldn't decide just closing the ready checkpoint there, but I saw we did this pattern else where so I thought it might be good for now. But what do you think of just closing the ready channel here?

pmtk: We don't have to worry about this now

pkg/controllers/kube-storage-version-migrator.go
```diff
@@ -0,0 +1,118 @@
...
+		migration,
+	)
+
+	c.Run(ctx)
+
+	select {
+	case <-ctx.Done():
+		return nil
+	default:
+		return fmt.Errorf("%s shutdown unexpeditly", kubeStorageVersionMigrator)
```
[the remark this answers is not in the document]
eggfoobar: Oops, updated!

pkg/controllers/kube-storage-version-migrator.go
```diff
@@ -0,0 +1,118 @@
+/*
+Copyright © 2021 MicroShift Contributors
```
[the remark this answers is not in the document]
eggfoobar: Fixed, thanks!

PR #2028, opened 2023-07-10
USHIFT-1422: use the regular RPM repositories for dependencies in image-builder script

dhellmann:
Instead of downloading dependencies and manipulating the local repo,
use the regular repositories normally. This allows composer to find the
pinned version of openvswitch3.1 needed right now.

Includes some other cleanup changes.

/assign @ggiguash @jogeo

### DIFF
--- a/scripts/image-builder/build.sh
+++ b/scripts/image-builder/build.sh
@@ -305,14 +305,15 @@ fi
 # Set the cleanup and elapsed time traps only if command line parsing was successful
 trap '${SCRIPTDIR}/cleanup.sh; echo "Execution time: $(( ($(date +%s) - STARTTIME) / 60 )) minutes"' EXIT
 
-title "Downloading local OpenShift and MicroShift repositories"
+title "Setting up local MicroShift repository"
 # Copy MicroShift RPM packages
 rm -rf microshift-local 2>/dev/null || true
 if [[ "${MICROSHIFT_RPM_SOURCE}" == http* ]] ; then
     wget -q -nd -r -L -P microshift-local -A rpm "${MICROSHIFT_RPM_SOURCE}"
 else
     [ ! -d "${MICROSHIFT_RPM_SOURCE}" ] && echo "MicroShift RPM path '${MICROSHIFT_RPM_SOURCE}' does not exist" && exit 1
-    cp -TR "${MICROSHIFT_RPM_SOURCE}" microshift-local
+    cp -TR "${MICROSHIFT_RPM_SOURCE}/RPMS" microshift-local
+    cp -TR "${MICROSHIFT_RPM_SOURCE}/SRPMS" microshift-local
 fi
 
 # Exit if no RPM packages were found
--- a/scripts/image-builder/build.sh
+++ b/scripts/image-builder/build.sh
@@ -337,24 +337,6 @@ if [ -z "${IMAGE_VERSION}" ] ; then
     exit 1
 fi
 
-# Download openshift local RPM packages (noarch for python and selinux packages)
-rm -rf openshift-local 2>/dev/null || true
-
-OCP_REPO_NAME="rhocp-4.13-for-rhel-${OSVERSION}-$(uname -m)-rpms"
-reposync -n -a "${BUILD_ARCH}" -a noarch --download-path openshift-local \
-    --repo="${OCP_REPO_NAME}" \
-    --repo="fast-datapath-for-rhel-${OSVERSION}-${BUILD_ARCH}-rpms" >/dev/null
-
-# Remove 'microshift' packages to avoid overrides from the remote repository
-find openshift-local -name \*microshift\* -exec rm -f {} \;
-# Exit if no RPM packages were found
-if [ "$(find openshift-local -name '*.rpm' | wc -l)" -eq 0 ] ; then
-    echo "No RPM packages were found at the '${OCP_REPO_NAME}' repository. Exiting..."
-    exit 1
-fi
-createrepo openshift-local >/dev/null
-open_repo_permissions openshift-local
-
 # Install prometheus process exporter
 install_prometheus_rpm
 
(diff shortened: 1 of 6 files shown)


dhellmann:
I was able to build an ISO using RPMs from source and boot the VM successfully.

```
[root@localhost ~]# rpm-ostree db list e211ad60c69b42703c430f88f64b07a1c20a2a2fe6a99287b41706266bd9c1a2 | grep openvswitch
 openvswitch-selinux-extra-policy-1.0-31.el9fdp.noarch
 openvswitch3.1-3.1.0-14.el9fdp.x86_64

[root@localhost ~]# oc get pods -A
NAMESPACE                  NAME                                       READY   STATUS    RESTARTS      AGE
kube-system                csi-snapshot-controller-75b9b8c9ff-wghtt   1/1     Running   0             22m
kube-system                csi-snapshot-webhook-84684f46f6-x2nxv      1/1     Running   0             22m
openshift-dns              dns-default-jc7pg                          2/2     Running   0             21m
openshift-dns              node-resolver-wdvjf                        1/1     Running   0             22m
openshift-ingress          router-default-77dbd6c6b6-qlzgx            1/1     Running   0             22m
openshift-ovn-kubernetes   ovnkube-master-k24j9                       4/4     Running   1 (21m ago)   22m
openshift-ovn-kubernetes   ovnkube-node-b7djt                         1/1     Running   1 (21m ago)   22m
openshift-service-ca       service-ca-ff55d8996-wzsw4                 1/1     Running   0             22m
openshift-storage          topolvm-controller-ffd4fbff9-l588d         5/5     Running   0             22m
openshift-storage          topolvm-node-bk84t                         4/4     Running   0             21m
```

scripts/image-builder/build.sh
```diff
@@ -89,7 +89,7 @@ waitfor_image() {
     local status
     status=$(sudo composer-cli compose status | grep "${uuid}" | awk '{print $2}')
     while [[ "${status}" = "RUNNING" ]] || [[ "${status}" = "WAITING" ]]; do
-        sleep 10
+        sleep 30
```
ggiguash, line 92: In the interactive mode, 30 seconds may be a bit long for users to wait.

scripts/image-builder/build.sh
```diff
@@ -336,23 +341,41 @@ if [ -z "${IMAGE_VERSION}" ] ; then
...
-# Download openshift local RPM packages (noarch for python and selinux packages)
-rm -rf openshift-local 2>/dev/null || true
-
-OCP_REPO_NAME="rhocp-4.13-for-rhel-${OSVERSION}-$(uname -m)-rpms"
-reposync -n -a "${BUILD_ARCH}" -a noarch --download-path openshift-local \
-    --repo="${OCP_REPO_NAME}" \
-    --repo="fast-datapath-for-rhel-${OSVERSION}-${BUILD_ARCH}-rpms" >/dev/null
+title "Ensure there is at least one osbuild worker running"
+sudo systemctl start osbuild-composer.socket
+sudo systemctl start "osbuild-worker@1.service"
```
ggiguash, line 346: I'm not sure this belongs here. If there's some problem with cleanup, maybe we should add code to the `cleanup.sh` script to make sure it starts the service?

scripts/image-builder/build.sh
```diff
@@ -336,23 +341,41 @@ if [ -z "${IMAGE_VERSION}" ] ; then
...
-    echo "No RPM packages were found at the '${OCP_REPO_NAME}' repository. Exiting..."
-    exit 1
-fi
-createrepo openshift-local >/dev/null
-open_repo_permissions openshift-local
+for name in $(sudo composer-cli sources list); do
+    echo
+    echo "Package source: ${name}"
+    sudo composer-cli sources info "${name}" | sed -e 's/gpgkeys.*/gpgkeys = .../g'
+done
```
ggiguash, line 378: This may generate a bit noisy output in the current (otherwise clean) sequence. If we'd like to display the enabled repos, can we just display name/url instead of all the parameters? 
```
sudo composer-cli sources info "${name}" | grep -E '^id|^url'
```

scripts/image-builder/build.sh
```diff
@@ -323,6 +324,10 @@ fi
 createrepo microshift-local >/dev/null
 open_repo_permissions microshift-local
 
+# Determine the version of microshift we have, for use in blueprint templates later.
+MICROSHIFT_VERSION=$(rpm -q --queryformat '%{version}' "${MICROSHIFT_RPM_SOURCE}/RPMS/noarch/microshift-release-info*.rpm")
```
ggiguash, line 328: This code may not work when RPMs are downloaded from a remote site because it assumes `RPMS/noarch` suffix exists.

We should instead use `find | tail -1` to discover the last version of the release-info. Note that `tail -1` is required because even in devenv, the local repo may contain multiple copies of RPMs and we need the last one.

scripts/image-builder/build.sh
```diff
@@ -336,23 +341,41 @@ if [ -z "${IMAGE_VERSION}" ] ; then
...
+id = "fast-datapath"
+name = "Fast Datapath for RHEL ${OSVERSION}"
+type = "yum-baseurl"
+url = "https://cdn.redhat.com/content/dist/layered/rhel${OSVERSION}/$(uname -m)/fast-datapath/os"
+check_gpg = true
+check_ssl = true
+system = false
+rhsm = true
+EOF
+sudo composer-cli sources add fast-datapath.toml
```
ggiguash, line 372: I think this code is best converted to .toml.template and expanded in the loop at [ line399](https://github.com/openshift/microshift/pull/2028/files#diff-bb6dd0287122cf14166d793e40a00a53e4cf43748f2318fad274ff4028020906R399)

In any case, we should first delete the source and then re-add it. Otherwise, fix of a toml does not get into osbuilder.

scripts/image-builder/build.sh
```diff
@@ -373,15 +396,16 @@ if [ -n "${CUSTOM_RPM_FILES}" ] ; then
...
     [ ! -d ${f} ] && continue
     sed "s;REPLACE_IMAGE_BUILDER_DIR;${BUILDDIR};g" "${SCRIPTDIR}/config/${f}.toml.template" > ${f}.toml
     sudo composer-cli sources delete ${f} 2>/dev/null || true
     sudo composer-cli sources add "${BUILDDIR}/${f}.toml"
 done
 
 title "Preparing blueprints"
-cp -f "${SCRIPTDIR}"/config/{blueprint_v0.0.1.toml,installer.toml} .
+cp -f "${SCRIPTDIR}"/config/installer.toml .
+envsubst <"${SCRIPTDIR}"/config/blueprint_v0.0.1.toml >blueprint_v0.0.1.toml
```
ggiguash, line 408: `envsubst` is definitely a better techinque than sed replacement. However, I think we'd better convert all the `sed` calls to `envsubst` in the same PR. At this time, can we keep one consistent technique for replacements?
In any case, there is a convention in the script that uses `REPLACE_` prefix for all such variables. It should help us avoid unwanted replacements from the user environment.

ggiguash (changes requested):

dhellmann:
@ggiguash I think I have addressed all of your comments.

scripts/image-builder/build.sh
```diff
@@ -323,6 +324,11 @@ fi
 createrepo microshift-local >/dev/null
 open_repo_permissions microshift-local
 
+# Determine the version of microshift we have, for use in blueprint templates later.
+MICROSHIFT_RELEASE_RPM=$(find "${MICROSHIFT_RPM_SOURCE}" -name 'microshift-release-info*.rpm' | tail -n 1)
+MICROSHIFT_VERSION=$(rpm -q --queryformat '%{version}' "${MICROSHIFT_RELEASE_RPM}")
+export MICROSHIFT_VERSION
```
ggiguash, line 330: No longer need the export
```suggestion
```

PR #2039, opened 2023-07-11
USHIFT-1395: Use file protocol for local ostree updates instead of Caddy HTTP server

ggiguash:
Using HTTP server for local `ostree` upgrades is not necessary, hence the simplification of the default ISO configuration.

Closes [USHIFT-1395](https://issues.redhat.com//browse/USHIFT-1395)

### DIFF
--- a/docs/contributor/rhel4edge_iso.md
+++ b/docs/contributor/rhel4edge_iso.md
@@ -60,7 +60,7 @@ Optional arguments:
   -embed_containers
           Embed the MicroShift container dependencies in the image
   -ostree_server_url URL
-          URL of the ostree server (default: http://127.0.0.1:8085/repo)
+          URL of the ostree server (default: file:///var/lib/ostree-local/repo)
   -build_edge_commit
           Build edge commit archive instead of an ISO image. The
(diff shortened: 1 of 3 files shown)


docs/contributor/rhel4edge_iso.md
```diff
@@ -299,14 +290,14 @@ The contents of the archive can be used for serving ostree updates:
 > `-embed_containers` arguments to customize the edge commit archive contents.
 
 Copy the produced `microshift-0.0.1-commit.tar` archive to the
-MicroShift server and unpack it at the `/var/lib/ostree-server-local` directory,
+MicroShift server and unpack it at the `/var/lib/ostree-local` directory,
```
dhellmann, line 293: Is that path something known to rpm-ostree, or is that a path name we're suggesting that has no special meaning?

ggiguash: It's a path that we're suggesting.

ggiguash: @dhellmann , please unhold the PR if you think the current approach is acceptable.

PR #2042, opened 2023-07-11
USHIFT-1387: Use storage pool for scenario VM creation

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/test/bin/ci_phase_iso_boot.sh
+++ b/test/bin/ci_phase_iso_boot.sh
@@ -23,6 +23,12 @@ cd ~/microshift/test
 # repository.
 bash -x ./bin/start_webserver.sh
 
+# set up the storage pool for VMs
+sudo virsh pool-define-as "${VM_STORAGE_POOL}" dir - - - - "${VM_DISK_DIR}"
+sudo virsh pool-build "${VM_STORAGE_POOL}"
+sudo virsh pool-start "${VM_STORAGE_POOL}"
+sudo virsh pool-autostart "${VM_STORAGE_POOL}"
+
 # Build all of the needed VMs
 for scenario in scenarios/*.sh; do
     scenario_name=$(basename "${scenario}" .sh)
--- a/test/bin/ci_phase_iso_boot.sh
+++ b/test/bin/ci_phase_iso_boot.sh
@@ -23,11 +23,8 @@ cd ~/microshift/test
 # repository.
 bash -x ./bin/start_webserver.sh
 
-# set up the storage pool for VMs
-sudo virsh pool-define-as "${VM_STORAGE_POOL}" dir - - - - "${VM_DISK_DIR}"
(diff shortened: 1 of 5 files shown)


test/bin/ci_phase_iso_boot.sh
```diff
@@ -27,6 +27,12 @@ cd ~/microshift/test
 # repository.
 bash -x ./bin/start_webserver.sh
 
+# set up the storage pool for VMs
```
dhellmann, line 30: I've been trying to retain the ability for a developer to run some setup steps and then run 1 scenario, since it is likely impossible to run all of the scenarios at once on a laptop. We should land this as-is if it works, but then move it to a separate script so we can have a "configure" phase. The ci build phase script works like that so far, but logically these steps don't make sense there.

test/bin/ci_phase_iso_boot.sh
```diff
@@ -27,6 +27,12 @@ cd ~/microshift/test
...
+sudo virsh pool-define-as "${VM_STORAGE_POOL}" dir - - - - "${VM_DISK_DIR}"
```
dhellmann, line 31: That's a lot of null arguments. :-)

pmtk: Would this work?
```suggestion
sudo virsh pool-define-as "${VM_STORAGE_POOL}" dir --target "${VM_DISK_DIR}"
```

test/bin/ci_phase_iso_boot.sh
pacevedom: Done.

test/bin/ci_phase_iso_boot.sh
pacevedom: Done.

test/bin/manage_vm_storage_pool.sh
```diff
@@ -0,0 +1,59 @@
...
+
+EOF
+}
+
+action_create() {
+    sudo sh -c '
+virsh pool-define-as '"${VM_STORAGE_POOL}"' dir --target '"${VM_DISK_DIR}"'
+virsh pool-build '"${VM_STORAGE_POOL}"'
+virsh pool-start '"${VM_STORAGE_POOL}"'
+virsh pool-autostart '"${VM_STORAGE_POOL}"'
```
pmtk, line 29: This shouldn't affect the result very much, but because of no `set -e` all these commands will run even if first fails. 
Although without `set -x` it might be hard to match stderr to command. 
Maybe they should be at least chained with `&&`?

dhellmann:
The `ostree refs` problem should be fixed by #2051, but this PR might need to be rebased to pick it up.

pmtk:
/lgtm
/hold

unhold at your convenience

pacevedom:
/unhold
Boot phase went through without errors and thats good enough for this PR, as it was the target.

PR #2047, opened 2023-07-12
USHIFT-1319: Add a message to the greenboot pod count check when it fails

ggiguash:
I simulated the pod "restart" by deleting pods in the `openshift-storage` namespace and the error reason was correctly p
printed.
```
Running Required Health Check Scripts...
STARTED
GRUB boot variables:
boot_success=0
boot_indeterminate=0
Greenboot variables:
...
...
Checking pod restart count in the 'openshift-ovn-kubernetes' namespace
Checking pod restart count in the 'openshift-service-ca' namespace
Checking pod restart count in the 'openshift-ingress' namespace
Checking pod restart count in the 'openshift-dns' namespace
Checking pod restart count in the 'openshift-storage' namespace
Pod restart count check failed for the 'openshift-storage' namespace
FAILURE
```

Closes: [USHIFT-1319](https://issues.redhat.com//browse/USHIFT-1319)

### DIFF
--- a/packaging/greenboot/microshift-running-check.sh
+++ b/packaging/greenboot/microshift-running-check.sh
@@ -128,15 +128,29 @@ for i in "${!PODS_NS_LIST[@]}"; do
 done
 
 # Verify that MicroShift core pods are not restarting
-pids=()
+declare -A pid2name
 for i in "${!PODS_NS_LIST[@]}"; do
     CHECK_PODS_NS=${PODS_NS_LIST[${i}]}
 
     echo "Checking pod restart count in the '${CHECK_PODS_NS}' namespace"
     namespace_pods_not_restarting "${CHECK_PODS_NS}" &
-    pids+=($!)
+    pid=$!
+
+    pid2name["${pid}"]="${CHECK_PODS_NS}"
 done
 
-for pid in "${pids[@]}"; do
-    wait "${pid}"
+# Wait for the restart check functions to complete, printing errors in case of a failure
+check_failed=false
+for pid in "${!pid2name[@]}"; do
+    if ! wait "${pid}" ; then
+        check_failed=true
+
+        name=${pid2name["${pid}"]}
+        echo "Pod restart count check failed for the '${name}' namespace"
+    fi
 done
+
+# Exit with an error code if the pod restart check failed
+if ${check_failed} ; then
+    exit 1
+fi


packaging/greenboot/microshift-running-check.sh
```diff
@@ -128,15 +128,22 @@ for i in "${!PODS_NS_LIST[@]}"; do
...
 done
 
-for pid in "${pids[@]}"; do
-    wait "${pid}"
+# Wait for the restart check functions to complete, printing errors in case of a failure
+for pid in "${!pid2name[@]}"; do
+    if ! wait "${pid}" ; then
+        name=${pid2name["${pid}"]}
+        echo "Pod restart count check failed for the '${name}' namespace"
+        exit 1
```
pmtk, line 147: if there's a problem with many namespaces, this will cause to print out first one.

do you think it might be better to print all problematic namespaces so it's easier to see if the problem is "global" or it's just one namespace?

packaging/greenboot/microshift-running-check.sh
```diff
@@ -128,15 +128,22 @@ for i in "${!PODS_NS_LIST[@]}"; do
...
+    pid2name["${pid}"]="${CHECK_PODS_NS}"
 done
 
-for pid in "${pids[@]}"; do
-    wait "${pid}"
+# Wait for the restart check functions to complete, printing errors in case of a failure
+for pid in "${!pid2name[@]}"; do
+    if ! wait "${pid}" ; then
+        name=${pid2name["${pid}"]}
+        echo "Pod restart count check failed for the '${name}' namespace"
```
pmtk, line 146: i don't want to put this flow on its head (maybe in another PR or tech debt?), but what do you think about providing info which pod actually failed, not just in which namespace?

ggiguash: The current implementation of restart only checks for counts on the namespace level. 
```
    count1=$(${OCGET_CMD} pods ${OCGET_OPT} -n "${ns}" -o 'jsonpath={..status.containerStatuses[].restartCount}' 2>/dev/null)
```

Tracking this on a per-pod level is going to complicate the implementation. I wonder why namespace pointer is not enough for troubleshooting? The next step is to request sos report, right?

dhellmann: There are a few issues with relying only on sos. By the time a user sees the greenboot failure, it's too late to collect sos details. That means reconfiguring the system to collect that report, which is itself very large. If, instead, the health check reported more specific detail then users may be able to avoid using sos at all in some cases. We don't need full logs, but it would be useful to have more detail than we have now.

ggiguash: Can you suggest what information might be useful here? For example, we could easily extract pod bootstrap or container logs?
Alternatively, we could run `sos report` if greenboot fails and point the user to that file as a source of the information.

pmtk: I think pod's name is bare minimum. We know the pod is restarting, so there's no problem pulling images, so maybe `tail -n10` of failing container?
I think we discussed running sos when greeboot fails some time ago and I vaguely recall we don't want to create that dependency (and gathering sos takes some time, right? it's not as quick as getting logs)

dhellmann: Right, @pmtk, adding sos all the time on a rollback impacts the recovery time. That should be a user choice, not a requirement from us.

Pod names, namespaces, actual vs. expected counts, and log tails are all good ideas for what to include. Anything that might not already be in a journal log to be collected after the reboot is a good candidate.

PR #2055, opened 2023-07-13
NO-ISSUE: add __pycache__ to .gitignore

pmtk:
It's a directory created when running robot framework tests, but we don't want to commit it to the repo.


pacevedom (commented):
/lgtm

PR #2057, opened 2023-07-13
USHIFT-1435: add script to report on go dependencies

dhellmann:


scripts/dependency_report.sh
```diff
@@ -0,0 +1,126 @@
...
+# Show any go dependencies in MicroShift's go.mod that do not appear
+# in other OpenShift components.
+
+set -o errexit
+set -o errtrace
+set -o nounset
+set -o pipefail
+
+REPOROOT="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
+WORKDIR="${REPOROOT}/_output/dependency_report"
```
ggiguash, line 12: If we're adding a new directory, we should also clean it in Makefile

scripts/dependency_report.sh
```diff
@@ -0,0 +1,126 @@
...
+    if [ ! -f "${name}/go.mod" ]; then
+        continue
+    fi
+
+    build_dep_list "${name}"
+
+    sed -e "s|^|${repo_url} |g" "${name}/require.txt" >> all_require.txt
+done
+
+sort -u microshift/require.txt | while read -r dep; do
```
ggiguash, line 108: Rather than dumping this output, maybe we should redirect it to a file? 
I think we may want to accept the output file name from the command line. 

Otherwise, the output of processing is intermixed with the dependency output.

scripts/dependency_report.sh
```diff
@@ -0,0 +1,126 @@
...
+
+set -o errexit
+set -o errtrace
+set -o nounset
+set -o pipefail
+
+REPOROOT="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
+WORKDIR="${REPOROOT}/_output/dependency_report"
+
+PULL_SECRET_FILE="${HOME}/.pull-secret.json"
```
ggiguash, line 14: Let's accept this as an argument to the script?

ggiguash: To work properly, it should also contain a key for registry.ci.openshift.org in addition to what we're getting in a pull secret. I think we might want to mention this in the usage message.

ggiguash (changes requested):

scripts/dependency_report.sh
dhellmann: I don't understand what you mean.

scripts/dependency_report.sh
dhellmann: What extra output are you seeing?

scripts/dependency_report.sh
dhellmann: Done

dhellmann (commented):
@ggiguash I really only submitted this PR as a way to record what I did to produce the report. Let's just land it so we can focus on more important things.

scripts/dependency_report.sh
ggiguash: `make clean` would not delete this new directory , so _output will remain.

PR #2064, opened 2023-07-14
USHIFT-1432: E2E RF "Rollback" test

pmtk:


pmtk:
/test microshift-metal-tests

Retest due to flaky snapshot test, this one passed :)

test/image-blueprints/rhel92-source-fake-next-minor.toml
```diff
@@ -0,0 +1,25 @@
+name = "rhel-9.2-microshift-source-fake-next-minor"
```
dhellmann, line 1: I might have just called this "rhel-9.2-microshift-next-minor" or "rhel-9.2-microshift-fake-next-minor", but we haven't really talked about naming conventions and that's a nit so let's not worry about it for now.

test/bin/build_images.sh
```diff
@@ -24,20 +24,22 @@ if [ ! -d "${LOCAL_REPO}" ]; then
     error "Run ${SCRIPTDIR}/create_local_repo.sh before building images."
     exit 1
 fi
-release_info_rpm=$(find "${LOCAL_REPO}" -name 'microshift-release-info-*.rpm')
+release_info_rpm=$(find "${LOCAL_REPO}" -name 'microshift-release-info-*.rpm' -a -not -name '*_fake_*')
```
dhellmann, line 27: We're going to have a few different RPMs now, especially when @copejon adds the "upgrade from HEAD to PR" test scenario. Maybe we can set up the RPM builds so they end up in different directories and different repos so it's easier to manipulate the files cleanly. Again, we can evolve this in another PR.

test/bin/ci_phase_test.sh
```diff
@@ -9,6 +9,9 @@ source "${SCRIPTDIR}/common.sh"
 
 cd "${TESTDIR}"
 
+# Start the web server to host ostree commit repository.
+bash -x ./bin/start_webserver.sh
```
dhellmann, line 13: The server should have been started in the boot phase script. This will restart it, which should not break anything, but I wonder if you ran into a problem where it wasn't running?

dhellmann (commented):
/lgtm

test/scenarios/rhel-9.2-microshift-source-rollback.sh
```diff
@@ -0,0 +1,18 @@
...
+    launch_vm host1
+}
+
+scenario_remove_vms() {
+    remove_vm host1
+}
+
+scenario_run_tests() {
+    run_tests host1 \
+        --variable "FAKE_NEXT_MINOR_REF:rhel-9.2-microshift-source-fake-next-minor" \
```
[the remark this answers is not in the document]
pmtk: No, but it's up to "operator" to set up "next fake minor" instead of another alias to "microshift-source" (which would work but current approach explicitly checks if it's in fact restored, "microshift-source" could be migrated)

test/bin/ci_phase_test.sh
pmtk: Yes, because of https://github.com/openshift/microshift/blob/main/test/bin/ci_phase_iso_boot.sh#L50
Since these are `ci_phase_*.sh` scripts, we could just keep first `start_webserver.sh` in iso_boot and let it run indefinitely

test/bin/ci_phase_iso_build.sh
```diff
@@ -50,7 +50,8 @@ bash -x ./bin/configure_hypervisor_firewall.sh
 # Re-build from source.
 cd ~/microshift/
 rm -rf ./_output/rpmbuild
-make rpm
+time make rpm
+time make -C test/ fake-next-minor-rpm
```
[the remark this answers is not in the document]
pmtk: Oh yes, missed the `rebuild_source_image.sh`, sorry

test/bin/build_images.sh
pmtk: I didn't notice any structure requirements for rpm repository. So maybe we could have one repository, but different subdirs?
Multiple repos would be just an overhead, if you feed them all to osbuilder, then you still need to specify versions to pick right one

test/bin/ci_phase_test.sh
dhellmann: Eventually I would like to merge the phases. There's no reason to run the scripts from different pods on the CI system, and I think it will be simpler. Now that we have it working, though, I think that sort of change is a bit of a distraction, so let's just keep this change.

PR #2066, opened 2023-07-14
USHIFT-1320 USHIFT-1431: E2E RF Test: "block downgrade"

pmtk:


test/resources/ostree.resource
```diff
@@ -30,15 +23,15 @@ Current Deployment Should Be
...
     [Documentation]    Deploys given ref and configures test agent for failing greenboot.
     ...    It expects the system to roll back.
-    [Arguments]    ${ref}
+    [Arguments]    ${ref}    ${write_agent_cfg}=${TRUE}
 
     ${initial_deploy_id}=    Get Booted Deployment Id
     ${deploy_id}=    Rebase System    ${ref}
-    TestAgent.Write
-    Write Greenboot Microshift Wait Timeout    60
+    IF    ${write_agent_cfg}    TestAgent.Write
```
dhellmann, line 30: Is the config always the same?

test/scenarios/rhel-9.2-microshift-source-block-downgrade.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+
+scenario_create_vms() {
+    prepare_kickstart host1 kickstart.ks.template rhel-9.2-microshift-source-fake-next-minor
```
dhellmann, line 6: We could combine the next-minor and previous-minor images for a similar test that ensures we block skipping 2 minor versions.

dhellmann (commented):
/lgtm

test/scenarios/rhel-9.2-microshift-source-block-downgrade.sh
pmtk: This is very good point and I have a feeling that we lost implementation of this when upgrading from 4.13... https://github.com/openshift/microshift/blob/main/pkg/admin/prerun/version.go#L36

test/resources/ostree.resource
pmtk: No, this line is only persist what test should set up, see [fdo.robot#L33](https://github.com/openshift/microshift/blob/main/test/suites-ostree/fdo.robot#L33) and [failed-upgrade.robot#L30](https://github.com/openshift/microshift/blob/main/test/suites-ostree/failed-upgrade.robot#L30)

dhellmann: Ah, right, I forgot the config was managed separately from the write step.

test/scenarios/rhel-9.2-microshift-source-block-downgrade.sh
dhellmann: We could attach a test to https://issues.redhat.com/browse/USHIFT-1088

PR #2070, opened 2023-07-15
USHIFT-1448: remove redundant ostree setup in kickstart

dhellmann:
The call to ostreesetup creates an edge remote with the correct URL,
so there is no need to delete it and re-create it.

/assign @ggiguash

### DIFF
--- a/test/kickstart-templates/kickstart.ks.template
+++ b/test/kickstart-templates/kickstart.ks.template
@@ -47,10 +47,6 @@ if [ -n "REPLACE_PUBLIC_IP" ]; then
     echo "    - REPLACE_PUBLIC_IP" >>/etc/microshift/config.yaml
 fi
 
-# Update the ostree server URL
-ostree remote delete edge
-ostree remote add --no-gpg-verify edge REPLACE_OSTREE_SERVER_URL
-
 # The pull secret is mandatory for MicroShift builds on top of OpenShift, but not OKD
 # The /etc/crio/crio.conf.d/microshift.conf references the /etc/crio/openshift-pull-secret file
 cat > /etc/crio/openshift-pull-secret <<EOF


scripts/image-builder/config/kickstart.ks.template
```diff
@@ -37,10 +37,6 @@ ostreesetup --nogpg --osname=rhel --remote=edge --url=file:///run/install/repo/o
 
 %post --log=/var/log/anaconda/post-install.log --erroronfail
 
-# Update the ostree server URL
-ostree remote delete edge
-ostree remote add --no-gpg-verify edge REPLACE_OSTREE_SERVER_URL
```
ggiguash, line 42: The ostreesetup command uses the `file:///run/install/repo/ostree/repo` path, which is necessary during the installation. In the post-install stage, we're replacing it by REPLACE_OSTREE_SERVER_URL.
I do not understand why this replacement is not necessary - where is  REPLACE_OSTREE_SERVER_URL set?

dhellmann: I keep forgetting this file is used in a different way. I've removed this change from the PR.

test/kickstart-templates/kickstart.ks.template
```diff
@@ -47,10 +47,6 @@ if [ -n "REPLACE_PUBLIC_IP" ]; then
     echo "    - REPLACE_PUBLIC_IP" >>/etc/microshift/config.yaml
 fi
 
-# Update the ostree server URL
-ostree remote delete edge
-ostree remote add --no-gpg-verify edge REPLACE_OSTREE_SERVER_URL
-
```
ggiguash, line 53: I think the ostreesetup should point to /run/install so that the OS installation finishes. In post install stage, we can point it to "our" repos. 
Let's see if the image can be created this way...

dhellmann: In the scenario tests, there is a web server with all of the commits and we install over the network, not from files in an ISO image. That's why this step is redundant.

ggiguash: /lgtm

PR #2073, opened 2023-07-16
NO-ISSUE: Review non-ostree test harness code and docs

ggiguash:
The following minor fixes were implemented:
* Add e2e directory cleanup to the Makefile clean target
* Reuse existing verify-rf target in Makefile
* Fix the run.sh non-ostree flow docs and usage information
* Enable shellcheck warning SC1091 and adjust all scripts including others
* Remove executable permission from version.robot


test/bin/build_images.sh
```diff
@@ -22,31 +23,31 @@ fi
...
-# Machine platform type ("x86_64")
-UNAME_M=$(uname -m)
-export UNAME_M
-export LOCAL_REPO              # defined in common.sh
-export NEXT_REPO               # defined in common.sh
-export SOURCE_VERSION          # defined earlier
-export FAKE_NEXT_MINOR_VERSION # defined earlier
-export MINOR_VERSION           # defined earlier
-export PREVIOUS_MINOR_VERSION  # defined earlier
+SUBST_UNAME_M=$(uname -m) # Machine platform type ("x86_64")
```
dhellmann, line 38: This feels like it is going to be harder to maintain. Part of the point of using the same variables in the template and in the script is that we don't need to create a bunch of extra variables.

test/bin/common.sh
```diff
@@ -35,11 +35,11 @@ NEXT_RPM_SOURCE="${ROOTDIR}/_output/rpmbuild-fake-next-minor"
 
 # Location of local repository used by composer
 # shellcheck disable=SC2034  # used elsewhere
-LOCAL_REPO="${IMAGEDIR}/rpm-repos/microshift-local"
+SUBST_LOCAL_REPO="${IMAGEDIR}/rpm-repos/microshift-local"
 
 # Location of local repository used by composer
 # shellcheck disable=SC2034  # used elsewhere
-NEXT_REPO="${IMAGEDIR}/rpm-repos/microshift-fake-next-minor"
+SUBST_NEXT_REPO="${IMAGEDIR}/rpm-repos/microshift-fake-next-minor"
```
dhellmann, line 42: This file, at least, should have "normal" variable names only. Scripts might need to use the values.

dhellmann (commented):
The readme and shellcheck changes look OK. I am not a fan of renaming the variables, but let's see what the rest of the team thinks.

test/bin/build_images.sh
ggiguash: OK, I reverted the commit.

test/bin/common.sh
ggiguash: OK, I reverted the commit.

test/run.sh
```diff
@@ -78,6 +78,7 @@ if [ -z "${TESTS}" ]; then
 fi
 
 set -x
+pushd "${ROOTDIR}/test" &>/dev/null
```
dhellmann, line 81: I think line 73 is already doing this?

dhellmann (commented):
/lgtm

test/bin/build_images.sh
```diff
@@ -6,6 +6,7 @@
 set -euo pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# shellcheck source=/dev/null
```
pmtk, line 9: Why this instead of properly configuring shellcheck to find the file?

test/Makefile
```diff
@@ -11,3 +12,8 @@ fake-next-minor-rpm:
 	MICROSHIFT_VERSION="$(MAJOR).$(NEXT_MINOR).$(PATCH)_fake_next_minor" \
 	RPMBUILD_DIR=$(RPMBUILD_DIR) \
 		$(MAKE) -C $(REPO) rpm
+
+.PHONY: robotidy
```
pmtk, line 16: How about moving logic of `scripts/verify/verify-rf.sh` here?

test/README.md
```diff
@@ -65,7 +75,7 @@ The `-h` option prints usage instructions:
 
 ```
 $ ./test/run.sh -h
-run.sh [-h] [-n] [-o output_dir] [test suite files]
+run.sh [-h host] [-n] [-o output_dir] [test suite files]
```
pmtk, line 78: Description suggests `-h` is help
```suggestion
run.sh [-h] [-n] [-o output_dir] [test suite files]
```

test/bin/build_images.sh
ggiguash: The problem is that shellcheck does not follow paths when variables are specified. I could not make it work in VSCode.

https://www.shellcheck.net/wiki/SC1090

test/Makefile
ggiguash: Good point. We already have that functionality, so let's reuse it.

test/bin/build_images.sh
pmtk: I don't think that problem with VSCode extension should be extended to `make verify-sh` thus affecting the CI.
Can you try changing setting of `shellcheck.customArgs`  to match args given to shellcheck in `verify-shell.sh` and see if that helps?

ggiguash: Looking at [this code](https://github.com/openshift/microshift/blob/main/scripts/verify/verify-shell.sh#L26), we're already excluding the SC1091, so there is not shell-check error.

My fix only comes to address the issue inside the VSCode.

ggiguash: @pmtk , I uploaded another fix. Is this what you had in mind?

pmtk:
I feel much better with current version. However, I noticed there's following option:
```
-P SOURCEPATHS      --source-path=SOURCEPATHS  Specify path when looking for sourced files ("SCRIPTDIR" for script's dir)
```
This would be worth exploring, but I don't want to drag this PR, we can handle it in another PR or create a ticket for later, whatever you think works best

/lgtm
/hold

test/agent/microshift-test-agent.sh
```diff
@@ -62,7 +62,7 @@ _get_current_boot_number() {
     local max_boot_attempts
 
     if test -f "${GREENBOOT_CONFIGURATION_FILE}"; then
-        # shellcheck source=/dev/null
+        # shellcheck source=test/bin/common.sh
```
dhellmann, line 65: This isn't the file being sourced, right?

pmtk: Oh, that's true. For this /dev/null is what I'd expect

ggiguash:
Need few minor fixed in "shellcheck source"
/lgtm cancel

ggiguash:
@copejon , this is LGTM, except you need to run `make verify-rf` and fix the errors.

Makefile
```diff
@@ -309,6 +309,7 @@ clean-cross-build:
 	if [ -d '$(OUTPUT_DIR)/staging' ]; then $(RM) -rf '$(OUTPUT_DIR)/staging'; fi
 	if [ -d '$(OUTPUT_DIR)/venv' ]; then $(RM) -rf '$(OUTPUT_DIR)/venv'; fi
 	if [ -d '$(OUTPUT_DIR)/robotenv' ]; then $(RM) -rf '$(OUTPUT_DIR)/robotenv'; fi
+	if ls $(OUTPUT_DIR)/e2e-* &>/dev/null ; then $(RM) -rf $(OUTPUT_DIR)/e2e-* ; fi
```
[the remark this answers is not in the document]
ggiguash: I agree. We could just delete the `_output` directory to simplify things. 
Let's do this in a separate PR.

PR #2080, opened 2023-07-17
USHIFT-1456: use more complete version numbers in release tagging

dhellmann:
We had 2 builds of 4.14.0~ec.3, which confused the release note tool. This
change adds the date and patch number values from the RPM version to the
tag and release name in case we have a similar issue in the future.

/assign @ggiguash


ggiguash:
@dhellmann, Looks like Python linter is complaining.

PR #2082, opened 2023-07-17
USHIFT-1458: update build scripts to make rpm builds consistent

dhellmann:
Consolidate several places where we build RPMs into a single script that
can be reused.

/assign @pmtk @jogeo @copejon


dhellmann:
I also took this opportunity to provide a little more context to the output of some of the scripts, at @dhensel-rh's suggestion.

PR #2083, opened 2023-07-17
[NO ISSUE] handle thinpool provisioning through rf

copejon:
Thin pools should be created and deleted as part of the test process, rather than during the infra spin up phase


ggiguash:
@copejon , this is LGTM, except you need to run make verify-rf and fix the errors.

ggiguash:
/retitle USHIFT-1444: handle thinpool provisioning through robot framework

ggiguash:
This fix also addresses [USHIFT-1447](https://issues.redhat.com/browse/USHIFT-1447)

pkg/components/storage.go
```diff
@@ -145,9 +144,5 @@ func startCSIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
 		klog.Warningf("Failed to apply sccs %v: %v", scc, err)
 		return err
 	}
-	if err := assets.ApplyVolumeSnapshotClass(ctx, kubeconfigPath, vcs, nil, nil); err != nil {
```
pmtk, line 148: So, now whole `volumeSnapshotClassApplier` and related are unused?

Why do we delete `assets/components/lvms/topolvm_default-volumeclass.yaml`?
Does removing provisioning of thin volume from infra steps cause a failure of MicroShift startup because it cannot apply this file?

copejon: > So, now whole volumeSnapshotClassApplier and related are unused?

Correct. I opted to leave them for the moment as we may have a need for them later.

> Why do we delete assets/components/lvms/topolvm_default-volumeclass.yaml?
This is a small fix to the early snapshot implementation. Thinpools aren't created automatically, and MicroShift does not check for them, so deploying a volumesnapshotclass object is superfluous and could create some confusion for users.  If a user deploys microshift with snapshotting enabled, they will have handled the thinpool creation, lvmd.yaml configuration, as well as creating a volumesnapclass for the device class.

dhellmann: Can that be put into its own commit with a message explaining all of that?

copejon:
Changes in this PR also address USHIFT-1447

test/suites/snapshot.robot
```diff
@@ -41,14 +41,24 @@ Snapshotter Smoke Test
 
 *** Keywords ***
 Test Suite Setup
-    [Documentation]    Setup test namespace, patch the lvmd for thin-volume support, and restart microshift
+    [Documentation]    Setup test namespace, patch the lvmd for thin-volume support, and restart microshift for
```
dhellmann, line 44: Could you remove the `robot:skip-on-failure` tag on line 30 as part of this PR to ensure we're landing the job in a green state? I see it's passing, so should be OK.

test/suites/snapshot.robot
```diff
@@ -83,3 +100,10 @@ Extend LVMD Config
     ${patch_cfg}=    OperatingSystem.Get File    ${LVMD_PATCH}
     ${merged_cfg}=    Lvmd Merge    ${default_cfg}    ${patch_cfg}
     Upload String To File    ${merged_cfg}    /etc/microshift/lvmd.yaml
+
+Delete LVMD Config
```
dhellmann, line 104: Should this work more like the MicroShift config and restore the "old" file contents? If not, maybe the `Lvmd Merge` keyword can be simplified, too?

copejon: The latter is probably the best approach.  I'll rework it to just write a hardcoded lvmd to disk without worrying about what's in the configMap.  Unless we add additional storage tests, there's no need to restore the lvmd - it's not there to begin with. It's enough to delete the file and restart the service.

copejon:
Sigh. SSHLibrary.PutFile cannot perform sudo actions, and so can't simply scp the lvmd.yaml file to /etc/microshift/lvmd.yaml.  Will take a more manual approach with SSHLibrary.Execute Command

copejon:
> > Sigh. SSHLibrary.PutFile cannot perform sudo actions, and so can't simply scp the lvmd.yaml file to /etc/microshift/lvmd.yaml. Will take a more manual approach with SSHLibrary.Execute Command
> 
> @copejon , you do not need to elevate SSH. Instead, can we run "sudo lvcreate", etc.? Do we need SSH at all?

Summarizing our slack convo: 

The issue isn’t actually with running the lvcreate command, but with copying over the lvmd.yaml from host to test vm.  I had used SSHLibrary.Put File, but since the destination is owned by root, it fails with a permissions error.  Instead I'll just use Upload String to File in a manner similar to how the microshift config is handled.

copejon:
ci/prow/microshift-e2e is blocked on https://github.com/openshift/release/pull/41518

test/resources/microshift-config.resource
```diff
@@ -58,3 +59,41 @@ Upload MicroShift Config
...
+    END
+
+Upload Lvmd Config
+    [Documentation]    Upload a test's lvmd.yaml file to the MicroShift host
+    [Arguments]    ${config_content}
+    Upload String To File    ${config_content}    /etc/microshift/lvmd.yaml
+
+Clear Lvmd Config
+    [Documentation]    Removes the LVMD configuration as part of restoring test environment
+    ${stderr}    ${rc}=    sshLibrary.Execute Command    rm -f /etc/microshift/lvmd.yaml
```
ggiguash, line 96: why do we need `sshLibrary` here, why not `OperatingSystem.Execute Command` like above?

test/resources/microshift-host.resource
```diff
@@ -55,3 +55,17 @@ Reboot MicroShift Host
     Sleep    30s
     Wait Until Keyword Succeeds    5m    15s
     ...    SSH Connection To MicroShift Host Should Be Functional
+
+Create Thin Storage Pool
+    [Documentation]    Create a new thin storage pool
+    ${stderr}    ${rc}=    SSHLibrary.Execute Command    lvcreate --size 4G --thin /dev/rhel --name thin
```
ggiguash, line 61: why do we need `sshLibrary` here, why not `OperatingSystem.Execute Command` like above?

test/resources/microshift-host.resource
```diff
@@ -55,3 +55,17 @@ Reboot MicroShift Host
...
+    ...    sudo=True    return_rc=True    return_stderr=True    return_stdout=False
+    Log    ${stderr}
+    Should Be Equal As Integers    0    ${rc}
+
+Delete Thin Storage Pool
+    [Documentation]    Delete the thin storage pool as part of restoring test env
+    ${stderr}    ${rc}=    SSHLibrary.Execute Command    lvremove -y rhel/thin
```
ggiguash, line 68: why do we need `sshLibrary` here, why not `OperatingSystem.Execute Command` like above?

test/resources/microshift-config.resource
copejon: In this instance, it's necessary to run the command on the remote host. I'm not sure which line your referring to above.  Although I see I could have removed the `SSHLibrary` dot prefix from here.

PR #2084, opened 2023-07-18
USHIFT-1399: Pre run filelogger

eggfoobar:
Added support for logging to both klog and file output during prerun step.

/assign @pmtk


pkg/util/multilogger/filelogger.go
```diff
@@ -0,0 +1,64 @@
...
+// NewFileLogger creates a logger that writes to both klog and the specified file path.
+// The intention is to be best effort logging, meaning that it will try to create
+// the full file path if it does not exist.
+func NewFileLogger(filePath string) (*fileLogger, error) {
+	err := os.MkdirAll(filepath.Dir(filePath), 0755)
+	if err != nil {
+		return nil, err
+	}
+
+	f, err := os.OpenFile(filePath, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0644)
```
dhellmann, line 44: I can't remember how these flags work. Is this going to append to an existing file? Is that what we want for this case? The idea is to get the information about why the most recent pre-run phase failed to emit them as part of the logs in greenboot-healthcheck to speed debugging.

I wonder if we can collect that same information by asking for the journal of the microshift service for the current boot? Something like `journalctl -u microshift --boot`?

eggfoobar: Oh, good catch, yeah that should be `TRUNC`. In regard to `journalctl`, I like that more, let me see if we can get the same info from that.

eggfoobar: Changed approach per our discussion.

dhellmann (commented):
This looks generally good. 

Do you think there's any benefit to catching the error higher up, say before main() exits?

Do you want to add the health check script update to this PR, or handle that separately?

eggfoobar:
I liked the approach of moving up the call site, makes more sense to catch it there and centralize writing out errors at the point we start services.

pkg/cmd/run.go
```diff
@@ -93,8 +115,11 @@ func RunMicroshift(cfg *config.Config) error {
 		klog.Fatalf("MicroShift must be run privileged")
 	}
 
+	cleanUpPreviousFileLogs()
+
 	if err := performPrerun(); err != nil {
 		klog.ErrorS(err, "Pre-run procedure failed")
+		writeFileLogError(preRunFailedLogPath, err)
```
dhellmann, line 122: This is good. I was thinking we might record the error from `RunMicroshift` in case something after the pre-run phase triggers the error, but let's start with this and see if it's useful to expand the scope later.

dhellmann (commented):
This looks good. I'd like Patryk to take a look to make sure we're all on the same page.

/assign @pmtk

pkg/cmd/run.go
eggfoobar: Sure, that's a trivial change if need be, since we were calling out pre-run specifically in this ticket that's the reason it's still oriented around that, but we can change it up to capture all start up errors and propagate that up through greenboot.

pmtk: Prerun is now split into two phases: data management and version checking/updating which happens in line 147 (`prerun.CheckAndUpdateDataVersion()`) which we also need to catch.

We might need to rename some functions to improve readability, but for now:
if we want to persist only prerun errors, we might wrap errors from these two functions and only persist if that's the error type, something like in RunE func:
```
if err := RunMicroShift(...); err != nil {
    if errors.Is(err, FailedPreRun) {
        writeFileLogError(...)
    }
}
```

pkg/cmd/run.go
```diff
@@ -67,6 +75,20 @@ func NewRunMicroshiftCommand() *cobra.Command {
...
+	for _, p := range cleanUpFileLogPaths {
+		if errLog := p.Remove(); errLog != nil {
+			klog.ErrorS(errLog, "failed to remove log file", "path", p)
+		}
+	}
+}
+
+func writeFileLogError(l util.LogFilePath, err error) {
+	if errLog := l.Write([]byte(err.Error())); errLog != nil {
+		klog.ErrorS(errLog, "failed to write log file", "path", l)
```
pmtk, line 88:
```suggestion
		klog.ErrorS(errLog, "Failed to write log file", "path", l)
```

pkg/cmd/run.go
```diff
@@ -67,6 +75,20 @@ func NewRunMicroshiftCommand() *cobra.Command {
 	return cmd
 }
 
+func cleanUpPreviousFileLogs() {
+	for _, p := range cleanUpFileLogPaths {
+		if errLog := p.Remove(); errLog != nil {
+			klog.ErrorS(errLog, "failed to remove log file", "path", p)
```
pmtk, line 81:
```suggestion
			klog.ErrorS(errLog, "Failed to remove log file", "path", p)
```

pkg/util/util.go
```diff
@@ -109,3 +110,24 @@ func HealthCheckServer(ctx context.Context, path, port string) (start func() err
 
 	return start, shutdown
 }
+
+type LogFilePath string
+
+// Write writes data to the desired filename, calling multiple times will ovewrite
```
pmtk, line 116:
```suggestion
// Write writes data to the desired filename, calling multiple times will overwrite
```

pkg/cmd/run.go
```diff
@@ -67,6 +75,20 @@ func NewRunMicroshiftCommand() *cobra.Command {
...
+		}
+	}
+}
+
+func writeFileLogError(l util.LogFilePath, err error) {
+	if errLog := l.Write([]byte(err.Error())); errLog != nil {
```
pmtk, line 87: Similarly to "log msg review", we might want to do a "error msg review" to make sure that what we write here makes sense, but that's another ticket

pkg/cmd/run.go
eggfoobar: I'm thinking we avoid custom type errors for now since that would be a bigger conversation. I think just capturing both `performPrerun` and `prerun. CheckAndUpdateDataVersion()` would be good enough?

pkg/cmd/run.go
eggfoobar: done

pkg/util/util.go
eggfoobar: lol every time, thanks!

pkg/cmd/run.go
eggfoobar: updated

pkg/cmd/run.go
eggfoobar: Yeah good point, didn't want to wrap it and have more repeated words in the output

pkg/cmd/run.go
eggfoobar: From slack conversation, keeping this simple and focusing only on pre-run errors.

PR #2091, opened 2023-07-18
USHIFT-1455: Verify that upgrading by 2 minor versions is blocked

pmtk:


pkg/admin/prerun/version.go
```diff
@@ -29,19 +29,37 @@ func CheckAndUpdateDataVersion() error {
...
-	if err != nil {
-		if !errors.Is(err, errDataVersionDoesNotExist) {
-			return fmt.Errorf("failed to get version of existing MicroShift data: %w", err)
-		}
-		klog.InfoS("Version file does not exist yet")
-	} else {
-		if err := checkVersionCompatibility(execVer, dataVer); err != nil {
-			return fmt.Errorf("checking version skew failed: %w", err)
-		}
+	dataVersionMissing := err != nil && errors.Is(err, errDataVersionDoesNotExist)
```
dhellmann, line 32: Does `errors.Is()` not handle nil?

test/bin/download_images.sh
```diff
@@ -71,5 +71,7 @@ title "Updating ostree references in ${IMAGEDIR}/repo"
 cd "${IMAGEDIR}"
 ostree refs --repo=repo --force \
        --create "rhel-9.2-microshift-failing" "rhel-9.2-microshift-source"
+ostree refs --repo=repo --force \
```
dhellmann, line 74: This is a bit simpler than managing the version number in the image reference within the scenario script. Should we update https://github.com/openshift/microshift/blob/main/test/scenarios/upgrade-minor-version-healthy.sh#L9 as well?

test/suites-ostree/block-upgrade-2-minor.robot
```diff
@@ -0,0 +1,61 @@
...
+    ...    X.Y-1 to X.Y+1 is blocked (skipping current X.Y).
+
+    Wait For Healthy System
+    ${initial_deploy_backup}=    Get Future Backup Name For Current Boot
+
+    Deploy Commit Expecting A Rollback    ${TOO_NEW_MICROSHIFT_REF}
+
+    Wait For Healthy System
+    Backup Should Exist    ${initial_deploy_backup}
+    Journal Should Have Information About Failed Version Comparison
```
dhellmann, line 33: I'm originally assumed we would reuse the existing failed upgrade suite. For test code I do lean towards a bit of repetition for clarity, so it's not a problem. How different are the 2 files aside from this keyword?

dhellmann (commented):
A couple of small questions, but this looks good overall.

pkg/admin/prerun/version.go
pmtk: Yes, changed

test/bin/download_images.sh
pmtk: I just noticed that `download_images.sh` does not get `$PREVIOUS_MINOR_VERSION`.
I'm not sure but I think it worked when using `test/bin/rebuild_source_images.sh`.

I'll delete this line and follow the way from file you linked

test/suites-ostree/block-upgrade-2-minor.robot
pmtk: `failed-upgrade.robot` uses test-agent to block the backup to check if data will be backed up after rollback.
This doesn't add anything extra beside providing two commits too far apart

dhellmann: I was thinking we should be able to have 1 test suite file that says "starting from image A and rebasing to image B, the upgrade will fail and that failure is detected and handled correctly". Then the _reason_ for the failure is built into image B.

Maybe that's too simplistic, and we need the tests to configure image B to ensure that failure, like setting the agent sometimes or a version number other times? If that's the case, perhaps we could at least reuse a lot of the keywords, like you're doing here.

I don't think we should hold anything up over this, but it's something to think about as we evolve over time.

test/resources/ostree.resource
```diff
@@ -90,3 +90,18 @@ Wait For Transaction To End
 
     Wait Until Keyword Succeeds    2m    15s
     ...    No Transaction In Progress
+
+Journal Should Have Information About Failed Version Comparison
+    [Documentation]    Assert that unhealthy deployment's journal contains
+    ...    information about failed version compatibility check
+
+    FOR    ${boot}    IN RANGE    -3    0
```
dhellmann, line 98: I see you're just moving this, but the loop looks a little odd. Should there be a sleep somewhere? Or should this be a "wait until keyword succeeds" loop?

dhellmann (commented):
It looks like the metal job timed out.

test/suites-ostree/block-upgrade-2-minor.robot
pmtk: This sounds very good, though "the reason for the failure is built into image B" is kind of problematic given how much time is needed to build a commit.

Is three-phase build correct assumption? Like `rhel92~microshift-src~fail-greenboot.toml`
Rebuilding RPMs will kick off builds of phase two and three. Seems like a choice between dev productivity and conciseness (not duplicating the test cases with just a minor difference).

Maybe we could use a test suite variable provided by the scenario.sh that would influence test agent set up. Idk, just some random thoughts

test/resources/ostree.resource
pmtk: Not sure, why?
I need to execute journalctl 3 times with different `--boot` values: `-3`, `-2, and `-1`. This happens when system rolls back and it's only `journalctl | grep`

pmtk:
> It looks like the metal job timed out.

Yeah, I need to investigate it but it's kind of surprising it took that long:

> Step microshift-metal-tests-openshift-microshift-infra-iso-boot failed after **2h10m8s**

pmtk:
Presubmits fail because snapshot.robot leaves lvmd config with thin volume which does not exists resulting in lvmd container failing consistently and microshift never getting ready

PR #2095, opened 2023-07-19
OCPBUGS-16192: Use no uplink br-ex in mutlinode mode

pliurh:
1. Enable no uplink mode for ovnkube.
2. Configure dummy IP for br-ex interface.

### DIFF
--- a/scripts/multinode/configure-pri.sh
+++ b/scripts/multinode/configure-pri.sh
@@ -43,6 +43,11 @@ function configure_microshift() {
     sleep 5
     sudo systemctl start microshift-ovs-init.service
 
+    # OVN-K expects br-ex to have IP address assigned, add dummy IP to br-ex.
+    if ! ip addr show br-ex 2>/dev/null | grep -q '10.44.0.0/32'; then
+        sudo ip addr add 10.44.0.0/32 dev br-ex
+    fi
+
     # Configure MicroShift to allow API server access by an IP address
     cat <<EOF | sudo tee /etc/microshift/config.yaml &>/dev/null
 apiServer:
--- a/scripts/multinode/configure-sec.sh
+++ b/scripts/multinode/configure-sec.sh
@@ -40,6 +40,11 @@ function configure_microshift() {
     sleep 5
     sudo systemctl start microshift-ovs-init.service
 
+    # OVN-K expects br-ex to have IP address assigned, add dummy IP to br-ex.
+    if ! ip addr show br-ex 2>/dev/null | grep -q '10.44.0.0/32'; then
+        sudo ip addr add 10.44.0.0/32 dev br-ex
+    fi
+
     # Stop and unload the kubelet service, cleaning up its old data
     if [ "$(systemctl is-active kubelet.service)" = "active" ] ; then
         sudo systemctl stop --now kubelet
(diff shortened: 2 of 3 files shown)


scripts/multinode/configure-pri.sh
```diff
@@ -43,6 +43,9 @@ function configure_microshift() {
     sleep 5
     sudo systemctl start microshift-ovs-init.service
 
+    # Add dummy IP to br-ex
+    sudo ip addr add 10.44.0.0/32 dev br-ex
```
ggiguash, line 47: Is this call idempotent? In other words, if we run the script on the same machine twice, does it work for the second time without failing?

scripts/multinode/configure-sec.sh
```diff
@@ -40,6 +40,9 @@ function configure_microshift() {
     sleep 5
     sudo systemctl start microshift-ovs-init.service
 
+    # Add dummy IP to br-ex
+    sudo ip addr add 10.44.0.0/32 dev br-ex
```
ggiguash, line 44: Why are we choosing the IP from 10.44.0.0 subnet? Is it related to some default configuration setting?

scripts/multinode/configure-pri.sh
ggiguash: Also, can we please add a comment in the script explaining why this is done?

scripts/multinode/configure-sec.sh
pliurh: We use this IP in the single node cluster as the default KAS advertise address. We can use any IP as long as it doesn't overlap with any other internal/external IP ranges.

scripts/multinode/configure-pri.sh
pliurh: fixed. PTAL

scripts/multinode/configure-pri.sh
```diff
@@ -43,6 +43,9 @@ function configure_microshift() {
     sleep 5
     sudo systemctl start microshift-ovs-init.service
 
+    # OVN-K expects br-ex to have IP address assigned, add dummy IP to br-ex.
+    if ! ip addr show br-ex | grep '10.44.0.0/32'; then sudo ip addr add 10.44.0.0/32 dev br-ex; fi
```
ggiguash, line 47:
```suggestion
    if ! ip addr show br-ex 2>/dev/null | grep -q '10.44.0.0/32'; then
        sudo ip addr add 10.44.0.0/32 dev br-ex
    fi
```

scripts/multinode/configure-sec.sh
```diff
@@ -40,6 +40,9 @@ function configure_microshift() {
     sleep 5
     sudo systemctl start microshift-ovs-init.service
 
+    # OVN-K expects br-ex to have IP address assigned, add dummy IP to br-ex.
+    if ! ip addr show br-ex | grep '10.44.0.0/32'; then sudo ip addr add 10.44.0.0/32 dev br-ex; fi
```
ggiguash, line 44:
```suggestion
    if ! ip addr show br-ex 2>/dev/null | grep -q '10.44.0.0/32'; then
        sudo ip addr add 10.44.0.0/32 dev br-ex
    fi
```

PR #2099, opened 2023-07-20
USHIFT-1467: Upgradeability contributor's doc

pmtk:


docs/contributor/updateability.md
```diff
@@ -0,0 +1,349 @@
...
+- `/var/lib/microshift-backups/`
+    - `health.json`
+    - backup directories
+- `/var/lib/microshift/version`
+
+### `/var/lib/microshift-backups/`
+
+New, additional directory for keeping health and backup related artifacts.
+For ostree systems it's a non-configurable directory where backups are created.
+For non-ostree systems it's a default (configurable) directory for backups.
```
dhellmann, line 18: I didn't realize this was configurable. Why is it configurable?

docs/contributor/updateability.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### `health.json` [ostree-only]
```
dhellmann, line 20: Greenboot runs on systems without ostree. Automated rollback is only supported with ostree, but some of the other greenboot features, especially reporting system health holistically, are supported. So, this file might exist on a regular system, too.

docs/contributor/updateability.md
```diff
@@ -0,0 +1,349 @@
...
+
+Schema:
+```
+MAJOR.MINOR.PATCH
+```
+
+All of MAJOR, MINOR, and PATCH are unsigned integers.
+For example: `4.14.0`
+
+## Greenboot [ostree-only]
```
dhellmann, line 73: Same comment, the greenboot scripts can and might be installed on systems not based on ostree.

docs/contributor/updateability.md
```diff
@@ -0,0 +1,349 @@
...
+
+[0] `.nodename` might be created by the "effective config generation" phase
+which happens before the prerun. This can create `.nodename` and result
+in `/var/lib/microshift` being present but, beside that one file, empty.
+
+##### `/var/lib/microshift` (data) does not exist
+
+> Lack of `/var/lib/microshift` implies that `/var/lib/microshift/version` does not exist.
+
+If both `data` and `health.json` are missing, then we assume it's a first run of MicroShift.
```
dhellmann, line 151: nit: Using "`data`" here makes me think that's the name of an actual file, but I think you mean `/var/lib/microshift` == `data`? Maybe use _data_ instead?

dhellmann (commented):
I like the way you're focusing on the contributor audience here. A few points inline, but I'll give it a deeper review tomorrow.

docs/contributor/updateability.md
pmtk: It made sense since it's a cli command. If this is manual process, we don't really care where user will store it.

```
Usage:
  microshift admin data [command]

Available Commands:
  backup      Backup MicroShift data

Flags:
  -h, --help             help for data
      --name string      Backup name (default "4.15__20230721_072807")
      --storage string   Directory with backups (default "/var/lib/microshift-backups")
```

I just noticed that we missed command to restore :smile:

docs/contributor/updateability.md
pmtk: We don't create it on non-ostree ([ref](https://github.com/openshift/microshift/blob/72e8e0aab81342e1db9f773dc5f1d67f0407540c/packaging/greenboot/microshift_set_healthy.sh#L13-L16)):
```sh
if [ ! -f /run/ostree-booted ]; then
    echo "System is not booted with ostree"
    exit 0
fi
```

docs/contributor/updateability.md
pmtk: Well, yes. But it's usefulness is greatly reduced. Also, you can perform many lifecycle operations on microshift on RPM without rerunning the greenboot.

I'll add a comment

docs/contributor/updateability.md
pmtk: Do you think we should change the cli? From `microshift admin data {backup,restore}` to something more concise?

docs/contributor/updateability.md
```diff
@@ -405,9 +405,59 @@ file is created with version of the executable.
...
+
+#### Kube Storage Version Trigger
+
+Trigger performs a discovery which is a process of gathering information about
+existing Kubernetes Resources and their versions.
+If the Trigger controller discovers that the Resource should be migrated,
+a `StorageVersionMigration` CR is created which is then picked up by 
+the Migrator.
+
+Trigger controller performs the discovery and creates `StorageVersionMigration`
```
eggfoobar, line 438: A bit more context here with this.

> Trigger controller performs the discovery and creates a `StorageState` and a `StorageVersionMigration` respectively. The `StorageState` contains the current discovered state of the API resource and the previous, if there is a version difference for that specific API resource, then a `StorageVersionMigration` is created by the trigger. This check happens on start and every 10 minutes (discovery period; hardcoded in upstream code).

docs/contributor/updateability.md
```diff
@@ -405,9 +405,59 @@ file is created with version of the executable.
...
+kube-apiserver running.
+This turned out to not be a viable options because migrating
+CustomResourceDefinitions might require webhooks which run as a Pods, 
+therefore requiring kubelet, CNI, DNS, scheduler, and other components.
+
+Right now, storage migration is comprised of two additional controllers
+embedded into the MicroShift: Kube Storage Version Trigger and
+Kube Storage Version Migrator. They make use of a `StorageVersionMigration` CR.
+
+### `StorageVersionMigration` Custom Resource
```
eggfoobar, line 423: `StorageState` and `StorageVersionMigration` Custom Resource

docs/contributor/updateability.md
```diff
@@ -405,9 +405,59 @@ file is created with version of the executable.
...
+
+Initially, the plan was to perform the migration with only etcd and 
+kube-apiserver running.
+This turned out to not be a viable options because migrating
+CustomResourceDefinitions might require webhooks which run as a Pods, 
+therefore requiring kubelet, CNI, DNS, scheduler, and other components.
+
+Right now, storage migration is comprised of two additional controllers
+embedded into the MicroShift: Kube Storage Version Trigger and
+Kube Storage Version Migrator. They make use of a `StorageVersionMigration` CR.
```
eggfoobar, line 421: They make use of the `StorageState` and `StorageVersionMigration` CRs.

eggfoobar (commented):
Just added some info on the `StorageState`.

docs/contributor/updateability.md
pmtk: Thanks! I haven't noticed that one when skimming through the code

docs/contributor/updateability.md
```diff
@@ -0,0 +1,650 @@
...
+For ostree systems it's a non-configurable directory where backups are created.
+For non-ostree systems it's a default (configurable) directory for backups.
+
+#### `health.json` [ostree-only]
+
+Contains information about health of the system.
+It's updated by green and red script supplied by MicroShift.
+
+Because MicroShift's healthcheck takes some time to assess the state of the MicroShift,
+this file might be updated after couple minutes into system boot.
```
ggiguash, line 26: Let's add a command here to display the status of greenboot, so that the user knows when the file is updated.

docs/contributor/updateability.md
```diff
@@ -0,0 +1,650 @@
...
+Greenboot heavily integrates with grub. In fact, it is the grub that decrements
+the `boot_counter` and actually rolls back the system (by selecting the other
+boot target). See [Grub](#grub-ostree) section for more information.
+
+MicroShift delivers these scripts in `microshift-greenboot` RPM package.
+
+Health check script for MicroShift is resides in the repository
+as `packaging/greenboot/microshift-running-check.sh`.
+
+Pre-updateability (and continues to do so), `microshift-greenboot` started to
```
ggiguash, line 107:
```suggestion
Before the updateability feature was implemented, `microshift-greenboot` started to
```

docs/contributor/updateability.md
```diff
@@ -0,0 +1,650 @@
...
+### Updateability additions to `microshift-greenboot`
+
+Updateability feature added two, new files - one "green" and one "red" script:
+- `packaging/greenboot/microshift_set_healthy.sh` -> `/etc/greenboot/green.d/40_microshift_set_healthy.sh`
+- `packaging/greenboot/microshift_set_unhealthy.sh` -> `/etc/greenboot/red.d/40_microshift_set_unhealthy.sh`
+
+They work only on ostree-based systems (by checking if `/run/ostree-booted` exists).
+Reason for this is two-fold: how greenboot works on non-ostree systems, and
+deployment ID not being available, leaving only boot ID and health, which
+are not as useful without the deployment as backup name.
```
ggiguash, line 124: I would simplify this sentence.

docs/contributor/updateability.md
```diff
@@ -0,0 +1,650 @@
...
+Primary responsibility of "set healthy" and "set unhealthy" scripts
+is to create or update `health.json`.
+
+`microshift_set_unhealthy.sh` has an additional logic that, before
+updating `health.json`, checks if backup matching deployment and boot IDs
+from the `health.json` exists, and does not update the file if not.
+This ensures that, when upgrading (from a healthy system) MicroShift fails to
+create a backup, in effect causing the system to be unhealthy, that the new health
+information does not overwrite existing one ("healthy" instructs MicroShift to
+create a backup).
```
ggiguash, line 138: It may be too much detail. Let's simplify and merge the 2 sentences?

docs/contributor/updateability.md
```diff
@@ -0,0 +1,650 @@
...
+Without this mechanism in place, healthy data of previous deployment, would not
+get backed up, `health.json` get updated with `unhealthy` causing subsequent
+starts of MicroShift want to restore a backup (which might not exist because it
+wasn't created, or it's outdated because it contains data of
+"second to last boot of previous deployment").
```
ggiguash, line 143: Why do I need to know this information? Seems like an implementation detail to me.

docs/contributor/updateability.md
```diff
@@ -0,0 +1,650 @@
...
+
+TODO
+
+### Useful `rpm-ostree` commands
+
+TODO
+
+## Grub [ostree]
+
+While greenboot sets and clears env vars such as `boot_counter` and `boot_success`,
```
ggiguash, line 181:
```suggestion
While greenboot sets and clears Grub env vars such as `boot_counter` and `boot_success`,
```

dhellmann (commented):
This is full of good detail, thank you.  I have a few suggestions for small changes inline, and I see some sections are still being worked on. We should talk about whether to take this PR and iterate or finish the write-up and then merge it all at once.

docs/contributor/updateability.md
```diff
@@ -0,0 +1,650 @@
...
+
+#### Backup directories
+
+> Note: Only ostree-based systems perform automated backups and restores.
+
+`/var/lib/microshift-backups/` also stores backups of MicroShift data.
+Backups are created by copying data directory with `--reflink`
+option to leverage Copy-on-Write functionality.
+
+Naming schema of backup directories: `deployment-id_boot-id[_unhealthy]`.
```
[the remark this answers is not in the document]
pmtk: Do you find `deploymentID_bootID` less confusing?

docs/contributor/updateability.md
pmtk: I'm not sure, it is implementation detail, but this is a contributor's doc.
I think it might be useful to at least mention it because I can imagine someone looking up the file and seeing it has `healthy` but the system rebooted and rolled back, but the greenboot didn't finish yet, so the `health.json` has old info. I can even imagine a bug: "health.json contains healthy, even though system was unhealthy"

docs/contributor/updateability.md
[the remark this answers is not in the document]
dhellmann: Yes, that works for me.

docs/contributor/updateability.md
dhellmann: For the sake of getting all of the "how and why" in one place, let's keep the detail here, for now. If we decide expanding the comments in the script is better we can move some content, but for the short term it's more important to have it written down at all than to get it into exactly the right place.

docs/contributor/updateability.md
pmtk: Changing in https://github.com/openshift/microshift/pull/2108

dhellmann:
/lgtm

This has lots of good detail. I'm sure we will continue to expand on it and refine it over time, so let's take this version now and update it when we have questions or make changes to the implementation.

PR #2100, opened 2023-07-20
USHIFT-1468: build CI images with parents

dhellmann:
Rewrite the build_images.sh script to make it possible to have
data-driven mechanism for creating system images with the right
parentage.

Merge the download step into the build script, since we need images
available on the web server to let them be used as parents of other
images.

Update the way the images containing MicroShift built from source are
managed and rebuilt.

Update the way installer images are identified.

Update the way image reference aliases are created.

/assign @pmtk


dhellmann:
The script changes in build_images.sh are significant enough that it might be easier to review that as though it is a new file.

PR #2106, opened 2023-07-24
Rebase 4.14 + renaming Route Controller Manager assets

pmtk:
Addresses failures from https://github.com/openshift/microshift/pull/2105


dhellmann:
Now that #2083 has landed, I think this will need to be rebased to pick up the changes and pass the tests.

PR #2107, opened 2023-07-24
USHIFT-1059: fix: rbac generator for topolvm from lvms bundle starting with 4.13

jakobmoellerdev:
In the ClusterServiceVersion for LVM Operator / TopoLVM within Microshift there are encoded RBAC information for OLM deployments instead of RBAC yamls in the bundled operator.
Since microshift skips this installation through the bundle and uses a custom one based on extracted YAMLs, we have to extract the RBAC out of the CSV and recover the RBAC (Service Account, Roles Role Assignments) as if they were installed through the bundle instead.

This PR creates the necessary changes in rebase.sh which recovers the RBAC information starting with 4.13 of the LVMS operator bundle. It also corrects `assets.yaml` which references no longer existing files for the rebase and updates it accordingly. Also touches `storage.go` as the CSI driver now needs to use aggregated roles. (breaks upgrade path, but expected)


scripts/auto-rebase/rebase_job_entrypoint.sh
```diff
@@ -27,7 +27,7 @@ release_arm64="$(oc image info --registry-config=/tmp/registry.json $OPENSHIFT_R
 #  can increment independently from LVMS, and will usually be 1 y-stream ahead of LVMS in-between OCP releases.
 #  For example, ocp-release at 4.13 will more often than not correspond to 4.12 LVMS, until the official 4.13 release when
 #  both components will be 4.13.
-release_lvms="v4.12"
+release_lvms="v4.13.1-5"
```
dhellmann, line 30: Does this cause `last_rebase.sh` to be updated, too?

scripts/auto-rebase/rebase.sh
```diff
@@ -166,16 +172,146 @@ update_lvms_images(){
...
-    local namespace_scoped_manifests=(
-        "topolvm-controller_v1_serviceaccount.yaml"
-        "topolvm-csi-provisioner_rbac.authorization.k8s.io_v1_role.yaml"
-        "topolvm-csi-provisioner_rbac.authorization.k8s.io_v1_rolebinding.yaml"
-        "topolvm-node_v1_serviceaccount.yaml"
-    )
-    for m in "${namespace_scoped_manifests[@]}"; do
-        yq -i '.metadata.namespace = "openshift-storage"' "${dest}/${m}" || return 1
-    done
+}
```
dhellmann, line 175: It looks like `update_lvms_manifests` is no longer useful.

scripts/auto-rebase/rebase_job_entrypoint.sh
jakobmoellerdev: when running `rebase_to` the last_rebase.sh will be updated with the latest manifest, however I seem to get stuck at a seemingly unrelated component: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/2107/pull-ci-openshift-microshift-main-test-rebase/1683554164476481536#1:build-log.txt%3A950

After this succeeds on the rebase, it should be updated in `last_rebase.sh` as well through `update_last_rebase` which is called in `rebase to`

scripts/auto-rebase/rebase.sh
jakobmoellerdev: Agreed, wanted to verify that rebase procedure does not need additional changes of namespaces after rebase is successful (see above comment on why I dont have the rebase running yet).

copejon:
Synced w/ Jakob on the implementation and code changes earlier today.  All looks good.  The changes have passed the snapshotting tests on the metal-e2e job.

/lgtm

scripts/auto-rebase/rebase_job_entrypoint.sh
dhellmann: https://github.com/openshift/microshift/pull/2106 may help with that issue. You could try rebasing to pull in those changes.

PR #2108, opened 2023-07-24
USHIFT-1470: Add subcommand to restore a backup

pmtk:
* Adds `restore` subcommand that takes fullpath to a backup: `sudo microshift admin restore /var/lib/microshift-backups/some-manual-backup/`
* Flattens the subcommand structure from `admin data backup,restore` to `admin backup,restore`.

### DIFF
--- a/pkg/cmd/admin.go
+++ b/pkg/cmd/admin.go
@@ -2,6 +2,9 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/openshift/microshift/pkg/admin/data"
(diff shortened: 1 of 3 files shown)


pkg/cmd/admin.go
```diff
@@ -54,22 +64,40 @@ func newAdminDataCommand() *cobra.Command {
 
 			return nil
 		},
+		RunE: backup,
 	}
 	v := version.Get()
-	data.PersistentFlags().String("storage", config.BackupsDir, "Directory with backups")
-	data.PersistentFlags().String("name",
+	cmd.PersistentFlags().String("storage", config.BackupsDir, "Directory with backups")
+	cmd.PersistentFlags().String("name",
```
dhellmann, line 71: Is it useful to have a default for the name of the thing being restored?

pmtk: These flags are for backup only, restore expects full path - not really sure if that's a good or bad UX... I'm not a fan.
Nice to have could be to provide a subcommand or opt (`--list`) that would output a list of backups in the `--storage`, or even have a completion for backup name.

To provide a default backup name to restore, we'd have to scan the `microshift-backups/` and pick a right one (last one?). I don't see a good default backup name for restoring when there's more than one backup

dhellmann: OK, I thought part of the reason they were changing was because they were being reused but I see now that's just a variable name change. Using the full path as input argument makes the most sense. I agree, having a meaningful default would be hard.

dhellmann (commented):
It would be nice to start including automated tests with new features, but this one is straightforward enough that I think we can do that in a separate PR.

/lgtm

dhellmann:
The restore test is failing to run

```
sudo microshift admin data backup --name "rhel-f524ee11a57ab63092d2a955d67ff43e09f6df4ae7baedcca41e10d381818e33.0_manual"
```

because this PR changes the command line syntax to remove the `admin`.

PR #2109, opened 2023-07-24
USHIFT-1471: Use data/version for default backup name

pmtk:


pkg/cmd/admin.go
```diff
@@ -55,11 +56,25 @@ func newAdminDataCommand() *cobra.Command {
 			return nil
 		},
 	}
-	v := version.Get()
+
+	defaultBackupName := ""
+	if dataVer, err := prerun.GetVersionOfData(); err == nil {
+		// If data/version exists and command is executed with enough
+		// privilege to read the file, then use it to construct default name
```
dhellmann, line 63: I like the idea of including version details in the default name, but I'm worried that silently switching between sources of version information when the values might be different could be confusing.

How about just leaving the version out if we can't read the file? That way the version in the backup filename is always the version of the data.

pmtk: I don't think that executable's version would be used for backup because it's only used when there is no data, so nothing to back up.

Also, it's important that to properly prepare default value, this needs to be run with `sudo` or as root, otherwise it cannot access `microshift/version` file... So in that case, it might be very confusing because the help doc might be different without sudo.
Does using `0.0.0` instead of exec's ver seem any better?

dhellmann: I think it's safe for us to assume the user must have the right privileges to run "admin" commands. So if the version file exists but cannot be read, let's just report that as an error and not back up.

Then if the file does not exist, we assume 4.13.0, as elsewhere.

Then we don't need lines 67-74, I think?

pmtk: > I think it's safe for us to assume the user must have the right privileges to run "admin" commands. So if the version file exists but cannot be read, let's just report that as an error and not back up.

Yes, that is what happens.
I just want to emphasize one thing: this default value is also relevant for the subcommand help.
I don't think throwing an error when running `microshift backup --help` without sudo is good UX.

So, my question was: what in that case? 
- `version` exists and can be opened -> real data
- otherwise, `MAJOR.MINOR.PATCH_$date_$time`? (`MAJOR.MINOR.PATCH` as literal string)
  - doesn't look as clear as `0.0.0` but is more descriptive (+ maybe extra hint that this relates to `data/version`?)

Ugh, I wish `version` would not be `0600`

pkg/cmd/admin.go
```diff
@@ -55,11 +56,25 @@ func newAdminDataCommand() *cobra.Command {
 			return nil
 		},
 	}
-	v := version.Get()
+
+	defaultBackupName := ""
+	if dataVer, err := prerun.GetVersionOfData(); err == nil {
```
pmtk, line 61: Huh, interesting. `GetVersionOfData()` returns unexported struct, but it works and linter is happy...

pkg/cmd/admin.go
```diff
@@ -55,11 +55,19 @@ func newAdminDataCommand() *cobra.Command {
 			return nil
 		},
 	}
-	v := version.Get()
+
+	versionString := prerun.GetVersionStringOfData()
```
dhellmann, line 59: If `GetVersionStringOfData` returns the error instead of treating an unknown error as version `""`, then this call could _also_ return the error and not take any action. If the version file exists but is empty or we can't read it, that data is broken and we don't want to mess with it.

pkg/admin/prerun/version.go
```diff
@@ -108,6 +108,23 @@ func getVersionOfData() (versionMetadata, error) {
...
+		if errors.Is(err, errDataVersionDoesNotExist) {
+			dataExists, err := util.PathExistsAndIsNotEmpty(config.DataDir, ".nodename")
+			if err == nil && dataExists {
+				// version does not exists, but data exists
+				return "4.13"
+			}
+		}
+		klog.ErrorS(err, "Failed to read version - ignoring error")
+		// couldn't access the file for whatever reason
+		return ""
```
ggiguash, line 123: Should we return "unknown" instead of the empty string? This would not cause weird `__<date>` directory name creation.

pmtk: Changed

PR #2116, opened 2023-07-25
USHIFT-1472: Deploy and boot ID in version file to avoid unnecessary restores

pmtk:


test/image-blueprints/group2/rhel92-source-fake-next-next-minor.toml
```diff
@@ -0,0 +1,39 @@
+name = "rhel-9.2-microshift-source-fake-next-next-minor"
```
dhellmann, line 1: This name is a bit subtle when scanning. Can we use YPLUS2 or some other more distinct thing other than "next next"?

test/image-blueprints/group2/rhel92-source-fake-next-minor.toml
```diff
@@ -1,5 +1,5 @@
 name = "rhel-9.2-microshift-source-fake-next-minor"
-description = "A RHEL 9.2 image with the RPMs built from source from the current PR but with the _version_ set to the next y-stream."
+description = "A RHEL 9.2 image with the RPMs built from source from the current PR but with the _version_ set to the current+2 y-stream."
```
dhellmann, line 2: I think you updated the description in the wrong file here?

pkg/admin/prerun/version.go
```diff
@@ -23,6 +24,23 @@ var (
 // CheckAndUpdateDataVersion checks version compatibility between data and executable,
 // and updates data version
 func CheckAndUpdateDataVersion() error {
+	// Q: Should we use separate version schema for ostree and non-ostree?
```
dhellmann, line 27: What do you have in mind for the different version schema?

dhellmann (commented):
One nit about naming, but that's probably not worth holding this up. Remove the hold if you're ready to proceed.

/lgtm
/hold

test/image-blueprints/group2/rhel92-source-fake-next-next-minor.toml
pmtk: Yes, thanks for a suggestion, I had no ideas at the moment of writing :stuck_out_tongue_winking_eye:

test/image-blueprints/group2/rhel92-source-fake-next-minor.toml
pmtk: Fixed

dhellmann:
The linter looks unhappy. Otherwise, this looks good.

PR #2122, opened 2023-07-26
rebase 4.14.0 + fix for router assets' source path

pmtk:
Paths changed in https://github.com/openshift/cluster-ingress-operator/pull/905


dhellmann:
I think the issues with the metal tests here will be fixed by #2123

PR #2123, opened 2023-07-26
USHIFT-1445: fix: prune test

eggfoobar:
This is to avoid errors in timing between file creation and when the system comes up, since the prune logic skips touching the directory if backup exists, this tests could fail sometimes if ran out of order.

Moving to one test to avoid issue.

/assign @pmtk

### DIFF
--- a/test/suites-ostree/backup-prune.robot
+++ b/test/suites-ostree/backup-prune.robot
@@ -20,22 +20,19 @@ ${UNKOWN_DIR_REG}       unknown_fake*
 
 
 *** Test Cases ***
-Rebooting Healthy System Should Result In Pruning Old Backups
-    [Documentation]    Check that MicroShift backup logic prunes old backups
+Rebooting Healthy System Should Result In Pruning Old Backups And Keeping Unknown Directories
+    [Documentation]    Check that MicroShift backup logic prunes old backups and keeps unknown directories
 
     Create Fake Backup Directories
     Validate Fake Backups Exist
-    Reboot MicroShift Host
-    Wait For MicroShift Service
-    Validate Fake Backups Have Been Deleted
 
-Rebooting Healthy System Should Not Delete Unknown Directories
-    [Documentation]    Check that MicroShift backup logic does not delete unknown directories
-
-    Create Fake Unknown Backup Directories
+    Create Unknown Fake Backup Directories
     Validate Unknown Fake Backups Exist
+
     Reboot MicroShift Host
-    Wait For MicroShift Service
+    Wait For Healthy System
+
+    Validate Fake Backups Have Been Deleted
     Validate Unknown Fake Backups Exist
 
 
@@ -52,7 +49,7 @@ Validate Fake Backups Have Been Deleted
     [Documentation]    Listing the fake backups should return an error code of 2 if no longer exists
     List Backups With Expected RC    2    ${FAKE_DIR_REG}
 
-Create Fake Unknown Backup Directories
+Create Unknown Fake Backup Directories
     [Documentation]    Create unknown fake backup folders
     Create Fake Backups    2    True
 
--- a/test/suites-ostree/backup-prune.robot
+++ b/test/suites-ostree/backup-prune.robot
@@ -23,15 +23,6 @@ ${UNKOWN_DIR_REG}       unknown_fake*
 Rebooting Healthy System Should Result In Pruning Old Backups And Keeping Unknown Directories
     [Documentation]    Check that MicroShift backup logic prunes old backups and keeps unknown directories
 
-    Create Fake Backup Directories
-    Validate Fake Backups Exist
-
-    Create Unknown Fake Backup Directories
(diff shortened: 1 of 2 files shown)


test/suites-ostree/backup-prune.robot
```diff
@@ -20,22 +20,19 @@ ${UNKOWN_DIR_REG}       unknown_fake*
...
-Rebooting Healthy System Should Not Delete Unknown Directories
-    [Documentation]    Check that MicroShift backup logic does not delete unknown directories
 
-    Create Fake Unknown Backup Directories
+    Create Unknown Fake Backup Directories
     Validate Unknown Fake Backups Exist
+
     Reboot MicroShift Host
-    Wait For MicroShift Service
+    Wait For Healthy System
```
dhellmann, line 33: Another way to solve this would be to move these steps into the suite setup. Then it would happen 1 time before each test case, and the test case could just check the expected results.

eggfoobar: That makes sense, I like it. I refactored but kept it as one test still, that seems more valid to me since it's a single prune action and we're not validating different reboot cycles to prune. What do you think?

test/suites-ostree/backup-prune.robot
```diff
@@ -20,22 +20,10 @@ ${UNKOWN_DIR_REG}       unknown_fake*
...
     Validate Fake Backups Have Been Deleted
-
-Rebooting Healthy System Should Not Delete Unknown Directories
-    [Documentation]    Check that MicroShift backup logic does not delete unknown directories
-
-    Create Fake Unknown Backup Directories
-    Validate Unknown Fake Backups Exist
-    Reboot MicroShift Host
-    Wait For MicroShift Service
     Validate Unknown Fake Backups Exist
```
dhellmann, line 27: I may be having some trouble with the naming of these keywords, but why would we expect any unknown backups to exist at this point?

eggfoobar: Fair, maybe we could reword that, this is validating our pruning logic when it does not delete directories it does not know (follow our backup directory naming convention), so maybe a more accurate wording would be "Unknown Directories". So if for some reason a user places a directory in there or we decide to place something else in there in the future, it does not get deleted by our pruning logic.

eggfoobar:
@dhellmann and @pmtk I updated the test but I also ran across why things were failing, the issue was that we prune `all` directories with an `underscore`, is that desired or should we post fix something to our backups?

dhellmann:
> @dhellmann and @pmtk I updated the test but I also ran across why things were failing, the issue was that we prune `all` directories with an `underscore`, is that desired or should we post fix something to our backups?

Is there any reason for us to tolerate anything in that directory that we don't create?

dhellmann:
/lgtm

Let's get CI working and come back to the question about how much we should retain or how aggressively we should delete things.

eggfoobar:
> > @dhellmann and @pmtk I updated the test but I also ran across why things were failing, the issue was that we prune `all` directories with an `underscore`, is that desired or should we post fix something to our backups?
> 
> Is there any reason for us to tolerate anything in that directory that we don't create?

Currently, we place a failed log there that we push up to greenboot, which does have the name `prerun_failed.log`, that won't be deleted before it's presented to greenboot, since it's a result of a backup failure and subsequent failed restarts mean that that file won't be removed, but just something to keep in mind for other future instances where we store files there.

dhellmann:
> > > @dhellmann and @pmtk I updated the test but I also ran across why things were failing, the issue was that we prune `all` directories with an `underscore`, is that desired or should we post fix something to our backups?
> > 
> > 
> > Is there any reason for us to tolerate anything in that directory that we don't create?
> 
> Currently, we place a failed log there that we push up to greenboot, which does have the name `prerun_failed.log`, that won't be deleted before it's presented to greenboot, since it's a result of a backup failure and subsequent failed restarts mean that that file won't be removed, but just something to keep in mind for other future instances where we store files there.

We put that in the backup directory? Could it live in /usr/lib/microshift instead?

pmtk:
> @dhellmann and @pmtk I updated the test but I also ran across why things were failing, the issue was that we prune `all` directories with an `underscore`, is that desired or should we post fix something to our backups?

Seems like I've skipped few thought steps lgtming previous PR :laughing: 
This is actually intentional, how do you differentiate between known and unknown? We take all backups matching `%s_%s` and remove them if deployment id is no longer on the system.

> Currently, we place a failed log there that we push up to greenboot, which does have the name prerun_failed.log, that won't be deleted before it's presented to greenboot, since it's a result of a backup failure and subsequent failed restarts mean that that file won't be removed, but just something to keep in mind for other future instances where we store files there.

Does `prerun_failed.log` gets deleted when pruning the backups? dataManager's GetBackupList() should only return directories.
I wouldn't rely on "backup pruning" to remove that file (even sounds like a side effect, if that works that is, but it shouldn't)

PR #2129, opened 2023-07-27
USHIFT-1487: refactor: rename prerun struct to datamanager

eggfoobar:
Initial PR to rename `prerun` struct to `datamanger`.

Since `PreRun` does much more now, we decided to do a refactor and change the struct name to `DataManager` and it's receiver functions.

/assign @pmtk

### DIFF
--- a/pkg/admin/prerun/prerun.go
+++ b/pkg/admin/prerun/prerun.go
@@ -4,23 +4,23 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/openshift/microshift/pkg/admin/data"
+	datadir "github.com/openshift/microshift/pkg/admin/data"
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util"
 	"k8s.io/klog/v2"
 )
 
 type DataManager struct {
-	dataManager data.Manager
+	dataManager datadir.Manager
 }
 
-func NewDataManager(dataManager data.Manager) *DataManager {
+func NewDataManager(dataManager datadir.Manager) *DataManager {
 	return &DataManager{
-		dataManager: dataManager,
+		dataManager,
 	}
 }
 
-func (pr *DataManager) Run() error {
+func (dm *DataManager) Run() error {
 	klog.InfoS("Starting pre-run")
 	defer klog.InfoS("Pre-run complete")
 
@@ -77,7 +77,7 @@ func (pr *DataManager) Run() error {
 		}
 
 		// 2
-		return pr.missingDataExistingHealth()
+		return dm.missingDataExistingHealth()
 	}
 
 	if !versionExists {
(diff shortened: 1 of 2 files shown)


eggfoobar:
/hold

This is part of the renaming, but discussions are still happening in the Jira ticket. This PR was created to validate nothing broke CI with this rename.

pkg/cmd/run.go
```diff
@@ -106,7 +106,7 @@ func performPrerun() error {
 		return err
 	}
 
-	return prerun.New(dataManager).Perform()
+	return management.New(dataManager).Perform()
```
dhellmann, line 109: We might want a new method name here. "management perform" isn't very specific about what's going to happen.

eggfoobar: Yeah, agreed. Thinking more along the lines of `management.NewPreRun().Validate()`

pmtk: I was more thinking about keeping `prerun` package, but rename the `PreRun` struct to e.g. `BackupManagement` since this is what it does - backs up, restores, or deletes the data, so maybe even `DataManagement`?

eggfoobar: Sounds good, based off of our slack convo, went ahead and changed it to `BackupManager`, that does make more sense since it's doing all the backup management tasks.

eggfoobar:
/unhold

Removing hold since now we have some more direction on where we want to take this. Consensus is to keep it simple and just do a small rename of `PreRun` to `BackupManager` where it's more descriptive of its functionality.

eggfoobar:
Had a quick chat with @pmtk, DataManager seems a bit more appropriate here given that this functionality does also alter the DataDir.

pkg/admin/prerun/prerun.go
```diff
@@ -4,22 +4,23 @@ import (
...
-type PreRun struct {
-	dataManager data.Manager
+type DataManager struct {
+	dataManager datadir.Manager
 }
 
-func New(dataManager data.Manager) *PreRun {
-	return &PreRun{
-		dataManager: dataManager,
+func NewDataManager(dataManager datadir.Manager) *DataManager {
```
eggfoobar, line 17: Not 100% sure on this alias, thoughts?

pmtk:
I'm not totally convinced with this, but since this is my idea after all and I'm out of naming ideas... it's better than a "PreRun" which says almost nothing.
I'm applying lgtm+hold, maybe @dhellmann has some thoughts :pleading_face: 

/lgtm
/hold

dhellmann:
It looks like the final decision was to rename the type, but not the package?

pmtk:
> It looks like the final decision was to rename the type, but not the package?

Looks like it. Package name `prerun` is still accurate and it contains: backup and version metadata management.
For me it's the `type PreRun struct` that is no longer accurate description since PreRun is no longer a single block of actions, and that struct deals with data/backup management

dhellmann:
> > It looks like the final decision was to rename the type, but not the package?
> 
> Looks like it. Package name `prerun` is still accurate and it contains: backup and version metadata management. For me it's the `type PreRun struct` that is no longer accurate description since PreRun is no longer a single block of actions, and that struct deals with data/backup management

OK, I like the direction but the PR description is a bit out of date, so I wanted to verify.

eggfoobar:
/unhold

Updated description to reflect changes.

PR #2133, opened 2023-07-27
USHIFT-1513: update rebase to work with kube-storage-version-migrator

dhellmann:
The rebase script had assumed that all image and git repo names would 
match, but they do not when we bring something from upstream into a fork.
Add support for a "via" argument to the "release" instruction in go.mod to
specify the git repository name. Then add the kube-storage-version-migrator
to the components included when we rebase.

/assign @pmtk @eggfoobar


eggfoobar:
LGTM

I'll leave it to @pmtk to add the label.

Thanks for doing this @dhellmann !

PR #2140, opened 2023-07-29
USHIFT-1510: Enhance multinode configuration scripts to be more resilient

ggiguash:
The secondary node is now tagged with "worker" label.
```
NAME             STATUS   ROLES                         AGE   VERSION
microshift-pri   Ready    control-plane,master,worker   79s   v1.27.1
microshift-sec   Ready    worker                        31s   v1.27.1
```

The following additional changes were also implemented:
- Keep images when cleaning MicroShift data to speed up pod reboot
- Wait until microshift-ovs-init.service is started to avoid race condition with IP setting
- Verify kubelet process is killed on the secondary node
- Unmount /var/lib/kubelet on the secondary node to allow directory deletion
- Wait until all MicroShift nodes are ready
- Wait until the pods are ready using the multinode pod counts

Closes [USHIFT-1510](https://issues.redhat.com//browse/USHIFT-1510)

### DIFF
--- a/scripts/multinode/configure-sec.sh
+++ b/scripts/multinode/configure-sec.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
 
+OC_CMD="sudo -i oc --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig"
+
 PRI_NODE_HOST=
 PRI_NODE_ADDR=
 SEC_NODE_HOST=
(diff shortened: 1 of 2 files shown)


dhellmann:
This looks OK to me, but I'd like someone more familiar with the 2 node setup to take a look.

/assign @jerpeter1

scripts/multinode/configure-sec.sh
```diff
@@ -50,8 +52,14 @@ function configure_microshift() {
         sudo systemctl stop --now kubelet
     fi
     sudo systemctl reset-failed kubelet || true
-    sudo pkill -9 --exact kubelet       || true
-    sudo rm -rf /var/lib/kubelet &> /dev/null || true
+    until ! pidof kubelet &>/dev/null ; do
+        sudo pkill -9 --exact kubelet || true
+        sleep 1
+    done
```
pacevedom, line 58: I am wondering if there is a need for a loop, given that sigkill is not catchable/ignorable.

scripts/multinode/configure-sec.sh
```diff
@@ -109,6 +117,49 @@ EOF
         --cluster-domain=cluster.local
 }
 
+function configure_node() {
+    local service_ready=false
+
+    echo "Waiting for MicroShift nodes to be ready"
+    for _ in $(seq 300) ; do
```
pacevedom, line 124: Is this equivalent to `oc wait --for=condition=Ready nodes --all --timeout=300s`?

pacevedom (approved):
Two nits. Tested to be working in a metal host running CNCF conformance.

scripts/multinode/configure-sec.sh
ggiguash: The idea was to make sure that the process is dead. Let me reqwrite the code not to send KILL more than once.

scripts/multinode/configure-sec.sh
ggiguash: I still have to add the loop because the command fails initially because `microshift-sec` takes time to be created.

ggiguash: For the same reason, I have to check each node separately, or I'm getting errors. 
I enhanced the checks though. Is this better?

PR #2143, opened 2023-07-29
USHIFT-1519: hard-code source of LVMS manifests

dhellmann:
The production pipeline is passing an old value. We want that input
parameter to be controlled explicitly by the development, so ignore
the parameter and hard-code a value in the rebase script.

/assign @copejon @ggiguash @pmtk

### DIFF
--- a/scripts/auto-rebase/rebase.sh
+++ b/scripts/auto-rebase/rebase.sh
@@ -40,6 +40,8 @@ cluster-csi-snapshot-controller-operator"
 declare -a ARCHS=("amd64" "arm64")
 declare -A GOARCH_TO_UNAME_MAP=( ["amd64"]="x86_64" ["arm64"]="aarch64" )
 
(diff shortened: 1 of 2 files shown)


scripts/auto-rebase/rebase.sh
```diff
@@ -40,6 +40,8 @@ cluster-csi-snapshot-controller-operator"
 declare -a ARCHS=("amd64" "arm64")
 declare -A GOARCH_TO_UNAME_MAP=( ["amd64"]="x86_64" ["arm64"]="aarch64" )
 
+LVMS_RELEASE_MULTI_ARCH="registry.access.redhat.com/lvms4/lvms-operator-bundle:v4.13.1-5"
```
ggiguash, line 43: Should we allow environment override here?

dhellmann: The point of this change is to *not* allow anything outside of the script to influence the source of LVMS. I'll provide a better fix as part of USHIFT-1521, but we need something quick so we can get an EC build this sprint.

ggiguash:
/lgtm
/hold

Please unhold the PR when ready to be merged

PR #2144, opened 2023-07-29
USHIFT-1521: separate rebase of OCP and LVMS

dhellmann:
Remove the LVMS rebase steps for the overall OCP rebase process and
provide separate, parallel, commands for updating LVMS.

/assign @copejon @pmtk @ggiguash

### DIFF
--- a/scripts/auto-rebase/rebase.sh
+++ b/scripts/auto-rebase/rebase.sh
@@ -40,8 +40,6 @@ cluster-csi-snapshot-controller-operator"
 declare -a ARCHS=("amd64" "arm64")
 declare -A GOARCH_TO_UNAME_MAP=( ["amd64"]="x86_64" ["arm64"]="aarch64" )
 
-LVMS_RELEASE_MULTI_ARCH="registry.access.redhat.com/lvms4/lvms-operator-bundle:v4.13.1-5"
-
 title() {
     echo -e "\E[34m$1\E[00m";
 }
@@ -158,7 +156,7 @@ write_lvms_images_for_arch(){
 }
 
 update_lvms_images(){
-    title "Updating LVMS manifests"
+    title "Updating LVMS images"
 
     local workdir="$STAGING_DIR/lvms"
     [ -d "$workdir" ] || {
@@ -1068,15 +1066,13 @@ update_changelog() {
 }
 
 
-# Runs each rebase step in sequence, commiting the step's output to git
+# Runs each OCP rebase step in sequence, commiting the step's output to git
 rebase_to() {
     local release_image_amd64=$1
     local release_image_arm64=$2
-    local lvms_operator_bundle_manifest="$3"
 
-    title "# Rebasing to ${release_image_amd64} and ${release_image_arm64} with LVMS bundle ${lvms_operator_bundle_manifest}"
+    title "# Rebasing to ${release_image_amd64} and ${release_image_arm64}"
     download_release "${release_image_amd64}" "${release_image_arm64}"
-    download_lvms_operator_bundle_manifest "$lvms_operator_bundle_manifest"
 
     ver_stream="$(cat ${STAGING_DIR}/release_amd64.json | jq -r '.config.config.Labels["io.openshift.release"]')"
     amd64_date="$(cat ${STAGING_DIR}/release_amd64.json | jq -r .config.created | cut -f1 -dT)"
@@ -1110,7 +1106,6 @@ rebase_to() {
     done
 
     update_images
-    update_lvms_images
     if [[ -n "$(git status -s pkg/release packaging/crio.conf.d)" ]]; then
         title "## Committing changes to pkg/release"
         git add pkg/release
@@ -1122,7 +1117,6 @@ rebase_to() {
 
     copy_manifests
     update_openshift_manifests
-    update_lvms_manifests
     if [[ -n "$(git status -s assets)" ]]; then
         title "## Committing changes to assets and pkg/assets"
         git add assets pkg/assets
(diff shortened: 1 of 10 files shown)


dhellmann:
The metal job failed to boot some of the VMs. None of the changes in this PR should have caused that, and we can recheck after reviews.

The pre-submit check for the rebase script passed. [The logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_microshift/2144/pull-ci-openshift-microshift-main-test-rebase/1685736363699212288/artifacts/test-rebase/openshift-microshift-rebase/build-log.txt) show both rebase phases working properly.

jakobmoellerdev (commented):
While I understand that these steps will separate the jobs out completely, I feel like this is quite a big clone of basically all rebase.sh steps just to get lvms out of the normal `rebase_to`. 

The existing workaround with a variable inside `rebase.sh` from you did the job quite well and we would only need to remove it from pipelines to get the job running fine without the FIXMEs in https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase.sh#L1162C7-L1171

That way (from what I understand) the process is already purely manual then. Automation could work by simply running the rebase normally and when we do an upstream change in LVMS with a release, propagate by issuing a change in microshift that changes this variable, runs the rebase with that variable (maybe as ENV) changed and then submits a new PR with the existing rebase script.

TLDR: Mainly hesistent that we keep rebase.sh / rebase.md but basically clone every single command/function to separate LVMS out from the rest. If you feel like there is more reasoning to split the two parts, then maybe it would be best if we completely remove it from `rebase.sh` and put it in a `rebase-lvms.sh` so that the domains are truly separate.

scripts/auto-rebase/rebase.sh
```diff
@@ -159,6 +156,8 @@ write_lvms_images_for_arch(){
 }
 
 update_lvms_images(){
+    title "Updating LVMS manifests"
```
jakobmoellerdev, line 159:
```suggestion
    title "Updating LVMS images"
```

scripts/auto-rebase/rebase.sh
```diff
@@ -1141,6 +1145,64 @@ rebase_to() {
...
+    git branch -D "${rebase_branch}" || true
+    git checkout -b "${rebase_branch}"
+
+    update_last_lvms_rebase "${lvms_operator_bundle_manifest}"
+
+    update_lvms_images
+    if [[ -n "$(git status -s pkg/release)" ]]; then
+        title "## Committing changes to pkg/release"
+        git add pkg/release
+        git add packaging/crio.conf.d
```
jakobmoellerdev, line 1188: is this necessary to be updated in lvms rebase (I thought it would only touch assets)? Surprised to see CRIO/pkg/release configuration here

dhellmann:
> While I understand that these steps will separate the jobs out completely, I feel like this is quite a big clone of basically all rebase.sh steps just to get lvms out of the normal `rebase_to`.

Yes, that's exactly what I was going for.

The LVMS and OCP rebase processes are entirely unrelated, except that they update some files in the same directories. I'm trying to untangle them here so they are easier to maintain. The work you did earlier was very helpful in that regard.

> The existing workaround with a variable inside `rebase.sh` from you did the job quite well and we would only need to remove it from pipelines to get the job running fine without the FIXMEs in https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase.sh#L1162C7-L1171
> 
> That way (from what I understand) the process is already purely manual then. Automation could work by simply running the rebase normally and when we do an upstream change in LVMS with a release, propagate by issuing a change in microshift that changes this variable, runs the rebase with that variable (maybe as ENV) changed and then submits a new PR with the existing rebase script.

I did consider that approach. Aside from trying to ease maintenance by separating the code paths, I thought this also made it clearer how to do the LVMS rebase: Run the command, passing the new value as input. 

There's no need for CI or other jobs to keep running the same rebase repeatedly, since there should not be any changes.

> 
> TLDR: Mainly hesistent that we keep rebase.sh / rebase.md but basically clone every single command/function to separate LVMS out from the rest. If you feel like there is more reasoning to split the two parts, then maybe it would be best if we completely remove it from `rebase.sh` and put it in a `rebase-lvms.sh` so that the domains are truly separate.

Eventually I would love to do that. I worried about the CI changes needed to do it all at once in this PR. Maybe we can continue the evolution in that direction next release?

scripts/auto-rebase/rebase.sh
dhellmann: Good catch, I'll fix that in the next revision.

scripts/auto-rebase/rebase.sh
dhellmann: No, that was a copy-and-paste error from duplicating the commit step. I'll clean that up in the next revision.

jakobmoellerdev (approved):
Agree with your sentiment, maybe just add one TODO/FIXME so we can go over it and extract it to a separate script next release. Other than that LGTM

dhellmann:
> Agree with your sentiment, maybe just add one TODO/FIXME so we can go over it and extract it to a separate script next release. Other than that LGTM

I opened a jira ticket to remember to do that and included you as a contributor.

scripts/auto-rebase/rebase.sh
```diff
@@ -1161,25 +1223,34 @@ case "$command" in
     to)
         # FIXME: Allow more than 3 arguments until pipelines stop passing LVMS value.
```
pmtk, line 1224: Is this for ART rebase procedure?

dhellmann: Yes, the ART pipeline is still passing the extra value.

scripts/auto-rebase/rebase.sh
dhellmann: Done

scripts/auto-rebase/rebase_job_entrypoint.sh
```diff
@@ -22,23 +22,30 @@ oc registry login --to=/tmp/registry.json
...
+#  correspond to 4.12 LVMS, until the official 4.13 release when both
+#  components will be 4.13.
+release_lvms="v4.13.1-5"
+
+# Since LVMS is not part of the release payload, it is not kept in
+# CI. Use the latest z-stream that coincides with the release
+# payload's X.Y version
+pullspec_release_lvms="registry.access.redhat.com/lvms4/lvms-operator-bundle:${release_lvms}"
+
+./scripts/auto-rebase/rebase.sh lvms-to "${pullspec_release_lvms}"
```
pmtk, line 51: Because this is after `rebase.py`, it won't be included in the PR.

Do we run this here, to make sure it continues to work instead of surprising us last minute when we actually need it?

pmtk:
/lgtm
/hold

Just one question

scripts/auto-rebase/rebase_job_entrypoint.sh
dhellmann: Yes, that's right. In practice, I expect a person to do the lvms rebase until we get that better automated, so they would submit their own PR.

PR #2145, opened 2023-07-30
NO-ISSUE: Review scenario part of the test harness

ggiguash:
The testing environment was comprised of a development machine separate from the hypervisor host.
The following changes were implemented:
* Reorganize the README.md into 4 main sections and fix title cases
* Fix all the commands and script references in the docs to be relative to the root microshift repo
* Fix local source changes script to run missing repo creation and web server startup
* The image build procedure now fails after downloading logs if at least one job failed
* Scenarios can be run with parameters from an arbirary directory
* Fixed `scripts/image-builder/cleanup.sh` script to properly clean multiple osbuild workers

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -128,14 +128,14 @@ $ ./_output/robotenv/bin/robot -h
 
 ## Test Scenarios
 
-The test scenario tools in the `bin` directory are useful for running
+The test scenario tools in the `./test/bin` directory are useful for running
 more complex test cases that require VMs and different images.
 
 ### Package Sources
 
-In order to build different images, Composer needs to be configured
+In order to build different images, Image Builder needs to be configured
 with all of the right package sources to pull the required RPMs. The
-sources used by all scenarios are in the `package-sources` directory.
+sources used by all scenarios are in the `./test/package-sources` directory.
 
 Package source definition files are templates using `envsubst`,
 which means the files can have shell variables embedded.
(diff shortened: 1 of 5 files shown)


test/README.md
```diff
@@ -406,31 +456,37 @@ are not saved to the default variables file by the framework.
...
+test infrastructure for a scenario.
 
 ```
-$ ./bin/scenario.sh cleanup  ./scenarios/rhel-9.2-microshift-source-standard-suite/
+$ ./test/bin/scenario.sh cleanup \
+      ./test/scenarios/rhel-9.2-microshift-source-standard-suite.sh
 ```
 
-## CI Integration Scripts
+### CI Integration Scripts
```
dhellmann, line 480: I think this makes this section part of "Troubleshooting" but it wasn't intended to be. Maybe it should have been?

test/bin/build_images.sh
```diff
@@ -214,6 +214,19 @@ do_group() {
         fi
     done
 
+    echo "Checking build statuses"
```
dhellmann, line 217: Instead of looping again, the variable could be set up around line 193-196 where the status condition is checked already and then this could just be a test and return.

dhellmann (commented):
These changes all look like good improvements. I had a couple of comments inline, but they're small and nothing to hold this up over. I'll leave the PR open until tomorrow in case other reviewers have comments.

PR #2150, opened 2023-07-31
USHIFT-1389: Enhance greenboot check script to print cluster debugging information on failures

ggiguash:
Closes [USHIFT-1389](https://issues.redhat.com//browse/USHIFT-1389)

### DIFF
--- a/packaging/greenboot/microshift-running-check.sh
+++ b/packaging/greenboot/microshift-running-check.sh
@@ -11,7 +11,7 @@ PODS_CT_LIST=(2                        1                    1                 2 
 source /usr/share/microshift/functions/greenboot.sh
 
 # Set the term handler to convert exit code to 1
-trap 'return_failure' TERM
+trap 'return_failure' TERM SIGINT
 
 # Set the exit handler to log the exit status
 trap 'script_exit' EXIT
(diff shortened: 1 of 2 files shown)


packaging/greenboot/microshift-running-check.sh
```diff
@@ -28,7 +28,43 @@ function return_failure() {
 # args: None
 # return: None
 function script_exit() {
-    [ "$?" -ne 0 ] && echo "FAILURE" || echo "FINISHED"
+    if [ "$?" -ne 0 ] ; then
+        # Always log the last 1000 MicroShift and etcd journal entries
```
dhellmann, line 32: Why do we want so much output?

I think we'll serve the user better by giving them a concise message and helping them understand where to go to debug next. Dumping all of the log output from MicroShift into the health check log is going to make it harder to debug failures unrelated to MicroShift.

packaging/greenboot/microshift-running-check.sh
```diff
@@ -28,7 +28,43 @@ function return_failure() {
...
+        for srv in microshift.service microshift-etcd.service ; do
+            log_failure_cmd "${srv}" "journalctl -xu ${srv} -n 1000 --no-pager"
+        done
+        # Always log the list of pods
```
dhellmann, line 36: Could we log this sort of information closer to the place where it is being checked? For example, `"Expected N pods in $namespace, found M: $(oc get pods -n $namespace)"`

packaging/greenboot/microshift-running-check.sh
ggiguash: Agreed, I overdid it - removing.

packaging/greenboot/microshift-running-check.sh
ggiguash: The problem is that it's difficult because the checks are running in background and there's a wait for status reports.
I think, however, that producing pod list and events before exit helps to identify the problem on the spot.

So, there will be a message that pod restart check failed and pod-list-snapshot taken 1s after that error.

packaging/greenboot/microshift-running-check.sh
```diff
@@ -41,8 +77,9 @@ function microshift_service_active() {
     local -r is_active=$(systemctl is-active microshift.service)
 
     # Terminate the script in case of a failed service - nothing to wait for
-    if [ "${is_failed}" = "failed" ] ; then
-        print_failure_logs
+    if [ "${is_failed}" = "failed" ] || [ "${is_failed}" = "inactive" ] ; then
+        # MicroShift upgrade failure logs
+        LOG_FAILURE_FILES+=("/var/lib/microshift-backups/prerun_failed.log")
```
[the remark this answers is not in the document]
ggiguash: I will keep it in the implementation file, but move it to the main function unconditionally.

packaging/greenboot/microshift-running-check.sh
dhellmann: How is a user supposed to know what pods should be present so they can compare that to the list of pods that are present?

dhellmann: In the loops where the script looks for pods in specific namespaces, it knows when it does not find them. It should report whatever details it can at that point. Messages like "Expected 5 ready pods in openshift-foo but found 4" point directly to issues within the namespace. "Deployment `embedded-component` has no ready pods" is even better because it points directly to the thing that is broken.

ggiguash: I added error messages in all the loops. Is this any better?

dhellmann (commented):
/lgtm

ggiguash:
I reverted the "optimization" of service status check [here](https://github.com/openshift/microshift/pull/2150/files#diff-f55716eb250be56950561b7dd1372286b90452e7acd9f5224c76408dec148028R69).

PR #2154, opened 2023-08-01
NO-ISSUE:  Add auto-attach to rhsm registration command in manage-vm.sh script

eslutsky:
Ensures that manage-vm.sh  bring the rhsm to a healthy state.

On certain cases  after manage-vm.sh  and  RHEL registration run  successful ,  the system status is Invalid:
```
microshift@microshift-dev ~]$ sudo subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Invalid

Red Hat Enterprise Linux for x86_64:
- Not supported by a valid subscription.

System Purpose Status: Not Specified
```


ggiguash:
@eslutsky , can we also fix this script?
https://github.com/openshift/microshift/blob/main/scripts/devenv-builder/configure-vm.sh#L84

scripts/devenv-builder/configure-vm.sh
```diff
@@ -80,8 +80,15 @@ echo -e "${USER}\tALL=(ALL)\tNOPASSWD: ALL" | sudo tee "/etc/sudoers.d/${USER}"
 
 # Check the subscription status and register if necessary
 if ${RHEL_SUBSCRIPTION}; then
+
+    # Handle invalid status
+    if sudo subscription-manager status | grep Invalid >&/dev/null ; then
+        sudo subscription-manager attach --auto
+    fi
+
+    # Assume its unregistered
```
ggiguash, line 89: This code would not work when the system is unregistered. I'm not sure we should parse the output at all:
```
# subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Unknown

System Purpose Status: Unknown

# echo $?
1
```

eslutsky: yeah, it wouldnt handle the case when its unregistered, only when its Invalid state (invalid) ,  the other option is to fail early if its Invalid with an error message.

ggiguash: So, let's remove `grep` now that we're handing the registration with `auto-attach`?

docs/contributor/devenv_setup.md
```diff
@@ -60,7 +60,7 @@ ssh-copy-id -f microshift@${VMIPADDR}
 
 scp ./scripts/devenv-builder/configure-vm.sh microshift@${VMIPADDR}:
 scp ~/.pull-secret.json microshift@${VMIPADDR}:
-ssh microshift@${VMIPADDR} "~/configure-vm.sh ~/.pull-secret.json"
+ssh -t microshift@${VMIPADDR} "~/configure-vm.sh ~/.pull-secret.json"
```
ggiguash, line 63: Why do we need a terminal allocation here?

ggiguash: Ah, I see. Terminal is necessary when we're hitting the subscription register case.
```
+ sudo subscription-manager status
+ sudo subscription-manager register
Registering to: subscription.rhsm.redhat.com:443/subscription
Error: --username is a required parameter in non-interactive mode.
```

PR #2156, opened 2023-08-01
NO-ISSUE:  set the correct permission for osbuild-composer

eslutsky:
when /home/microshift  using default permissions (700)  `build.sh`  fails with an error: 
```bash
# Loading microshift blueprint v0.0.1
+ sudo composer-cli blueprints delete microshift
+ true
+ sudo composer-cli blueprints push /home/microshift/microshift/_output/image-builder/blueprint_v0.0.1.toml
+ sudo composer-cli blueprints depsolve microshift
ERROR: BlueprintsError: microshift: DNF error occurred: RepoError: There was a problem reading a repository: Failed to download metadata for repo 'de3d95914e632f2aee75dc0a200ac02106835607234a2ba2e4b3c2451b1efcad' [microshift-local: file:///home/microshift/microshift/_output/image-builder/microshift-local/]: Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
+ /home/microshift/microshift/scripts/image-builder/cleanup.sh

```

### DIFF
--- a/scripts/image-builder/configure.sh
+++ b/scripts/image-builder/configure.sh
@@ -25,9 +25,14 @@ mkdir "${TEST_FILE}.dir"
 HOME_PERM=$(stat -c 0%a ~)
 FILE_PERM=$(stat -c 0%a "${TEST_FILE}.file")
 DIR_PERM=$(stat -c 0%a "${TEST_FILE}.dir")
-rm -rf "${TEST_FILE}"*
 
-if [ "${HOME_PERM}" -lt 0755 ] || [ "${FILE_PERM}" -lt 0644 ] || [ "${DIR_PERM}" -lt 0755 ] ; then
-    echo "Check home directory permissions and umask. The settings must allow read to group and others"
+# Set the Correct Permissions for osbuild-composer
+[ "${HOME_PERM}" -lt 0711 ]  && chmod go+x ~
+
+if [ "${FILE_PERM}" -lt 0644 ] || [ "${DIR_PERM}" -lt 0711 ] ; then
+    echo "Check ${TEST_FILE}.dir permissions. The umask setting must allow execute to group/others"
+    echo "Check ${TEST_FILE}.file permissions. The umask setting must allow read to group/others"    
     exit 1
 fi
+
+rm -rf "${TEST_FILE}"*


scripts/image-builder/configure.sh
```diff
@@ -17,6 +17,9 @@ sudo dnf install -y "https://dl.fedoraproject.org/pub/epel/epel-release-latest-$
 sudo dnf install -y mock caddy tomcli
 sudo usermod -a -G mock "$(whoami)"
 
+# Set the Correct Permissions for osbuild-composer
+sudo chmod 0755 /home/microshift
```
dhellmann, line 21: Does the whole home directory actually need to be world readable or can it be executable so that subdirectories can be accessed but someone can't read my dot files?

ggiguash: @eslutsky , can you test if osbuilder works in the following configuration?
```
chmod 700 ~microshift
chmod ugo+x ~microshift
```

ggiguash: Also, since we're testing HOME_PERM below, let's only change the permissions is it's necessary - something like the following:
```
[ "${HOME_PERM}" -lt 0755 ]  && chmod 755 ~
if [ "${FILE_PERM}" -lt 0644 ] || [ "${DIR_PERM}" -lt 0755 ] ; then
    echo "Check your umask. The settings must allow read to group and others"
    exit 1
fi
```

eslutsky: > Does the whole home directory actually need to be world readable or can it be executable so that subdirectories can be accessed but someone can't read my dot files?
@dhellmann , i agree that  ch-modding 700 to  home dir is too permissive and not  a good idea , 
@ggiguash   `+x`  does solve our problem and its  much safer.

scripts/image-builder/configure.sh
```diff
@@ -17,6 +17,9 @@ sudo dnf install -y "https://dl.fedoraproject.org/pub/epel/epel-release-latest-$
 sudo dnf install -y mock caddy tomcli
 sudo usermod -a -G mock "$(whoami)"
 
+# Set the Correct Permissions for osbuild-composer
+[ "${HOME_PERM}" -lt 0711 ]  && chmod 0711 ~
```
dhellmann, line 21: I'd rather if we did delta changes instead of absolute changes. Who knows what other permissions someone might need for other reasons?

```suggestion
[ "${HOME_PERM}" -lt 0711 ]  && chmod go+x ~
```

scripts/image-builder/configure.sh
```diff
@@ -27,7 +27,11 @@ FILE_PERM=$(stat -c 0%a "${TEST_FILE}.file")
...
 rm -rf "${TEST_FILE}"*
 
-if [ "${HOME_PERM}" -lt 0755 ] || [ "${FILE_PERM}" -lt 0644 ] || [ "${DIR_PERM}" -lt 0755 ] ; then
-    echo "Check home directory permissions and umask. The settings must allow read to group and others"
+
+# Set the Correct Permissions for osbuild-composer
+[ "${HOME_PERM}" -lt 0711 ]  && chmod go+x ~
+
+if [ "${FILE_PERM}" -lt 0644 ] || [ "${DIR_PERM}" -lt 0711 ] ; then
+    echo "Check ${TEST_FILE}.dir  permissions and umask. The settings must allow execute to others"
```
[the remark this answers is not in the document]
eslutsky: > 

also moving `rm -rf "${TEST_FILE}"*` to last line.

PR #2157, opened 2023-08-01
USHIFT-1528: Don't throw an error when there is no rollback deploy

pmtk:
System without rollback deployment is valid system.
By not throwing an error in this situation, MicroShift will continue start up (without restoring anything) to give cluster a chance to recover instead of just failing.

### DIFF
--- a/pkg/admin/prerun/system.go
+++ b/pkg/admin/prerun/system.go
@@ -90,7 +90,8 @@ func getRollbackDeploymentID() (string, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


dhellmann:
This looks OK, but really could use some more "why" in the commit message to explain why the change is needed.

Based on the bug ticket, I think we end up in the state where the system is unhealthy but there is no rollback commit if the first image booted fails somehow, right? Don't we want users to clean up that data and restart us? Maybe not because there's unlikely to be anything of value?

PR #2163, opened 2023-08-02
USHIFT-1392: feat: add e2e test for staged failed update scenario

eggfoobar:
add test that validates when an unhealthy first boot happens during upgrade, a restore is performed.

### DIFF
--- a/test/suites-ostree/staged-failed-upgrade.robot
+++ b/test/suites-ostree/staged-failed-upgrade.robot
@@ -58,9 +58,9 @@ Validate Backup Is Restored
 
     Backup Should Exist    ${backup_name}
 
-    ${stdout}    ${rc}=    Execute Command
+    ${stdout}=    Execute Command
     ...    cd ${BACKUP_STORAGE} && ls -d1 rhel-*
-    ...    sudo=False    return_rc=True
+    ...    sudo=False    return_rc=False
 
     ${mark_stdout}    ${mark_stderr}    ${mark_rc}=    Execute Command
     ...    stat ${DATA_DIR}/marker
@@ -73,8 +73,8 @@ Expected Boot Count
     [Documentation]    Validate that the host rebooted only the specified number of times
     [Arguments]    ${reboot_count}
 
-    ${stdout}    ${rc}=    Execute Command
+    ${stdout}=    Execute Command
     ...    journalctl --list-boots --quiet | wc -l
-    ...    sudo=True    return_rc=True
+    ...    sudo=True    return_rc=False
 
     Should Be Equal As Integers    ${reboot_count}    ${stdout}
--- a/test/suites-ostree/staged-failed-upgrade.robot
+++ b/test/suites-ostree/staged-failed-upgrade.robot
@@ -26,7 +26,7 @@ First Failed Staged Upgrade Should Restore Backup
     Wait For Healthy System
 
     ${initial_backup}=    Get Future Backup Name For Current Boot
-    Create Backup With Marker    ${initial_backup}
+    Create Backup    ${initial_backup}    ${TRUE}
 
     TestAgent.Add Action For Next Deployment    1    fail_greenboot
 
@@ -47,29 +47,10 @@ Setup
     Check Required Env Variables
     Should Not Be Empty    ${TARGET_REF}    TARGET_REF variable is required
     Login MicroShift Host
-    Setup Kubeconfig
 
 Teardown
     [Documentation]    Test suite teardown
     Logout MicroShift Host
-
-Create Backup With Marker
-    [Documentation]    Stops MicroShift and creates manual backup that
-    ...    masquerades as automated one and creates marker file for validating restore.
-    [Arguments]    ${backup_name}
-    Systemctl    stop    microshift.service
-
-    ${stdout}    ${stderr}    ${rc}=    Execute Command
-    ...    microshift backup --name "${backup_name}"
-    ...    sudo=True    return_stderr=True    return_rc=True
-
-    # create a marker that we expect to show up in data directory after restore
-    ${mark_stdout}    ${mark_stderr}    ${mark_rc}=    Execute Command
-    ...    touch ${BACKUP_STORAGE}/${backup_name}/marker
-    ...    sudo=True    return_stderr=True    return_rc=True
-
-    Should Be Equal As Integers    0    ${rc}
-    Should Be Equal As Integers    0    ${mark_rc}
 
 Validate Backup Is Restored
     [Documentation]    Validate that the desired backup exists and is the only one that exists
--- a/test/suites-ostree/staged-failed-upgrade.robot
+++ b/test/suites-ostree/staged-failed-upgrade.robot
@@ -27,11 +27,16 @@ First Failed Staged Upgrade Should Restore Backup
     ${initial_backup}=    Get Future Backup Name For Current Boot
 
     TestAgent.Add Action For Next Deployment    1    fail_greenboot
-    Deploy Commit Not Expecting A Rollback    ${TARGET_REF}
+
+    # We don't expect a rollback here since after the first failure
+    # things should continue as desired. This test validates the
+    # restore function is working when the first boot is unhealthy.
+    Deploy Commit Not Expecting A Rollback    ${TARGET_REF}    ${TRUE}
 
     Wait For Healthy System
 
-    Validate Backup    ${initial_backup}
+    Validate Backup Is Restored    ${initial_backup}
+    Expected Boot Count    3
 
 
 *** Keywords ***
@@ -46,14 +51,29 @@ Teardown
     [Documentation]    Test suite teardown
     Logout MicroShift Host
 
-Validate Backup
+Validate Backup Is Restored
     [Documentation]    Validate that the desired backup exists and is the only one that exists
-    [Arguments]    ${initial_backup}
+    [Arguments]    ${backup_name}
 
-    Backup Should Exist    ${initial_backup}
+    Backup Should Exist    ${backup_name}
 
     ${stdout}    ${rc}=    Execute Command
     ...    cd ${BACKUP_STORAGE} && ls -d1 rhel-*
     ...    sudo=False    return_rc=True
 
-    Should Be Equal As Strings    ${stdout}    ${initial_backup}
+    ${grep_stdout}    ${grep_rc}=    Execute Command
+    ...    journalctl -u microshift.service | grep "Starting restore"
+    ...    sudo=True    return_rc=True
+
+    Should Be Equal As Strings    ${stdout}    ${backup_name}
+    Should Be Equal As Integers    0    ${grep_rc}
+
+Expected Boot Count
+    [Documentation]    Validate that the host rebooted only the specified number of times
+    [Arguments]    ${reboot_count}
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    sudo journalctl --list-boots --quiet | wc -l
+    ...    sudo=True    return_rc=True
+
+    Should Be Equal As Integers    ${reboot_count}    ${stdout}
--- a/test/suites-ostree/staged-failed-upgrade.robot
+++ b/test/suites-ostree/staged-failed-upgrade.robot
@@ -1,0 +1,59 @@
+*** Settings ***
+Documentation       Tests related to upgrading MicroShift
+
+Resource            ../resources/common.resource
+Resource            ../resources/ostree.resource
+Library             Collections
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+Test Tags           ostree
+
+
+*** Variables ***
+${USHIFT_HOST}      ${EMPTY}
+${USHIFT_USER}      ${EMPTY}
+
+${TARGET_REF}       ${EMPTY}
+
+
+*** Test Cases ***
+First Failed Staged Upgrade Should Restore Backup
+    [Documentation]    Performs an upgrade, when first unhealthy a restore is performend
+
+    Wait For Healthy System
+
+    ${initial_backup}=    Get Future Backup Name For Current Boot
+
+    TestAgent.Add Action For Next Deployment    1    fail_greenboot
+    Deploy Commit Not Expecting A Rollback    ${TARGET_REF}
+
+    Wait For Healthy System
+
+    Validate Backup    ${initial_backup}
+
+
+*** Keywords ***
+Setup
+    [Documentation]    Test suite setup
+    Check Required Env Variables
+    Should Not Be Empty    ${TARGET_REF}    TARGET_REF variable is required
+    Login MicroShift Host
+    Setup Kubeconfig
+
+Teardown
+    [Documentation]    Test suite teardown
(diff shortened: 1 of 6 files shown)


test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,59 @@
...
+*** Test Cases ***
+First Failed Staged Upgrade Should Restore Backup
+    [Documentation]    Performs an upgrade, when first unhealthy a restore is performend
+
+    Wait For Healthy System
+
+    ${initial_backup}=    Get Future Backup Name For Current Boot
+
+    TestAgent.Add Action For Next Deployment    1    fail_greenboot
+    Deploy Commit Not Expecting A Rollback    ${TARGET_REF}
```
dhellmann, line 30: Why don't we expect a rollback here?

eggfoobar: In this situation we're verifying that if the first deployment is set to unhealthy just once, we expect the restore to be triggered, but after the first boot, things should roll as intended with out rollback. However, I think I might have a mis-conception here, added another validation to make sure the log statements are printing but I'm not seeing the restore logs.

pmtk: Ah, take a look at:
```
Deploy Commit Expecting A Rollback
    [Documentation]    Deploys given ref and configures test agent for failing greenboot.
    ...    It expects the system to roll back.
    [Arguments]    ${ref}    ${write_agent_cfg}=${TRUE}

    ${initial_deploy_id}=    Get Booted Deployment Id
    ${deploy_id}=    Rebase System    ${ref}
    IF    ${write_agent_cfg}    TestAgent.Write
```

`Deploy Commit Not Expecting A Rollback` does not persist the TestAgent's config, so the test didn't create the scenario we wanted - after staging and booting into a new deployment, there should be one auto-reboot, so on 2nd boot MicroShift should restore the backup of the first (rollback) deploy

pmtk: I implemented following change
```
diff --git a/test/resources/ostree.resource b/test/resources/ostree.resource
index 0efdc4b9b..d1b8ebeec 100644
--- a/test/resources/ostree.resource
+++ b/test/resources/ostree.resource
@@ -41,10 +41,11 @@ Deploy Commit Expecting A Rollback
 Deploy Commit Not Expecting A Rollback
     [Documentation]    Deploys given ref and configures test agent for failing greenboot.
     ...    It expects the system to roll back.
-    [Arguments]    ${target_ref}
+    [Arguments]    ${target_ref}     ${write_agent_cfg}=${FALSE}    
 
     ${initial_deploy_id}=    Get Booted Deployment Id
     ${deploy_id}=    Rebase System    ${target_ref}
+    IF    ${write_agent_cfg}    TestAgent.Write
     Reboot MicroShift Host
 
     Log To Console    "System rebased to ${target_ref} and rebooted - starting health checking"
diff --git a/test/suites-ostree/staged-failed-upgrade.robot b/test/suites-ostree/staged-failed-upgrade.robot
index 3f6ba1085..3b44bb3df 100644
--- a/test/suites-ostree/staged-failed-upgrade.robot
+++ b/test/suites-ostree/staged-failed-upgrade.robot
@@ -27,7 +27,7 @@ First Failed Staged Upgrade Should Restore Backup
     ${initial_backup}=    Get Future Backup Name For Current Boot
 
     TestAgent.Add Action For Next Deployment    1    fail_greenboot
-    Deploy Commit Not Expecting A Rollback    ${TARGET_REF}
+    Deploy Commit Not Expecting A Rollback    ${TARGET_REF}    ${TRUE}
 
     Wait For Healthy System
```

but I see that `test-agent` doesn't get to remove the artifacts:
```
Aug 03 03:12:00 rhel-9-2-microshift-source-failed-staged-host1 bash[771]: + sleep infinity
Aug 03 03:14:42 rhel-9-2-microshift-source-failed-staged-host1 bash[771]: Terminated
Aug 03 03:14:42 rhel-9-2-microshift-source-failed-staged-host1 bash[771]: ++ _cleanup
Aug 03 03:14:42 rhel-9-2-microshift-source-failed-staged-host1 bash[771]: ++ for cmd in "${CLEANUP_CMDS[@]}"
Aug 03 03:14:42 rhel-9-2-microshift-source-failed-staged-host1 bash[771]: ++ rm -v /etc/greenboot/check/required.d/99_microshift_test_failure.sh
Aug 03 03:14:42 rhel-9-2-microshift-source-failed-staged-host1 systemd[1]: Stopping MicroShift Test Agent...
Aug 03 03:14:42 rhel-9-2-microshift-source-failed-staged-host1 bash[771]: Terminated
```

but it should be (`-v` to `rm`):
```
+ sleep infinity
^C++ _cleanup
++ for cmd in "${CLEANUP_CMDS[@]}"
++ rm -v /etc/greenboot/check/required.d/99_microshift_test_failure.sh
removed '/etc/greenboot/check/required.d/99_microshift_test_failure.sh'
```

pmtk: After adding `exit 0` to `_cleanup()` in `microshift-test-agent.sh` it seems that the file gets deleted properly.
Spoke to Egli and he'll include that as well

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,82 @@
...
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    cd ${BACKUP_STORAGE} && ls -d1 rhel-*
+    ...    sudo=False    return_rc=True
+
+    ${grep_query}=    Catenate    SEPARATOR=
+    ...    "Restoring backup for a rollback deployment to perform migration and try starting again"
+
+    ${grep_stdout}    ${grep_rc}=    Execute Command
+    ...    journalctl -u microshift.service -b0 | grep ${grep_query}
```
dhellmann, line 68: journalctl accepts a `--grep` option. I don't know if that's better than piping to grep, or just different.

Does grep in a pipe like this fail in the way we would expect? In bash, we would have to set the pipefail option, for example.

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,82 @@
...
+
+    Should Be Equal As Strings    ${stdout}    ${backup_name}
+    Should Be Equal As Integers    0    ${grep_rc}
+
+Expected Boot Count
+    [Documentation]    Validate that the host rebooted only the specified number of times
+    [Arguments]    ${reboot_count}
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    sudo journalctl --list-boots --quiet | wc -l
```
dhellmann, line 79: Shouldn't need the extra sudo here, since the keyword already has `sudo=True` set.

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,82 @@
...
+    [Arguments]    ${backup_name}
+
+    Backup Should Exist    ${backup_name}
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    cd ${BACKUP_STORAGE} && ls -d1 rhel-*
+    ...    sudo=False    return_rc=True
+
+    ${grep_query}=    Catenate    SEPARATOR=
+    ...    "Restoring backup for a rollback deployment to perform migration and try starting again"
```
dhellmann, line 65: Looking for this message tells us that MicroShift thinks it is restoring the content, but it doesn't tell us that the content was actually restored. What can we do to prove it's the same data?

Could we put a marker file in the data directory, for example?

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: Oh nice, to my knowledge and from testing it, it spits out a non zero status code if `grep` fails in both the pipe and the `journalctl` flag. It looks cleaner with the flag, so I went ahead and updated it.

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: Good catch, updated it.

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: > Looking for this message tells us that MicroShift thinks it is restoring the content, but it doesn't tell us that the content was actually restored. What can we do to prove it's the same data?

In this test the way we try and validate that, is by checking what MicroShift logs has happened and what the state of the backup directory is after. We record the initial backup, and make sure that is the only backup that exists in the filesystem, plus the log statement from MicroShift, we can be sure that things progressed as intended. If there was a failure in the restore MicroShift wouldn't start and that would break this test.

> Could we put a marker file in the data directory, for example?

I'm not opposed to this, however, in the backup/restore logic we put a lot of effort to be self healing. Presenting a file marker might be too much of an implementation detail where it's usefulness might be limited to just this test. Maybe a consistent log statement would suffice for this? (ex: `step: prerun, backup: created | restored...`)

pmtk: I think the marker Doug is mentioning is created in test, TAL:
- https://github.com/openshift/microshift/blob/main/test/suites-ostree/backup-restore.robot#L41 
- https://github.com/openshift/microshift/blob/main/test/suites-ostree/rollback.robot#L35-L38

We could do that here, we just need to get the timing right, but that should be hard, there's plenty of time: after rebooting to newly staged deployment, shortly after starting MicroShift should create a backup, that's our window until the auto-reboot

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,82 @@
...
+*** Variables ***
+${USHIFT_HOST}      ${EMPTY}
+${USHIFT_USER}      ${EMPTY}
+
+${TARGET_REF}       ${EMPTY}
+
+
+*** Test Cases ***
+First Failed Staged Upgrade Should Restore Backup
+    [Documentation]    Performs an upgrade, when first unhealthy a restore is performend
```
pmtk, line 23: WDYT?

```suggestion
MicroShift Restores Backup To Reattempt Upgrade
    [Documentation]    Staged deployment is unhealthy on first boot. After auto-greenboot-reboot MicroShift should restore rollback deployment's backup to reattempt the upgrade from healthy data.
```

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: Ah see, thats way easier. Let me update the test to use that then. Do we even need to worry about the timing if we do what the restore does and just manually create a backup and create the marker there and kick off the deployment?

pmtk: I think that creating backup manually will cause MicroShift to fail, because it's not expecting it to exist (even for the same deployment, each boot produces different backup based on previous boot ID, we just keep the most recent one)

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: Not needed anymore, using file marker instead for identifying restores.

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: I like it, updated .

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,99 @@
...
+    Check Required Env Variables
+    Should Not Be Empty    ${TARGET_REF}    TARGET_REF variable is required
+    Login MicroShift Host
+    Setup Kubeconfig
+
+Teardown
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+
+Create Backup With Marker
```
pmtk, line 56: This is very similar to what we have in `backup-restore.robot`, do you think we should join these functions as one in common file?
We can also do it as a follow up

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,99 @@
...
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    microshift backup --name "${backup_name}"
+    ...    sudo=True    return_stderr=True    return_rc=True
+
+    # create a marker that we expect to show up in data directory after restore
+    ${mark_stdout}    ${mark_stderr}    ${mark_rc}=    Execute Command
+    ...    touch ${BACKUP_STORAGE}/${backup_name}/marker
+    ...    sudo=True    return_stderr=True    return_rc=True
+
+    Should Be Equal As Integers    0    ${rc}
```
pmtk, line 71: I would check this before creating the marker - what if making a backup fails and you try to create marker anyway? Not a big deal, but might be confusing when debuggin

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,99 @@
...
+    Expected Boot Count    3
+
+
+*** Keywords ***
+Setup
+    [Documentation]    Test suite setup
+    Check Required Env Variables
+    Should Not Be Empty    ${TARGET_REF}    TARGET_REF variable is required
+    Login MicroShift Host
+    Setup Kubeconfig
```
pmtk, line 50: Teardown is missing `Remove Kubeconfig`

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: Alrighty, I like that, changing it up now, just testing it real quick.

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: Good call out, yeah this just happened now where it failed due to the backup creation failed due to the change in the subcommand for backup. Updating to fail early here. Thanks!

test/suites-ostree/staged-failed-upgrade.robot
eggfoobar: Oh, that's not needed anymore, removed it

test/suites-ostree/staged-failed-upgrade.robot
```diff
@@ -0,0 +1,80 @@
...
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+
+Validate Backup Is Restored
+    [Documentation]    Validate that the desired backup exists and is the only one that exists
+    [Arguments]    ${backup_name}
+
+    Backup Should Exist    ${backup_name}
+
+    ${stdout}    ${rc}=    Execute Command
```
pmtk, line 61: You're not checking the ${rc}

eggfoobar: Oh, yeah, I didn't really care about the `${rc}` here, the `${stdout}` is matched against the `${backup_name}`, so it'll fail if `${rc}` is non zero regardless.

eggfoobar: Since we're not using the RC here, decided to remove it since it's not needed information, the logs by default state the RC.

PR #2149, opened 2023-07-31
USHIFT-1459: Add CNCF conformance test scenario

pacevedom:
**Which issue(s) this PR addresses**:

Closes #<Issue Number>

PR #2164, opened 2023-08-02
USHIFT-1097 Test upgrade from Engineering Candidate

pmtk:


test/scenarios/rhel-9.2-microshift-source-from-latest-ec.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
pmtk, line 1: Planning to move the file to `scenarios-periodics` after https://github.com/openshift/microshift/pull/2149/ merges.
@dhellmann @pacevedom I guess we'll run all periodics in a single job?

dhellmann: Yes, that's what I had in mind. I think @pacevedom 's work will make that easy.

test/image-blueprints/group2/rhel92-microshift-latest-ec.toml
```diff
@@ -0,0 +1,44 @@
...
+description = "RHEL 9.2 with latest MicroShift Engineering Candidate."
+version = "0.0.1"
+modules = []
+groups = []
+distro = "rhel-92"
+
+# FIXME: Temporarily pin the openvswitch version.
+[[packages]]
+name = "openvswitch3.1"
+version = "3.1.0-14.el9fdp"
```
dhellmann, line 11: We can remove this pin for 4.14 ECs now.

test/package-sources/microshift-latest-ec.toml
```diff
@@ -0,0 +1,7 @@
+id = "microshift-latest-ec"
+name = "Latest MicroShift Engineering Candidate"
+type = "yum-baseurl"
+url = "https://mirror.openshift.com/pub/openshift-v4/${UNAME_M}/microshift/ocp-dev-preview/latest-4.14/elrhel-9/os/"
```
dhellmann, line 4: Just FYI, we _might_ need to change the `elrhel-9` portion of this path back to just `el9` later. @vfreex was going to look at fixing the pipeline change that introduced that path change for EC4.

pmtk:
/unhold

Changed composer's source to use $MINOR_VERSION in the url:
```
url = "https://mirror.openshift.com/pub/openshift-v4/${UNAME_M}/microshift/ocp-dev-preview/latest-4.${MINOR_VERSION}/elrhel-9/os/"
```

pmtk:
The test currently fails due to recent change of `version` schema (from `X.Y.Z` to json):
```
failed to get version of existing MicroShift data: failed to read version file: unmarshalling \"4.14.0\" failed: invalid character '.' after top-level value
```

@dhellmann is it okay that we don't implement "migration" from one schema to the other and just wait for new EC/RC?

pmtk:
Let's wait for https://github.com/openshift/release/pull/41823 so we can test this as presubmit
/hold

dhellmann:
> The test currently fails due to recent change of `version` schema (from `X.Y.Z` to json):
> 
> ```
> failed to get version of existing MicroShift data: failed to read version file: unmarshalling \"4.14.0\" failed: invalid character '.' after top-level value
> ```
> 
> @dhellmann is it okay that we don't implement "migration" from one schema to the other and just wait for new EC/RC?

I think right now that's safe. After we have a GA release we would have to deal with format conversions like that.

If we don't make the change now, how long will it be before we have a candidate release that will pass the job with the right data format?

pmtk:
I'm planning to revisit this next sprint with some dynamic logic to pick EC, RC, or even released RPMs.

PR #2179, opened 2023-08-06
USHIFT-1512: Prerequisites for automated testing of disconnected use-case

ggiguash:
Implemented prerequisites for testing container image embedded in isolated network.

The following changes were implemented to support the new functionality.
- Fix image-builder/configure.sh script to include the osbuilder pull secret configuration
- The `build_images.sh` script support for rendering embedded container list via `RENDER_CONTAINER_IMAGES`
  - Use first gost IP instead of VM_BRIDGE_IP in ostree parent URL
- The `manage_hypervisor_config.sh` script is a combination of `manage_vm_storage_pool.sh` and `configure_hypervisor_firewall.sh`
  - There is also a cleanup option to revert the configuration
- Deleted `setup_local.sh` script as it seems to overlap with `ci_phase_iso_build.sh`
- Updated README.md with:
  - `RENDER_CONTAINER_IMAGES` feature
  -  New `manage_hypervisor_config.sh` script replacing others
  - The `launch_vm` optional arguments
- Added configuration of an "isolated" network from `isolated-network.xml` file
  - Added an option to specify a network optional argument in `launch_vm` function of `scenario.sh`
- Fix `configure-vm.sh` script not to drag the libvirt dependency via cockpit VM plugin
- Refactored CI scripts:
  - Allow running `ci_phase_iso_build.sh` on a devhost without libvirt installed
  - Fix `ci_phase_iso_boot.sh` to call the new hypervisor config script
- Fixes in common.sh
  - Added a notion of "isolated" network
  - Do not fail when no libvirt bridge interfaces exist

Partially implements: [USHIFT-1512](https://issues.redhat.com//browse/USHIFT-1512)


test/bin/manage_hypervisor_config.sh
```diff
@@ -0,0 +1,104 @@
...
+        sudo virsh pool-define-as "${VM_STORAGE_POOL}" dir --target "${VM_DISK_DIR}"
+        sudo virsh pool-build "${VM_STORAGE_POOL}"
+        sudo virsh pool-start "${VM_STORAGE_POOL}"
+        sudo virsh pool-autostart "${VM_STORAGE_POOL}"
+    fi
+
+    # Isolated network
+    if ! sudo sudo virsh net-info "${VM_ISOLATED_NETWORK}" &>/dev/null ; then
+        local -r netconfig_tmpl="${SCRIPTDIR}/../assets/isolated-network.xml"
+        local -r netconfig_file="$(mktemp /tmp/isolated-network-XXXXX.xml)"
```
dhellmann, line 54: Can we put this rendered file somewhere under $IMAGEDIR so we can collect it as part of the output of the script if we need to? Maybe as a file in $IMAGEDIR/infrastructure for example? Then the output file can have a constant name, too.

test/bin/setup_local.sh
```diff
@@ -1,32 +0,0 @@
-#!/bin/bash
```
dhellmann, line 1: Why delete this file?

test/README.md
```diff
@@ -285,8 +301,8 @@ images that use RPMs created from source (not already published releases).
 #### Rebuilding from Sources Easily
 
 If you build new RPMs, you need to re-run several steps (build the
-RPMs, build the local repos, build the images, download the
-images). Use `./bin/rebuild_source_images.sh` to automate all of those
+RPMs, build the local repos, build the images, download the images).
+Use `./bin/dev_rebuild_source_images.sh` to automate all of those
```
dhellmann, line 305: I'm not sure why this script needed to be renamed?

test/bin/build_images.sh
```diff
@@ -56,6 +56,47 @@ configure_package_sources() {
...
+
+    # Extract the version string
+    local -r version="$(grep ^RENDER_CONTAINER_IMAGES= "${blueprint_file}" | cut -d'=' -f2 | xargs)"
+    if [ -z "${version}" ] ; then
+        echo "Error: missing version string for the RENDER_CONTAINER_IMAGES macro in ${blueprint_file}"
+        exit 1
+    fi
+
+    # Find the microshift-release-info RPM with the specified version
+    local -r release_info_rpm=$(find "${IMAGEDIR}/rpm-repos" -name "microshift-release-info-${version}*.rpm" | tail -1)
```
dhellmann, line 77: That output from find should probably be sorted, too.

I wonder if it would be less error prone to use the repository name instead of the version string? How do I express that I want the RPM being built from the PR, for example?

test/assets/isolated-network.xml
```diff
@@ -0,0 +1,9 @@
+<network>
+  <name>${VM_ISOLATED_NETWORK}</name>
```
dhellmann, line 2: Does the network name need to change?

test/README.md
ggiguash: It does not need to be renamed. I will revert this.
The original idea was to note that this script is for dev flows.

test/assets/isolated-network.xml
ggiguash: The idea was to have this name defined in `commons.sh` and propagate everywhere. 
It does not need to change at the moment.

test/bin/setup_local.sh
ggiguash: There are quite a few files under `test/bin` that run command sequences. This one called `ci_phase_iso_build` (a composition of other commands by itself) and `start_webserver.sh`.
Note that web server is already started from the `phase` script, so I thought this one is redundant.
Finally, it is not documented on when it should be called.

test/assets/isolated-network.xml
dhellmann: Makes sense.

test/bin/build_images.sh
ggiguash: I fixed the `find` command output to be sorted. 
For using repository names, I think the versions are covering most of the cases we need because they are "always" explicitly mentioned in the blueprints. So, the macro would be something like:
```
RENDER_CONTAINER_IMAGES="${SOURCE_VERSION}"
```

test/bin/setup_local.sh
dhellmann: OK, we should document it.

The intent was to provide a single script for a developer to use to build their images and set up to the point of being able to run a single scenario of their choice on a local system.

test/bin/build_images.sh
dhellmann: Sounds good, I didn't notice that the templating took 2 passes.

dhellmann (commented):
Everything here is OK except removing the setup_local.sh script. If you keep that, I'll update the docs to mention it.

test/bin/setup_local.sh
ggiguash: I thought that was `ci_phase_iso_build` for a full build, or just build_images -s

ggiguash:
> Everything here is OK except removing the setup_local.sh script. If you keep that, I'll update the docs to mention it.

@dhellmann , following the offline discussion we had, it was agreed that we can delete the script.

test/bin/common.sh
```diff
@@ -123,11 +138,10 @@ get_vm_bridge_ip() {
...
     #        valid_lft forever preferred_lft forever
-
     ip -f inet addr show "${bridge}" | grep inet | awk '{print $2}' | cut -d/ -f1
 }
 
 # The IP address of the current host on the bridge used for the
 # default network for libvirt VMs.
 # shellcheck disable=SC2034  # used elsewhere
-VM_BRIDGE_IP="$(get_vm_bridge_ip)"
+VM_BRIDGE_IP="$(get_vm_bridge_ip "default")"
```
pmtk, line 147: This var is used in `scenario.sh`: `WEB_SERVER_URL="http://${VM_BRIDGE_IP}:${WEB_SERVER_PORT}"`.
Does it mean "isolated network" scenarios will be installed from an "all-in-one ISO" containing installer+commit+embedded images? (Just like the `scripts/image-builder/`). If yes, looks like we'll have another blueprint group (or atleast group "2.5"):
- `rhel-92` edge-commit
- `rhel-92 + microshift + images` edge-commit
- `all-in-one ISO` edge-installer

It cannot be without an effect on how long `ci_phase_iso_build.sh` takes to complete (buildiong edge-installers is kind of slow...).
Could we use "regular" install procedure just change the network post install?
I'd say we don't care how the system gets installed, but we care about the environment when it boots

test/bin/manage_hypervisor_config.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
pmtk, line 1:
```suggestion
#!/usr/bin/env bash
```

test/bin/common.sh
ggiguash: We decided to split the prerequisite work from the implementation, so I see it can be confusing. 
What's going to happen in the implementaton scenario file is the following.
```
# Redefine network-related settings to use the isolated network bridge
VM_BRIDGE_IP="$(get_vm_bridge_ip "${VM_ISOLATED_NETWORK}")"
WEB_SERVER_URL="http://${VM_BRIDGE_IP}:${WEB_SERVER_PORT}"

scenario_create_vms() {
    prepare_kickstart host1 kickstart.ks.template rhel-9.2-microshift-source
    # Use the isolated network when creating a VM
    launch_vm host1 "" "${VM_ISOLATED_NETWORK}"
    verify_no_internet "host1"
}
```

Note that I'm providing the network argument to the `launch_vm` function, so it's going to use the same ISO, just different boot parameters. I hope it answers your questions.

test/bin/manage_hypervisor_config.sh
ggiguash: I created the https://issues.redhat.com/browse/USHIFT-1565 issue to have a discussion and make this change in all the scripts if we decide it's a better way to go.

test/README.md
```diff
@@ -138,8 +136,11 @@ In order to build different images, Image Builder needs to be configured
 with all of the right package sources to pull the required RPMs. The
 sources used by all scenarios are in the `./test/package-sources` directory.
 
-Package source definition files are templates using `envsubst`,
-which means the files can have shell variables embedded.
+Package source definition files are templates that support embedded shell
+variables. Image build procedure expands those variables using `envsubst`.
```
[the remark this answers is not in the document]
ggiguash: The statements are similar, but they're taking about `envsubst` of different things:
- Package source definition files
- Image blueprint definition files

PR #2187, opened 2023-08-08
WIP ci_phase_iso_build.sh: add timestamp to PS4

pmtk:


pmtk:
Let's see what's taking so long

PR #2184, opened 2023-08-07
USHIFT-1544 Reorganize testing scenarios

pmtk:
Reorganization:
- drop `microshift-source` from scenario filenames
- shorten `rhel-9.2` to `el92`
- introduce MicroShift-distinct-source tag in filenames: `src`, `base`, `prev`
- rename scenarios and suites to start with "what we're testing" (e.g. upgrade, downgrade) and then "what flow" to group scenarios together
- added "failed upgrade" scenarios with end up with host rolling back:
  - `src` to `src` due to greenboot failure (previously reason was MicroShift not being able to create a backup, which is kept as `el92_src_upgrade-failing-cannot-backup.sh`)
  - `base` to `src`
  - `prev` to `src` (which we might postpone until next release because 4.13 will probably run with 4.14 data, but it's "happy accident")

PR #2190, opened 2023-08-08
[USHIFT-1108] Add periodic scenario to test upgrade from latest release to head of next y-stream branch or main

copejon:
This PR is part of the implementation of [USHIFT-1108](https://issues.redhat.com//browse/USHIFT-1108), the rest of the work being in the openshift/release repo.

Changes here tried to reuse existing logic as much as possible.  
- A ./test/Makefile target is added to build the 4.y-1 branch.
- The PREVIOUS_MINOR_VERSION is reused.  More var related to building prior release were added where they were needed to complete the pipeline.
-

### DIFF
--- a/test/scenarios-periodics/el92-src@upgrade-from-release-fails-and-rollsback.sh
+++ b/test/scenarios-periodics/el92-src@upgrade-from-release-fails-and-rollsback.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+
+scenario_create_vms() {
+    prepare_kickstart host1 kickstart.ks.template rhel-9.2-microshift-4."$(previous_minor_version)"
+    launch_vm host1
+}
+
+scenario_remove_vms() {
+    remove_vm host1
+}
+
+# CI will execute these tests on a set release branch, so source version will always be the latest code for that branch.
+# I.e. CI jobs testing release-4.14 will always compile the latest 4.14 code as the source_version.
+scenario_run_tests() {
+    run_tests host1 \
+        --variable "FAILING_REF:rhel-9.2-microshift-source" \
+        --variable "REASON:fail_greenboot" \
+        suites/upgrade/upgrade-fails-and-rolls-back.robot
+}


test/suites-ostree/release-y-stream-upgrade-and-rollback.robot
```diff
@@ -0,0 +1,39 @@
+*** Settings ***
```
copejon, line 1: Work is currently stuck on failures in the build phase.  This file was copied from another, but hasn't been touched yet.

test/Makefile
```diff
@@ -35,6 +35,17 @@ build-base-branch:
 	mv $(TMP_REPO)/_output/rpmbuild $(REPO)/_output/rpmbuild-base
 	rm -rf "$(TMP_REPO)"
 
+
+.PHONY: build-previous-minor-version
+build-previous-minor-version: YMINUS1_MINOR =release-$(MAJOR).$(shell expr $(MINOR) - 1)
+build-previous-minor-version: TMP_REPO !=mktemp -d
+build-previous-minor-version:
+	if ! [ -d "$(REPO)/_output" ]; then mkdir $(REPO)/_output; fi
+	git clone --single-branch --branch $(YMINUS1_MINOR) https://github.com/openshift/microshift.git $(TMP_REPO)
```
pmtk, line 44: Why do we build previous minor from code? So far we were using already built and released 4.13 RPMs - I think we should continue on doing that

copejon: Good point. I'll TAL at that approach and work it in here

test/suites-ostree/healthy-upgrade-with-forced-rollback.robot
```diff
@@ -0,0 +1,71 @@
+*** Settings ***
```
dhellmann, line 1: I think https://github.com/openshift/microshift/blob/main/test/suites-ostree/staged-failed-upgrade.robot tests this, too. How are the suites different? Do we need both?

copejon: Staged-failed-upgrade is really similar. The difference is that staged-failed-upgrade forces the rollback without verifying the health of the upgraded os, which is half of what we want to ensure.

dhellmann: We phrased some of the tickets for tests imprecisely, so I want to make sure that's a real requirement. I worry the scenario might turn out flakey if we try to require the initial boot to be healthy, but if there's a specific reason we need to require that for it to be a valid test we can just keep an eye on it.

cc @pmtk

copejon: As I understood them, the test should deploy the y-1 release, upgrade to release-4.y, verify health, rollback, and again verify health. Is that correct?

Also, IIUC, an unhealthy state will block an upgrade. Is that accurate?

copejon:
e2e and conformance are timing out waiting for the cluster to deploy

/retest

test/bin/scenario.sh
```diff
@@ -343,7 +343,7 @@ run_tests() {
         if [ ! -f "${f}" ]; then
             error "Cannot read ${f}"
             exit 1
-        fi
+            fi
```
ggiguash, line 346: Is this an accidental change?

copejon: darn, yes. will fix

test/suites-ostree/healthy-upgrade-with-forced-rollback.robot
dhellmann: That's not quite the scenario described in USHIFT-1108. The goal is to test that rollback works from y+1 to y. So we need y to be healthy, then upgrade to y+1, then trigger a rollback regardless of how the upgrade goes, then verify y is healthy again.

That other RF suite file does all of that.

copejon:
The upgrade test is succeeding but the cncf suites is consistently failing.

/retest-required

copejon:
conformance-reduced is also seeing some timeouts within the openshift-tests execution

copejon:
rebased, hoping that solves this

test/suites-ostree/healthy-upgrade-with-forced-rollback.robot
pmtk: I don't think `staged-failed-upgrade.robot` rolls back: it fails on first boot of new deployment, but on second it becomes healthy.

It's `failed-upgrade.robot` that consistently fails, causing system to roll back.
However, currently the test does that by preventing backup.
In https://github.com/openshift/microshift/pull/2184 I change that to have two separate ([1](https://github.com/openshift/microshift/pull/2184/files#diff-f7bef174572b9938e8a98a36522c46782a66013bd4ce4c77b2363f536731225f), [2](https://github.com/openshift/microshift/pull/2184/files#diff-23f9418f94001d3da07283921bc1cdb8a252f2a99b4ac6b18f9a54799186bbb1)) scenarios based on [single RF file with a reason variable](https://github.com/openshift/microshift/pull/2184/files#diff-e6691d89f9a5bc89929fe7c6f5561b29f52c2aae793a7d21598e0420715b3059)

copejon:
rhel-9.2-microshift-source-failed-staged is failing in the meta-tests job.  Looks like the cluster fails to come back after reboot. Rebased to see if there's a recent fix.

test/suites-ostree/failed-upgrade.robot
```diff
@@ -30,7 +30,9 @@ Staged Deployment Consistently Fails To Back Up The Data
     TestAgent.Add Action For Next Deployment    every    prevent_backup
```
pmtk, line 30: Please take a look at [current version of this file](https://github.com/openshift/microshift/blob/main/test/suites-ostree/failed-upgrade.robot#L31). I think you need to rebase these changes

test/suites-ostree/failed-upgrade.robot
```diff
@@ -31,7 +31,8 @@ New Deployment Is Consistently Unhealthy And Rolls Back
     TestAgent.Add Action For Next Deployment    every    ${REASON}
     Deploy Commit Expecting A Rollback    ${FAILING_REF}
     Wait For MicroShift Service
-    Backup Should Exist    ${backup}
+    # backup logic is only present in 4.14+
+    IF    ${backup} != "4.13"    Backup Should Exist    ${backup}
```
pmtk, line 35: Looks like CI fails because:
```
Evaluating IF condition failed: Evaluating expression 'rhel-16a972f4e385541d7cdc2c4e9071ce6032f1042753eaab53e2a1eca084aa996c.0_fc34f14811a846b7a91a111f9dfab4ad != "4.13"' failed: SyntaxError: invalid syntax (<string>, line 1)
```
Maybe we need to wrap the var into quotes.

However, I think we can drop "Backup Should Exist" only in case of 4.13 & "prevent backup".
For `4.14`, even if the 2nd deployment fails to create a backup, it should be created after rolling back.


```suggestion
    # backup logic is only present in 4.14+
    IF    "${backup}" != "4.13" and "${REASON}" != "prevent_backup"    Backup Should Exist    ${backup}
```

BTW. I see I missed the filename of https://github.com/openshift/microshift/blob/main/test/scenarios/el92-src%40upgrade-fails-cannot-backup.sh, it should be "upgrade-fails.sh" without cannot-backup...

test/scenarios-periodics/rhel-9.2-microshift-release-forced-source-rollback.sh
```diff
@@ -0,0 +1,21 @@
...
+    remove_vm host1
+}
+
+# CI will execute these tests on a set release branch, so source version will always be the latest code for that branch.
+# I.e. CI jobs testing release-4.14 will always compile the latest 4.14 code as the source_version.
+scenario_run_tests() {
+    run_tests host1 \
+        --variable "FAILING_REF:rhel-9.2-microshift-source" \
+        --variable "REASON:fail_greenboot" \
+        suites-ostree/failed-upgrade.robot
```
pmtk, line 20: Yesterday a PR was merged that changed the filenames
```suggestion
        suites-ostree/upgrade-fails-and-rolls-back.robot
```

copejon:
/hold
not to self, remember to squash

copejon:
Greenboot complaining that pods are restarting too frequently in openshift-ingress namespace.

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/2190/pull-ci-openshift-microshift-main-microshift-rpm-upgrade/1694578630484561920#1:build-log.txt%3A764

/retest

copejon:
microshift-rpm-upgrade test is flaky.  A fix is posetd at #2255

copejon:
/unhold

squashed commits

ggiguash:
/hold
@pmtk , can we run this test as part of another existing upgrade scenario?
Otherwise, please unhold to merge this test.

pmtk:
/unhold

@ggiguash I don't think we have necessary functionality for reusing VMs yet.
I think we don't have to worry about this scenario - it's periodic

PR #2193, opened 2023-08-09
USHIFT-1555: add timeout for long running command

eslutsky:
This change safeguards the  wait_for function  against long running commands.
if the  command is not finished within the arg_1 `timeout`  , command (arg_2)  will be killed   and `return 1`.


ggiguash:
These tests are not relevant for the change. Overriding them to save CI cycles
/override ci/prow/microshift-e2e

PR #2194, opened 2023-08-09
USHIFT-1551: Select a host IP as nodeIP when no default route

pliurh:


pkg/sysconfwatch/sysconfwatch_linux.go
```diff
@@ -103,16 +103,22 @@ func (c *SysConfWatchController) Run(ctx context.Context, ready chan<- struct{},
 		select {
 		case <-ticker.C:
 			// Check the IP change
-			currentIP, _ := util.GetHostIP()
-			if c.NodeIP != currentIP {
-				klog.Warningf("IP address has changed from %q to %q, restarting MicroShift", c.NodeIP, currentIP)
+			currentIP, err := util.GetHostIP()
+			if err != nil {
+				klog.Warningf("cannot find an host IP: %v", err)
 				os.Exit(0)
```
dhellmann, line 109: If there's an error, we should probably exit with an error code.

pkg/util/net.go
```diff
@@ -142,3 +146,37 @@ func addNoProxyEnvVarEntries(entries map[string]struct{}, envVar string) {
...
+		if err != nil {
+			klog.Warningf("failed to get IPs for interface %s: %v", i.Name, err)
+			continue
+		}
+
+		for _, addr := range addrs {
+			var ip tcpnet.IP
+			switch v := addr.(type) {
+			case *tcpnet.IPAddr:
+				// ignore IPv6
```
dhellmann, line 168: This is OK for now, but at some point we won't be able to ignore IPv6.

pkg/util/net.go
```diff
@@ -142,3 +146,37 @@ func addNoProxyEnvVarEntries(entries map[string]struct{}, envVar string) {
...
 	}
 }
+
+func selectV4IPFromHostInterface() (string, error) {
+	ifaces, err := tcpnet.Interfaces()
+	if err != nil {
+		return "", err
+	}
+
+	// get list of interfaces
```
dhellmann, line 156: Are these lists sorted in anyway? The loops return the first time they find any IP at all, so a host that has several might get different values at different times if the list of interfaces and then addresses on the interface are not sorted. I don't know which direction we would want to sort, but we do want the results to be deterministic.

pkg/sysconfwatch/sysconfwatch_linux.go
```diff
@@ -103,16 +103,22 @@ func (c *SysConfWatchController) Run(ctx context.Context, ready chan<- struct{},
...
 				return nil
+			} else {
```
ggiguash, line 111: A nit: No need for else here?

pkg/util/net.go
ggiguash: We have a separate epic for supporting IPv6, so we can handle it then.

pkg/util/net.go
ggiguash: Yes, I agree. Since we are implementing a fallback in case of non-existing default route, we may only want to use it in case we have one and only IPv4 address. Otherwise, we're risking non-deterministic behavior.

pliurh: This list is sorted by the index of interfaces, from little to big. The result is deterministic if no new netdevice is added.

pkg/sysconfwatch/sysconfwatch_linux.go
```diff
@@ -103,16 +103,21 @@ func (c *SysConfWatchController) Run(ctx context.Context, ready chan<- struct{},
...
+			currentIP, err := util.GetHostIP()
+			if err != nil {
+				klog.Warningf("cannot find an host IP: %v", err)
+				os.Exit(1)
+				return nil
+			}
 			if c.NodeIP != currentIP {
 				klog.Warningf("IP address has changed from %q to %q, restarting MicroShift", c.NodeIP, currentIP)
-				os.Exit(0)
+				os.Exit(1)
```
dhellmann, line 114: If we automatically detect the default node IP differently in a way that allows that IP to change when the host's network settings change, then when the host goes on or off of a network MicroShift will exit when the default route IP changes and then fail to start, exiting here, because the previous node IP does not match the "current" IP value.

We have a way for the user to specify the node IP using the configuration file already. Can they use that setting with an IP that does not have the default route? Will kubelet accept that IP as the node IP?

If so, do we need to change the way we automatically detect the node IP? Maybe I'm misunderstanding the intent of the change?

pliurh: This patch doesn't fix the conflict between the configured NodeIP and the detected one. It's only for fixing [USHIFT-1551](https://issues.redhat.com//browse/USHIFT-1551).

> We have a way for the user to specify the node IP using the configuration file already. Can they use that setting with an IP that does not have the default route? Will kubelet accept that IP as the node IP?

Kubelet can use any non-loopback host IP as the NodeIP. The problem is on the `sysconfigwatch` side. If the IP detected by `sysconfigwatch` differs from the NodeIP in the Microshift config file, `sysconfigwatch` will always restart microshift.

I can update the PR to also fix the `NodeIP` flag issue.

dhellmann: > I can update the PR to also fix the `NodeIP` flag issue.

Let me re-read this today, now that I see your answer. I may have misunderstood what you were trying to do.

pliurh: I've added the missing part. Now if users config `NodeIP` flag, `sysconfwatch` will only check the existence of the IP on the host. If yes, the provided NodeIP will always be used. Otherwise, Microshift will exit.

pkg/config/config.go
```diff
@@ -35,7 +35,7 @@ type Config struct {
 
 	// Internal-only fields
 	Ingress      IngressConfig `json:"-"`
-	userSettings *Config       `json:"-"` // the values read from the config file
+	UserSettings *Config       `json:"-"` // the values read from the config file
```
dhellmann, line 38: I'm a little worried about exposing this entire data structure. It was meant to be an implementation detail. Let's continue with this approach for now, but we probably want to refactor it once we have the basic code working (we can do that in another PR so we don't hold up progress).

I see a few ways to keep the details hidden.

1. Add a public string field with a name like `UserNodeIP` and fill the value in during `incorporateUserSettings()`.
2. Add a method with a name like `UserNodeIP` and have it look inside the `userSettings` structure and return `""` if there is no node IP specified in the configuration file.
3. Add a public bool field with a name like `UserSetNodeIP` and fill it in during the `incorporateUserSetting()`. That function already copies the user's nodeIP value from the config file struct to the real struct.

dhellmann:
The unit tests are failing with a compilation error because of the field rename

```
# github.com/openshift/microshift/pkg/config [github.com/openshift/microshift/pkg/config.test]
pkg/config/config_test.go:248:12: config.userSettings undefined (type *Config has no field or method userSettings, but does have UserSettings)
```

dhellmann:
I think the general approach is good. I know you have other things to do, so I can take on the refactoring if you don't have time.

pliurh:
@dhellmann I refactored the patch as the second approach you suggested. PTAL.

pkg/util/net.go
```diff
@@ -34,16 +34,33 @@ import (
...
+	var hostIP string
+	var err error
+
+	if nodeIP != "" {
+		klog.Infof("try to find nodeIP %s on host", nodeIP)
+		hostIP, err = selectV4IPFromHostInterface(nodeIP)
+		if err != nil {
+			return "", fmt.Errorf("failed to find nodeIP %s on host: %v", nodeIP, err)
+		}
+		goto found
```
dhellmann, line 47: Oh, my, a goto?!  ;-)

That probably could have been done with a defer, but this is clear, too.

pkg/util/net.go
```diff
@@ -34,16 +34,33 @@ import (
...
+			return "", err
+		}
+	}
+
+found:
 	if hostIP != previousGatewayIP {
 		previousGatewayIP = hostIP
-		klog.V(2).Infof("host gateway IP address: %s", hostIP)
 	}
+	klog.V(2).Infof("host gateway IP address: %s", hostIP)
```
dhellmann, line 63: I think we moved this line up inside the `if` block where it was because we saw the same message, with the same IP, every time the check ran and it made the logs very noisy. I feel like we had a bug ticket for that change a while back.

dhellmann (commented):
This looks good. I had one comment about logging, but we can address that in a follow-up so we can keep moving forward.

/lgtm

pkg/util/net.go
pliurh: I used to use an `if / else` here. But it cannot pass our nestedif check. So...

PR #2202, opened 2023-08-11
OCPBUGS-11829: rebase + react to changes from route-controller-manager 

atiratree:
- remove leader election which fixes https://issues.redhat.com/browse/OCPBUGS-11829
- add system:openshift:authentication-reader-route-controller-manager
  rolebinding for initializing delegating authentication, otherwise we are getting following error when starting the server

```
service route-controller-manager exited with error: error initializing delegating authentication: unable to load configmap based request-header-client-ca-file: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:openshift-route-controller-manager:route-controller-manager-sa" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```


pkg/controllers/openshift-route-controller-manager.go
```diff
@@ -88,15 +84,46 @@ func (s *OCPRouteControllerManager) writeConfig() *openshiftcontrolplanev1.OpenS
...
+		WithComponentNamespace(namespace).
+		// Without an explicit owner reference, the builder will try using POD_NAME or the
+		// first pod in the target namespace (and fail because we have no pod).
+		WithComponentOwnerReference(&corev1.ObjectReference{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "Namespace",
+			Name:       namespace,
+			Namespace:  namespace,
+		}).
+		WithServer(*rcmConfig.ServingInfo, v1alpha1.DelegatedAuthentication{}, v1alpha1.DelegatedAuthorization{})
```
atiratree, line 113: not sure if microshift needs the server, but it is serving `metrics` and `healthz` endpoints

scripts/auto-rebase/assets.yaml
```diff
@@ -149,20 +149,21 @@ assets:
...
-      - file: route-controller-manager-leader-rolebinding.yaml
-      - file: route-controller-manager-leader-role.yaml
       - file: route-controller-manager-ns.yaml
       - file: route-controller-manager-sa.yaml
       - file: route-controller-manager-separate-sa-rolebinding.yaml
       - file: route-controller-manager-separate-sa-role.yaml
       - file: route-controller-manager-tokenreview-clusterrolebinding.yaml
       - file: route-controller-manager-tokenreview-clusterrole.yaml
+      - file: route-controller-manager-authentication-reader-rolebinding.yaml
+        ignore: "required for initializing delegating authentication when starting a server"
```
atiratree, line 166: is this the correct way how to introduce a new custom RBAC?

assets/controllers/route-controller-manager/route-controller-manager-authentication-reader-rolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: kube-system
+  name: system:openshift:authentication-reader-route-controller-manager
+roleRef:
+  kind: Role
+  name: extension-apiserver-authentication-reader
```
dhellmann, line 8: Where does this role come from? It looks like it's built into some of the kube code?

pkg/controllers/openshift-route-controller-manager.go
```diff
@@ -46,7 +52,7 @@ const (
 
 func NewRouteControllerManager(cfg *config.Config) *OCPRouteControllerManager {
 	s := &OCPRouteControllerManager{}
-	s.configure(cfg)
+	s.configErr = s.configure(cfg)
```
dhellmann, line 55: It looks like maybe we should change the API of the constructors to allow returning errors. I'm surprised none of the others have errors they might return. We'll deal with that separately from this PR.

pkg/controllers/openshift-route-controller-manager.go
dhellmann: I think we'd want healthz now, and at some point we'll want metrics, too. So I think it's OK to leave it in.

Can we control the port used? We're likely to have a couple of endpoints like this for embedded components, and I would either want to have them on separate ports or merge the results so we only need to expose one endpoint for metrics and one for health.

dhellmann (commented):
Thanks, @atiratree !

assets/controllers/route-controller-manager/route-controller-manager-authentication-reader-rolebinding.yaml
atiratree: yes, it comes from kube: https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/namespace_policy.go#L77

scripts/auto-rebase/assets.yaml
atiratree: thanks for the explanation, hopefully it will be better now (but have not tested it...)

pkg/controllers/openshift-route-controller-manager.go
atiratree: ok, yeah the port can be changed in the config

dhellmann:
I'm not sure what was leading to the e2e test failures. I've downloaded the PR locally and MicroShift starts up fine, but it looks like in CI the system health check is timing out.

pkg/controllers/openshift-route-controller-manager.go
atiratree: I have coppied this concept from cluster-policy-controller so there are at least two places  :) . +1 for changing the API for this

pmtk:
/lgtm
/hold

Looks good to me, but maybe someone else wants to take a look

PR #2203, opened 2023-08-13
USHIFT-1555: add timeout for long running command

eslutsky:
This change safeguards the wait_for function against long running commands.
if the command is not finished within the arg_1 timeout , command (arg_2) will be killed and return 1.

### DIFF
--- a/packaging/greenboot/functions.sh
+++ b/packaging/greenboot/functions.sh
@@ -88,6 +88,7 @@ function get_wait_timeout() {
     echo "${wait_timeout}"
 }
 
+
 # Run a command with a second delay until it returns a zero exit status
 #
 # arg1: Time in seconds to wait for a command to succeed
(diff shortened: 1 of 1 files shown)


packaging/greenboot/functions.sh
```diff
@@ -98,7 +98,8 @@ function wait_for() {
     shift 1
 
     local -r start=$(date +%s)
-    until ("$@"); do
+    until (timeout "${timeout}" bash -x -c $@); do
```
ggiguash, line 101: Since this code runs in production, I think we may want to remove `bash -x`. If the caller wants tracing, they will pass `bash -x` as part of the command.

packaging/greenboot/microshift-running-check.sh
```diff
@@ -103,6 +103,9 @@ function any_pods_running() {
 # Main
 #
 
+# export function for subshell access
+export -f microshift_service_active
```
ggiguash, line 107: While this provides a solution for the current problem, the `functions.sh` code is generic, so this local solution is not generic enough. 

While, it is possible to write a generic code in BASH checking if a string is a function and it's exported, I'm not sure if we can add this code in `functions.sh` because we do not know if some part of the command is an unexported function.
```
if type "string_name" | grep -q 'function'; then
  echo "It is a function."
else
  echo "Not a function."
fi

if declare -F | grep -q 'function_name'; then
  echo "The function is exported."
else
  echo "The function is not exported."
fi
```

ggiguash: Can we think about some other way to implement the timeout that does not complicate the function handling?

ggiguash: Would something like this be better?
```
command_to_run &
pid=$!
sleep 5s && kill -HUP $pid &   # 5 seconds timeout
wait $pid
```

eslutsky: > 

this wont work, because `wait `  only work if that pid is a child process of the current process,and timeout is invoked in a new shell.
```
function test(){ echo "calling function"; sleep 5; echo done; }
{ test & pid=$!; echo "waiting for ${pid}"; timeout 30 wait ${pid}; }

wait: pid 2865116 is not a child of this shell
```

but we can do something like  this:
```
function test(){ echo "calling function"; sleep 5; echo done; }
{ test & pid=$!; echo "waiting for ${pid}"; timeout 30 tail --pid=$pid -f /dev/null; }
```

ggiguash:
@eslutsky , I wonder if we should simply add a sensible timeout to the `oc` command options [here](https://github.com/openshift/microshift/blob/main/packaging/greenboot/functions.sh#L10)

eslutsky:
> @eslutsky , I wonder if we should simply add a sensible timeout to the `oc` command options [here](https://github.com/openshift/microshift/blob/main/packaging/greenboot/functions.sh#L10)?

i didnt see that OCGET_CMD is using the wait_for function , so its not related to this change ..
anyway i'm not very comfortable with with the current `tail` implementation , and it appears there is no a `cleaner` way to implement this without exposing the functions.

eslutsky:
closing for now

PR #2208, opened 2023-08-14
USHIFT-1580: Consolidate cleanup of composer / hypervisor and optimize VM boot sequence

ggiguash:
- Consolidate devenv and hypervisor cleanup procedures
- Add system status information to the CI boot script for debugging
- Added delay of 5 seconds between subsequent VM startups

Closes [USHIFT-1580](https://issues.redhat.com//browse/USHIFT-1580)


test/bin/ci_phase_test.sh
```diff
@@ -14,6 +14,8 @@ if [ ! -d "${RF_VENV}" ]; then
     "${ROOTDIR}/scripts/fetch_tools.sh" robotframework
 fi
 
+bash -x ./bin/manage_vm_connections.sh local
```
dhellmann, line 17: This is no longer necessary. The scenario script does the same thing as it builds each VM.

dhellmann (commented):
Every time we rename these scripts it breaks up my workflow. I'd _really_ like to just settle on the names we have and improve the scripts in place for the future.

/lgtm

test/bin/ci_phase_iso_boot.sh
```diff
@@ -25,6 +25,9 @@ cd "${ROOTDIR}"
 # other scripts use virsh.
 bash -x ./scripts/devenv-builder/manage-vm.sh config
 
+# Clean up the image builder cache to free disk for virtual machines
+bash -x ./scripts/image-builder/cleanup.sh -full
```
eslutsky, line 29: does it contain `bash -x`  for debugging only?

test/bin/ci_phase_iso_build.sh
```diff
@@ -52,6 +52,6 @@ MAX_WORKERS=5
 CUR_WORKERS="$( [ "${CPU_CORES}" -lt  $(( MAX_WORKERS * 2 )) ] && echo $(( CPU_CORES / 2 )) || echo ${MAX_WORKERS} )"
 
 bash -x ./bin/start_osbuild_workers.sh "${CUR_WORKERS}"
-time bash -x ./bin/build_images.sh
+bash -x ./bin/build_images.sh
```
eslutsky, line 55: does it contain bash -x for debugging only?

test/bin/ci_phase_iso_boot.sh
ggiguash: Yes, we do this for most of the scripts we launch in CI. Otherwise, debugging requires applying special techniques.

ggiguash:
> Every time we rename these scripts it breaks up my workflow. I'd _really_ like to just settle on the names we have and 
> improve the scripts in place for the future.

Sorry for that. I think we now have consistent naming and clear separation between devenv and hypervisor scripts.

dhellmann (commented):
/lgtm

PR #2209, opened 2023-08-14
USHIFT-1578: Optimize virt-install and greenboot sequences in scenario.sh

ggiguash:
- Fix VM boot timeout by using --wait timeout option instead of timeout
- Implement virt-install retries due to https://github.com/virt-manager/virt-manager/issues/498
- Moved greenboot check from the launch to test phase

**Tests on c5n.metal at eu-west-1b**
|Phase|Greenboot in launch|Greenboot in test|
|---------|------------------------------|-------------------------|
|boot   |16m                             |9m                         |
|test     |20m                             |20m                       |

Closes [USHIFT-1578](https://issues.redhat.com//browse/USHIFT-1578)

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -177,17 +177,14 @@ EOF
 #                    should be used to boot the VM. This is _not_
 #                    necessarily the image to be installed (see
 #                    prepare_kickstart).
-# network_name -- The name of the network used when creating the VM.
+#  network_name -- The name of the network used when creating the VM.
 launch_vm() {
-    local vmname="$1"
-    local boot_blueprint="${2:-${DEFAULT_BOOT_BLUEPRINT}}"
-    local network_name="${3:-default}"
+    local -r vmname="$1"
+    local -r boot_blueprint="${2:-${DEFAULT_BOOT_BLUEPRINT}}"
+    local -r network_name="${3:-default}"
 
-    local full_vmname
-    local kickstart_url
-
-    full_vmname="$(full_vm_name "${vmname}")"
-    kickstart_url="${WEB_SERVER_URL}/scenario-info/${SCENARIO}/vms/${vmname}/kickstart.ks"
+    local -r full_vmname="$(full_vm_name "${vmname}")"
+    local -r kickstart_url="${WEB_SERVER_URL}/scenario-info/${SCENARIO}/vms/${vmname}/kickstart.ks"
 
     # Checking web server for kickstart file
     if ! curl -o /dev/null "${kickstart_url}" >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


test/bin/scenario.sh
```diff
@@ -196,21 +193,46 @@ launch_vm() {
...
+
+            # Check if the command exited within 15s due to a failure
+            local vm_create_end
+            vm_create_end=$(date +%s)
+            if [ $(( vm_create_end -  vm_create_start )) -lt 15 ] ; then
+                echo "Error running virt-install on attempt ${attempt}: retrying in 5s"
+                sleep 5
+                continue
+            fi
+            # Stop retrying on timeout error
```
dhellmann, line 227: How is the timeout induced? There's no longer a `timeout` on line 207.

test/bin/scenario.sh
```diff
@@ -241,35 +263,27 @@ launch_vm() {
         echo "${PUBLIC_IP}" > "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/public_ip"
     else
         echo "${ip}" > "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/public_ip"
-        # Set the defaults for the various ports so that connections
-        # from the hypervisor to the VM work.
-        echo "22" > "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/ssh_port"
-        echo "6443" > "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/api_port"
-        echo "5678" > "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/lb_port"
     fi
+    # Set the defaults for the various ports so that connections
```
dhellmann, line 267: These values are only correct when PUBLIC_IP is not set. They need to move back into the else clause above.

test/bin/scenario.sh
ggiguash: The timeout is induced by the `--wait  ${vm_wait_timeout}` option.

test/scenario_settings.sh.example
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 
 # Set the PUBLIC_IP to the IP address used to access the hypervisor.
+# Enable this setting to connect to VMs via the specified IP using
+# port-forwarded connections. Otherwise, IP addresses assigned by
+# libvirt are used for accessing VMs.
+# hypervisor is used by
```
dhellmann, line 7: I think this line is a copy and paste error, but I'm not sure.

test/bin/scenario.sh
dhellmann: Ah, I missed that you added the time argument there.

dhellmann (commented):
I'm happy with this version. Let's let some of the dependencies merge first. That will also give the rest of the team a chance to  look at the changes.

test/bin/scenario.sh
ggiguash: Both me and Patryk saw cases when `virt-install` was not killed by timeout, so we thought it would be better to use the native functionality.

PR #2212, opened 2023-08-15
USHIFT-1564: Selinux update

eggfoobar:
This PR does a small change to our existing SELinux policy, we're still missing some context around some of these folders but given our proximity to feature freeze, this PR errs on the side of caution and only modifies things we know to do safely and move us closer to OCP policy.

Note, this PR will be followed by more changes given our engagement with the SELinux team. ([see](https://issues.redhat.com/browse/SELINUX-3165))

1. Keep more inline with OCP policy, remove allowing `container_t` to read `container_var_lib_t` also see article on [defaults](https://danwalsh.livejournal.com/81269.html)
2. `/etc/microshift` is in spirit similar to what `/etc/kubernetes` is and therefore we are applying the same label to that folder
3. Add missing labels to `/var/lib/microshift-backups` 
4. `/var/lib/microshift-backups` should not be accessed by containers and as such we are applying the `container_var_lib_t` following the change in number 1.


eggfoobar:
/cc @wrabcak 
for visibility

wrabcak:
@eggfoobar, Do you see any SELinux denials with the current SELinux policy ? As you're trying to expand the existing SELinux policy for container_runtime_t and container_t so I understand it as no new policy from scratch specifically for the Microshift is needed?

FYI @zpytela @5umm3r15

packaging/selinux/microshift.fc
```diff
@@ -1,6 +1,8 @@
 # vim: sw=8:ts=8:et
-/var/run/secrets/kubernetes.io/serviceaccount(/.*)?				gen_context(system_u:object_r:container_runtime_tmpfs_t,s0)
-/var/lib/microshift/certs/ca-bundle(/.*)?                                       gen_context(system_u:object_r:container_file_t,s0)
-/usr/local/bin/microshift                                                  --	gen_context(system_u:object_r:container_runtime_exec_t,s0)
-/usr/bin/microshift                                                        --	gen_context(system_u:object_r:container_runtime_exec_t,s0)
-/var/hpvolumes(/.*)?            						gen_context(system_u:object_r:container_file_t,s0)
+/var/run/secrets/kubernetes.io/serviceaccount(/.*)?             gen_context(system_u:object_r:container_runtime_tmpfs_t,s0)
+/var/lib/microshift/certs/ca-bundle(/.*)?                       gen_context(system_u:object_r:container_file_t,s0)
+/var/lib/microshift-backups(/.*)?                           	gen_context(system_u:object_r:container_var_lib_t,s0)
```
dhellmann, line 4: I see the certs paths on line 3 are given different settings. Do we need to worry about that for backups, too, or will the files get the right settings when they're restored to the running location?

Do we need to worry that the backups are reflink copies of the original files? I think if the files are changed the links are broken and new data is written, but does the reflink allow _reading_ data that might not otherwise be readable?

Do we need other files in /var/lib/microshift to have specific contexts?

eggfoobar: >I see the certs paths on line 3 are given different settings. Do we need to worry about that for backups, too, or will the files get the right settings when they're restored to the running location?

Ah good call out, let me verify the restore for the context, given that we didn't have a transition rule prior I don't know if that desired context was even being applied during restore before this change.

> Do we need to worry that the backups are reflink copies of the original files? I think if the files are changed the links are broken and new data is written, but does the reflink allow reading data that might not otherwise be readable?

Regarding the reflink, given that SELinux operates with just labels I don't know if reflinks change at all. We're just enforcing who can read what files, so here we're saying that containers can't read files in microshift-backups. Right now I don't see any policy denials in our backup and restore test or our failed staged test (which does a restore).

> Do we need other files in /var/lib/microshift to have specific contexts?

I don't even fully think we need to have the granularity we currently do around the certs bundle, that might be a hold out but I don't have the full context on that one just yet. I think we'll need to add granularity as we identify situations where access needs to be granted, but given the current policy doesn't seem to be causing denials, I'm not sure it's needed yet.

eggfoobar:
> @eggfoobar, Do you see any SELinux denials with the current SELinux policy ? As you're trying to expand the existing SELinux policy for container_runtime_t and container_t so I understand it as no new policy from scratch specifically for the Microshift is needed?
> 
> FYI @zpytela @5umm3r15

Hey @wrabcak, currently in all our tests we generate an `sos` report, in there we do collect the `/var/log/audit/audit.log` file and currently we're seeing denials for things like `firewalld` and `modprobe`. However, that denial was happening prior to this change as well so it doesn't seem to be critical.

```
type=AVC msg=audit(1692083252.243:50): avc:  denied  { getattr } for  pid=844 comm="firewalld" path="/usr/sbin/xtables-nft-multi" dev="tmpfs" ino=6 scontext=system_u:system_r:firewalld_t:s0 tcontext=system_u:object_r:init_var_run_t:s0 tclass=file permissive=0
...
type=AVC msg=audit(1692083252.508:56): avc:  denied  { search } for  pid=1090 comm="modprobe" name="events" dev="tracefs" ino=1027 scontext=system_u:system_r:openvswitch_load_module_t:s0 tcontext=system_u:object_r:tracefs_t:s0 tclass=dir permissive=0
```

packaging/selinux/microshift.te
```diff
@@ -4,11 +4,21 @@ type microshift_t;
...
 
+# When microshift creates backup folders in /var/lib/microshift-backups, the correct labels are applied
+# Note: filetrans_pattern rules states;
+#   Process running as `container_runtime_t` that creates a `dir` called `microshift-backups` 
+#   in a folder labeled `var_lib_t` get's the label `container_var_lib_t`
+filetrans_pattern(container_runtime_t, var_lib_t, container_var_lib_t, dir, "microshift-backups")
+
+# greenboot-healthcheck runs as an bin_t context, here we make sure the correct label is applied when 
+# greenboot creates the health.json file.
+filetrans_pattern(bin_t, var_lib_t, container_var_lib_t, dir, "microshift-backups")
```
jerpeter1, line 20: This is the same rule as the previous one, except applied to processes running in the bin_t domain - since health.json will inherit the label of it's parent, this is fine, but the comment is a little misleading, since it sounds like you intend to make a filetrans_pattern that applies to the health.json file specifically.

packaging/selinux/microshift.fc
```diff
@@ -1,6 +1,8 @@
...
+/etc/microshift(/.*)?                                       	gen_context(system_u:object_r:kubernetes_file_t,s0)
+/usr/local/bin/microshift                                   --	gen_context(system_u:object_r:container_runtime_exec_t,s0)
+/usr/bin/microshift                                         --	gen_context(system_u:object_r:container_runtime_exec_t,s0)
```
jerpeter1, line 7: Have we determined why `container_runtime_exec_t` appropriate for microshift, when e.g. kubelet is unconfined on OCP?

packaging/selinux/microshift.te
```diff
@@ -13,9 +13,6 @@ allow container_t container_runtime_tmpfs_t:file read_file_perms;
 allow container_t container_runtime_tmpfs_t:lnk_file read_file_perms;
```
jerpeter1, line 23: Removing the unwanted permissions is good, but have we confirmed that these other permissions (e.g. `allow container_t container_runtime_tmpfs_t:file read`), are actually being used? Given our investigations last week, I would have expected needing additional permissions (e.g. getattr) to be able actually read a file using standard libraries.

dhellmann:
This looks like a good update, but I'm going to hold this until we have consensus on that.

/hold

packaging/selinux/microshift.te
eggfoobar: Good point, updated. I'm annoyingly running into an issue here where green boot creates the `/var/lib/microshift-backups` folder but it seems it runs under a context I can't identify to get this `filetrans_pattern` to trigger. Luckily there is no difference in access since `container_t` types are not allowed to access `var_lib_t`

packaging/selinux/microshift.fc
eggfoobar: This has been updated, we are going to be a little more restrictive that OCP and run as `kubelet_exec_t`, That seems to make more sense for our use case.

packaging/selinux/microshift.te
eggfoobar: It seems these aren't needed, in my local testing I can't seem to get audit logs to show denials when removing these, for now I'm removing them to keep us more inline with OCP

eggfoobar:
After some conversations we decided to match OCP a lot more closely in our policy definitions. We have lined up things as closely as we could and we have called out the instances where OCP and MicroShift are different in their SELinux policy.

1. The MicroShift folder `/var/lib/microshift-backups` has no comparable alternative on OCP, we are applying the context `container_var_lib_t` to keep inline with sensetive information that shouldn't be accessed by containers.
2. The MicroShift folder `/var/lib/microshift` shares some design files with `/etc/kubernetes` but is different enough where we're applying the `container_var_lib_t` to keep it inline with OCP `/var/lib/kubelet` which shares more in common.
3. Due to the design of MicroShift in regard to kubelet/etcd/api-server, we have applied `kubelet_exec_t` context to the `microshift` and `microshift-etcd` binaries. This differes from OCP where etcd runs as an super container, and kubelet runs unconfined. The OCP context defines `kubelet` binary as `kubelet_exec_t` but in practice it runs as unconfined, that might be in error on the OCP side.
4. Another difference in OCP is the `/var/lib/kubelet/pods`, In MicroShift the context is respected and the files under `/var/lib/kubelet/pods` are correctly labeled `container_file_t`. In OCP it seems those files are labeled with `container_var_lib_t`, this might be in error since a `restorecon` command will correctly label them as `container_file_t`

----

### Below are some validation commands and the diffs between between an OCP and MicroShift instance.

Note, no denials were observed in the audit logs during testing.

#### Context Validation

**Command:**

```sh
$ sudo semanage fcontext -l | grep -E  "(kubernetes_file_t|container_var_lib_t|kubelet_exec_t)" | awk '{print $1 }'
```

**Diff:**

With the new policy, the only difference between the contexts from an OCP and MicroShift policy was the addition of the MicroShift specific folders to the same contexts used by OCP with the provided reasoning above.

```sh
$ diff ocp.txt ushift.txt 
1a2
> /etc/microshift(/.*)?
2a4,7
> /usr/bin/microshift
> /usr/bin/microshift-etcd
> /usr/local/bin/microshift
> /usr/local/bin/microshift-etcd
15a21,22
> /var/lib/microshift(/.*)?
> /var/lib/microshift-backups(/.*)?
```
#### Allow Validation

**Command:**

```sh
sudo sesearch --allow | grep -E  "(kubernetes_file_t|container_var_lib_t|kubelet_exec_t)"
```

**Diff:**

I am not 100% where this allow statement is being generated from, but given that MicroShift creates and alters files in `/var/lib/microshift` and `/var/lib/microshift-backups`, this makes sense.
```sh
diff ocp-sesearch.txt ushift-sesearch.txt
10a11
> allow kubelet_exec_t var_lib_t:dir { add_name getattr ioctl lock open read remove_name search write };
```

jerpeter1:
> > allow kubelet_exec_t var_lib_t:dir { add_name getattr ioctl lock open read remove_name search write };

This doesn't look right - `kubelet_exec_t` is an `exec_t` label used for executables for domain transitions (in this case, to the `kubelet_t` domain) - an `_exec_t` label is not supposed to be used as a domain type (a process should transition to another domain after it starts, and that domain should have the permissions the process needs).

packaging/rpm/microshift.spec
```diff
@@ -167,8 +167,8 @@ install -p -m755 scripts/microshift-cleanup-data.sh %{buildroot}%{_bindir}/micro
 restorecon -v %{buildroot}%{_bindir}/microshift
```
ggiguash, line 167: Should we move these to the macro as well?

packaging/rpm/microshift.spec
```diff
@@ -19,14 +19,14 @@
 %global selinuxtype targeted
 %define selinux_policyver 3.14.3-67
 %define microshift_relabel_files() \
-   mkdir -p /var/hpvolumes; \
    mkdir -p /var/run/kubelet; \
    mkdir -p /var/lib/kubelet/pods; \
    mkdir -p /var/run/secrets/kubernetes.io/serviceaccount; \
-   restorecon -R /var/hpvolumes; \
+   mkdir -p /var/lib/microshift-backups; \
```
ggiguash, line 25: What about `/var/lib/microshift`? Should we also create and selinux-tag it here?

eggfoobar: That's actually created by MicroShift at startup, so no worries there, it should be captured by the rules and we'll have the e2e tests in place to make sure the dirs are labeled correctly

eggfoobar:
This is good to go for now. After some extensive testing, I think we're in a good state with this PR. We are matching the behavior or OCP as closely as possible and the folders we create also follow the convention described in these SELinux policy. We also have a PR on hold for e2e tests to help verify these rules going forward. https://github.com/openshift/microshift/pull/2218

packaging/rpm/microshift.spec
dhellmann: A comment inline about why we need to have the RPM create the backup directory to avoid the race condition would be good.

dhellmann (commented):
/hold cancel

packaging/rpm/microshift.spec
eggfoobar: Done, thanks!

PR #2213, opened 2023-08-15
USHIFT-1120: migrate greenboot tests to RF

eslutsky:
this test covering 3 use-cases:
- adding user workload  and making sure greenboot runs it.
- simulates service failure and making sure greenboot reports a failure.
- simulates pod failure and making sure greenboot reports a failure.

it covers both regular and ostree based deployments.


test/suites/greenboot.robot
```diff
@@ -0,0 +1,123 @@
...
+Suite Setup         Setup Suite 
+Suite Teardown      Teardown Suite
+
+
+*** Variables ***
+${LOCAL_CONFIG_FILE}                  ../docs/config/busybox_running_check.sh
+${REMOTE_CONFIG_FILE}                 /etc/greenboot/check/required.d/50_busybox_running_check.sh
+${GREENBOOT_CONFIG_FILE}              /etc/greenboot/greenboot.conf
+${GREENBOOT_CONFIG_CONTENT}           MICROSHIFT_WAIT_TIMEOUT_SEC=180
+${MANIFEST_SOURCE_DIR}                assets/kustomizations/greenboot/
```
ggiguash, line 16:
```suggestion
${MANIFEST_SOURCE_DIR}                ./assets/kustomizations/greenboot/
```

test/suites/greenboot.robot
```diff
@@ -0,0 +1,123 @@
...
+
+Greenboot Restart
+    Systemctl    restart  greenboot-healthcheck.service
+
+Microshift Cleanup And Start
+    Cleanup MicroShift and Keep Images
+    Systemctl    enable    --now microshift
+    Greenboot Healthy
+
+Cleanup MicroShift and Keep Images
```
ggiguash, line 117: This is a duplication of an existing [Cleanup MicroShift](https://github.com/openshift/microshift/blob/main/test/resources/microshift-process.resource#L75) function. 

Can we add an optional argument to the existing function instead of writing a new local one?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,123 @@
...
+
+Microshift Disrupt Pod Network
+    ${configuration}=    Catenate    SEPARATOR=\n
+    ...    network:
+    ...    \ serviceNetwork:
+    ...    \ - 10.66.0.0/16
+    ...
+    Upload String To File    ${configuration}    ${MICROSHIFT_CONFIG}
+
+Microshift Restore Pod Network
```
ggiguash, line 103: We already have [Clear MicroShift Config](https://github.com/openshift/microshift/blob/main/test/resources/microshift-config.resource#L52) function that can be used here.

test/suites/greenboot.robot
```diff
@@ -0,0 +1,123 @@
...
+    Should Be Equal As Integers    0    ${rc}
+
+Greenboot Healthy
+    [Documentation]    check
+    Greenboot Restart
+    Wait Until Greenboot Health Check Exited
+
+Microshift Disrupt Service
+    ${stdout}    ${rc}=    Execute Command
+    ...    dnf remove -y hostname
```
ggiguash, line 85: This assumes non-ostree OS. We need to have a condition here for the code to run `rpm-ostree` command instead of `dnf`. 
Alternatively, we can choose to `sudo chmod 000 $(which hostname)`. It can also work on ostree systems after `sudo rpm-ostree usroverlay`

test/suites/greenboot.robot
```diff
@@ -0,0 +1,123 @@
...
+    [Documentation]    Upload a new configuration file to the MicroShift host
+    Put File    ${LOCAL_CONFIG_FILE}    /tmp/busybox_running_check.sh
+    ${stdout}    ${rc}=    Execute Command
+    ...    [ -f ${REMOTE_CONFIG_FILE} ] || sudo ln -s /tmp/busybox_running_check.sh ${REMOTE_CONFIG_FILE} 
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+
+    Put Directory    ${MANIFEST_SOURCE_DIR}     /tmp/manifests
+    ${stdout}    ${rc}=    Execute Command
+    ...    cp -f /tmp/manifests/* /etc/microshift/manifests/ && sudo rm -rf /tmp/manifests
```
ggiguash, line 68: If the command is already elevated, do we need the sudo?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,123 @@
...
+
+Update Greenboot Config
+    [Documentation]    Upload a new configuration file to the MicroShift host
+    Upload String To File   ${GREENBOOT_CONFIG_CONTENT}    ${GREENBOOT_CONFIG_FILE}
+
+Upload User Config
+    [Documentation]    Upload a new configuration file to the MicroShift host
+    Put File    ${LOCAL_CONFIG_FILE}    /tmp/busybox_running_check.sh
+    ${stdout}    ${rc}=    Execute Command
+    ...    [ -f ${REMOTE_CONFIG_FILE} ] || sudo ln -s /tmp/busybox_running_check.sh ${REMOTE_CONFIG_FILE}
```
ggiguash, line 62: If the command is already elevated, do we need the sudo?

test/suites/greenboot.robot
ggiguash: This is [how you can detect](https://github.com/openshift/microshift/blob/main/packaging/greenboot/microshift_set_healthy.sh#L13) if it's an ostree system.

Let's add a function to resources/ostree.resource?

test/suites/greenboot.robot
eslutsky: > If the command is already elevated, do we need the sudo?

yes , the built-in `sudo=True` option elevates  only the  first command expression,that it the only option if we want to keep it one-liner.

test/suites/greenboot.robot
```diff
@@ -0,0 +1,109 @@
...
+
+*** Keywords ***
+Setup Suite
+    [Documentation]    Set up all of the tests in this suite
+    Check Required Env Variables
+    Login MicroShift Host
+    Setup Kubeconfig
+    Upload String To File   ${GREENBOOT_CONFIG_CONTENT}    ${GREENBOOT_CONFIG_FILE} # change the default timeout
+
+Upload User Config
```
dhellmann, line 60: The files being manipulated here aren't really "configuration" files. They're the workload and check script, right? This might be better named "Add Workload" or something along those lines. That would lead to changing the suite variable names, too.

test/suites/greenboot.robot
```diff
@@ -0,0 +1,109 @@
...
+
+    Put Directory    ${MANIFEST_SOURCE_DIR}     /tmp/manifests
+    ${stdout}    ${rc}=    Execute Command
+    ...    cp -f /tmp/manifests/* ${MANIFEST_DIR} && sudo rm -rf /tmp/manifests
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+
+Cleanup User Config
+    ${stdout}    ${rc}=    Execute Command
+    ...    rm -rf ${REMOTE_CONFIG_FILE} ; sudo rm -rf ${MANIFEST_DIR}/*
```
dhellmann, line 76: This could be one command, I think?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,109 @@
+*** Settings ***
+Documentation       Tests related to greenboot
+Resource            ../resources/common.resource
+Resource            ../resources/systemd.resource
+Resource            ../resources/ostree-health.resource
+Resource            ../resources/microshift-config.resource
+Library    SSHLibrary
+Suite Setup         Setup Suite 
+Suite Teardown      Teardown Suite
+
```
dhellmann, line 10: Since this triggers MicroShift restarting, please apply the tags "restart" and "slow". See https://github.com/openshift/microshift/blob/main/test/suites/etcd.robot#L13 for an example.

test/suites/greenboot.robot
```diff
@@ -0,0 +1,109 @@
...
+${GREENBOOT_CONFIG_CONTENT}           MICROSHIFT_WAIT_TIMEOUT_SEC=180
+${MANIFEST_SOURCE_DIR}                ./assets/kustomizations/greenboot/
+${MANIFEST_DIR}                       /etc/microshift/manifests
+${MICROSHIFT_CONFIG}                  /etc/microshift/config.yaml
+*** Test Cases ***
+
+# user-workload-health
+
+GreenbootHealth Run with User Workload
+    [Documentation]    check
```
dhellmann, line 25: I think this test is verifying that the workload starts and then greenboot is successful?

dhellmann (commented):
It's not entirely clear how this test is different from some of the other greenboot test scenarios we have. Maybe @pmtk or @pacevedom can help explain that?

ggiguash:
> It's not entirely clear how this test is different from some of the other greenboot test scenarios we have. Maybe @pmtk or @pacevedom can help explain that?

@dhellmann, other tests expect greenboot to be successful at all times. This test simulates service and pod failure to verify that greenboot script actually reports errors.

test/suites/greenboot.robot
eslutsky: since we are using `Put File` keyword here which is using SCP behind the scenes (non-root user), it cannot copy directly into the system folder with out escalating to root.

test/suites/greenboot.robot
eslutsky: i like the `chmod 000`  idea better and it doesnt require fiddling with RPMs

eslutsky: handled in https://github.com/openshift/microshift/pull/2213/commits/1abb4ea635acac6b3073c67e0047e8fcfa7e2881

test/suites/greenboot.robot
```diff
@@ -0,0 +1,135 @@
+*** Settings ***
+Documentation       Tests related to greenboot
+Resource            ../resources/common.resource
+Resource            ../resources/systemd.resource
+Resource            ../resources/ostree-health.resource
+Resource            ../resources/ostree.resource
+Resource            ../resources/microshift-config.resource
+#Library    SSHLibrary
```
ggiguash, line 8: Leftover comment?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,135 @@
...
+    Check Required Env Variables
+    Login MicroShift Host
+    Setup Kubeconfig
+    Upload String To File   ${GREENBOOT_CONFIG_CONTENT}    ${GREENBOOT_CONFIG_FILE} # change the default timeout
+
+Add User Workload
+    [Documentation]    Upload a new configuration file to the MicroShift host
+    Put File    ${LOCAL_WORKLOAD_FILE}    /tmp/busybox_running_check.sh
+    ${stdout}    ${rc}=    Execute Command
+    ...    [ -f ${GREENBOOT_WORKLOAD_FILE} ] || sudo ln -s /tmp/busybox_running_check.sh ${GREENBOOT_WORKLOAD_FILE}
```
ggiguash, line 77: Is there a special reason for soft-linking this file? Can we just `mv -f` it?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,135 @@
...
+    END
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    ${cmd}
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+    
+Microshift Restore Service
+    ${stdout}    ${rc}=    Execute Command
+    ...    sudo chmod 755 /usr/bin/hostname
```
ggiguash, line 113: The `which` command does not work here, so can we save this in a variable and use here and in other places?

test/suites/greenboot.robot
eslutsky: well I've taken this part from the original script, i guess the  idea was to retain the original copy even after the teardown.

ggiguash: Let's use move this time. I cannot think of a reason why to use `ln -s`

test/suites/greenboot.robot
```diff
@@ -0,0 +1,144 @@
...
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    ${cmd}
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+
+Microshift Restore Service
+    [Documentation]    Restore Microshift service to the correct form.
+    ${stdout}    ${rc}=    Execute Command
+    ...    sudo chmod 755 ${HOSTNAME_BIN_PATH}
```
ggiguash, line 120: No need for sudo here?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,144 @@
...
+    Systemctl    restart    greenboot-healthcheck.service
+    Wait Until Greenboot Health Check Exited
+
+Microshift Disrupt Service
+    [Documentation]    Prevent Microshift service from starting correctly.
+    ${is_ostree}=    Check If OStree
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    which hostname
+    ...    sudo=True    return_rc=True
```
ggiguash, line 104: Probably no need to sudo here?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,144 @@
...
+    Check Required Env Variables
+    Login MicroShift Host
+    Setup Kubeconfig
+    Upload String To File    ${GREENBOOT_CONFIG_CONTENT}    ${GREENBOOT_CONFIG_FILE} # change the default timeout
+
+Add User Workload
+    [Documentation]    Upload User workload files to the MicroShift host
+    Put File    ${LOCAL_WORKLOAD_FILE}    /tmp/busybox_running_check.sh
+    ${stdout}    ${rc}=    Execute Command
+    ...    [ -f ${GREENBOOT_WORKLOAD_FILE} ] || sudo mv /tmp/busybox_running_check.sh ${GREENBOOT_WORKLOAD_FILE}
```
pmtk, line 76: Do we need `sudo` here? Does Robot's `sudo=True` no longer applies at this point?

test/suites/greenboot.robot
```diff
@@ -0,0 +1,144 @@
...
+${GREENBOOT_WORKLOAD_FILE}      /etc/greenboot/check/required.d/50_busybox_running_check.sh
+${GREENBOOT_CONFIG_FILE}        /etc/greenboot/greenboot.conf
+${GREENBOOT_CONFIG_CONTENT}     MICROSHIFT_WAIT_TIMEOUT_SEC=180
+${MANIFEST_SOURCE_DIR}          ./assets/kustomizations/greenboot/
+${MANIFEST_DIR}                 /etc/microshift/manifests
+${HOSTNAME_BIN_PATH}            ""
+
+
+*** Test Cases ***
+GreenbootHealth Run with User Workload
```
pmtk, line 27: Do we need `GreenbootHealth`? All these tests will be grouped by the `greenboot.robot` test suite, so I think we might drop the prefix

test/suites/greenboot.robot
```diff
@@ -0,0 +1,144 @@
...
+    ...    rm -rf ${GREENBOOT_WORKLOAD_FILE} ${MANIFEST_DIR}/*
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    0    ${rc}
+
+Greenboot Healthy
+    [Documentation]    Restart the greenboot-healthcheck service and check its status
+    Systemctl    restart    greenboot-healthcheck.service
+    Wait Until Greenboot Health Check Exited
+
+Microshift Disrupt Service
```
pmtk, line 98: I'd put verb first, so it reads naturally in the test case
```suggestion
Disrupt MicroShift
```
Same goes for other keywords starting with `MicroShift`

test/suites/greenboot.robot
eslutsky: No .. Robo't only prepend `sudo` at the beginning of the expression.

test/suites/greenboot.robot
eslutsky: i'll replace it with verb `Simulate`

test/suites/greenboot.robot
```diff
@@ -0,0 +1,142 @@
...
+    ${is_ostree}=    Is System OSTree
+
+    ${stdout}    ${rc}=    Execute Command
+    ...    which hostname
+    ...    sudo=False    return_rc=True
+    IF    ${rc} == 0    Set Suite Variable    \${HOSTNAME_BIN_PATH}    ${stdout}
+
+    ${cmd}=    Set Variable    sudo chmod 000 ${HOSTNAME_BIN_PATH}
+    IF    ${is_ostree}
+        ${cmd}=    Set Variable    sudo rpm-ostree usroverlay && sudo chmod 000 $(which hostname)
```
pmtk, line 107: Sorry, just noticed: you're using both `${HOSTNAME_BIN_PATH}` and `$(which hostname)`. Could we use `${HOSTNAME_BIN_PATH}` in both cases or are there any obstacles?

eslutsky: Yeap good catch :)

PR #2217, opened 2023-08-16
USHIFT-1578: Increase backoff between virt-install

pmtk:


test/bin/scenario.sh
```diff
@@ -220,8 +220,8 @@ launch_vm() {
             local vm_create_end
             vm_create_end=$(date +%s)
             if [ $(( vm_create_end -  vm_create_start )) -lt 15 ] ; then
-                echo "Error running virt-install on attempt ${attempt}: retrying in 5s"
-                sleep 5
+                echo "Error running virt-install on attempt ${attempt}: retrying in 10s"
+                sleep 10
```
ggiguash, line 224: Let's try to back off using `$(( retry * 5 ))`?

PR #2218, opened 2023-08-16
USHIFT-1568: feat: added initial smoke tests for selinux

eggfoobar:
Added initial smoke tests for SELinux

Holding until we decide on SELinux policy change, these are expected to fail until then.

/hold


e2e/tests/0060-selinux-smoke-test.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
pmtk, line 1: Sorry, but we're obsoleting the `e2e/` tests in favor of RF. This needs to be rewritten :fearful:

eggfoobar: lol no worries, earlier we had discussed placing these not in a RF due to the VM concerns, but I think we're good there for now.

eggfoobar:
/unhold
/retest-required

Unholding now that https://github.com/openshift/microshift/pull/2212 has merged.

ggiguash:
/hold

@eggfoobar, if we place this test in the standard directory, it is run as part of the standart suite automatically because [this scenario](https://github.com/openshift/microshift/blob/main/test/scenarios/el92-src%40standard-suite.sh#L15) works on the dir contents.

Since you added your own scenario file, it create yet another VM and runs this test for the second time in parallel.

In general, since the test runs over 9m, we may want to run it as part of another suite because `standard` periodic would go higher than 20m then. We should either piggy-back it on another suite, or move to periodics.

From the presubmits, can you discuss with @pmtk if it makes sense to run this together with an upgrade or backup tests? Another option is the isolated network test. Otherwise, let's move it to presubmit as a separate VM.

eggfoobar:
Ideally this should be part of a successful upgrade, if we're okay with that, I would prefer it. My original idea was to have this be a function you can call from any test as an added check. That'll give us the chance to validate the different things we do in tests. I've pushed up a commit with that change if we think it's simpler. I left the standard since theres no need to create a backup that should speed up test time (MicroShift doesn't need to be turned off). Otherwise, we can continue to have it as it's own seperate VM. Ideally I like

test/resources/selinux.py
```diff
@@ -0,0 +1,119 @@
...
+    context_list = "kubernetes_file_t|container_var_lib_t|kubelet_exec_t|container_t"
+    semanage_filter_cmd = f"semanage fcontext -l | grep -E  \"({context_list})\" | awk '{{print $1 }}'"
+    return remote_sudo(semanage_filter_cmd).split('\n')
+
+
+def get_denial_audit_log() -> List[str]:
+    ausearch_filter_cmd = "ausearch --input-logs -m avc | grep microshift"
+    stdout, rc = remote_sudo_rc(ausearch_filter_cmd)
+    if rc == 0:
+        return stdout.split('\n')
```
dhellmann, line 67: `splitlines()` should avoid an empty blank entry in the list due to a trailing newline.

test/resources/selinux.py
```diff
@@ -0,0 +1,119 @@
...
+    "system_u:object_r:kubelet_exec_t:s0": [
+        "/usr/bin/microshift",
+        "/usr/bin/microshift-etcd",
+    ],
+    "system_u:object_r:kubernetes_file_t:s0": [
+        "/etc/microshift",
+    ],
+}
+
+EXPECTED_FCONTEXT_LIST = [
```
dhellmann, line 24: Testing the effect, that the files and directories have the right labels, seems like the most important bit. If managing this list requires a lot of maintenance over time, I'd consider dropping the test step that uses it and focus on the labels on files unique to MicroShift.

dhellmann (commented):
I like the approach. I have a couple of small comments inline. I'll leave it for another member of the team to give the final approval.

test/resources/selinux.py
eggfoobar: That's a good point, I can add more access checks on our files and directories. 

In regard to this list, it should only ever change if we alter the SELinux policy or upstream container linux package changes something in these contexts, those events should be rare. Regardless, I think this is a good canary check if anything changes these contexts, so we can decide what that means for MicroShift.

test/resources/selinux.py
eggfoobar: Oh thank you, updating it

test/resources/selinux.py
ggiguash: @eggfoobar , can you add this to the comments in the file?

test/resources/selinux.py
```diff
@@ -0,0 +1,119 @@
+from robot.libraries.BuiltIn import BuiltIn
+from libostree import remote_sudo_rc, remote_sudo
+from typing import List
+
+ACCESS_CHECK_MAP = {
+    "/var/lib/microshift/version": ["cat"],
+    "/etc/microshift/config.yaml.default": ["cat"],
```
ggiguash, line 7: We have 2 more config default files. Should they be in the list?

test/resources/selinux.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def get_expected_ocp_microshift_fcontext_list() -> List[str]:
+    return EXPECTED_FCONTEXT_LIST
+
+
+def get_fcontext_list() -> List[str]:
+    context_list = "kubernetes_file_t|container_var_lib_t|kubelet_exec_t|container_t"
+    semanage_filter_cmd = f"semanage fcontext -l | grep -E  \"({context_list})\" | awk '{{print $1 }}'"
+    return remote_sudo(semanage_filter_cmd).splitlines()
```
ggiguash, line 60: Can we add comments here on why we are filtering on this selected list of context?

test/resources/selinux.py
```diff
@@ -0,0 +1,119 @@
...
+def get_denial_audit_log() -> List[str]:
+    ausearch_filter_cmd = "ausearch --input-logs -m avc | grep microshift"
+    stdout, rc = remote_sudo_rc(ausearch_filter_cmd)
+    if rc == 0:
+        return stdout.splitlines()
+    return []
+
+
+def run_access_check(access_check_map: dict[str, List[str]]) -> List[str]:
+    runcon_cmd = "runcon -u system_u -r system_r -t container_t"
```
ggiguash, line 72: Can we explain in a comment why we chose these specific users and roles?

test/resources/selinux.resource
```diff
@@ -0,0 +1,66 @@
...
+Library             Collections
+Library             selinux.py
+
+
+*** Keywords ***
+Validate SELinux With Backup
+    [Documentation]    Wrapper call for all SELinux checks with extra check inside backup folder
+    [Arguments]    ${backup_name}
+
+    ${result}=    Run Container Access Check On    "${BACKUP_STORAGE}/${backup_name}/resources/kubeadmin/kubeconfig"
```
ggiguash, line 14: Can you explain the idea a bit more on why we are only checking one file? 
Should we check the entire backup tree here?

test/suites/upgrade/upgrade-successful.robot
```diff
@@ -21,14 +22,15 @@ ${TARGET_REF}       ${EMPTY}
...
-    ...    and verifies if it was successful
+    ...    and verifies if it was successful, with SELinux validation
 
     MicroShift 413 Should Not Have Upgrade Artifacts
     Wait For Healthy System
 
     ${future_backup}=    Get Future Backup Name For Current Boot
     Deploy Commit Not Expecting A Rollback    ${TARGET_REF}
     Backup Should Exist    ${future_backup}
+    Validate SELinux With Backup    ${future_backup}
```
ggiguash, line 33: I wonder if we'd better change `Backup Should Exist` library function to call SELinux check?
This should cover all the cases where it's called.

@pmtk , WDYT?

test/resources/selinux.py
eggfoobar: Sure thing. Added the comment

test/resources/selinux.py
eggfoobar: Sure thing, added.

test/resources/selinux.py
eggfoobar: Sure thing, added.

test/resources/selinux.resource
eggfoobar: I intended to make this better, now its updated to check all files in a directory.

eggfoobar:
/unhold 

Deciding to handle 4.13 upgrades in a different PR.

PR #2219, opened 2023-08-16
USHIFT-1591: Optimize VM creation process by creating multiple libvirt storage pools

ggiguash:
When running in CI, virtual machine creation hits the https://github.com/virt-manager/virt-manager/issues/498 libvirt bug.

```
sudo virt-install --noautoconsole --name rhel-9.2-microshift-source-rollback-host1 \
--vcpus 2 --memory 4092 --disk pool=vm-storage,size=20 --network network=default,model=virtio \
--events on_reboot=restart --location /home/microshift/microshift/_output/test-images/vm-storage/rhel-9.2.iso \
--extra-args inst.ks=http://192.168.122.1:8080/scenario-info/rhel-9.2-microshift-source-rollback/vms/host1/kickstart.ks \
--wait 15
ERROR    Error: --disk pool=vm-storage,size=20: internal error: pool 'vm-storage' has asynchronous jobs running.
```

Closes [USHIFT-1591](https://issues.redhat.com//browse/USHIFT-1591)

### DIFF
--- a/test/bin/manage_hypervisor_config.sh
+++ b/test/bin/manage_hypervisor_config.sh
@@ -15,7 +15,7 @@ ${BASH_SOURCE[0]} (create|cleanup)
   -h           Show this help.
 
 create: Set up firewall, storage pool and network.
-        Uses the VM_STORAGE_POOL, VM_DISK_DIR and
+        Uses the VM_POOL_BASENAME, VM_DISK_BASEDIR and
         VM_ISOLATED_NETWORK variables.
 
 cleanup: Undo the settings made by 'create' command.
@@ -44,11 +44,12 @@ firewall_settings() {
 
 action_create() {
     # Storage pool
-    if ! sudo virsh pool-info "${VM_STORAGE_POOL}" &>/dev/null; then
-        sudo virsh pool-define-as "${VM_STORAGE_POOL}" dir --target "${VM_DISK_DIR}"
-        sudo virsh pool-build "${VM_STORAGE_POOL}"
-        sudo virsh pool-start "${VM_STORAGE_POOL}"
-        sudo virsh pool-autostart "${VM_STORAGE_POOL}"
+    # Only create the base pool - the rest are defined before each VM creation
+    if ! sudo virsh pool-info "${VM_POOL_BASENAME}" &>/dev/null; then
+        sudo virsh pool-define-as "${VM_POOL_BASENAME}" dir --target "${VM_DISK_BASEDIR}"
+        sudo virsh pool-build "${VM_POOL_BASENAME}"
+        sudo virsh pool-start "${VM_POOL_BASENAME}"
+        sudo virsh pool-autostart "${VM_POOL_BASENAME}"
     fi
 
     # Isolated network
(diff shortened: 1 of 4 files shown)


test/bin/manage_hypervisor_config.sh
```diff
@@ -44,12 +44,25 @@ firewall_settings() {
 
 action_create() {
     # Storage pool
-    if ! sudo virsh pool-info "${VM_STORAGE_POOL}" &>/dev/null; then
-        sudo virsh pool-define-as "${VM_STORAGE_POOL}" dir --target "${VM_DISK_DIR}"
-        sudo virsh pool-build "${VM_STORAGE_POOL}"
-        sudo virsh pool-start "${VM_STORAGE_POOL}"
-        sudo virsh pool-autostart "${VM_STORAGE_POOL}"
-    fi
+    for scenario in "${SCENARIO_SOURCES}"/*.sh; do
```
dhellmann, line 47: Instead of making this script know which scenarios are being run, could we just have scenario.sh create the storage pool before it creates each VM?

test/bin/manage_hypervisor_config.sh
```diff
@@ -79,11 +92,17 @@ action_cleanup() {
     fi
 
     # Storage pool
-    if sudo virsh pool-info "${VM_STORAGE_POOL}" &>/dev/null; then
-        sudo virsh pool-destroy "${VM_STORAGE_POOL}"
-        sudo virsh pool-undefine "${VM_STORAGE_POOL}"
-    fi
-
+    for scenario in "${SCENARIO_SOURCES}"/*.sh; do
```
dhellmann, line 95: This should loop over the pools that actually exist by looking for them in the expected location. That way we only clean up what is actually created.

test/bin/manage_hypervisor_config.sh
ggiguash: I wanted to keep pool management in a single location, but I can change it according to your suggestion if you think it's better.

dhellmann: What you have works great in CI, but for a local developer running 1 scenario it's going to end up creating a lot more pools than needed. It also means as the developer adds a new scenario, they have to re-run this script, they can't just run scenario.sh.

test/bin/scenario.sh
```diff
@@ -194,6 +194,19 @@ launch_vm() {
...
     echo "Creating ${full_vmname}"
     local -r vm_wait_timeout=$(( VM_BOOT_TIMEOUT / 60 ))
+    local -r vm_pool_name="${VM_POOL_BASENAME}-${SCENARIO}"
+    local -r vm_pool_dir="${VM_DISK_BASEDIR}/${vm_pool_name}"
+
+    # Create the pool if it does not exist
+    if [ ! -d "${vm_pool_dir}" ] ; then
+        mkdir -p "${vm_pool_dir}"
+    fi
+    if ! sudo virsh pool-info "${vm_pool_name}" &>/dev/null; then
```
copejon, line 204: Nit  - Should the `action_create` function be symmetric with `action_cleanup` in pool management? I see test/bin/manage_hypervisor_config.sh deletes all pools itself which makes sense for CI, but less so in a dev env where I would be executing `scenario.sh {create|cleanup} ...`.

copejon:
If my [comment](https://github.com/openshift/microshift/pull/2219#discussion_r1296375999) isn't a concern, feel free to unhold.

/hold
/lgtm

test/bin/scenario.sh
ggiguash: Yes, this makes a lot of sense. I will unhold this PR to unblock other merges and implement this suggestion separately in no time. 
/unhold

PR #2225, opened 2023-08-17
USHIFT-1544: Test harness: rename scenarios and discover coverage gaps

pmtk:
This PR is extracted from https://github.com/openshift/microshift/pull/2184/ for easier review and reducing risk of merging big refactor.

Changes:
- `rhel-9.2` was shortened to `el92`
- `microshift-source` was replaced with "RPM origin tag". It aims to succinctly mark a scenario that uses RPMs different than "source" (code under test; PR):
  - `src`
  - `base`
  - `prev`
- renamed scenario filenames to be more descriptive
- added a "scenario coverage" table to README which aims to provide a testing matrix of test-induced-failed and successful upgrade scenarios between different RPM versions


ggiguash:
@pmtk, I'm not sure if `_` and `-` distinction in the scenario names can be easily interpreted by all readers.

If we want to separate between `system-ref` and `description`, maybe we can use other char, i.e. `system-ref@descr-details`?

test/README.md
```diff
@@ -444,6 +444,26 @@ All of the functions that act as the scenario API are run in the
 context of `./test/bin/scenario.sh` and can therefore use any functions
 defined there.
 
+#### Scenarios coverage
+
+Scenarios utilize following distinct MicroShift sources:
+- `src`: built from source (code in PR)
+- `base`: built from base branch (PR's target branch)
+- `prev`: previous MicroShift minor release
```
copejon, line 452: Just a thought: `rel` or `release` might imply more clearly what the source is.

pmtk: Yeah, `prev` might not be most descriptive in this case, but it originates from `prev-minor` and it's meant to point to a previous Z stream, so for `4.14` it's `4.13`.

In the readme change I mentioned `released`/`rel` which I want to use together with ECs, RCs, and finally Z-stream releases of $current release, like `4.14~ec.N`, etc.

If we can come up with something short and sweet then we can change it :) But for now, we're facing one of most difficult challenges - naming things

ggiguash: What about `prel`, `crel`, etc.?

pmtk: Sure, we can agree on these for now and if we'll discover some new reqs in 4.15, we'll rethink

ggiguash:
This looks good to me. Approving and putting on hold in case others object.
/lgtm
/hold

PR #2239, opened 2023-08-22
USHIFT-1599:  test scenario optimization and reorganization

eslutsky:


pmtk:
Nice, looking at [history of microshift-metal-tests](https://prow.ci.openshift.org/?repo=openshift%2Fmicroshift&type=presubmit&job=pull-ci-openshift-microshift-main-microshift-metal-tests&state=success) looks like we shaved off around 30 minutes :)
Went down from `2h1m ~ 2h15m` to `1h33m ~ 1h40m` (sample of two successful presubmits on this PR). And since cncf test takes like an hour, moving scenarios there didn't affect the time.

/lgtm

PR #2240, opened 2023-08-22
USHIFT-1544: Change OSTree suites filenames to improve descriptiveness 

pmtk:
Rename RF test suites for OSTree to start with "what area it's testing" (upgrade, downgrade, etc.) and then "what flow" to visually group tests together and make filenames more descriptive


eggfoobar:
I was just testing this, we need to update how we create our `vm_hostname` the `@` is not allowed https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L65

PR #2241, opened 2023-08-22
NO-ISSUE: Remove "greenboot-complete" from scenario.sh junit

pmtk:
junit is recorded during "create" phase and waiting for greenboot was moved to "run" phase.

That resulted in adding an extra <testcase> after the </testsuite> which violates XML spec which says that XML document can only have one root element.

Malformed XML is available in `artifacts/microshift-metal-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/$SCENARIO/vms/junit.xml`


ggiguash:
Test not relevant for the change.

PR #2242, opened 2023-08-22
USHIFT-1606: Run parallel make jobs in build_rpm.sh script

ggiguash:
Closes [USHIFT-1606](https://issues.redhat.com//browse/USHIFT-1606)


dhellmann:
How much time does this actually save?

ggiguash:
> How much time does this actually save?

The measuring I've done is in the [USHIFT-1606](https://issues.redhat.com//browse/USHIFT-1606) issue description.

ggiguash:
This test is not relevant for the change.

ggiguash:
test microshift-periodics-metal

ggiguash:
/assign @dhellmann @pmtk
Is this implementation according to our previous discussion?

test/bin/build_rpms.sh
```diff
@@ -13,22 +13,78 @@ cd "${ROOTDIR}"
...
+    echo "The RPM build log file does not exist"
+    exit 1
+fi
+jq < "${BUILD_RPMS_LOG}"
+
+# Determine the script exit code
+if [ "$(jq '.Exitval==0' "${BUILD_RPMS_LOG}" | grep -wc false)" != 0 ] ; then
+    echo "Failed to build all RPMs"
+    exit 1
+fi
```
pmtk, line 87: Parallel exits with non zero if any of the jobs fail, so we might not even get to this step (so we wouldn't even print out results)

pmtk: Unless, this is a consequence of wrapping the commands in bash functions?
I'd rather run the commands directly without intermediate functions (in this case, if we decide to use parallel in other parts of the code, it might make more sense as we'd reuse existing functions)

ggiguash: I uploaded a fix to use commands directly. Let me know if this is how we want to keep it.

pmtk:
This looks good to me. Let's wait for presubmit to finish and see what's the difference in time

ggiguash:
> This looks good to me. Let's wait for presubmit to finish and see what's the difference in time

The pre-submits completed successfully. Here is the link to the [log file](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_microshift/2242/pull-ci-openshift-microshift-main-microshift-metal-tests/1703770748604125184/artifacts/microshift-metal-tests/openshift-microshift-infra-iso-build/build-log.txt).
Look for `RPM build complete` string - it took 134s this time.

PR #2247, opened 2023-08-23
USHIFT-1590: Event cache configuration update

pacevedom:
Default configuration for event cache allows only 1 event every 5min. This triggers event aggregation which collapses several events into a single one, losing information about the event (aggregation happens when the reason is the same, regardless of the message). By allowing 1 event per second this problem is greatly reduced (if not completely).

Closes #<Issue Number>

### DIFF
--- a/pkg/controllers/cluster-policy-controller.go
+++ b/pkg/controllers/cluster-policy-controller.go
@@ -21,6 +21,7 @@ import (
 	openshiftcontrolplanev1 "github.com/openshift/api/openshiftcontrolplane/v1"
 	clusterpolicycontroller "github.com/openshift/cluster-policy-controller/pkg/cmd/cluster-policy-controller"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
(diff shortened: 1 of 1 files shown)


pacevedom:
Event invariant does not fail now according to https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/2247/pull-ci-openshift-microshift-main-e2e-openshift-conformance-reduced/1694291170806468608

pacevedom:
As events are generated they are placed into a cache. This cache has different filters and aggregators to avoid spamming the api server with events. The default configuration allows only 1 event every 5min to come out of the token bucket from the cache. If the same event (same reason but different message) is generated 10+ times in the latest 10min, it is aggregated into a single one with a higher `count`.
With this change some of those parameters are changed to allow events go through to the api server more frequently and not get aggregated (saving the original messages and not losing information).
The cache allows 4k entries, so this change does not have a significant impact in the memory footprint.

PR #2257, opened 2023-08-24
USHIFT-1623: various improvements to development VM configuration

dhellmann:
* Add getty to tty0 so `virsh console` will work with the VMs.
* Add a VNC console to VMs created by the dev scripts to make it easy to
  access them via cockpit.
* ignore subscription-manager error on teardown of vm
* add 'console' command to manage-vm.sh
* add 'ssh' command to manage-vm.sh

/assign @ggiguash @pmtk @pacevedom


scripts/devenv-builder/create-vm.sh
```diff
@@ -63,6 +63,7 @@ virt-install \
     --location ${ISOFILE} \
     --initrd-inject=${KICKSTART_FILE} \
     --extra-args \"inst.ks=file:/$(basename "${KICKSTART_FILE}")\" \
+    --graphics vnc,listen=0.0.0.0 \
```
ggiguash, line 66: How do we connect to this VM using VNC? Should add RPMs to some configure script (i.e. `manage-vm.sh config`, etc.)?

scripts/devenv-builder/manage-vm.sh
```diff
@@ -193,13 +210,15 @@ function action_help {
 function usage {
     local -r script_name=$(basename "$0")
     cat - <<EOF
-${script_name} (create|ip|delete|rm|help) [options]
+${script_name} (create|ip|ssh|console|delete|rm|help) [options]
```
ggiguash, line 213: Looks like `config` is missing here - can we fix it in this PR?

scripts/devenv-builder/create-vm.sh
dhellmann: This enables the VNC console in cockpit, so you don't need any other software.

scripts/devenv-builder/manage-vm.sh
dhellmann: Done

PR #2258, opened 2023-08-24
USHIFT-1623: add support for vnc consoles to scenario vms

dhellmann:
It can be useful to access test VMs through the console. This patch
adds an option to the scenario test harness to add a VNC console that
can be used with cockpit to monitor or login to the VM.

/assign @ggiguash @pmtk @pacevedom


dhellmann:

The scenario script is not used in the RPM upgrade job and the metal test job where it is used did pass.

PR #2262, opened 2023-08-25
USHIFT-1622: Change method to obtain K8s version information

pmtk:
Repository [openshift/kubernetes](https://github.com/openshift/kubernetes) no longer creates new tags when rebasing to newer Kubernetes version.

Instead, [openshift-hack/images/hyperkube/Dockerfile.rhel](https://github.com/openshift/kubernetes/blob/master/openshift-hack/images/hyperkube/Dockerfile.rhel#L16) is updated with the Kubernetes version stored inside a label:
```
io.openshift.build.versions="kubernetes=1.27.4"
```


scripts/auto-rebase/rebase.sh
```diff
@@ -986,10 +986,15 @@ update_buildfiles() {
     title "Rebasing Makefile"
     source hack/lib/version.sh
     kube::version::get_version_vars
+
+    local -r kube_version=$(awk -F'[="]' '/io.openshift.build.versions/ {print $4}' openshift-hack/images/hyperkube/Dockerfile.rhel)
```
dhellmann, line 990: This introduces a sort of API between the rebase processes. Let's make sure the team that owns the kube rebase understands that so we can minimize the times this breaks. Is there a document we can update, for example?

dhellmann (commented):
The implementation looks good. One comment about maintainability inline.

scripts/auto-rebase/rebase.sh
dhellmann: This information is also available as a tag on the hypercube image, so it's in the JSON files created in the staging directory as part of downloading the image. Those files might be a more reliable way to get the information since they are structured and we wouldn't have to parse a Dockerfile.

```
$ cat _output/staging/release_amd64.json | jq -j '.references.spec.tags[] | select(.name == "hyperkube") | .annotations["io.openshift.build.versions"]'
kubernetes=1.27.4
```

pmtk: That is much better, thank you!

PR #2263, opened 2023-08-25
USHIFT-1108: Rename failing upgrade from prev release scenario

pmtk:


pmtk:
I missed the filename during the review

/assign @copejon

PR #2266, opened 2023-08-26
healthcheck deployments and daemonsets

dhellmann:
Instead of checking individual pod state, including restart count, check the ready status of the deployment or daemonset that controls the pod. Let the controllers for deployment and daemonset be responsible for getting their pods into that ready state.

Because the old implementation applied the timeout to several different phases, introduce a retry loop for each namespace to allow each deployment or daemonset several attempts to get to ready.

Closes #<Issue Number>

### DIFF
--- a/packaging/greenboot/microshift-running-check.sh
+++ b/packaging/greenboot/microshift-running-check.sh
@@ -156,8 +156,14 @@ function deployment_ready() {
     local ready
     local reps
 
-    ready=$(${OCGET_CMD} -n "${ns}" "${dep}" -o jsonpath='{.status.readyReplicas}')
+    ready=$(${OCGET_CMD} -n "${ns}" "${dep}" -o jsonpath='{.status.availableReplicas}')
+    if [ -z "${ready}" ]; then
+        return 1
+    fi
     reps=$(${OCGET_CMD} -n "${ns}" "${dep}" -o jsonpath='{.status.replicas}')
+    if [ -z "${reps}" ]; then
+        return 1
+    fi
 
     echo "${ns}/${dep#*/} ${ready}/${reps}"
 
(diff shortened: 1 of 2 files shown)


packaging/greenboot/microshift-running-check.sh
```diff
@@ -148,60 +149,95 @@ fi
...
+    local ready
+    local reps
 
-    echo "Waiting ${WAIT_TIMEOUT_SECS}s for pod image(s) from the '${CHECK_PODS_NS}' namespace to be downloaded"
-    if ! wait_for "${WAIT_TIMEOUT_SECS}" namespace_images_downloaded ; then
-        echo "Error: Timed out waiting for pod image(s) from the '${CHECK_PODS_NS}' namespace to be downloaded"
-        exit 1
-    fi
-done
+    ready=$(${OCGET_CMD} -n "${ns}" "${dep}" -o jsonpath='{.status.readyReplicas}')
```
pmtk, line 159: I'm wondering if `availableReplicas` would be a better fit, but currently we set `minReadySeconds` only for `router-default` Pod, so they work mostly the same (perhaps something in scope of Pod Probes epic)
```
   availableReplicas	<integer>
     Total number of available pods (ready for at least minReadySeconds)
     targeted by this deployment.

   readyReplicas	<integer>
     readyReplicas is the number of pods targeted by this Deployment with a
     Ready Condition.
```

packaging/greenboot/microshift-running-check.sh
```diff
@@ -148,60 +149,95 @@ fi
...
-for i in "${!PODS_NS_LIST[@]}"; do
-    CHECK_PODS_NS=${PODS_NS_LIST[${i}]}
-    CHECK_PODS_CT=${PODS_CT_LIST[${i}]}
+    echo "${ns}/${dep#*/} ${ready}/${reps}"
 
-    echo "Waiting ${WAIT_TIMEOUT_SECS}s for ${CHECK_PODS_CT} pod(s) from the '${CHECK_PODS_NS}' namespace to be in 'Ready' state"
-    if ! wait_for "${WAIT_TIMEOUT_SECS}" namespace_pods_ready ; then
-        echo "Error: Timed out waiting for ${CHECK_PODS_CT} pod(s) in the '${CHECK_PODS_NS}' namespace to be in 'Ready' state"
-        exit 1
+    if [ "${ready}" -ne "${reps}" ]; then
```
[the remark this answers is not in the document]
dhellmann: Fixed

packaging/greenboot/microshift-running-check.sh
dhellmann: Good call, I changed both to look at the available count.

PR #2274, opened 2023-08-28
USHIFT-1601: Add an option to test harness for starting non-ostree VMs

ggiguash:
Closes [USHIFT-1601](https://issues.redhat.com//browse/USHIFT-1601)

### DIFF
--- a/test/bin/create_local_repo.sh
+++ b/test/bin/create_local_repo.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 #
-# This script should be run on the image build server (usually the
-# hypervisor) to create a local RPM repository containing the RPMs
-# built from the source under test.
+# This script should be run on the image build server to
+# create a local RPM repository containing the RPMs built
+# from the source under test.
 
 set -euo pipefail
 
@@ -10,9 +10,18 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=test/bin/common.sh
 source "${SCRIPTDIR}/common.sh"
 
+create_repo() {
+    local -r repodir="$1"
+    createrepo "${repodir}"
+
+    echo "Fixing permissions of RPM repo contents"
+    find "${repodir}" -type f -exec sudo chmod a+r  {} \;
+    find "${repodir}" -type d -exec sudo chmod a+rx {} \;
+}
+
 make_repo() {
-    local repodir="$1"
-    local builddir="$2"
+    local -r repodir="$1"
+    local -r builddir="$2"
 
     title "Creating RPM repo at ${repodir}"
     if [ -d "${repodir}" ]; then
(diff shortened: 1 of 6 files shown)


test/kickstart-templates/kickstart.ks.template
```diff
@@ -35,8 +35,30 @@ rootpw --lock
 # Configure ostree
 ostreesetup --nogpg --osname=rhel --remote=edge --url=REPLACE_OSTREE_SERVER_URL --ref=REPLACE_BOOT_COMMIT_REF
 
+# Configure a non-ostree system
+#liveimg --url file:///run/install/repo/liveimg.tar.gz
```
pmtk, line 39: Where does it come from?

pmtk: Ah, it's part of the ISO itself, right?
`sed` only uncomments this line, it doesn't change the value of `--url`

ggiguash: Yes, exactly. The value is "hardcoded".
If that changes in the future, we can extract this value from the ISO and fill in kickstart dynamically.

test/bin/create_local_repo.sh
```diff
@@ -40,3 +74,5 @@ make_repo "${LOCAL_REPO}" "${RPM_SOURCE}"
 make_repo "${NEXT_REPO}" "${NEXT_RPM_SOURCE}"
 make_repo "${YPLUS2_REPO}" "${YPLUS2_RPM_SOURCE}"
 make_repo "${BASE_REPO}" "${BASE_RPM_SOURCE}"
+
+download_repo "${YMINUS1_REPO}" "4.$(previous_minor_version)"
```
dhellmann, line 79: What's this for?

test/bin/scenario.sh
```diff
@@ -83,9 +79,87 @@ prepare_kickstart() {
...
+        sed -i \
+            -e "s|^ostreesetup|#ostreesetup|g" \
+            -e "s|^#liveimg|liveimg|g" \
+            "${output_file}"
+    fi
+
     record_junit "${vmname}" "prepare_kickstart" "OK"
 }
 
+<<<<<<< HEAD
```
dhellmann: This looks like a rebase issue.

test/scenarios/el92-src@upgrade-rpm-ok.sh
```diff
@@ -0,0 +1,20 @@
...
+scenario_create_vms() {
+    prepare_kickstart host1 kickstart.ks.template ""
+    launch_vm host1
+}
+
+scenario_remove_vms() {
+    remove_vm host1
+}
+
+scenario_run_tests() {
```
dhellmann, line 14: This implementation is not going to produce junit.xml output. If the test is that we can install the RPM, it needs to be done in an RF test. If the test is that the system can run, the RPM should be added to the image via a blueprint for consistency with the other scenarios and then an RF test should be run to verify the installed system.

dhellmann (changes requested):

test/bin/create_local_repo.sh
ggiguash: To re-implement `ci/prow/microshift-rpm-upgrade` job, we need Y-1 microshift RPMs. 
Since we have no rshm at the moment, we cannot pull the RPMs directly from the repo and install on the host.
Hence, we copy the RPMs to a local repo alongside others that we build.

Such repos can then be installed on a VM assuming that all the dependencies are furnished beforehand.

ggiguash:
Working on a slightly different implementation
/close

PR #2277, opened 2023-08-29
USHIFT-1538: Verify contents of a backup before restoring

pmtk:
Adds a naive verification if provided path is
in fact a backup of MicroShift data.
It should prevent some accidental overwrites of
MicroShift data with contents of random directories.

### DIFF
--- a/pkg/admin/data/data_manager.go
+++ b/pkg/admin/data/data_manager.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util"
+	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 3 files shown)


pkg/admin/data/data_manager.go
```diff
@@ -181,6 +195,28 @@ func (dm *manager) RemoveData() error {
 	return nil
 }
 
+// isMicroShiftBackup verifies if given path is a valid MicroShift backup
+// by checking if all expected subdirs exists
+func (dm *manager) isMicroShiftBackup(path string) error {
```
dhellmann, line 200: This function could use some unit tests.

pmtk: Had to split the function because it interacts with the filesystem, but the functional part is now unit-tested

PR #2279, opened 2023-08-29
USHIFT-1601: Implement MicroShift RPM upgrade tests on a non-ostree system

ggiguash:
Implement a substitute test for `prow/microshift-rpm-upgrade` using test harness framework, adjusting it to support booting non-ostree images.

* Use `liveimg` directive to boot from the ISO instead of `ostreesetup` call
* Create an ISO image from `rhel92-microshift-previous-minor` blueprint
* Boot the mentioned ISO as a starting point of the test
  * Verify greenboot and upgrade MicroShift RPMs to the current version
  * Verify greenboot to ensure the upgrade was successful

P.S. This PR also fixes the `wait_for_ssh` and `wait_for_greenboot` issue I encountered. I noticed that long timeouts often seem to get stuck with `ssh`, so I replaced the `timeout` function by a loop.

Closes [USHIFT-1601](https://issues.redhat.com//browse/USHIFT-1601)

### DIFF
--- a/test/suites/upgrade/upgrade-rpm-successful.robot
+++ b/test/suites/upgrade/upgrade-rpm-successful.robot
@@ -1,0 +1,52 @@
+*** Settings ***
+Documentation       Tests related to upgrading MicroShift on a non-ostree system
+
+Resource            ../../resources/common.resource
+Resource            ../../resources/microshift-rpm.resource
+Resource            ../../resources/microshift-host.resource
+Resource            ../../resources/microshift-process.resource
+Resource            ../../resources/ostree-data.resource
+Resource            ../../resources/ostree-health.resource
+Library             Collections
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+Test Tags           rpm-based-system
+
+
+*** Variables ***
+${TARGET_REPO_URL}      ${EMPTY}
+
+
+*** Test Cases ***
+Upgrade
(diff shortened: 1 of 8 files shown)


test/suites/upgrade/upgrade-rpm-successful.robot
```diff
@@ -0,0 +1,51 @@
...
+Upgrade
+    [Documentation]    Performs an upgrade from a given RPM repository URL
+    ...    and checks if it was successful
+
+    # Health of the system is implicitly checked by greenboot successful exit
+    MicroShift 413 Should Not Have Upgrade Artifacts
+    Wait Until Greenboot Health Check Exited
+
+    Uninstall MicroShift RPM Packages
+    Reboot MicroShift Host
```
pmtk, line 32: I don't think we need this here.
Original job removed RPM because our CI steps install $current version during provisioning which is 4.14, but we wanted to start from 4.13 and upgrade to 4.14 (hence the removal of initial 4.14 microshift).
Here, we're already at right starting point which is 4.13, so we should just update to new RPMs

test/suites/upgrade/upgrade-rpm-successful.robot
```diff
@@ -0,0 +1,51 @@
...
+
+    Install MicroShift RPM Packages    ${TARGET_REPO_URL}
+    Enable MicroShift
```
pmtk, line 35: It should be needed when updating the RPMs, right?

PR #2282, opened 2023-08-30
USHIFT-1637: `microshift {backup,restore}` UX improvements

pmtk:
Changes `microshift backup` to look and feel the same as `microshift restore`.
Both commands now take one, non-optional argument which is a fullpath to a backup.

In case of backup - the destination should not exist, MicroShift will create a backup directory.


ggiguash:
Approving and putting on hold, in case someone else wants to review it.
/lgtm
/hold

pkg/cmd/admin.go
```diff
@@ -5,91 +5,90 @@ import (
...
 
 	"github.com/openshift/microshift/pkg/admin/data"
-	"github.com/openshift/microshift/pkg/admin/prerun"
-	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/util"
 
 	"github.com/spf13/cobra"
 )
 
+func backupPathToStorageAndName(path string) (data.StoragePath, data.BackupName, error) {
```
dhellmann, line 15: It would be good to have some unit tests for this function. 

What happens if the path is a relative path?

dhellmann: It might be easier to force the path to an absolute path using `filepath.Abs()` instead of trying to deal with the different cases directly.

dhellmann: `filepath.Clean()` could also be useful.

pkg/cmd/admin_test.go
```diff
@@ -0,0 +1,90 @@
...
+		},
+		{
+			name:               "Relative path with trailing slash",
+			input:              "../microshift-backups/custom-backup/",
+			errExpected:        false,
+			expectedStorage:    "../microshift-backups",
+			expectedBackupName: "custom-backup",
+		},
+		{
+			name:               "Relative path without trailing slash",
```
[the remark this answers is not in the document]
pmtk: That's `Storage is .` from line 54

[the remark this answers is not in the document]
dhellmann: Aha, I missed that, ok.

pkg/cmd/admin.go
pmtk: Used `filepath.Clean()`, but skipped the `filepath.Abs()` because:
- it uses CWD, so values we expect in UT would become dynamic,
- it returns `(string, error)` so it's not as elegant so including it would mean: either adding it to 3 places (PreRun and 2* Run) before `backupPathToStorageAndName`) or using global vars (parse once in PreRun, use in Run) and I'm not a fan of any of them

I removed checking for `.` and `..` but had to add `/` because `filepath.{Dir,Base}("/")` returns `/`.

WDYT?

pkg/cmd/admin.go
```diff
@@ -21,92 +20,140 @@ func shouldRunPrivileged() error {
...
+
+	return storage, name, nil
+}
+
+// backupRestorePreRun contains necessary checks before attempting to perform
+// backup or restore. It is meant to be used as PersistentPreRunE.
+// It is not a part of RunE because k8s.io/component-base/cli.Run() wrapper
+// sets up klog  which considerably decreases readability of errors because
+// they're hidden in logging information such as pid, datetime,
+// source file and line.
```
dhellmann, line 98: I like that you gave so much "why" in this comment. +1

pkg/cmd/admin.go
dhellmann: It seems like the unit test could figure out what the expected values should be, but this is definitely an improvement.

dhellmann (commented):
/lgtm

PR #2283, opened 2023-08-30
USHIFT-1632: Early check of elevated privileges

pmtk:
Adds an early check if microshift `backup` or `restore` commands are executed as root or with sudo to quickly fails otherwise and provide succinct output to user.


pmtk:
/unhold

All jobs we care about are required, PR won't be merged if metal-tests fail

PR #2292, opened 2023-09-01
OCPBUGS-19457: Test update from current release

pmtk:
Closes: [USHIFT-1097](https://issues.redhat.com//browse/USHIFT-1097) [USHIFT-1107](https://issues.redhat.com//browse/USHIFT-1107) OCPBUGS-19457

### DIFF
--- a/test/bin/build_images.sh
+++ b/test/bin/build_images.sh
@@ -14,23 +14,25 @@ shopt -s nullglob
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=test/bin/common.sh
 source "${SCRIPTDIR}/common.sh"
+# shellcheck source=test/bin/get_crel_version_repo.sh
+source "${SCRIPTDIR}/get_crel_version_repo.sh"
 
 configure_package_sources() {
     ## TEMPLATE VARIABLES
-    #
-    # Machine platform type ("x86_64")
-    UNAME_M=$(uname -m)
-    export UNAME_M
+    export UNAME_M                 # defined in common.sh
     export LOCAL_REPO              # defined in common.sh
     export NEXT_REPO               # defined in common.sh
     export BASE_REPO               # defined in common.sh
     export YPLUS2_REPO             # defined in common.sh
+    export CURRENT_RELEASE_REPO
+
     export SOURCE_VERSION
     export FAKE_NEXT_MINOR_VERSION
     export FAKE_YPLUS2_MINOR_VERSION
     export MINOR_VERSION
     export PREVIOUS_MINOR_VERSION
     export SOURCE_VERSION_BASE
+    export CURRENT_RELEASE_VERSION
 
     # Add our sources. It is OK to run these steps repeatedly, if the
     # details change they are updated in the service.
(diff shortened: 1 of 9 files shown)


test/bin/build_images.sh
```diff
@@ -14,29 +14,62 @@ shopt -s nullglob
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=test/bin/common.sh
 source "${SCRIPTDIR}/common.sh"
+# shellcheck source=test/bin/get_crel_version_repo.sh
+source "${SCRIPTDIR}/get_crel_version_repo.sh"
+
+test_envsubst() {
```
dhellmann, line 20: I like this idea. I would split it from the new test to make reviews easier. See also https://issues.redhat.com/browse/USHIFT-1546

test/bin/build_images.sh
```diff
@@ -14,29 +14,62 @@ shopt -s nullglob
...
     export YPLUS2_REPO             # defined in common.sh
+    export CREL_REPO
+
     export SOURCE_VERSION
     export FAKE_NEXT_MINOR_VERSION
     export FAKE_YPLUS2_MINOR_VERSION
     export MINOR_VERSION
     export PREVIOUS_MINOR_VERSION
     export SOURCE_VERSION_BASE
+    export CREL_VERSION
```
dhellmann, line 57: "CREL" is "current release"?

pmtk: Yes, just like we have "prel" (previous release) in scenarios. Though I guess that might've been a puzzle. I'll expand the abbrev and use `crel` only in filenames

test/bin/build_images.sh
pmtk: Sure, I'll create separate PR for this one.

Re templating: Unless we'd like to address "potential non-existence of current release (e.g. early 4.15)" in another way, then we have additional requirement from a templating engine to tell us if the template could not be fully rendered (it would be nice if it was separate from an error about syntax problems etc).

Because this is what I need here: if could not fully render (because CREL_VERSION will be empty (!) in early 4.15) then I don't want to load this osbuild source and don't want the test to even run

pmtk: On the other hand: given our tests will only grow and we introduce optional behavior such as this, should we start thinking about moving away from bash? Logs from bash -x are like last resort because of poor readability. With Python we could replace script by script. I wish we had a log from infrastructure part that looks like robot's log.html

pmtk:
Until new EC/RC, the tests will fail/work incorrectly:
```
"FAIL reading version file" err="unmarshalling \"4.14.0\" failed: invalid character '.' after top-level value"
```

dhellmann:
@pmtk are we going to have an issue with this test for the first sprint of every release before there is an EC for the current version?

pmtk:
> @pmtk are we going to have an issue with this test for the first sprint of every release before there is an EC for the current version?

Because the source/blueprint are optional if we cannot determine the version, it should work.

scripts/image-builder/configure.sh
```diff
@@ -27,6 +27,8 @@ sudo dnf install -y "https://dl.fedoraproject.org/pub/epel/epel-release-latest-$
 sudo dnf install -y mock nginx tomcli
 sudo usermod -a -G mock "$(whoami)"
 
+go install github.com/hairyhenderson/gomplate/v4/cmd/gomplate@latest # TODO: Fetch versioned binary
```
dhellmann, line 30: Do we want this done in fetch-tools.sh?

pmtk:
I moved the gomplate POC to separate PR: https://github.com/openshift/microshift/pull/2351
Let's decide if we want to utilize gomplate for this (looks nice)

PR #2306, opened 2023-09-05
USHIFT-1529: Storage upgrade remove trigger

eggfoobar:
Remove storage migration trigger service

With this change, storage state checks are no longer needed so the e2e tests were updated to correctly account for the change and execute manual migration requests.

This PR keeps the removal of the service code in a separate commit to make it easier to restore if we ever need to re-add this to MicroShift.

/assign @dhellmann


pmtk:
Could you also update https://github.com/openshift/microshift/blob/main/docs/contributor/updateability.md#kubernetes-storage-migration?

eggfoobar:
> Could you also update https://github.com/openshift/microshift/blob/main/docs/contributor/updateability.md#kubernetes-storage-migration?

Hey @pmtk Should be all set now

PR #2308, opened 2023-09-05
USHIFT-1663: Add missing psa label syncing rbac manifests

copejon:


pmtk:
I checked [logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_microshift/2308/pull-ci-openshift-microshift-main-microshift-metal-tests/1699214819103608832/artifacts/microshift-metal-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el92-base@upgrade-ok/vms/host1/sos/) and it looks like the message about forbidden namespaces still appears:
```
privileged_namespaces_controller.go:105] namespaces "default" is forbidden: User "system:serviceaccount:openshift-infra:privileged-namespaces-psa-label-syncer" cannot patch resource "namespaces" in API group "" in the namespace "default"
privileged_namespaces_controller.go:105] namespaces "kube-public" is forbidden: User "system:serviceaccount:openshift-infra:privileged-namespaces-psa-label-syncer" cannot patch resource "namespaces" in API group "" in the namespace "kube-public"
privileged_namespaces_controller.go:105] namespaces "kube-system" is forbidden: User "system:serviceaccount:openshift-infra:privileged-namespaces-psa-label-syncer" cannot patch resource "namespaces" in API group "" in the namespace "kube-system"
```

My memory on SA and RBAC is little rusty, but could it be that we run [cluster-policy-controller](https://github.com/openshift/microshift/blob/main/pkg/controllers/cluster-policy-controller.go#L74) in a different namespace (`"openshift-kube-controller-manager"`) but the
RBAC for it (CR and CRB you're adding, but also what we already have) are meant to give privileges to a SA in `openshift-infra` namespace?
```
subjects:
- kind: ServiceAccount
  name: privileged-namespaces-psa-label-syncer
  namespace: openshift-infra
```

copejon:
@pmtk Ah, that would be because I had the wrong contents for the clusterrole.  That's fixed now

copejon:
The message has vanished in local testing. Dropping the hold now

/unhold

copejon:
After the fix, the log lines will still appear between the start of the cluster-policy-controller service up until the infra-services-controller service creates the necessary RBAC. I played around with having the cluster-policy-controller deploy the RBAC itself, but it's service account does not have the permissions.

After the infra-services-controller starts, the RBAC failure logs for the privileged_namespace_controller will stop.

dhellmann:
> After the fix, the log lines will still appear between the start of the cluster-policy-controller service up until the infra-services-controller service creates the necessary RBAC. 

Are they showing up on an upgrade, or on a fresh install?

> I played around with having the cluster-policy-controller deploy the RBAC itself, but it's service account does not have the permissions.

Yeah, we want to keep the controller RBAC limited to what they need.

> After the infra-services-controller starts, the RBAC failure logs for the privileged_namespace_controller will stop.

Should we think about changing any of the start up ordering in the future (not this PR or release)?

copejon:
> > After the fix, the log lines will still appear between the start of the cluster-policy-controller service up until the infra-services-controller service creates the necessary RBAC.
> 
> Are they showing up on an upgrade, or on a fresh install?

Both. Because of the load order of the services, the logs will still appear a handful of times in either case. 

> > I played around with having the cluster-policy-controller deploy the RBAC itself, but it's service account does not have the permissions.
> 
> Yeah, we want to keep the controller RBAC limited to what they need.
> 
> > After the infra-services-controller starts, the RBAC failure logs for the privileged_namespace_controller will stop.
> 
> Should we think about changing any of the start up ordering in the future (not this PR or release)?

I don't think that would help in this case.  The only order imposed during startup is based on a particular service's dependencies.  The infra-services-controller has more dependencies than the cluster-policy-controller, so it starts later organically.  That said, I still think the right answer is to have each controller service deploy the RBAC respective of that controller. I.e. the cluster-policy-controller really ought to deploy these manifests.

Since this was just a spin off of a higher priority ticket, and is just to clean up logging, I didn't want to spend more time untangling the rbac perms.

dhellmann:
> > Should we think about changing any of the start up ordering in the future (not this PR or release)?
> 
> I don't think that would help in this case. The only order imposed during startup is based on a particular service's dependencies. The infra-services-controller has more dependencies than the cluster-policy-controller, so it starts later organically. That said, I still think the right answer is to have each controller service deploy the RBAC respective of that controller. I.e. the cluster-policy-controller really ought to deploy these manifests.
> 
> Since this was just a spin off of a higher priority ticket, and is just to clean up logging, I didn't want to spend more time untangling the rbac perms.

OK, maybe queue up a separate conversation about that, then. We don't want controllers having blanket permissions, but there might be other solutions. Let's get the team to brainstorm a bit.

copejon:
Well, still seeing the lines appear in the metal-tests' journal.  My local test env isn't producing them after the infrastructure-services-manager starts. Putting back on hold until I can dig into why that is.

/hold

pmtk:
> I don't think that would help in this case. The only order imposed during startup is based on a particular service's dependencies. The infra-services-controller has more dependencies than the cluster-policy-controller, so it starts later organically.

Just FYI, we don't create any dynamic tree of services/components/controllers startup based on the dependencies. Ordering is still rather manual and affected by the code order in run.go.

I see it more as a safety mechanisms so we don't try to start svc without preconditions, but it's just that (i.e. `m.AddService` is not a bucket we add to which re-orders the services based on dependencies and if it did, then ClusterPolicyController should have InfraSvcs as dependency [at least for time being until we separate that RBAC deployment]). The `topological sorting of directed acyclic graphs via DFS traversal` is commented out.

While I agree it would be best to have `pkg/controllers/cluster-policy-controller.go` deploy RBAC for itself, in current case (tested on my local env), moving
```
	util.Must(m.AddService(controllers.NewClusterPolicyController(cfg)))
```

after
```
	util.Must(m.AddService(controllers.NewInfrastructureServices(cfg)))
```
helps with the messages.

pkg/controllers/infra-services-controller.go
```diff
@@ -70,11 +70,13 @@ func applyDefaultRBACs(ctx context.Context, cfg *config.Config) error {
 			"controllers/kube-controller-manager/csr_approver_clusterrole.yaml",
 			"controllers/cluster-policy-controller/namespace-security-allocation-controller-clusterrole.yaml",
 			"controllers/cluster-policy-controller/podsecurity-admission-label-syncer-controller-clusterrole.yaml",
+			"controllers/cluster-policy-controller/podsecurity-admission-label-privileged-namespaces-syncer-controller-clusterrole.yaml",
```
dhellmann, line 73: As we discussed on slack, let's move these RBAC changes to cluster-policy-controller.go so they're closer to the code they affect.

dhellmann: Nevermind, these are cluster roles and we want them all managed in one place.

copejon:
I'm not able to reproduce the storage test failures in local test runs.  In local runs, all tests pass consistently.

/retest

PR #2311, opened 2023-09-06
USHIFT-1664: Introduce microshift-sos-report helper

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/scripts/microshift-sos-report.sh
+++ b/scripts/microshift-sos-report.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -euo pipefail
+
+SCRIPT_NAME=$(basename "$0")
+PROFILES="network,security,microshift"
+PLUGINS="container_log,crio,rpmostree,rpm"
+TEMPDIR="/tmp"
+
+function usage() {
+    echo "Collect an sos report including all the relevant information for MicroShift"
+    echo "Includes"
+    echo "  profiles: ${PROFILES}"
+    echo "  plugins: ${PLUGINS}"
+    echo ""
+    echo "Usage: ${SCRIPT_NAME} [--tmp-dir TMP-DIR]"
+    echo "   --tmp-dir TMP-DIR  Temporary directory for saving the report. Defaults to ${TEMPDIR}."
+    exit 1
+}
+
+if [ $# -ge 1 ]; then
+    case $1 in
+    --tmp-dir)
+        [ $# -ne 2 ] && usage
+        TEMPDIR="$2"
+        ;;
+    *)
+        usage
+        ;;
+    esac
+fi
+
+if [ "$(id -u)" -ne 0 ] ; then
+    echo "The '${SCRIPT_NAME}' script must be run with the 'root' user privileges"
+    exit 1
+fi
+
+if [ ! -d "${TEMPDIR}" ]; then
+    echo "Directory ${TEMPDIR} does not exist"
+    exit 1
+fi
+
+sos report \
+  --quiet \
+  --batch \
+  --all-logs \
+  --tmp-dir "${TEMPDIR}" \
+  --profiles "${PROFILES}" \
+  --only-plugins "${PLUGINS}"
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -36,7 +36,10 @@ sos_report() {
             continue
         fi
         ip=$(cat "${vmdir}/ip")
-        ssh "redhat@${ip}" "sudo sos report --quiet --batch --all-logs --tmp-dir /tmp && sudo chmod +r /tmp/sosreport*"
+        # Copy the sos helper for compatibility, it is only available in 4.14 RPMs
+        scp "${ROOTDIR}/scripts/microshift-sos-report.sh" "redhat@${ip}":/tmp
+        ssh "redhat@${ip}" \
+            "[ -f /usr/bin/microshift-sos-report ] && sudo /usr/bin/microshift-sos-report || sudo /tmp/microshift-sos-report.sh ; sudo chmod +r /tmp/sosreport*"
         mkdir -p "${vmdir}/sos"
         scp "redhat@${ip}:/tmp/sosreport*.tar.xz" "${vmdir}/sos/"
     done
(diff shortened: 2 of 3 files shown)


scripts/microshift-sos-report.sh
```diff
@@ -0,0 +1,50 @@
...
+
+set -euo pipefail
+
+SCRIPT_NAME=$(basename "$0")
+PROFILES="network,security,microshift"
+PLUGINS="container_log,crio,rpmostree,rpm"
+TEMPDIR="/tmp"
+
+function usage() {
+    echo "Collect an sos report including all relevant information for MicroShift"
```
ggiguash, line 11:
```suggestion
    echo "Collect an sos report including all the relevant information for MicroShift"
```

scripts/microshift-sos-report.sh
```diff
@@ -0,0 +1,50 @@
...
+    echo "Includes"
+    echo "  profiles: ${PROFILES}"
+    echo "  plugins: ${PLUGINS}"
+    echo ""
+    echo "Usage: ${SCRIPT_NAME} [--tmp-dir TMP-DIR]"
+    echo "   --tmp-dir TMP-DIR  Temporary directory where report will be written. Defaults to ${TEMPDIR}."
```
ggiguash, line 17:
```suggestion
    echo "   --tmp-dir TMP-DIR  Temporary directory for saving the report. Defaults to ${TEMPDIR}."
```

scripts/microshift-sos-report.sh
```diff
@@ -0,0 +1,50 @@
...
+    exit 1
+}
+
+if [ $# -ge 1 ]; then
+    case $1 in
+    --tmp-dir)
+        TEMPDIR="$2"
```
ggiguash, line 24: This would generate uninitialized variable error if run with one argument.
I think we need to change the "if" to be `$# -eq 2` else `usage`

scripts/microshift-sos-report.sh
```diff
@@ -0,0 +1,50 @@
...
+
+sos report \
+  --quiet \
+  --batch \
+  --all-logs \
+  --tmp-dir "${TEMPDIR}" \
+  --profiles "${PROFILES}" \
+  --only-plugins "${PLUGINS}"
+
+chmod +r "${TEMPDIR}"/sosreport*
```
ggiguash, line 50: This is a bit dangerous because we are changing permissions of all the files that start with `sosreport`. This may not be intended by the user for all sos reports in /tmp.

ggiguash (changes requested):

scripts/microshift-sos-report.sh
pacevedom: Done

scripts/microshift-sos-report.sh
pacevedom: Done

scripts/microshift-sos-report.sh
pacevedom: Was thinking mainly in CI, but it makes sense to require further permissions update after running the script if it runs as sudo.

scripts/microshift-sos-report.sh
ggiguash: Sorry, for the incomplete suggestion. I guess, we need to check for 0 or 2 arguments to the script because `--tmp-dir` is optional, right?

test/bin/scenario.sh
```diff
@@ -36,7 +36,7 @@ sos_report() {
             continue
         fi
         ip=$(cat "${vmdir}/ip")
-        ssh "redhat@${ip}" "sudo sos report --quiet --batch --all-logs --tmp-dir /tmp && sudo chmod +r /tmp/sosreport*"
+        ssh "redhat@${ip}" "sudo /bin/microshift-sos-report && sudo chmod +r /tmp/sosreport*"
```
pmtk, line 39:
```suggestion
        ssh "redhat@${ip}" "sudo /usr/bin/microshift-sos-report && sudo chmod +r /tmp/sosreport*"
```

scripts/microshift-sos-report.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ggiguash: I'd rather have it as a script, which support can look into and see what is run.

test/bin/scenario.sh
pacevedom: Done

scripts/microshift-sos-report.sh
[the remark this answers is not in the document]
pacevedom: Being an external tool I am not entirely sure we should embed that logic, as it doesnt belong to microshift itself.

scripts/microshift-sos-report.sh
```diff
@@ -0,0 +1,50 @@
...
+TEMPDIR="/tmp"
+
+function usage() {
+    echo "Collect an sos report including all relevant information for MicroShift"
+    echo "Includes"
+    echo "  profiles: ${PROFILES}"
+    echo "  plugins: ${PLUGINS}"
+    echo ""
+    echo "Usage: ${SCRIPT_NAME} [--tmp-dir TMP-DIR]"
+    echo "   --tmp-dir TMP-DIR  Temporary directory for saving the report. Defaults to ${TEMPDIR}."
```
ggiguash, line 17: Should we add a command line option for specifying plugins by user?

scripts/microshift-sos-report.sh
pacevedom: Done

scripts/microshift-sos-report.sh
pacevedom: Done

scripts/microshift-sos-report.sh
pacevedom: The script is mainly a one line command only, I thought of it as a helper for non-advanced users. If you need to customize the command it looks easier to me to just run `sos report` directly instead, wdyt?

pacevedom:
sos reports running with reduced plugins now, even if the job failed (for different reasons).

PR #2313, opened 2023-09-06
USHIFT-1400: Don't ignore errors when creating changelog and last_*rebase.sh

pmtk:


dhellmann (commented):
I see a pattern here that could become a function, but we can do that separately.

/lgtm

PR #2319, opened 2023-09-07
OCPBUGS-18548: Stop all MicroShift components when signaled to exit

pmtk:
### make KCM use context we give to command 

By using `cmd.Context()`, KCM will use a context MicroShift passes down through all the components.

When using `context.Background()` it won't ever stop, because it's never cancelled, and result in blocking `ServiceManager` which waits for all components to stop and make MicroShift use its whole internal `gracefulShutdownTimer`.

---

### override KAS' `shutdown-delay-duration` to `5s` (from `70s`)

`shutdown-delay-duration` tells KAS how long to serve after receiving a stop signal.

Its value is provided to `time.Sleep()`, so it's used in full; there is not short-circuit mechanism that would let KAS exit earlier.

Default value of `70s` is set in `assets/controllers/kube-apiserver/defaultconfig.yaml`


eggfoobar:
/lgtm

nice find!

PR #2322, opened 2023-09-08
OCPBUGS-18706: Don't wait before applying a CRD

pmtk:
By changing `immediate` from `false` to `true`, `PollUntilContextTimeout` will no longer wait for specified interval (5 seconds in this case) before running given `func` for the first time.

This results in drastic speed up of MicroShift's start up: 40 seconds because currently we have 8 CRDs to apply.


pacevedom (approved):
/lgtm

PR #2325, opened 2023-09-10
OCPBUGS-19296: Fix microshift-cleanup-data script and test its functionality

ggiguash:
Write an e2e test for testing the functionality of the `microshift-cleanup-data` script.
The following problems were found in the test and fixed:
* When deleting pods, no need to delete containers. Only delete images when `--keep-images` option is absent
* When cleaning OVN data, delete the `/var/run/ovn-kubernetes` directory instead of `/var/lib/ovnk` which does not exist

Closes [USHIFT-1509](https://issues.redhat.com//browse/USHIFT-1509)

### DIFF
--- a/test/resources/ostree-health.resource
+++ b/test/resources/ostree-health.resource
@@ -35,3 +35,8 @@ System Should Be Healthy
 
     ${health}=    Get Persisted System Health
     Should Be Equal As Strings    ${health}    healthy
+
+Restart Greenboot And Wait For Success
+    [Documentation]    Restart the greenboot-healthcheck service and check its status
+    Systemctl    restart    greenboot-healthcheck.service
+    Wait Until Greenboot Health Check Exited
--- a/test/suites/standard/cleanup-data.robot
+++ b/test/suites/standard/cleanup-data.robot
@@ -1,0 +1,185 @@
+*** Settings ***
+Documentation       Tests verifying microshift-cleanup-data script functionality
+
+Resource            ../../resources/systemd.resource
+Resource            ../../resources/ostree.resource
+Resource            ../../resources/ostree-health.resource
+Resource            ../../resources/microshift-config.resource
+
+Suite Setup         Setup Suite
+
+Test Tags           slow
+
+
+*** Test Cases ***
+Verify Invalid Command Line
+    [Documentation]    Verify invalid command line combinations
+
+    # Usage message
+    ${rc}=    Run MicroShift Cleanup Data    ${EMPTY}
+    Should Not Be Equal As Integers    ${rc}    0
+
+    # Invalid option combination
+    ${rc}=    Run MicroShift Cleanup Data    --ovn    --keep-images
+    Should Not Be Equal As Integers    ${rc}    0
+
+    ${rc}=    Run MicroShift Cleanup Data    --all    --ovn
+    Should Not Be Equal As Integers    ${rc}    0
+
+    ${rc}=    Run MicroShift Cleanup Data    --keep-images
+    Should Not Be Equal As Integers    ${rc}    0
+
+Verify Full Cleanup Data
+    [Documentation]    Verify full data clean scenarios
+
+    ${rc}=    Run MicroShift Cleanup Data    --all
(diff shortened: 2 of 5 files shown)


pliurh:
The OVN-related part looks good to me.

test/suites/standard/cleanup-data.robot
```diff
@@ -0,0 +1,158 @@
...
+    Should Be Equal As Integers    ${stdout}    1
+
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    crictl images | wc -l
+    ...    sudo=True    return_stdout=True    return_stderr=True    return_rc=True
+    Should Be Equal As Integers    ${rc}    0
+    IF    ${after_image_clean}
+        Should Be Equal As Integers    ${stdout}    1
+    ELSE
+        Should Be True    ${stdout} > 1
```
dhellmann, line 112: How does this comparison work? Does stdout get converted to an integer automatically somehow?

test/suites/standard/cleanup-data.robot
```diff
@@ -0,0 +1,158 @@
...
+Run MicroShift Cleanup Data
+    [Documentation]    Run the microshift-cleanup-data script and
+    ...    return its exit code
+    [Arguments]    ${cmd}    ${opt}=${EMPTY}
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    echo 1 | sudo microshift-cleanup-data ${cmd} ${opt}
+    ...    return_stdout=True    return_stderr=True    return_rc=True
+    RETURN    ${rc}
+
+Verify Crio Resource Count
```
dhellmann, line 88: If this was written as an assertion, I think it would be obvious it needs to be split into a few keywords.

- Crio should have 1 pod
- Crio should have 1 process
- Crio should have 1 image

Then the caller would use those instead of this jumbo keyword, and instead of an argument that triggers the conditional check on line 109 there would be a separate keyword for

- Crio should have 0 images

test/suites/standard/cleanup-data.robot
```diff
@@ -0,0 +1,158 @@
...
+    ...    crictl images | wc -l
+    ...    sudo=True    return_stdout=True    return_stderr=True    return_rc=True
+    Should Be Equal As Integers    ${rc}    0
+    IF    ${after_image_clean}
+        Should Be Equal As Integers    ${stdout}    1
+    ELSE
+        Should Be True    ${stdout} > 1
+    END
+
+Verify MicroShift Data Directory And Processes
```
dhellmann, line 115: Similarly here

- MicroShift Data Directory Should Exist
- MicroShift Data Directory Should Not Exist
- MicroShift Should Be Stopped
- MicroShift-etcd Should Be Stopped

etc.

test/suites/standard/cleanup-data.robot
ggiguash: Made it more explicit.

test/resources/ostree-health.resource
```diff
@@ -35,3 +35,8 @@ System Should Be Healthy
 
     ${health}=    Get Persisted System Health
     Should Be Equal As Strings    ${health}    healthy
+
+Restart Greenboot And Wait For Success
+    [Documentation]    Restart the greenboot-healthcheck service and check its status
+    Systemctl    restart    greenboot-healthcheck.service
+    Wait Until Greenboot Health Check Exited
```
dhellmann, line 42: This only waits for greenboot to finish, right, without checking the results? But line 39 says "wait for success".

dhellmann (commented):
/lgtm
/hold

One nit on naming that I'm not sure about. If I'm wrong, feel free to remove the hold.

test/resources/ostree-health.resource
ggiguash: The `Greenboot Health Check Exited` keyword is checking the status of the service, which is supposed to be `exited`
See https://github.com/openshift/microshift/pull/2325/files#diff-91546dab3e6e25e6ab48a09f02c8040d107f5cf9f8e40b06cd1babd6cb25b395R20

PR #2326, opened 2023-09-10
USHIFT-1671:  systemctl dont restart etcd.

eslutsky:
when `systemctl restart microshift` is invoked, microshift.service stopped  correctly but microshift-etcd.scope remains running , so after microshift is started it fails to start (second) microshift-etcd.scope  and  systemctl returns an  error.

this PR will:
- ensure that `microshift-etcd.scope`  is stopped.
- fix the RF restart test, to properly simulate the user action.


test/resources/microshift-process.resource
```diff
@@ -80,17 +80,9 @@ Wait For MicroShift Service
...
-    # whether we should count it as back up yet. Forcing the service
-    # down, then back up, ensures that when we do things like check
-    # systemd settings for the running process we get the values we
-    # expect without worrying about the race. Any test that modifies
-    # the MicroShift configuration file will be more reliable as a
-    # result.
-    Stop MicroShift
-    Sleep    10 seconds
-    Start MicroShift
+    ${cur_pid}=    MicroShift Process ID
```
ggiguash, line 83: We should also make sure that microshift-etcd is restarting.

eslutsky:
it's not very clear why  failed,trying metal again:
/test microshift-metal-tests

pkg/controllers/etcd.go
```diff
@@ -116,6 +116,15 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
 		os.Exit(0)
 	}()
 
+	// Handle microshift-etcd termination after microshift process exits
+	defer func() {
+		cmd := exec.Command("systemctl", "stop", "microshift-etcd.scope")
+
+		if err = cmd.Start(); err != nil {
```
pmtk, line 123: `Start()` should be paired with `Wait()`. Is lack of `Wait()` a decision?
Otherwise, we can use `Run()` which will wait for completion.

How does this impact shutdown ordering?

eslutsky: > `Start()` should be paired with `Wait()`. Is lack of `Wait()` a decision to make it "async and forget - don't wait for it"? Otherwise, we can use `Run()` which will wait for completion.
> 
> How does this impact shutdown ordering? Is the problem here that MicroShift starts too soon after restart? i.e. `microshift-etcd.scope` hasn't exited yet, but `microshift.service` is already up and running again?
the problem is that systemd doesnt restart `microshift-etcd.scope` although it depends on microshift.service. 
so when restart happens `microshift` is stopped but `microshift-etcd.scope` remains active, then `microshift` tries to run another instance with `systemd-run` while first is running and it fails.

eslutsky: > `Start()` should be paired with `Wait()`. Is lack of `Wait()` a decision to make it "async and forget - don't wait for it"? Otherwise, we can use `Run()` which will wait for completion.

I've replaced Start() with Run() to make it synchronous.

test/resources/microshift-process.resource
eslutsky: done

eslutsky:
need to make commits squashed..
/hold

pmtk:
Which test verifies the behavior? If it's the greenboot test, then we might want to run metal periodics job

eslutsky:
> Which test verifies the behavior? If it's the greenboot test, then we might want to run metal periodics job

its verified in the [standard/etcd](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_microshift/2326/pull-ci-openshift-microshift-main-microshift-metal-tests/1701213963905142784/artifacts/microshift-metal-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el92-src@standard-suite/log.html) test
image

ggiguash:
/retitle USHIFT-1671: systemctl does not restart etcd

pkg/controllers/etcd.go
```diff
@@ -116,6 +116,16 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
 		os.Exit(0)
 	}()
 
+	// Ensures microshift-etcd unit stopped after microshift
+	defer func() {
+		cmd := exec.Command("systemctl", "stop", "microshift-etcd.scope")
+
+		if err = cmd.Run(); err != nil {
```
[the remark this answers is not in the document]
eslutsky: @pmtk - great catch !
it appears that systemd  blocks `systemctl stop microshift-etcd.scope'  while `microshift` is also stopping. adding `--no-block` help in this case.

pmtk:
/lgtm
/hold

Unhold when you convert the ticket to OCPBUGS, set version to 4.15.0, and refresh jira ticket

PR #2331, opened 2023-09-11
USHIFT-1682: Update kickstart template to have GPT disk label

praveenkumar:
ISO which is created using the image builder script is not able to run on the windows hyperv-gen2 which need gpt disk label instead msdos and UEFI support.

- https://learn.microsoft.com/en-us/windows-server/security/guarded-fabric-shielded-vm/guarded-fabric-create-a-linux-shielded-vm-template

### DIFF
--- a/docs/contributor/rhel4edge_iso.md
+++ b/docs/contributor/rhel4edge_iso.md
@@ -108,6 +108,8 @@ The artifact of the build is the `_output/image-builder/microshift-installer-${V
 The `kickstart.ks` file is configured to partition the main disk using `Logical Volume Manager` (LVM). Such partitioning is required for the data volume to be utilized by the MicroShift CSI driver and it allows for flexible file system customization if the disk space runs out.
 
 By default, the following partition layout is created and formatted with the `XFS` file system:
+* BIOS boot partition (1MB)
+  * It is required to boot ISO to systems with legacy BIOS while using GPT as the partitioning scheme.
 * EFI partition with EFI file system (200MB)
 * Boot partition is allocated on a 1GB volume
 * The rest of the disk is managed by the `LVM` in a single volume group named `rhel`
@@ -124,9 +126,10 @@ As an example, a 20GB disk is partitioned in the following manner by default.
 $ lsblk /dev/sda
 NAME          MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
 sda             8:0    0   20G  0 disk
-├─sda1          8:1    0  200M  0 part /boot/efi
-├─sda2          8:2    0  800M  0 part /boot
-└─sda3          8:3    0   19G  0 part
+├─sda1          8:1    0    1M  0 part
+├─sda2          8:2    0  200M  0 part /boot/efi
+├─sda3          8:3    0  800M  0 part /boot
+└─sda4          8:4    0   19G  0 part
   └─rhel-root 253:0    0   10G  0 lvm  /sysroot
 
 $ sudo vgdisplay -s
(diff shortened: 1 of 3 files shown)


praveenkumar:
> I don't have an issue taking this change if it doesn't break anything for the developer workflow, 

I tested these changes on linux (RHEL-9.2) and on fedora which I think we are targeting as developer workflow.

> but these tools are really mean to be examples rather than interfaces that we support for other teams or customers

Right, those are merely examples to allow ramp up a developer for right environment. We thought of creating our copies but this way we also test out those examples and keep them up to date.

ggiguash:
/retitle USHIFT-1682: Update kickstart template to have GPT disk label

docs/contributor/rhel4edge_iso.md
```diff
@@ -108,6 +108,7 @@ The artifact of the build is the `_output/image-builder/microshift-installer-${V
 The `kickstart.ks` file is configured to partition the main disk using `Logical Volume Manager` (LVM). Such partitioning is required for the data volume to be utilized by the MicroShift CSI driver and it allows for flexible file system customization if the disk space runs out.
 
 By default, the following partition layout is created and formatted with the `XFS` file system:
+* BIOS boot partition (1MB)
```
ggiguash, line 111: Can we add a short note on why this is required.

scripts/image-builder/config/kickstart.ks.template
```diff
@@ -14,13 +14,15 @@ network --bootproto=dhcp --device=link --activate --onboot=on
 #
```
ggiguash, line 14: Need to fix a comment here as well. Let's keep it as short as possible though.

ggiguash:
I think we have no reason not to allow running on Windows HyperV by default for development purposes, assuming that the existing CI environment does not break with this change.

I'm approving the PR to see if all the tests pass.
/approve

ggiguash:
@praveenkumar, Let's also fix the https://github.com/openshift/microshift/blob/main/docs/config/microshift-starter.ks kickstart file with a similar change to remove the barrier of running the Getting Started procedure on Windows HyperV.

docs/contributor/rhel4edge_iso.md
praveenkumar: @ggiguash Added.

scripts/image-builder/config/kickstart.ks.template
praveenkumar: done.

praveenkumar:
> @praveenkumar, Let's also fix the https://github.com/openshift/microshift/blob/main/docs/config/microshift-starter.ks kickstart file with a similar change to remove the barrier of running the Getting Started procedure on Windows HyperV.

Updated `microshift-starter.ks` also

praveenkumar:
@ggiguash I also want to backport it till 4.13, let me know what I need to do.

ggiguash:
> @ggiguash I also want to backport it till 4.13, let me know what I need to do.

Can you explain why we need this in 4.13? It's an unsupported tech-preview version.

praveenkumar:
> Can you explain why we need this in 4.13? It's an unsupported tech-preview version.

@ggiguash even it is unsupported but these script have value for our CRC project side so either we patch them in our project or directly consume it from upstream (in this case microshift) as mention in https://github.com/openshift/microshift/pull/2331#issuecomment-1713981275 one. I want to avoid local patching and have it here itself. Also does the main sync with `release-4.14` or we need to cherry-pick for it also?

ggiguash:
@praveenkumar, we need to cherrypick separately for any release. Let me ask the team if there are any objections for backporting these changes. If there are no objections, we will open issues in the OCPBUGS project and run the backport.

praveenkumar:
> @praveenkumar, we need to cherrypick separately for any release. Let me ask the team if there are any objections for backporting these changes. If there are no objections, we will open issues in the OCPBUGS project and run the backport.

Sure, Thank you.

PR #2303, opened 2023-09-05
USHIFT-1593: Add MicroShift external certificate verification test on IP change

ggiguash:
Imlements a test for USHIFT-1593

PR #2343, opened 2023-09-13
USHIFT-1593: Add MicroShift external certificate verification test on IP change

pacevedom:
Based on https://github.com/openshift/microshift/pull/2303, includes all its commits

Closes #<Issue Number>


test/resources/microshift-process.resource
```diff
@@ -55,11 +55,7 @@ MicroShift Process ID Changed
     [Documentation]    Compare the specified and current MicroShift process ID
     [Arguments]    ${old_pid}
     ${cur_pid}=    MicroShift Process ID
-    IF    "${cur_pid}"=="${old_pid}"
-        RETURN    False
-    ELSE
-        RETURN    True
-    END
+    Should Not Be Equal    ${cur_pid}    ${old_pid}
```
ggiguash, line 58: This statement will fail the test if not true. The original idea here is to return the current status and retry the function in a loop. 
Why do we need this change?

pacevedom: The function is enclosed in `Wait Until Keyword Succeeds` builtin. According to the [docs](https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Wait%20Until%20Keyword%20Succeeds) this will retry on almost any failure.
The original function was only returning True/False strings, which are not checked by the builtin so it succeeded on the first try, everytime.

ggiguash: Ah, the function did not work as intended. Can you please rebase on top of the latest code in `main`? 
We have added another similar function for Etcd checks.

pacevedom: Fixed in etcd process too. Thanks!

test/resources/DataFormats.py
```diff
@@ -68,6 +68,17 @@ def _merge(dest, addition):
...
 
 
+def _replace(dest, replacement):
+    """Implements data structure field replacement
+
+    The dest value is modified in place by replacing all
+    the values found in replacement parameter.
+    """
+    for key, value in replacement.items():
+        dest[key] = value
```
pmtk, line 78: Can we use https://docs.python.org/3/library/stdtypes.html#dict.update or `d | other` or `d |= other` instead?

test/resources/microshift-process.resource
```diff
@@ -86,6 +82,8 @@ Restart MicroShift
     Systemctl With Retry    restart    microshift.service
     Wait Until MicroShift Process ID Changes    ${ushift_pid}
     Wait Until MicroShift Etcd Process ID Changes    ${ushift_etcd_pid}
+    # Kubeconfig may change between restarts, ensure we are taking the latest one.
+    Setup Kubeconfig
```
pmtk, line 86: This is interesting, is this because of IP change that we want to fetch new kubeconfig?
I'm not sure if that should be a rule for every test.

Changing IP in the kubeconfig is one thing, but kubeconfig getting changed completely is another.

Would you consider this a bug? I would be pretty annoyed if my kubeconfig stopped working all of the sudden on some remote device

test/suites/network/multi-nic.robot
```diff
@@ -113,50 +119,94 @@ Login Switch To IP1
...
+        Wait For MicroShift
+    END
 
-    # Wait for MicroShift API readiness and run verification
-    Wait For MicroShift
     Verify Hello MicroShift LB
-    Verify Hello MicroShift NodePort    ${verify_ip1}    ${verify_ip2}
+    Verify Hello MicroShift NodePort    ${login_ip}
+    Check IP Present In External Certificate    ${login_ip}
+    Check IP Not Present In External Certificate    ${removed_ip}
```
pmtk, line 166: Just to be sure: this verifies that our `config -> certs` functions are correct,
so it's kind of orthogonal to the IP change itself, right? Feels like a unit test that checks if `config.apiServer.subjectAltNames` gets reflected correctly in `server.crt`.

What do you think about, instead of verifying the file, using `openssl` to call KAS' HTTPS endpoint and verifying the certificate that is in fact served?
It verify that how we setup KAS is correctly (I know, it's just a filepath string...) and we'd have full end to end coverage from config to the endpoint.

ggiguash: Not instead, maybe, in addition?

test/resources/microshift-process.resource
pacevedom: This is related to the IP change, yes. A simple restart in between different IPs and calling this keyword may yield an unusable kubeconfig because the `server` part changed.
If the CAs would get rotated (or client certs) the same thing could happen. I would not say its a bug, its like a safety net. Setting up kubeconfig at the beginning does not cover everything.

test/resources/DataFormats.py
pacevedom: Done

test/suites/network/multi-nic.robot
```diff
@@ -31,15 +34,57 @@ Verify MicroShift Runs On Both NICs
...
+    [Setup]    Save Default MicroShift Config
+
+    Setup MicroShift Config    ${USHIFT_HOST_IP1}
+
+    ${cur_pid}=    MicroShift Process ID
+
+    Login Switch To IP    ${USHIFT_HOST_IP1}
+    Nmcli Connection Control    down    ${NIC2_NAME}
+
+    Restart MicroShift
```
pmtk, line 53: Hmm, do we need to restart microshift in we disable one of the interfaces? Does sysconf controller only checks single (primary? how it was selected then) IP?

test/suites/network/multi-nic.robot
```diff
@@ -31,15 +34,57 @@ Verify MicroShift Runs On Both NICs
...
+    Verify MicroShift On One NIC    ${USHIFT_HOST_IP1}    ${USHIFT_HOST_IP2}
+
+    # Rebooting MicroShift host restores the network configuration
+    [Teardown]    Run Keywords
+    ...    Restore Default MicroShift Config
+    ...    Reboot MicroShift Host
+    ...    Login Switch To IP1
+    ...    Wait For Healthy System
 
 Verify MicroShift Runs On Second NIC
```
pmtk, line 64: I'm wondering what is functional difference between two tests so we could name them better (or improve documentation).

They have different flows, but it's not obvious what exactly we're testing: why in one test we need to explicitly restart microshift, but in another it restarts itself.
It would be nice that this difference would be named other than "First NIC" and "Second NIC"

test/suites/network/multi-nic.robot
```diff
@@ -113,50 +158,65 @@ Login Switch To IP1
...
-        Wait Until Keyword Succeeds    30x    10s
-        ...    Access Hello Microshift    ${NP_PORT}    ${ip2}
-    END
+    Wait Until Keyword Succeeds    30x    10s
+    ...    Access Hello Microshift    ${NP_PORT}    ${ip}
 
     [Teardown]    Run Keywords
     ...    Delete Hello MicroShift Pod And Service
 
 Verify MicroShift On One NIC
```
pmtk, line 173:
```suggestion
Verify MicroShift On Single NIC
```

test/suites/network/multi-nic.robot
```diff
@@ -113,50 +158,65 @@ Login Switch To IP1
...
 
-    ${cur_pid}=    MicroShift Process ID
+    Verify Hello MicroShift LB
+    Verify Hello MicroShift NodePort    ${login_ip}
+    Check IP Present In External Certificate    ${login_ip}
+    Check IP Not Present In External Certificate    ${removed_ip}
 
-    Login Switch To IP    ${login_ip}
-    Nmcli Connection Control    down    ${down_nic}
+Setup MicroShift Config
```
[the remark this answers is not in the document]
pacevedom: Done

test/suites/network/multi-nic.robot
```diff
@@ -113,50 +158,65 @@ Login Switch To IP1
...
+
+Check IP Present In External Certificate
+    [Documentation]    Check the specified IP presence in the external
+    ...    certificate file
+    [Arguments]    ${ip}
+
+    ${grep_opt}=    Set Variable    ${EMPTY}
+    Check IP Certificate    ${ip}    ${grep_opt}
+
+Check IP Not Present In External Certificate
```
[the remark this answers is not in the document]
pacevedom: Done

test/suites/network/multi-nic.robot
```diff
@@ -113,50 +158,65 @@ Login Switch To IP1
...
+    ...    alternative names in ${CERT_FILE}.
+    [Arguments]    ${ip}    ${grep_opt}
+
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    ${OSSL_CMD} ${CERT_FILE} | ${GREP_SUBJ_IPS} | grep -w ${grep_opt} 'DNS:${ip}'
+    ...    sudo=True    return_stdout=True    return_stderr=True    return_rc=True
+    Log    ${stderr}
+    Should Be Equal As Integers    ${rc}    0
+
+Check IP Present In External Certificate
```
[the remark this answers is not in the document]
pacevedom: Done

test/suites/network/multi-nic.robot
```diff
@@ -31,15 +34,57 @@ Verify MicroShift Runs On Both NICs
...
     [Documentation]    Verify MicroShift can run on the first NIC
-    Verify MicroShift On One NIC    ${USHIFT_HOST_IP1}    ${NIC2_NAME}    ${USHIFT_HOST_IP1}    ${EMPTY}
+    [Setup]    Save Default MicroShift Config
+
+    Setup MicroShift Config    ${USHIFT_HOST_IP1}
+
+    ${cur_pid}=    MicroShift Process ID
+
+    Login Switch To IP    ${USHIFT_HOST_IP1}
+    Nmcli Connection Control    down    ${NIC2_NAME}
```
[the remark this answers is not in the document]
pacevedom: Done

test/suites/network/multi-nic.robot
```diff
@@ -113,50 +158,65 @@ Login Switch To IP1
...
-    ...    Wait For Healthy System
+Check IP Certificate
+    [Documentation]    Checks whether the ${ip} is present in the subject
+    ...    alternative names in ${CERT_FILE}.
+    [Arguments]    ${ip}    ${grep_opt}
+
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    ${OSSL_CMD} ${CERT_FILE} | ${GREP_SUBJ_IPS} | grep -w ${grep_opt} 'DNS:${ip}'
+    ...    sudo=True    return_stdout=True    return_stderr=True    return_rc=True
+    Log    ${stderr}
```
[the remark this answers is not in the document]
pacevedom: Done

test/suites/network/multi-nic.robot
pacevedom: The restart is required so that the new configuration is picked up again and certificates get renegerated. In this test case the host ip does not change because the interface is untouched.

test/suites/network/multi-nic.robot
pacevedom: Done

test/suites/network/multi-nic.robot
pacevedom: Added some documentation comments and swapped first/second with primary/secondary.

test/suites/network/multi-nic.robot
pmtk: Yes yes, what I meant is the difference in behavior of primary and other interfaces, but I see in code it depends on default route (which says which interface is "primary") - and that's the only one we watch for changes

test/suites/network/multi-nic.robot
```diff
@@ -31,15 +34,66 @@ Verify MicroShift Runs On Both NICs
...
+
+    ${cur_pid}=    MicroShift Process ID
+
+    Login Switch To IP    ${USHIFT_HOST_IP2}
+    Disable Interface    ${NIC1_NAME}
+
+    Wait Until MicroShift Process ID Changes    ${cur_pid}
+    Wait For MicroShift Service
+    Setup Kubeconfig
+    Wait For MicroShift
```
pmtk, line 87: Nit: we could probaby do following
```suggestion
    Wait Until MicroShift Process ID Changes    ${cur_pid}
    Wait For MicroShift
    Setup Kubeconfig
```

We could extract them into another kw: `Wait For Automatic Restart Of MicroShift`

test/suites/network/multi-nic.robot
```diff
@@ -31,15 +34,66 @@ Verify MicroShift Runs On Both NICs
...
+
+    Verify MicroShift On Single NIC    ${USHIFT_HOST_IP2}    ${USHIFT_HOST_IP1}
+
+    # Rebooting MicroShift host restores the network configuration
+    [Teardown]    Run Keywords
+    ...    Restore Default MicroShift Config
+    ...    Reboot MicroShift Host
+    ...    Login Switch To IP1
+    ...    Wait For Healthy System
```
pmtk, line 96: Maybe we could extract these keywords into another keyword: `Restore All Interfaces By Rebooting Host`.
Then we could get rid of `Login Switch To IP1` and call `Login Switch To IP    ${USHIFT_HOST_IP1}` directly

test/suites/network/multi-nic.robot
```diff
@@ -79,24 +132,21 @@ Initialize Nmcli Variables
...
     ${stdout}    ${stderr}    ${rc}=    Execute Command
     ...    ${NMCLI_CMD} | tail -1
     ...    return_stdout=True    return_stderr=True    return_rc=True
-    Log    ${stderr}
     Should Be Equal As Integers    ${rc}    0
     Set Suite Variable    \${NIC2_NAME}    ${stdout}
 
-Nmcli Connection Control
+Disable Interface
     [Documentation]    Run nmcli connection command with arguments
```
pmtk, line 145: Documentation went out of sync a little bit. Also: should we include information that this change is not persistent, i.e. reversible by reboot?

test/suites/network/multi-nic.robot
```diff
@@ -113,50 +163,64 @@ Login Switch To IP1
...
+    ${grep_opt}=    Set Variable    ${EMPTY}
+    Check IP Certificate    ${ip}    ${grep_opt}
+
+IP Should Not Be Present In External Certificate
+    [Documentation]    Check the specified IP absence in the external
+    ...    certificate file
+    [Arguments]    ${ip}
+
+    ${grep_opt}=    Set Variable    '--invert-match'
+    Check IP Certificate    ${ip}    ${grep_opt}
```
pmtk, line 226: Would this work?

```suggestion
    Check IP Certificate    ${ip}    ${EMPTY}

IP Should Not Be Present In External Certificate
    [Documentation]    Check the specified IP absence in the external
    ...    certificate file
    [Arguments]    ${ip}

    Check IP Certificate    ${ip}    '--invert-match'
```

pmtk (commented):
I think the logic is good and it's much more readable now.
I have couple more nits about style mostly; they're more of a question/suggestion than a request

test/suites/network/multi-nic.robot
pacevedom: Thought of this but `Wait For MicroShift` makes use of the kubeconfig, so you need to set it up before that keyword (server IP is changing upon restart). Thats the reasoning behind the current order.
About the keyword, its only used here, is it worth of a new one for a single test?

test/suites/network/multi-nic.robot
pacevedom: Done

test/suites/network/multi-nic.robot
pacevedom: Done

test/suites/network/multi-nic.robot
pacevedom: Done

pmtk:
this looks pretty nice
/lgtm

PR #2345, opened 2023-09-13
USHIFT-1681: Inject target branch name into rebase branch name

copejon:
When the rebase script executes against the latest release-4.X branch and main, it does not generate a unique branch name. This creates a situation where, if 1 of the 2 is pushed, the rebase process running in parallel on the other may pull the remote rebase-branch.  This means that if a rebase onto branch A pushes rebase-branch A', the rebase process running in parallel on branch B can mistakenly pull branch A', rather than create a branch B', because the names of A' == B'. This effectively means branch A is being rebased onto branch B, and includes the commits of the parant branch.  Injecting the onto-branch's name (release-4.x | main) will detangle this situation.

### DIFF
--- a/scripts/auto-rebase/rebase.sh
+++ b/scripts/auto-rebase/rebase.sh
@@ -1098,7 +1098,7 @@ rebase_to() {
     # PULL_BASE_REF is expected in CI environments and should identify the active branch.
     if [[ "${PULL_BASE_REF:-}" == "" ]]; then
         >&2 echo 'Warning: env var PULL_BASE_REF not found or empty, falling back to local active branch.'
-        onto_branch=$(git branch rev-parse --abbrev-ref HEAD)
+        onto_branch=$(git rev-parse --abbrev-ref HEAD)
         if [[ "${onto_branch}" == 'HEAD' ]]; then
             >&2 echo 'Warning: detected HEAD in detached state, falling back to current abbreviated commit'
             onto_branch="$(git rev-parse --short HEAD)"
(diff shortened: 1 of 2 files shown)


scripts/auto-rebase/rebase.sh
```diff
@@ -1094,7 +1094,20 @@ rebase_to() {
     amd64_date="$(cat ${STAGING_DIR}/release_amd64.json | jq -r .config.created | cut -f1 -dT)"
     arm64_date="$(cat ${STAGING_DIR}/release_arm64.json | jq -r .config.created | cut -f1 -dT)"
 
-    rebase_branch="rebase-${ver_stream}_amd64-${amd64_date}_arm64-${arm64_date}"
+    local onto_branch=""
+    # PULL_BASE_REF is expected in CI environments and should identify the active branch.
+    if [[ "${PULL_BASE_REF:-}" == "" ]]; then
+        >&2 echo 'Warning: env var PULL_BASE_REF not found or empty, falling back to local active branch.'
+        onto_branch=$(git branch rev-parse --abbrev-ref HEAD)
```
dhellmann, line 1101:
```suggestion
        onto_branch=$(git rev-parse --abbrev-ref HEAD)
```

This command failed for me locally, but removing the "branch" part gave me what looks like a valid branch name.

It would be good to show the onto_branch value after it is computed, to go with the warning on line 1100.

copejon: Thanks, still running the script in local testing

copejon:
Should be backported so that rebase automation in release branches also applies the naming change.

/cherrypick release-4.14

/cherrypick release-4.13

PR #2351, opened 2023-09-14
OCPBUGS-19384: Use `gomplate` templating engine

pmtk:


pmtk:
Putting on hold because to get this to 4.14, we need OCPBUGS tickets
/hold

dhellmann:
This looks like a good change. The tool offers some options for passing inputs, so we could consider moving some values out of environment variables into a config file if that would make anything easier. There are some other types of files that use sed for replacing values (the kickstart templates are the main ones that come to mind), and we should convert those, too, but that can happen in another PR.

What does the team think about when to merge this?

pmtk:
> This looks like a good change. The tool offers some options for passing inputs, so we could consider moving some values out of environment variables into a config file if that would make anything easier. There are some other types of files that use sed for replacing values (the kickstart templates are the main ones that come to mind), and we should convert those, too, but that can happen in another PR.
> 
> What does the team think about when to merge this?

I kept envvars so when we backport the change to 4.14, it doesn't change too much.
For 4.15 onwards, I'd definitely would try to ditch the envvars and use some explicit args instead

pacevedom (approved):
a nice addition that allows more complex logic, will hold to allow others to take a look/review.
/lgtm
/hold

eggfoobar:
/unhold

Looks great, thanks @pmtk !

PR #2356, opened 2023-09-15
USHIFT-1695: support building a single image in test harness

dhellmann:
Add -t argument to test/bin/build_images.sh so the caller can rebuild
a single image by passing the template file as input.


ggiguash:
Great feature! Happy to approve as soon as the "verify" and build errors are fixed.

PR #2361, opened 2023-09-18
USHIFT-1666: Selinux domain transition test 

eggfoobar:
Updated our context check to include /var/log/kube-apiserver

Adds a domain transition test to our selinux tests to validate that executing a binary can't transition to more privileged files.

### DIFF
--- a/test/resources/selinux.py
+++ b/test/resources/selinux.py
@@ -61,6 +61,10 @@ EXPECTED_FCONTEXT_LIST = [
     "/var/lib/registry(/.*)?",
 ]
 
+SOURCE_TARGET_TRANSITION = {
+    "container_t": ["container_var_lib_t"]
+}
+
 
 def get_expected_ocp_microshift_fcontext_list() -> List[str]:
     return EXPECTED_FCONTEXT_LIST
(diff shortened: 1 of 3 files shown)


jerpeter1:
It looks this like this test doesn't check that executing a binary can't transition to *any* more privileged domain, but it does check if a transition is possible to `container_var_lib_t`. Are there other transition rules where `parent_source` is `container_t`? If so, what are the target domains? Are there corresponding executables that processes running under the `container_t` are allowed to execute which are entrypoints for those target domains?

jerpeter1:
Still needs work.

I ran a few checks on my microshift VM. First, checking what `get_all_traversal_transition_contexts` would return:

```
[microshift@microshift ~]$ sudo sesearch --type_trans -c process -s container_t | awk '{{ print $3 }}'
[microshift@microshift ~]$ sudo semanage fcontext -l | grep abrt_helper_exec_t
/usr/bin/abrt-pyhook-helper                        regular file       system_u:object_r:abrt_helper_exec_t:s0:process
```

So that function would currently return a list e.g. ['abrt_helper_exec_t'].

`abrt_helper_exec_t` is the label of the executable file, not the domain that's transitioned into when running a binary labeled with abrt_helper_exec_t is executed - you should be comparing the permissions of the abrt_helper_t to the permissions of container_t in this example. Spot checking the two, I see that these two domains do have substantially different permissions.

[microshift@microshift ~]$ sudo sesearch -A -s abrt_helper_t | wc -l
345
[microshift@microshift ~]$ sudo sesearch -A -s container_t | wc -l
269

In this case however I can see that policy just has one file being labeled with `abrt_helper_exec_t` 

```
[microshift@microshift ~]$ sudo semanage fcontext -l | grep abrt_helper_exec_t
/usr/bin/abrt-pyhook-helper                        regular file       system_u:object_r:abrt_helper_exec_t:s0
```

... and `/usr/bin/abrt-pyhook-helper` does not exist on microshift systems, so adding a check that looks for labeled files like that might be a good way to close out this PR.

test/resources/selinux.py
```diff
@@ -112,6 +150,127 @@ def run_default_access_check() -> List[str]:
...
+        if "_exec_t" in domain_context:
+            domain_context = source.replace('_exec', '')
+
+        source_number_of_permissions, rc = remote_sudo_rc(f"{allow_cmd} -s {domain_context} | wc -l")
+        BuiltIn().should_be_equal_as_integers(rc, 0)
+
+        if source_number_of_permissions != parent_number_of_permissions:
+            file_path, rc = remote_sudo_rc(f"{semanage_cmd} | grep {source} | awk '{{ print $1 }}'")
+            BuiltIn().should_be_equal_as_integers(rc, 0)
+            file_exists, rc = remote_sudo_rc(f"[ -e {file_path} ] && echo found || echo safe")
```
jerpeter1, line 231: I think this part of the logic will fail if `semanage fcontext -l` returns more than one path that's labeled with {source}.

eggfoobar: Ah, good catch, let me fix that real quick

eggfoobar: Updated

PR #2367, opened 2023-09-19
USHIFT-335: Test mDNS resolution from the hypervisor/test runner

pmtk:


scripts/image-builder/configure.sh
```diff
@@ -16,7 +16,7 @@ fi
 sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
     cockpit-composer bash-completion podman runc genisoimage \
     createrepo yum-utils selinux-policy-devel jq wget lorax rpm-build \
-    containernetworking-plugins expect
+    containernetworking-plugins expect avahi-tools
```
dhellmann, line 19: Why in the image-builder dependencies instead of the developer dependencies?

test/suites/standard/networking-smoke.robot
```diff
@@ -41,3 +43,29 @@ Expose Hello MicroShift Service Via Route
 Delete Hello MicroShift Route
     [Documentation]    Delete route for cleanup.
     Oc Delete    route/hello-microshift -n ${NAMESPACE}
+
+Resolve Hello MicroShift MDNS
+    [Documentation]    Resolve hello-microshift route via mDNS from the hypervisor/RF runner.
+
+    # Open port 5353 on the hypervisor to be able to resolve mDNS hostnames
+    ${result}=    Run Process
+    ...    sudo firewall-cmd --zone\=libvirt --add-service\=mdns
```
dhellmann, line 52: This does not work when running the test against localhost (inside the VM running MicroShift). This is a setup step changing the hypervisor, not a part of the test. We have `manage_hypervisor_config.sh` for requirements like this.

scripts/image-builder/configure.sh
pmtk: Neither of the script felt like a 100% match since to me image-builder == test runner, but changed to another one (guess makes sense when you run tests from dev VM -> dev VM)

test/suites/standard/networking-smoke.robot
pmtk: Moved there

test/suites/standard/networking-smoke.robot
```diff
@@ -67,3 +69,22 @@ Create Hello MicroShift Ingress
 Delete Hello MicroShift Ingress
     [Documentation]    Delete ingress for cleanup.
     Oc Delete    -f ${HELLO_USHIFT_INGRESS} -n ${NAMESPACE}
+
+Resolve Hello MicroShift MDNS
```
dhellmann, line 73: I think I would have called this something like "DNS entry should resolve" and passed the name as an argument. Let's take this version, and I can propose an update and see if you like it.

dhellmann (commented):
/lgtm

test/suites/standard/networking-smoke.robot
dhellmann: See #2386

pmtk: I'm fine with the change :)

pmtk:
/test microshift-metal-tests

Fail in unrelated test (`suites/network/multi-nic.robot`)

test/suites/standard/networking-smoke.robot
pmtk: Included them in this PR

PR #2369, opened 2023-09-19
USHIFT-1711: do not close jira tickets if there are no PRs

dhellmann:
The tool for automatically managing jira tickets from the CLI should
not close a ticket if there is no PR. We can manage those by hand in
the GUI.

/assign @ggiguash


pacevedom (approved):
/lgtm

PR #2370, opened 2023-09-19
OCPBUGS-19422: add test to look for errors in logs

dhellmann:
This commit adds a new test to check for some common
permission-related errors that should never appear in the MicroShift
logs.

/assign @copejon @ggiguash @pmtk @pacevedom

### DIFF
--- a/test/suites/standard/log-scan.robot
+++ b/test/suites/standard/log-scan.robot
@@ -1,0 +1,55 @@
+*** Settings ***
+Documentation       Tests for various log messages we do or do not want.
+
+Resource            ../../resources/common.resource
+Resource            ../../resources/microshift-process.resource
+Library             ../../resources/journalctl.py
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+Test Tags           restart    slow
+
+
+*** Variables ***
+${CURSOR}       ${EMPTY}    # The journal cursor before restarting MicroShift
+
+
+*** Test Cases ***
+Should Not Find Forbidden
+    [Documentation]    Logs should not say "forbidden"
+    Pattern Should Not Appear In Log Output    ${CURSOR}    forbidden
+
+Should Not Find Cannot Patch Resource
+    [Documentation]    Logs should not say "cannot patch resource"
+    Pattern Should Not Appear In Log Output    ${CURSOR}    cannot patch resource
+
+Services Should Not Timeout When Stopping
+    [Documentation]    Logs should not say "Timed out waiting for services to stop"
+    Pattern Should Not Appear In Log Output    ${CURSOR}    Timed out waiting for services to stop
+
+Should Find Etcd Is Ready
+    [Documentation]    Logs should say "etcd is ready"
+    Pattern Should Appear In Log Output    ${CURSOR}    etcd is ready
+
+Should Find MicroShift Is Ready
+    [Documentation]    Logs should say "microshift is ready"
+    Pattern Should Appear In Log Output    ${CURSOR}    microshift is ready
+
+
+*** Keywords ***
+Setup
+    [Documentation]    Test suite setup
+    Check Required Env Variables
+    Login MicroShift Host
+    Setup Kubeconfig    # for readiness checks
+    # Save the journal cursor then restart MicroShift so we capture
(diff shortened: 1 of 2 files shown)


test/suites/standard/log-scan.robot
```diff
@@ -0,0 +1,49 @@
...
+
+*** Keywords ***
+Setup
+    [Documentation]    Test suite setup
+    Check Required Env Variables
+    Login MicroShift Host
+    Setup Kubeconfig    # for readiness checks
+    ${cursor}=    Get Journal Cursor
+    Set Suite Variable    \${CURSOR}    ${cursor}
+    Restart MicroShift
```
pmtk, line 44: Is this sequence of actions used to also get shutdown logs?

dhellmann: Honestly, I was just trying to avoid writing separate stop and start keywords, but you make a good point. It will have the effect of capturing the shutdown and startup messages so I've added a [retcon](https://en.wikipedia.org/wiki/Retroactive_continuity) comment indicating that it's on purpose.

test/suites/standard/log-scan.robot
```diff
@@ -0,0 +1,51 @@
...
+    [Documentation]    Logs should not say "cannot patch resource"
+    Pattern Should Not Appear In Log Output    ${CURSOR}    cannot patch resource
+
+Should Find Etcd Is Ready
+    [Documentation]    Logs should say "etcd is ready"
+    Pattern Should Appear In Log Output    ${CURSOR}    etcd is ready
+
+Should Find MicroShift Is Ready
+    [Documentation]    Logs should say "microshift is ready"
+    Pattern Should Appear In Log Output    ${CURSOR}    microshift is ready
```
pacevedom, line 33: Do we want to run these at the end of the other suites? Asking because I think it would be interesting as some sort of pre-suite-teardown instead of standalone tests. You would catch a wider variety of logs, as right now the test suite execution order is randomized.

test/suites/standard/log-scan.robot
```diff
@@ -0,0 +1,51 @@
...
+Suite Teardown      Teardown
+
+Test Tags           restart    slow
+
+
+*** Variables ***
+${CURSOR}       ${EMPTY}    # The journal cursor before restarting MicroShift
+
+
+*** Test Cases ***
```
pmtk, line 18: I'd add `Timed out waiting for services to stop` ([src](https://github.com/openshift/microshift/blob/69ca1dd7625190814b97d70091c6d9b62be391a6/pkg/cmd/run.go#L265)) and perhaps `systemd[1]: microshift.service: start operation timed out. Terminating.` to catch when we start exceeding deadlines. Though the latter is more suitable for Pablo's idea.
We should be more strict in tests than in prod (e.g.: checking for pod restarts via greenboot in tests is fine to catch problems, but maybe overkill in prod)

test/suites/standard/log-scan.robot
dhellmann: Today I'm mostly interested in looking for errors just from starting the service itself, not things triggered by the tests. I expect any test that might trigger an error to do its own check (the keywords here are all reusable for that).

test/suites/standard/log-scan.robot
dhellmann: I can do the first one, since we emit that. The other is a systemd message and I'll have to do some research about how to collect that since it won't be in the MicroShift logs. Can we do that in a separate ticket?

pmtk: Of course

test/suites/standard/log-scan.robot
pacevedom: What about executing this one right after https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L454-L461 ? That way you always ensure you are catching both the first start and the whole test. A test doing restarts might escape these checks because of randomization.

dhellmann: I'm trying to avoid the global invariant pattern. This suite checks that we don't generate a few specific errors unrelated to any tests. After we merge it, we will have the keywords in place to allow each test suite to check conditions relevant to it, so that if we start seeing failures we know why we're checking those messages.

PR #2376, opened 2023-09-20
OCPBUGS-19488: Gracefully handle legacy version schema

pmtk:


dhellmann (commented):
We don't need an RF test, but a unit test would be good.

pkg/admin/prerun/version.go
```diff
@@ -248,14 +248,28 @@ func getVersionFile() (versionFile, error) {
...
 	verFile := versionFile{}
-	if err := json.Unmarshal(versionFileContents, &verFile); err != nil {
+	jsonErr := json.Unmarshal(contents, &verFile)
+	if jsonErr == nil {
+		return verFile, nil
+	}
+
+	// Unmarshalling version as json failed - fallback to using previous version file schema
+	verMetadata, fallbackErr := versionMetadataFromString(string(contents))
+	if fallbackErr != nil {
```
dhellmann, line 263: This error case could use a test, otherwise everything looks good.

dhellmann (commented):
/lgtm

PR #2378, opened 2023-09-20
USHIFT-1628: Test MicroShift if offline environment

copejon:


test/resources/microshift-host.resource
```diff
@@ -109,3 +109,12 @@ Change Hostname
     Should Be Equal As Integers    0    ${rc}
 
     RETURN    ${old_hostname}
+
+Get Nic Name For Ip
+    [Documentation]     Uses pattern matching to find and return the name of a NIC for a given IP address
+    [Arguments]     ${ip}
+    ${stdout}  ${rc}=     Execute Command    ip -o a | awk '/${ip}/{sub(/:/,"", $2); print $2}'
```
dhellmann, line 116: We have access to python in RF, can we use that instead of awk?

test/resources/serial_console.resource
```diff
@@ -0,0 +1,43 @@
...
+    [Arguments]     ${command}
+    Send Line    sudo ${command}
+    ${index}=   Expect    ${{ [$PROMPT_PATTERN, pexpect.TIMEOUT, pexpect.EOF] }}   timeout=60s
+    Should Be Equal As Integers     ${index}    0
+
+Login To Guest
+    ${index}=   Expect   ${{ ['^.*login:', r'$PROMPT_PATTERN', pexpect.TIMEOUT, pexpect.EOF] }}
+    Should Not Be Equal As Integers     ${index}    2
+    Should Not Be Equal As Integers     ${index}    3
+    IF    ${index} == 0
```
dhellmann, line 22: The case for index = 1 isn't handled.

What do all of these index values mean? What's special about non-zero values?

test/resources/serial_console.resource
```diff
@@ -0,0 +1,43 @@
+*** Settings ***
+Library     PexpectLibrary
+Library    serial_console.py
+
+Resource    serial_console.resource
+
+
+*** Variables ***
+${PROMPT_PATTERN}   \\[["\\w"@\\-~\\ ]+\\]\\$\\ *
```
dhellmann, line 9: What if we set the PS1 value for the redhat user in our images to simplify this?

test/image-blueprints/group3/rhel92-source-isolated.toml
```diff
@@ -38,6 +42,14 @@ enabled = ["mdns", "ssh", "http", "https"]
...
 sources = ["10.42.0.0/16", "169.254.169.1"]
 
+[[customizations.files]]
+path = "/etc/sysconfig/qemu-ga"
+data = """
+# This is a systemd environment file, not a shell script.
+# It provides settings for "/lib/systemd/system/qemu-guest-agent.service".
+BLOCK_RPCS=
+FSFREEZE_HOOK_PATHNAME=/etc/qemu-ga/fsfreeze-hook"""
+
```
ggiguash, line 52: Can we comment in the file on why we need these settings?

test/scenarios/el92-src@offline.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+guest_name=host1
```
ggiguash, line 4: Nit: can we use uppercase since this is a global variable? 
i.e. OFFLINE_VM_NAME

copejon:
The ci_iso_build_phase.sh script is not composing `group3` for some reason.  There's not error that indicates why.  The ci_iso_build_phase.sh is also reporting a false positive success status for the offline scenario launch_vm call, which fails with a 'file not found' error.  That error doesn't cause the script to exit however.  I'll TAL at the ci scripts and see what's up.

copejon:
Oh, ofc. It's not building the ISO b/c group3 is built conditionally, when `CI_JOB_NAME =~ .*periodic.*`.  I'm going to add a temporary line to the ci_phase_iso_build.sh script to unconditionally build group3.  I'll post a follow up PR that we can merge immediately after this PR merges.

test/scenarios/el92-src@offline.sh
ggiguash, line 1: This should be moved to scenarios-periodic, right? 
Unless we'd like to add offline test as a presubmit, but I would not recommend it.

ggiguash: Also, it does not agree with the current setup of groups for building the offline ISO.

copejon: +1 I'll rework this

test/image-blueprints/group4/rhel92-source-isolated.toml
```diff
@@ -38,6 +42,14 @@ enabled = ["mdns", "ssh", "http", "https"]
 name = "trusted"
 sources = ["10.42.0.0/16", "169.254.169.1"]
 
+[[customizations.files]]
+path = "/etc/sysconfig/qemu-ga"
+data = """
+# This is a systemd environment file, not a shell script.
+# It provides settings for "/lib/systemd/system/qemu-guest-agent.service".
+BLOCK_RPCS=
+FSFREEZE_HOOK_PATHNAME=/etc/qemu-ga/fsfreeze-hook"""
```
ggiguash, line 51: Can we add to the comment the reason for setting these variables?

test/bin/scenario.sh
```diff
@@ -533,6 +533,32 @@ EOF
         "$@"
 }
 
+# Public function to run the tests for an offline scenario.  Greenboot checks should be handled
+# in RF suite setup, using the qemu-guest-agent.py library.
+run_offline_tests(){
```
ggiguash, line 538: We now have a way to switch off greenboot checks from the scenario. Do we still need this function?

test/scenarios-periodics/el92-src@offline.sh
```diff
@@ -1,18 +1,19 @@
...
 
 # Sourced from scenario.sh and uses functions defined there.
-
 scenario_create_vms() {
-    prepare_kickstart host1 kickstart-offline.ks.template ""
-    # Create a VM with embedded container images and 0 NICs
-    launch_vm host1 rhel-9.2-microshift-source-isolated "" "" "" "" 0
+    prepare_kickstart "host1" kickstart-offline.ks.template ""
+    #         vmname         boot_blueprint                      network_name vm_vcpus vm_memory vm_disksize vm_nics
+    launch_vm "host1"  rhel-9.2-microshift-source-isolated ""           ""       ""        ""          0
```
ggiguash, line 7: Nit: we do not usually align and comment on arguments. We have a task to make those position independent

test/bin/scenario.sh
copejon: The comment could be better worded.  It's just that `run_tests` executes a greenboot check, but also that it expects a number of files contain network data to exist.  I'd opted to add the simpler function that doesn't include any of the network dependencies rather than nest most of the `run_test` func under a large `if` block.

test/scenarios-periodics/el92-src@offline.sh
copejon: +1, I'll remove them

test/image-blueprints/group4/rhel92-source-isolated.toml
copejon: +1

test/bin/scenario.sh
ggiguash: I would not want to hold the PR off because of this, but it's "easy" to detect the offline case by checking for an empty "ip" file, and, then just skipping the unnecessary settings. We do this in run scenario function. 
We can implement this enhancement in the next PR. Please resolve this comment.

test/image-blueprints/group4/rhel92-source-isolated.toml
ggiguash: I'm not sure the comment explains why we need those settings. Could we please elaborate a bit more?
The idea is for someone to look at the code and understand why these 2 are applied.

test/resources/libvirt.resource
```diff
@@ -10,6 +10,7 @@ Reboot MicroShift Host
     [Documentation]    Reboot the MicroShift host.
     [Arguments]    ${vm_name}
     ${result}=    Virsh    ${vm_name}    reboot
+    Wait For MicroShift Host To Reboot    ${vm_name}
```
ggiguash, line 13: Do we have a race condition here? 
The `MicroShift Host Should Be Running` checks for the `running` value of `domstate`. Does it change immediately before the `virsh reboot` command returns?
P.S. We had this problem when rebooting hosts using SSH. We solved it by verifying that the `bootid` is different.

test/suites/network/offline.robot
```diff
@@ -0,0 +1,212 @@
...
+    END
+
+Restart MicroShift
+    [Documentation]    Restart the MicroShift systemd service
+    ${result}    ${ignore}=    Wait Until Keyword Succeeds    5x    5s
+    ...    Run Guest Process    ${GUEST_NAME}    systemctl    restart    microshift.service
+    Log Many    ${result["stdout"]}    ${result["stderr"]}    ${result["rc"]}
+    Should Be Equal As Integers    ${result["rc"]}    0
+
+Restart Greenboot Healthcheck
```
ggiguash, line 151: Is this better named as `Restart Greenboot And Wait Health Check To Exit`?
Otherwise, I had to look in the implementation to understand if there's a wait for success.

test/resources/libvirt.resource
copejon: Good observation.  I think you're right. If the state changes, it's too quick (or too slow) for it to be useful for detecting reboot.  The test follows this keyword with a poll on the qemu-ga guest-agent endpoint, which is probably why a race hasn't cropped up. The window b/w the reboot call and the start of the polling is small enough to not have been hit (yet).

Will drop the "wait" portion of this and look at using the bootid.

test/image-blueprints/group4/rhel92-source-isolated.toml
pmtk, line 1: Not quite sure why are these files getting moved from group3 to group4?

test/bin/scenario.sh
```diff
@@ -580,6 +580,32 @@ EOF
...
+
+    "${rf_binary}" \
+        --name "${SCENARIO}" \
+        --randomize all \
+        --loglevel TRACE \
+        --outputdir "${SCENARIO_INFO_DIR}/${SCENARIO}" \
+        -x junit.xml \
+        "$@"
+}
+
```
pmtk, line 608: This isn't exactly the place but: I don't think we'll get sos report for the test, right? Any plans to include it?

test/resources/qemu-guest-agent.py
```diff
@@ -306,6 +312,115 @@ def start_guest_process(vm_name: str, cmd: str, *args) -> int:
...
+    information on the guest-file-read API call. See https://qemu-project.gitlab.io/qemu/interop/qemu-ga-ref.html#qapidoc-54
+    for more information on the guest-file-close API call.
+    """
+    handle = _open_file(vm_name, path, 'r')
+    try:
+        content = _execute(vm_name, {
+            'execute': 'guest-file-read',
+            'arguments': {
+                'handle': handle,
+                'count': 40960
```
dhellmann, line 365: We may need to expose this parameter, but if we're just reading small configuration files for now maybe not yet.

dhellmann (commented):
This is coming along nicely

test/image-blueprints/group4/rhel92-source-isolated.toml
ggiguash: Yes, the comments look good. I requested the update before it was added.

test/bin/scenario.sh
copejon: This is true. It'll take some additional logic to collect it without the network.  I don't think it'd be difficult.  I agree it's not in the scope of this pr.

test/bin/scenario.sh
copejon: done

test/suites/network/offline.robot
```diff
@@ -0,0 +1,212 @@
...
+    Restore Default MicroShift Config
+    Restore Guest Environment
+
+Configure MicroShift For Offline
+    [Documentation]    Update the MicroShift config to use the offline-network endpoint
+    Save Default MicroShift Config
+    ${patch}=    Catenate    SEPARATOR=\n
+    ...    node:
+    ...    \ \ hostnameOverride: ${GUEST_NAME}
+    ...    \ \ nodeIP: 10.44.0.1
```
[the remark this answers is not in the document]
copejon: It's a good question.  I don't have the networking knowledge to provide a good answer though.
@dhellmann, is using the 10.44.01/32 cidr a requirement, or is like Patryk suggests and MicroShift is piggy-backing off a behavior he and I aren't aware of?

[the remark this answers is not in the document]
dhellmann: I think @pliurh just picked a CIDR range that was not already in use by MicroShift. I'm not sure that it actually matters what the value is, though.

PR #2391, opened 2023-09-22
OCPBUGS-19563: Move healthcheck script to /usr/lib/greenboot

pmtk:
When greenboot runs scripts, it considers two locations `/usr/lib/greenboot` and `/etc/greenboot`. It also runs `required.d` and `wanted.d` scripts from each path in
sequence, that is:
- (1) `/usr/lib/greenboot/checks/required.d`
- (2) `/usr/lib/greenboot/checks/wanted.d`
- (3) `/etc/greenboot/checks/required.d`
- (4) `/etc/greenboot/checks/wanted.d`

At the moment of writing this commit, greenboot shipped in RHEL unconditionally exits with rc=1 when required.d scripts failed.

Because MicroShift's scripts were placed in the (3) directory, they would not run if any script from (1) failed.

In upstream greenboot the behavior has already changed: all scripts will run, regardless of the path or importance (required vs wanted), and then greenboot will exit with appropriate return code.

Only healthcheck script is moved, because greenboot considers two mentioned paths for checks, it only uses `/etc/greenboot` for green and red scripts (that is until fix to run all checks before exiting makes into package for RHEL, as it also extends possible paths for green and red scripts to `/usr/lib/greenboot`).

Related greenboot PR: https://github.com/fedora-iot/greenboot/pull/97

Additional checks used in tests are not relocated, because `/usr/lib/` is immutable endpoint and would require remounting the filesystem to place files there. For that reason, these test assets will continue to reside in `/etc/greenboot/checks` as it's root writable.

### DIFF
--- a/docs/contributor/greenboot.md
+++ b/docs/contributor/greenboot.md
@@ -23,9 +23,9 @@ at this time.
 
 Proceed by creating a health check script in the `/etc/greenboot/check/required.d`
 directory.
-> The name prefix of the user script should be chosen to make sure it runs after
-> the `40_microshift_running_check.sh` script, which implements the MicroShift
-> health check procedure for its core services.
+> The name prefix of the user script should be chosen appropriately.
+> Because MicroShift's healthcheck resides in `/usr/lib/greenboot/check/require.d`,
+> it will run before any checks in `/etc/greenboot/check/require.d`.
 
 ```
 SCRIPT_FILE=/etc/greenboot/check/required.d/50_busybox_running_check.sh
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -244,8 +244,8 @@ install -m644 packaging/selinux/microshift.pp.bz2 %{buildroot}%{_datadir}/selinu
 install -d -m755 %{buildroot}%{_datadir}/microshift/functions
 install -p -m644 packaging/greenboot/functions.sh %{buildroot}%{_datadir}/microshift/functions/greenboot.sh
 
-install -d -m755 %{buildroot}%{_sysconfdir}/greenboot/check/required.d
-install -p -m755 packaging/greenboot/microshift-running-check.sh %{buildroot}%{_sysconfdir}/greenboot/check/required.d/40_microshift_running_check.sh
+install -d -m755 %{buildroot}%{_prefix}/lib/greenboot/check/required.d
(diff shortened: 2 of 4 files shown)


pmtk:
```
$ rpm -qlp microshift-greenboot-4.14.0_0.nightly_2023_09_12_195514_20230922084623_fd3310ea3-1.el9.noarch.rpm

/usr/lib/greenboot/check/required.d/40_microshift_running_check.sh
/usr/lib/greenboot/green.d/40_microshift_set_healthy.sh
/usr/lib/greenboot/red.d/40_microshift_pre_rollback.sh
/usr/lib/greenboot/red.d/40_microshift_set_unhealthy.sh
/usr/share/microshift/functions/greenboot.sh
```

pacevedom (approved):
just as clarification, greenboot exits when any of the required.d scripts fail. The execution order means that if any of the scripts in (1) fail, microshift scripts wont even run, as it sits in (3).

/lgtm

packaging/rpm/microshift.spec
```diff
@@ -244,8 +244,8 @@ install -m644 packaging/selinux/microshift.pp.bz2 %{buildroot}%{_datadir}/selinu
 install -d -m755 %{buildroot}%{_datadir}/microshift/functions
 install -p -m644 packaging/greenboot/functions.sh %{buildroot}%{_datadir}/microshift/functions/greenboot.sh
 
-install -d -m755 %{buildroot}%{_sysconfdir}/greenboot/check/required.d
-install -p -m755 packaging/greenboot/microshift-running-check.sh %{buildroot}%{_sysconfdir}/greenboot/check/required.d/40_microshift_running_check.sh
+install -d -m755 %{buildroot}%{_prefix}/lib/greenboot/check/required.d
```
dhellmann, line 247: You should update the changelog section of this file, too.

pmtk: You're right. Updated

docs/contributor/greenboot.md
```diff
@@ -23,9 +23,9 @@ at this time.
 
 Proceed by creating a health check script in the `/etc/greenboot/check/required.d`
 directory.
-> The name prefix of the user script should be chosen to make sure it runs after
-> the `40_microshift_running_check.sh` script, which implements the MicroShift
-> health check procedure for its core services.
+> The name prefix of the user script should be chosen appropriately.
+> Because MicroShift's healthcheck resides in `/var/lib/greenboot/check/require.d`,
```
pacevedom, line 27: Should this be /usr/lib instead of /var/lib?

pmtk: Yes!

PR #2393, opened 2023-09-22
OCPBUGS-19632: Rework automated backup and restore

pmtk:


pkg/admin/prerun/prerun.go
```diff
@@ -34,6 +34,108 @@ type dataManagement struct {
...
+	if err != nil {
+		return fmt.Errorf("loading version metadata failed: %w", err)
+	}
+
+	if currentDeploymentID == versionFile.DeploymentID {
+		// Previous and current boots are the same deployment, do not restore
+		return nil
+	}
+
+	// Deployment changed between boots - restore the backup
```
dhellmann, line 134: We assume that if the deployment changed and we have a backup that means this is a rollback? Or at least if it isn't a rollback, the user manually re-deployed an old image that we've seen before?

What would it look like if we are more strict about checking the ostree status for an actual rollback via greenboot rather than the manual case? What are the failure scenarios?

pkg/admin/prerun/version.go
```diff
@@ -27,6 +28,10 @@ type versionFile struct {
 	BootID       string          `json:"boot_id"`
 }
 
+func (hi *versionFile) BackupName() data.BackupName {
+	return data.BackupName(fmt.Sprintf("%s_%s", hi.DeploymentID, hi.BootID))
+}
```
pmtk, line 33: There'a an alternative for getting IDs for previous boot. boot ID from journalctl. Deployment ID is a bit controversial though:
```
[redhat@repro-host1 ~]$ sudo journalctl  -u ostree-prepare-root
Sep 22 05:18:50 localhost systemd[1]: Starting OSTree Prepare OS/...
Sep 22 05:18:50 localhost ostree-prepare-root[600]: preparing sysroot at /sysroot
Sep 22 05:18:50 localhost ostree-prepare-root[600]: Resolved OSTree target to: /sysroot/ostree/deploy/rhel/deploy/7c5bfecf8acf6e547d39efc9ae15f4e1b32df134cb4d2727ffae1c786784cc4f.0
Sep 22 05:18:50 localhost ostree-prepare-root[600]: filesystem at /sysroot currently writable: 0
Sep 22 05:18:50 localhost ostree-prepare-root[600]: sysroot.readonly configuration value: 0
Sep 22 05:18:50 localhost systemd[1]: Finished OSTree Prepare OS/.
```

pkg/admin/prerun/prerun.go
pmtk: > the user manually re-deployed an old image that we've seen before?

If it was gone from the system, then backups would also be gone. But if deployment was pinned and re-staged, then, right now, it would be treated as rollback and restore a backup.

I'm not sure we could make it more strict by using rpm-ostree status alone:
after re-staging pinned deployment, it would become active, and formerly active would be now a rollback. So version/health.deploymentID == rollback and we have backup for active.

We could have some script executed when deployment is staged ([this is how greenboot does it](https://github.com/fedora-iot/greenboot/blob/main/usr/lib/systemd/system/greenboot-grub2-set-counter.service#L13)) so we could differentiate between staging pinned deployment and rollback (it doesn't stage, grub2 selects 2nd boot entry, and after booting greenboot fixes the deployment with `rpm-ostree rollback` without reboot), but that's new scripts, new services, etc.

There is a command `rpm-ostree ex history` but is warns: `NOTICE: Experimental commands are subject to change.` so I'm not sure we should use it... [Beside it's parsing journal :)](https://github.com/coreos/rpm-ostree/blob/main/rust/src/history.rs) so maybe it's not so bad?

> What would it look like if we are more strict about checking the ostree status for an actual rollback via greenboot rather than the manual case?

[greenboot knows system rolled back by checking if](https://github.com/fedora-iot/greenboot/blob/main/usr/libexec/greenboot/greenboot-rpm-ostree-grub2-check-fallback#L6) `boot_counter == -1`, but clears that var shortly after.

Given `Before=ostree-finalize-staged.service` greenboot vs manual rollback:
- greenboot's first boots 2nd boot entry, then fixes the deployments
- manual first stages (?) rollback deployment (so ostree-finalize-staged runs), then reboots

pmtk: Extra info on `greenboot-rpm-ostree-grub2-check-fallback`. Greenboot rollback vs normal boot (perhaps also manual rollback)

```
-- Boot abfa7a2fb8174618a8d4a47842168804 --
Sep 22 14:55:11 repro-host1 systemd[1]: Starting Check for fallback boot...
Sep 22 14:55:11 repro-host1 systemd[1]: Finished Check for fallback boot.
Sep 22 14:55:22 repro-host1 systemd[1]: greenboot-rpm-ostree-grub2-check-fallback.service: Deactivated successfully.
Sep 22 14:55:22 repro-host1 systemd[1]: Stopped Check for fallback boot.


-- Boot 68c40b837ac745f982bc92ec734c975d --
Sep 22 14:55:33 repro-host1 systemd[1]: Starting Check for fallback boot...
Sep 22 14:55:33 repro-host1 greenboot-rpm-ostree-grub2-check-fallback[780]: Moving '42805fb9cc04ceb834ee66eb176ad1ef6ac4fca47db2c829f7951190653f4f9f.0' to be first deployment
Sep 22 14:55:36 repro-host1 greenboot-rpm-ostree-grub2-check-fallback[780]: Transaction complete; bootconfig swap: yes; bootversion: boot.1.1, deployment count change: 0
Sep 22 14:55:36 repro-host1 greenboot-rpm-ostree-grub2-check-fallback[780]: Changes queued for next boot. Run "systemctl reboot" to start a reboot
Sep 22 14:55:36 repro-host1 greenboot-rpm-ostree-grub2-check-fallback[763]: FALLBACK BOOT DETECTED! Default rpm-ostree deployment has been rolled back.
Sep 22 14:55:36 repro-host1 greenboot-rpm-ostree-grub2-check-fallback[763]: Health check logs from previous boot:
Sep 22 14:55:36 repro-host1 greenboot-rpm-ostree-grub2-check-fallback[763]: Script '40_microshift_running_check.sh' FAILURE (exit code '1'). Continuing...
Sep 22 14:55:36 repro-host1 greenboot-rpm-ostree-grub2-check-fallback[763]: Script 'fail.sh' FAILURE (exit code '1'). Continuing...
Sep 22 14:55:36 repro-host1 systemd[1]: Finished Check for fallback boot.
Sep 22 14:58:13 repro-host1 systemd[1]: greenboot-rpm-ostree-grub2-check-fallback.service: Deactivated successfully.
Sep 22 14:58:13 repro-host1 systemd[1]: Stopped Check for fallback boot.

```

pmtk: One more thought: we cannot assume that devices will be configured with persistent journal - we might have only logs from current boot, right?

dhellmann: > > the user manually re-deployed an old image that we've seen before?
> 
> If it was gone from the system, then backups would also be gone. But if deployment was pinned and re-staged, then, right now, it would be treated as rollback and restore a backup.

OK, I think we want to try to detect the difference between the automated rollback and the manual change and only restore when the rollback was automated. Think about all the reasons other than MicroShift that the images might be different. Preserving the data will allow a user to force the system back to a specific commit and save the database if the MicroShift version is the same, or manually revert if they need to because the database is not compatible.

> I'm not sure we could make it more strict by using rpm-ostree status alone: after re-staging pinned deployment, it would become active, and formerly active would be now a rollback. So version/health.deploymentID == rollback and we have backup for active. I.e. by using `rpm-ostree status` alone we cannot know if deployment switch happen just now, or in the past. By adding our version/health.json, we can decide that it happened just now, but I don't see how we'd know if it was greenboot rollback, manual rollback, or re-staging pinned deploy.

OK, maybe the ostree team can help?

> We could have some script executed when deployment is staged ([this is how greenboot does it](https://github.com/fedora-iot/greenboot/blob/main/usr/lib/systemd/system/greenboot-grub2-set-counter.service#L13)) so we could differentiate between staging pinned deployment and rollback (it doesn't stage, grub2 selects 2nd boot entry, and after booting greenboot fixes the deployment with `rpm-ostree rollback` without reboot), but that's new scripts, new services, new files to persist data across boots etc.

Yeah, not ideal at this stage in development. Let's not rule it out, though.

Could we use our existing red script to record the rollback action? I guess that's pretty similar to the health status we're recording now, isn't it?

> There is a command `rpm-ostree ex history` but is warns: `NOTICE: Experimental commands are subject to change.` so I'm not sure we should use it... [Beside it's parsing journal :)](https://github.com/coreos/rpm-ostree/blob/main/rust/src/history.rs) so maybe checking journal by ourselves is not so bad?

If they change the log message format I would expect the command to be updated to cope with that, so it might provide a more consistent API over time.

dhellmann: > One more thought: we cannot assume that devices will be configured with persistent journal - we might have only logs from current boot, right?

Possibly, although it's going to be very hard to administer a system with no persistent logging. We do document that users should enable it.

pmtk: I just remembered with already have [this](https://github.com/openshift/microshift/blob/54a850c951d17d96130c433414c1f98ffdbe8e04/packaging/greenboot/microshift-pre-rollback.sh#L38-L44) implemented.
If we only want to restore a backup in case of automated rollback and just proceed with the data we have, then this should be good for this.
For manual, admin issued `rpm-ostree rollback`, we would leave the decision in hand of admin: either restore backup before rolling back (because that might be 4.15 -> 4.14, or the embedded manifests changed significantly), or don't restore it (because it might be 4.14.5 -> 4.14.4, and they want to keep the etcd and other data, but there's bug in the microshift binary itself)

pkg/admin/prerun/health.go
```diff
@@ -26,42 +21,31 @@ type HealthInfo struct {
...
-		return nil, errHealthFileDoesNotExist
+// updateHealthInfo updates health.json files with hardcoded "healthy" and
+// deployment and boot IDs from provided argument.
+func updateHealthInfo(vf versionFile) error {
+	// health.json in kept in place to support rollback to 4.14.0~rc.1 and earlier,
+	// as lack of the file would result in MicroShift neither backing up the data
+	// nor restoring a backup and we want to always create a backup.
+	// We should be able to remove the health.json file in 4.16 or even 4.15
+	// if we state that upgrading from rc.0 and rc.1 directly to 4.15 is not supported,
+	// but should be done with intermediate step to released 4.14.0.
```
dhellmann, line 32: We should not plan on being able to have a restriction like that, unfortunately.

test/suites/upgrade/upgrade-block-2-minor.robot
```diff
@@ -61,7 +65,7 @@ Journal Should Have Information That MicroShift Skipped Restoring
     IF    ${version.minor} == 13    RETURN
 
     ${stdout}    ${rc}=    Execute Command
-    ...    journalctl --unit=microshift | grep "Restore skipped - data directory already matches backup to restore"
+    ...    journalctl --unit=microshift | grep "${EXPECTED_LOG}"
```
dhellmann, line 68: journalctl has a --grep option, will that work here?

dhellmann (commented):
I'll look at this in more detail tomorrow. I have a few small comments inline. I'm also a little bit worried that changing from waiting for greenboot to exit instead of being healthy will allow race conditions or flakes in the test suite that will be hard to debug because an unhealthy system will rollback while the test suite is still running.

eslutsky:
this PR will be continued here https://github.com/openshift/microshift/pull/2424

dhellmann:
/close

We will take #2424 instead

PR #2397, opened 2023-09-25
OCPBUGS-19719: Use cgo net for name resolution

pacevedom:
When using .local hostnames and mDNS the native golang resolver does not work properly as it only reads the /etc/resolv.conf The cgo based resolver calls the C library routines instead. See https://pkg.go.dev/net#hdr-Name_Resolution

Closes #<Issue Number>

### DIFF
--- a/test/suites/standard/hostname.robot
+++ b/test/suites/standard/hostname.robot
@@ -12,8 +12,8 @@ Test Tags           restart    slow
 
 
 *** Variables ***
-${NEW_HOSTNAME}    microshift.local
-${OLD_HOSTNAME}    ${EMPTY}
+${NEW_HOSTNAME}     microshift.local
+${OLD_HOSTNAME}     ${EMPTY}
 
 
 *** Test Cases ***
@@ -25,6 +25,7 @@ Verify local name resolution
     Oc Logs    namespace="openshift-ingress"    opts="deployment/router-default"
 
     [Teardown]    Restore Old Hostname
+
 
 *** Keywords ***
 Setup
--- a/test/suites/standard/hostname.robot
+++ b/test/suites/standard/hostname.robot
@@ -1,0 +1,57 @@
+*** Settings ***
+Documentation       Tests verifying hostname resolution
+
+Resource            ../../resources/microshift-process.resource
+Resource            ../../resources/microshift-host.resource
+Resource            ../../resources/oc.resource
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+Test Tags           restart    slow
+
+
+*** Variables ***
+${NEW_HOSTNAME}    microshift.local
+${OLD_HOSTNAME}    ${EMPTY}
+
+
+*** Test Cases ***
+Verify local name resolution
+    [Documentation]    Verify correct name resolution through mDNS
+    [Setup]    Configure New Hostname
+
+    Oc Wait For    deployment/router-default    condition\=Available    120s    ns=openshift-ingress
+    Oc Logs    namespace="openshift-ingress"    opts="deployment/router-default"
+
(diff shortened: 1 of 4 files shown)


pacevedom:
/hold
Test pending

eslutsky:
will this test work without any mdns configuration on the  node ?

test/suites/standard/hostname.robot
```diff
@@ -0,0 +1,58 @@
...
+Test Tags           restart    slow
+
+
+*** Variables ***
+${NEW_HOSTNAME}     microshift.local
+${OLD_HOSTNAME}     ${EMPTY}
+
+
+*** Test Cases ***
+Verify local name resolution
```
eslutsky, line 20: it would be more descriptive  to include mdns in the  test name.

pacevedom:
> will this test work without any mdns configuration on the node ?

Yes, this is entirely handled by microshift.

test/suites/standard/hostname.robot
pacevedom: Was trying to refer more to the local domain the hostname has in this test. The mDNS detail is within the documentation part. WDYT?

eslutsky: well adding `.local` to the domain will trigger chain of events in the cluster (exposing mDNS service ie), if we only wants to test the end result (.local resolution) that its fine as it is.

test/suites/standard/hostname.robot
```diff
@@ -0,0 +1,58 @@
...
+    [Documentation]    Verify correct name resolution through mDNS
+    [Setup]    Configure New Hostname
+
+    Oc Wait For    deployment/router-default    condition\=Available    120s    ns=openshift-ingress
+    Oc Logs    namespace="openshift-ingress"    opts="deployment/router-default"
```
ggiguash, line 25: Is this just for debugging?

ggiguash: Or, we should be expecting a certain output?

test/suites/standard/hostname.robot
```diff
@@ -0,0 +1,58 @@
...
+    [Documentation]    Configure old hostname again in the MicroShift host.
+    Setup Hostname    ${OLD_HOSTNAME}
+
+Setup Hostname
+    [Documentation]    Setup a new hostname and return the old one.
+    [Arguments]    ${hostname}
+    IF    "${hostname}"=="${EMPTY}"    RETURN
+    ${old}=    Change Hostname    ${hostname}
+    Cleanup MicroShift    --all    --keep-images
+    Restart MicroShift
```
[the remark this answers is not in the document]
pacevedom: MicroShift honors the first hostname as node name and keeps it through restarts, ignoring changes. To force a different hostname you need to remove the file where it stores this info, which is `/var/lib/microshift/.nodename`.
By cleaning up and then restarting you ensure the new node name will be picked up.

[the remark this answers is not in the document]
ggiguash: We have functions to make sure that the service PID changes

test/suites/standard/hostname.robot
pacevedom: The actual bug consists in not being able to retrieve logs from a host configured with `.local` suffix hostname.
When issuing a request to the apiserver asking for logs, it will go check the node IP using the name and thats when mDNS resolution needs to happen.
We dont really care about the logs contents, we care about being able to retrieve them (any container, actually).

test/suites/standard/hostname.robot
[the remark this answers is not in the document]
pacevedom: There is no automatic restart on hostname change, hence the cleanup and then the restart. Actually, that restart is already checking the pid change.

PR #2401, opened 2023-09-26
Never remove data, allow upgrade from unhealthy

pmtk:


pmtk:
Decided to go with: https://github.com/openshift/microshift/pull/2393

PR #2412, opened 2023-09-29
USHIFT-1737: fix: add wait for microshift serice to make sure health check succeeds

eggfoobar:
Seems like we might have a timing problem between when the teardown of kustomize happens and when we validate microshift is up

### DIFF
--- a/test/suites/standard/kustomize.robot
+++ b/test/suites/standard/kustomize.robot
@@ -241,6 +241,7 @@ Restore Default Config
     ${is_ostree}=    Is System OSTree
     IF    ${is_ostree}
         Reboot MicroShift Host
+        Wait For MicroShift Service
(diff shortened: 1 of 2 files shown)


test/suites/standard/kustomize.robot
```diff
@@ -241,6 +241,7 @@ Restore Default Config
     ${is_ostree}=    Is System OSTree
     IF    ${is_ostree}
         Reboot MicroShift Host
+        Wait For MicroShift Service
```
dhellmann, line 244: Maybe we should do this inside "Wait For MicroShift" so we do it everywhere we restart MicroShift?

eggfoobar: That make sense to me, updated.

PR #2416, opened 2023-09-29
USHIFT-1705: hostname/kustomize tests: re-enable microshift.service after cleanup

pmtk:
Because `hostname.robot` was cleaning up MicroShift, but not re-enabling it, when kustomize.robot rebooted the ostree host MicroShift was not starting


pacevedom (approved):
/lgtm
For a subsequent PR I would make that enable part of the cleanup keyword too.

PR #2423, opened 2023-10-02
NO-ISSUE: Enahance ci image build debuggability with osbuild logs

copejon:
Adds a generic trap to ci_phase_iso_build.sh to grab the osbuild-worker@[0-9]+.service and osbuild-composer.service journals.  At present, build failures in CI are opaque and difficult to identify.


copejon:
strike that, work was done already in build_images.sh :D

PR #2424, opened 2023-10-02
OCPBUGS-19632: Rework automated backup and restore

eslutsky:
follow-up  PR to https://github.com/openshift/microshift/pull/2393


pkg/admin/prerun/health.go
```diff
@@ -26,42 +21,31 @@ type HealthInfo struct {
...
-		return nil, errHealthFileDoesNotExist
+// updateHealthInfo updates health.json files with hardcoded "healthy" and
+// deployment and boot IDs from provided argument.
+func updateHealthInfo(vf versionFile) error {
+	// health.json in kept in place to support rollback to 4.14.0~rc.1 and earlier,
+	// as lack of the file would result in MicroShift neither backing up the data
+	// nor restoring a backup and we want to always create a backup.
+	// We should be able to remove the health.json file in 4.16 or even 4.15
+	// if we state that upgrading from rc.0 and rc.1 directly to 4.15 is not supported,
+	// but should be done with intermediate step to released 4.14.0.
```
dhellmann, line 32: We should not plan on being able to have a restriction like that, unfortunately. It's quite likely that in the future we are going to drop the requirement to upgrade one minor release at a time, too. We should update this comment to say that we need to keep this code in place until 4.14 is no longer supported.

test/resources/ostree.resource
```diff
@@ -65,7 +65,6 @@ System Is Running Right Ref And Healthy
 
     Should Be Equal As Strings    ${expected_deploy}    ${current_deploy}
     Greenboot Health Check Exited
-    System Should Be Healthy
```
dhellmann, line 68: I'm a little bit worried that removing this health check will allow race conditions or flakes in the test suite that will be hard to debug because an unhealthy system will rollback while the test suite is still running.

Do we need different behaviors in different test cases?

test/suites/upgrade/rollback-manual.robot
```diff
@@ -12,50 +12,56 @@ Test Tags           ostree
...
-    Run With Kubeconfig    oc create configmap -n default rollback-test
+    Oc Create    configmap -n default rollback-test
 
     Rollback System And Verify    ${upgrade_deploy_backup}    ${initial_deploy_id}
     ${rolled_back_deploy_backup}=    Get Future Backup Name For Current Boot
 
-    # Check that `oc get`'s error code is 1.
-    Run Keyword And Expect Error    1 != 0
-    ...    Oc Get    configmap    default    rollback-test
+    Oc Get    configmap    default    rollback-test
```
dhellmann, line 49: The intent here isn't very clear. A keyword like "ConfigMap Should Exist" wrapping the oc get call would improve that.

test/suites/upgrade/rollback-manual.robot
```diff
@@ -12,50 +12,56 @@ Test Tags           ostree
...
+    Oc Create    configmap -n default rollback-test-2
 
     Rollback System And Verify    ${rolled_back_deploy_backup}    ${upgrade_deploy_id}
     Oc Get    configmap    default    rollback-test
+    Oc Get    configmap    default    rollback-test-2
```
dhellmann, line 54: Same as above.

test/suites/upgrade/upgrade-block-2-minor.robot
```diff
@@ -61,7 +65,7 @@ Journal Should Have Information That MicroShift Skipped Restoring
     IF    ${version.minor} == 13    RETURN
 
     ${stdout}    ${rc}=    Execute Command
-    ...    journalctl --unit=microshift | grep "Restore skipped - data directory already matches backup to restore"
+    ...    journalctl --unit=microshift | grep "${EXPECTED_LOG}"
```
dhellmann, line 68: The `--grep` option to journalctl would make this 1 command.

test/suites/upgrade/upgrade-fails-and-rolls-back.robot
```diff
@@ -25,13 +25,18 @@ New Deployment Is Consistently Unhealthy And Rolls Back
...
 
+    Oc Create    configmap -n default unhealthy-upgrade-test
+
     TestAgent.Add Action For Next Deployment    every    ${REASON}
     Deploy Commit Expecting A Rollback    ${FAILING_REF}
+
     Wait For MicroShift Service
     Backup Should Exist    ${backup}
+    Expected Boot Count    5
+    Oc Get    configmap    default    unhealthy-upgrade-test
```
dhellmann, line 39: Same comment about intent

test/suites/upgrade/upgrade-block-2-minor.robot
eslutsky: it appears that `--grep`  in journalctl will ignore the log case, adding also `--case-sensitive=true`

test/resources/ostree.resource
eslutsky: after this change   greenboot-healthcheck will  be still expected to be successfull, it removes the extra step of checking the `health.json`  state

dhellmann: "Greenboot Health Check Exited" does not check that the system it health, only that the health checks are no longer running. https://github.com/openshift/microshift/blob/main/test/resources/ostree-health.resource#L10-L20

eslutsky: it expects substate of [exited](https://github.com/openshift/microshift/blob/main/test/resources/ostree-health.resource#L20)  which considers that check is succefful.

dhellmann: OK, I didn't assume that "exited" meant "exited successfully"

dhellmann (commented):
/lgtm

I like this simplification and focus on trying to detect the automatic rollback state rather than all of the other potential edge cases.

PR #2430, opened 2023-10-03
OCPBUGS-20037: Fix journal entries to belong to greenboot unit

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/packaging/greenboot/functions.sh
+++ b/packaging/greenboot/functions.sh
@@ -201,7 +201,7 @@ function print_failure_logs() {
         if [ -f "${file}" ]; then
             echo "Failure log in: '${file}'"
             echo "------"
-            echo "$(<${file})"
+            echo "$(<"${file}")"
             echo "------"
         else
             echo "Info: Log file '${file}' does not exist"
--- a/packaging/greenboot/functions.sh
+++ b/packaging/greenboot/functions.sh
@@ -3,6 +3,7 @@
 # Functions used by MicroShift in Greenboot health check procedures.
 # This library may also be used for user workload health check verification.
 #
+SCRIPT_NAME=$(basename "$0")
 SCRIPT_PID=$$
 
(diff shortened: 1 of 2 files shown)


packaging/greenboot/microshift-running-check.sh
```diff
@@ -19,6 +19,7 @@ trap 'script_exit' EXIT
 # The term handler to override the default behavior and have a uniform and
 # homogeneous exit code in all controlled situations.
 function return_failure() {
+    echo "Signal received, terminating."
```
ggiguash, line 22: From the handler name, it looks like it's also invoked on "return 1" from functions. 
I'm not sure it's an appropriate message in all cases.

dhellmann (commented):
It would be good to have a test case for this change.

pacevedom:
/hold
Repurposing the bug because more failures were found

packaging/greenboot/functions.sh
```diff
@@ -145,7 +156,7 @@ function namespace_pods_ready() {
 
     # Terminate the script in case more pods are ready than expected - nothing to wait for
     if [ "${tcount}" -gt "${ct}" ] ; then
-        echo "The number of ready pods in the '${ns}' namespace is greater than the expected '${ct}' count. Terminating..."
+        echo "The number of ready pods in the '${ns}' namespace is greater than the expected '${ct}' count. Terminating..." | tee >(systemd-cat -t "${SCRIPT_NAME}")
```
dhellmann, line 159: Can you add a comment explaining why we need this here for this echo, but not for any of the other echo calls?

pacevedom: I included a comment in the same file, above, as its something general: https://github.com/openshift/microshift/pull/2430/files#diff-63bfc037903c0ae357d2459b936ee63776f356d70dd3c4921e2d8e0608009770R13-R21

dhellmann: Is it possible to do this at the point where this function is called, instead of inside the function?

PR #2441, opened 2023-10-05
OCPBUGS-19363: No update of CoreDNS configmap for systemd-resolved

ggiguash:
Derived from #2362 for fixing the e2e test to work both on ostree and non-ostree systems.

Closes: OCPBUGS-19363

Co-Authored With @pliurh


test/suites/standard/systemd-resolved.robot
```diff
@@ -0,0 +1,120 @@
...
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+
+*** Variables ***
+${KUBELET_CONFIG_FILE}      /var/lib/microshift/resources/kubelet/config/config.yaml
+
+
+*** Test Cases ***
+Verify Kubelet Config With and Wihtout Systemd-Resolved
```
dhellmann, line 19: Can we split this into 2 tests so we can tell from the failure which mode is not working?

ggiguash:
/hold
Putting this PR on hold until we get guidelines from our Networking experts (cc @pliurh)

The current test checks the following scenarios:
- `systemd-resolved` installed and enabled
- `systemd-resolved` not installed

Do we need to check for the case when `systemd-resolved` is installed, but disabled (default when installing the RPM)?
If so, how do we expect MicroShift to behave in this case?

test/suites/standard/systemd-resolved.robot
ggiguash: Currently, the test needs to be run in this order because we start with a situation with `systemd-resolved` installed. 
Can you recommend on how to enforce the order if we split the tests?

dhellmann: We usually writes tests to enforce their own preconditions with setup steps. So in the test where we want to ensure it works when the package is installed, have a setup to install it that does not fail if it is already installed. In the test where we want to ensure it works when the package is not installed, do the opposite. Then the tests can call a common keyword to ensure that DNS resolution works as expected.

ggiguash: The problem is that we cannot really install this package unless we create a special commit for it for ostree configuration. I did not want to further slow down the image build procedure because of this simple test.

eggfoobar:
/retest-required

The initial failure in metal was due to the python package we saw. That seems to have been resolved.

test/suites/standard/systemd-resolved.robot
dhellmann: Why can we uninstall the package, but not install it?

pliurh:
> Do we need to check for the case when `systemd-resolved` is installed, but disabled (default when installing the RPM)? If so, how do we expect MicroShift to behave in this case?

Microshift only checks the existence of file `/run/systemd/resolve/resolv.conf`. This file will be created when systemd-resolved.service is enabled, but it will not be removed when the service is disabled or the rpm is uninstalled. So if the package is installed but not enabled, kubelet will keep using `/etc/resolv.conf` of pods.

test/suites/standard/systemd-resolved.robot
ggiguash: - We can easily install/uninstall on a non-ostree system using `dnf`.
- On an ostree system:
  - we can easily uninstall using `rpm-ostree override`.
  - we can also install using the same `rpm-ostree override`, but we need the RPM file, which we cannot get easily using "standard" repos

So, we should either include the package in a commit, or copy RPM to the system without installing it.
Is there anything I'm missing?

ggiguash:
> Microshift only checks the existence of file `/run/systemd/resolve/resolv.conf`. This file will be created when systemd-resolved.service is enabled, but it will not be removed when the service is disabled or the rpm is uninstalled. So if the package is installed but not enabled, kubelet will keep using `/etc/resolv.conf`.

Here are the results of my tests. 
- When `systemd-resolved` is installed but not enabled (first boot)
```
$ systemctl is-enabled systemd-resolved
disabled
$ ls -l /run/systemd/resolve/resolv.conf
ls: cannot access '/run/systemd/resolve/resolv.conf': No such file or directory
```

- After `systemd-resolved` is enabled
```
$ systemctl is-enabled systemd-resolved
enabled
$ ls -l /run/systemd/resolve/resolv.conf
-rw-r--r--. 1 systemd-resolve systemd-resolve 819 Oct  7 04:09 /run/systemd/resolve/resolv.conf
```

- After `systemd-resolved` is disabled again, the file remains
```
$ systemctl is-enabled systemd-resolved
disabled
$ ls -l /run/systemd/resolve/resolv.conf
-rw-r--r--. 1 systemd-resolve systemd-resolve 819 Oct  7 04:09 /run/systemd/resolve/resolv.conf
```

- After reboot with `systemd-resolved` disabled, the file is removed
```
$ systemctl is-enabled systemd-resolved
disabled
[redhat@el92-src-standard-suite-host1 ~]$ ls -l /run/systemd/resolve/resolv.conf
ls: cannot access '/run/systemd/resolve/resolv.conf': No such file or directory
```

ggiguash:
@pliurh , based on the above tests, it seems that `systemd-resolved` service keeps the `/run/systemd/resolve/resolv.conf` file consistent after the reboot.

I will add this as a test case before unholding the PR.

test/suites/standard/systemd-resolved.robot
ggiguash: I was able to split the test sequence into 3 distinct tests now that we understand what we should expect.

ggiguash:
/unhold
Fixed the e2e test to implement 3 distinct checks as requested.

test/suites/osconfig/systemd-resolved.robot
```diff
@@ -0,0 +1,167 @@
...
+    Login MicroShift Host
+
+Teardown
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+
+Systemd-Resolved Must Be Installed And Disabled
+    [Documentation]    Verify the systemd-resolved package is installed
+
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
```
dhellmann, line 89: This step will fail if the tests are run out of order, right? I don't see anything installing the package if it is missing.

dhellmann: Could we just run this test on an RPM-based system where we can add and remove packages easily?

ggiguash: It's a keyword, so when it is called, the environment should be appropriate both for ostree and rpm systems. 
I do not think we need to fall back to rpm-only in this case.

ggiguash: The assumption is that the package is installed because of this change: 
https://github.com/openshift/microshift/pull/2441/files/52e579fd21ce61369ba14bfbd42c9383fdd59354#diff-fa38d74133cee0980c67ee278c2e94c103b03d54fff633f4825b978c7d88707aR29

dhellmann: The tests run in a random order, though, and some of the other tests remove the package.

dhellmann: OK, I missed line 72.

PR #2443, opened 2023-10-05
USHIFT-1745: testing instructions for disconnected operation

dhellmann:


docs/user/howto_test_offline.md
```diff
@@ -0,0 +1,126 @@
...
+
+1. Start by preparing a VM or host in the normal way and installing
+   the version of MicroShift you want to test.
+
+2. Stop MicroShift and clean up the OVN data.
+
+```
+systemctl stop microshift
+systemctl stop kubepods.slice
+/usr/bin/microshift-cleanup-data --ovn
```
ggiguash, line 29: Do we need to explicitly stop microshift and pods? Aren't they stopped by the script?

docs/user/howto_test_offline.md
```diff
@@ -0,0 +1,126 @@
...
+
+```
+nmcli connection down $IFACE
+nmcli connection modify $IFACE connection.autoconnect no
+```
+
+5. Ensure the host has a persistent name
+
+```
+hostnamectl hostname $NAME
```
ggiguash, line 50: Shoud we set `NAME` and `IP` variables here as they are used a few times below?

docs/user/howto_test_offline.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+6. Add a fake IP on the loopback interface.
+
+```
+nmcli con add type loopback con-name stable-microshift ifname lo ip4 10.44.0.1/32
```
ggiguash, line 56: Do you think `loopback-microshift` would better depict the purpose of this interface?

docs/user/howto_test_offline.md
```diff
@@ -0,0 +1,126 @@
...
+9. Add the name from step 5 to the MicroShift config file as the
+   node.hostnameOverride setting.
+
+10. Add the IP from step 6 to the MicroShift config file as the
+    node.nodeIP setting.
+
+11. Restart microshift
+
+```
+systemctl start microshift
```
ggiguash, line 83: We should probably use `systemctl enable --now microshift` because it has been disabled by the cleanup script.

docs/user/howto_test_offline.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+8. Add an entry for the node's hostname from step 5 in /etc/hosts
+   using the IP from step 6.
+
+```
+echo “$IP $NAME” >> /etc/hosts
+```
+
+9. Add the name from step 5 to the MicroShift config file as the
```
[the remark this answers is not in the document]
dhellmann: I don't really want to try to predict the right way to edit YAML from the command line. Text editors are a thing, I think it's fine to just tell them to set the necessary value.

docs/user/howto_test_offline.md
```diff
@@ -0,0 +1,126 @@
...
+   node.hostnameOverride setting.
+
+10. Add the IP from step 6 to the MicroShift config file as the
+    node.nodeIP setting.
```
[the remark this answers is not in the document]
dhellmann: Same as above.

docs/user/howto_test_offline.md
dhellmann: I'll add some text to explain what they are.

dhellmann (commented):
@ggiguash This PR is just to record the manual testing instructions that were actually used in a public place. I don't actually expect anyone to follow these steps exactly as written. I expect a user will configure their system with the end-result of these commands, using either a kickstart or a blueprint. We will cover that approach in the product docs.

dhellmann:
@ggiguash I've address the comments and this is ready for another review.

PR #2454, opened 2023-10-09
USHIFT-1725: warn the user when the logging level is unrecognized

dhellmann:
Be more explicit about warning the user about a logging configuration
mistake.

/assign @eggfoobar

### DIFF
--- a/pkg/config/debugging.go
+++ b/pkg/config/debugging.go
@@ -1,6 +1,13 @@
 package config
 
-import "strings"
+import (
+	"fmt"
+	"strings"
+)
+
+// Use the upper-case version of the word to match the kubebuilder
+// default.
+const defaultLogLevel = "Normal"
 
 type Debugging struct {
 	// Valid values are: "Normal", "Debug", "Trace", "TraceAll".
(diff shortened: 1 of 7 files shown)


pkg/config/debugging.go
```diff
@@ -9,15 +16,32 @@ type Debugging struct {
...
+var logLevelNames = map[string]int{
+	"normal":   2,
+	"debug":    4,
+	"trace":    6,
+	"traceall": 8,
+}
+
+// cleanLoggingSetting validates the logging setting and saves a
+// warning if there is an issue.
+func (c *Config) cleanLoggingSetting() {
```
eggfoobar, line 28: What do you think about changing this to `validateLoggingSetting()` or something similar? Cleaning gives the impression of something being removed

eggfoobar (commented):
Looks good, just small rename suggestion

pkg/config/debugging.go
dhellmann: That's a good point. I renamed it `computeLoggingSetting()` to match the other compute function and I moved the invocation into `updateComputedValues()` to align with when recomputed things end up changing the content of the data structure.

PR #2469, opened 2023-10-13
NO-ISSUE: Correct minor issues found while reviewing howto_test_offline

dhensel-rh:

Closes #<Issue Number>

### DIFF
--- a/docs/user/howto_test_offline.md
+++ b/docs/user/howto_test_offline.md
@@ -11,6 +11,12 @@ Deployments](howto_offline_containers.md) for instructions for
 preparing the ostree image with the necessary container images
 embedded when they cannot be pulled over the network.
 
+## Setup
+
+The commands below are meant to be exeuted as the root user. Please
+login as root. Make sure the appropriate kubeconfig file is available
+in your environment so the oc client commands can run as expected
+
 ## MicroShift Configuration
 
 Kubernetes expects to have some basic network settings on the host for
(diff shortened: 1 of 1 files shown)


docs/user/howto_test_offline.md
```diff
@@ -27,7 +33,8 @@ operation.
 /usr/bin/microshift-cleanup-data --ovn
 ```
 
-3. Determine the interface on the host that has the default route
+3. Determine the interface on the host that has the default route.
+   Make sure you only have one default route configured
```
dhellmann, line 37: I think it just takes the first one if there are multiples, so maybe we should adjust for that by making the command on line 40 something like 

```
IFACE=$(ip route | head -n 1 | grep default | awk '{print $5}')
```

dhellmann (commented):
While you're in here, maybe we can fix the IFACE detection logic properly?

dhellmann:
This looks good. I'll give a lgtm when you remove the draft flag.

dhensel-rh:
@dhellmann It is not mentioned in the writeup, but what guidance do we provide the end user to execute these commands ? I use a 'virsh console <VM>' session to perform the steps. If someone is unaware, they might use SSH, and saw on the wrong end of the branch.

docs/user/howto_test_offline.md
```diff
@@ -11,6 +11,12 @@ Deployments](howto_offline_containers.md) for instructions for
 preparing the ostree image with the necessary container images
 embedded when they cannot be pulled over the network.
 
+## Setup
+
+The commands below are meant to be exeuted as the root user. Please
+login as root. Make sure the appropriate kubeconfig file is available
+in your environment so the oc client commands can run as expected
```
dhellmann, line 18:
```suggestion
in your environment so the oc client commands can run as expected.

Warning: These instructions involve disabling the network interface on the host. You will need to access the host directly, not via ssh. For VMs, `virsh console` is useful.
```

dhellmann (commented):
/lgtm

PR #2475, opened 2023-10-17
USHIFT-1766: Use mirror for unreleased RPMs for prev release

pmtk:
Reuse existing `get_crel_version_repo.sh` to obtain previous release most recent version and right mirror repository if RPMs are not published yet to RHEL repos.

### DIFF
--- a/test/bin/get_rel_version_repo.sh
+++ b/test/bin/get_rel_version_repo.sh
@@ -13,6 +13,8 @@ get_candidate_repo() {
 
 check_if_repo_exists() {
 	local -r repo="${1}"
+
+	local code
 	code=$(curl --silent --location --output /dev/null --write-out "%{http_code}" "${repo}/repodata/repomd.xml")
 	if [ "${code}" == "404" ]; then
 		return 1
@@ -24,12 +26,14 @@ check_if_repo_exists() {
 get_current_release_from_candidates() {
 	local -r minor="${1}"
 
+	local rc_repo
 	rc_repo=$(get_candidate_repo "${minor}" false)
 	if check_if_repo_exists "${rc_repo}"; then
 		echo "${rc_repo}"
 		return 0
 	fi
 
+	local ec_repo
 	ec_repo=$(get_candidate_repo "${minor}" true)
 	if check_if_repo_exists "${ec_repo}"; then
 		echo "${ec_repo}"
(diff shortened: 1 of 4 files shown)


test/bin/get_rel_version_repo.sh
```diff
@@ -58,30 +60,29 @@ get_current_release_from_sub_repos() {
...
 # http mirrors, and packages in the ocp repository.
 #
 # Failed attempt to obtain version/repo is not a failure, at the start of each release,
 # there might not be an EC yet. In such case variables exported below might be empty,
 # and such osbuild source, blueprint, and test scenarios should be skipped.
-get_crel_version_repo() {
+get_rel_version_repo() {
 	local -r minor="${1}"
 
 	CURRENT_RELEASE_VERSION=$(get_current_release_from_sub_repos "${minor}")
```
dhellmann, line 73: If this script is sourced and used for different settings, we should probably make all of the variables in these functions locals to avoid stomping over the values in the script that imports this code.

test/package-sources/microshift-prel.toml
```diff
@@ -0,0 +1,9 @@
+{{- if env.Getenv "PREVIOUS_RELEASE_REPO" "" -}}
```
dhellmann, line 1: Will we ever have a situation where we can't set up this package source?

dhellmann:
/hold

See my comments

test/package-sources/microshift-prel.toml
pmtk: Yes, currently, when running 4.14 tests we get 4.13 from `rhocp-4.13`.

When 4.14 is released, we'll switch to `rhocp-4.14` to get newer `oc`, `crun/runc`, and probably some other deps, and ushift 4.14 as well.
This file is only used to add `mirror.openshift.com` repo, but when `rhocp-4.14` includes the version we want, it won't be added anymore to osbuild (until 4.16 and no 4.15 releases).

It's worth noting that `get_rel_version_repo.sh` checks if `rhocp-4.${minor}-for-rhel-9-x86_64-rpms` repo is enabled on the build host, so dev/configure_vm.sh also needs to be updated. So there's a bit of weird dependency going on.

One more thing: dnf reports that rhocp-4.14 exists, but when I enabled it and tried fetching the metadata I got 403. I'd like to break that dependency since osbuild repos are supposed to be independent from system's, but dynamically selecting right `rhocp-4.Y` for osbuild might be too much to fix current problem

dhellmann: I need more time to think through the long-term strategy, but let's do that separately. I think this change improves the situation.

dhellmann:
/lgtm

Remove the hold when you're ready to merge

PR #2487, opened 2023-10-19
USHIFT-1778: update golang to 1.20.10

dhellmann:
* update golang check in rpm spec file to look at each part of the
  version string separately
* update rpm spec file to require go 1.20.10
* update golang in dev tool setup to 1.20.10

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -9,7 +9,8 @@
 }
 
 # golang specifics
-%global golang_version 1.20.3
+# Needs to match go.mod go directive
+%global golang_version 1.20
 #debuginfo not supported with Go
 %global debug_package %{nil}
 # modifying the Go binaries breaks the DWARF debugging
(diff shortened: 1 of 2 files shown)


pacevedom:
Test is failing because target build-base-branch will clone and `make rpm` release-4.15 branch, which does not have this change yet.

I tested the code to work with the failing combination:
```bash
$ cat /tmp/version.sh
version_at_least () {
    if [[ $1 == $2 ]]
    then
        return 0
    fi
    local IFS=.
    local i ver1=($1) ver2=($2)
    # fill empty fields in ver1 with zeros
    for ((i=${#ver1[@]}; i<${#ver2[@]}; i++))
    do
        ver1[i]=0
    done
    for ((i=0; i<${#ver1[@]}; i++))
    do
        if [[ -z ${ver2[i]} ]]
        then
            # fill empty fields in ver2 with zeros
            ver2[i]=0
        fi
        if ((10#${ver1[i]} > 10#${ver2[i]}))
        then
            return 0
        fi
        if ((10#${ver1[i]} < 10#${ver2[i]}))
        then
            return 1
        fi
    done
    return 0
}

version_at_least "1.20.10" "1.20.3" && echo ok || echo failed

$ bash /tmp/version.sh 
ok
```

dhellmann:
This will depend on #2489

dhellmann:
> Test is failing because target build-base-branch will clone and `make rpm` release-4.15 branch, which does not have this change yet.

I think you mean 4.14? See #2489

dhellmann:
I've updated the PR to remove the check in the spec file entirely. The compiler should tell us if the go.mod file has a version it can't support.

pacevedom (commented):
/lgtm

PR #2495, opened 2023-10-20
USHIFT-1778: Fix rpm spec version comparison

pacevedom:
Since the go directive in go.mod is advisory only until 1.21 the enforcement happens when building the rpm.

Closes #<Issue Number>

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -9,7 +9,8 @@
 }
 
 # golang specifics
-%global golang_version 1.20.3
+# Needs to match go.mod go directive
+%global golang_version 1.20
 #debuginfo not supported with Go
 %global debug_package %{nil}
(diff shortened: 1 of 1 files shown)


pmtk:
> Since the go directive in go.mod is advisory only until 1.21 the enforcement happens when building the rpm.

So when we update to 1.21, we plan to remove this check altogether and defer to `go build` for verification?

dhellmann (commented):
/lgtm

dhellmann:
> > Since the go directive in go.mod is advisory only until 1.21 the enforcement happens when building the rpm.
> 
> So when we update to 1.21, we plan to remove this check altogether and defer to `go build` for verification?

Yes

dhellmann:
The metal job passed but Prow isn't noticing that it's done.

/override microshift-metal-tests

PR #2502, opened 2023-10-23
USHIFT-1777: Add rpm checks after building

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/packaging/rpm/make-rpm.sh
+++ b/packaging/rpm/make-rpm.sh
@@ -24,6 +24,7 @@ GITHUB_SHA="${GIT_SHA:0:7}"
 TARBALL_FILE="microshift-${GITHUB_SHA}.tar.gz"
 DEFAULT_RPMBUILD_DIR="$(git rev-parse --show-toplevel)/_output/rpmbuild/"
 RPMBUILD_DIR="${RPMBUILD_DIR:-${DEFAULT_RPMBUILD_DIR}}"
+RPM_INFO_DIRS=""
 
 title() {
     echo -e "\E[34m\n# $1\E[00m";
@@ -68,6 +69,13 @@ EOF
   rpmbuild --quiet ${RPMBUILD_OPT} --define "_topdir ${RPMBUILD_DIR}" --define "_binary_payload w19T8.zstdio" "${RPMBUILD_DIR}"SPECS/microshift.spec
 }
 
+print_info() {
+  title "RPM info"
+  for dir in ${RPM_INFO_DIRS}; do
+    find "${RPMBUILD_DIR}${dir}" -type f -exec sh -c 'i=$1; echo "${i}" && rpm -qip --dump "${i}" && echo' shell {} \;
+  done
+}
+
 usage() {
   echo "Usage: $(basename "$0") <all | rpm | srpm> < local | commit <commit-id> >"
   exit 1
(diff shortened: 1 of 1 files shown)


packaging/rpm/make-rpm.sh
```diff
@@ -68,6 +69,13 @@ EOF
   rpmbuild --quiet ${RPMBUILD_OPT} --define "_topdir ${RPMBUILD_DIR}" --define "_binary_payload w19T8.zstdio" "${RPMBUILD_DIR}"SPECS/microshift.spec
 }
 
+print_info() {
+  title "RPM info"
+  for dir in ${RPM_INFO_DIRS}; do
+    find "${RPMBUILD_DIR}${dir}" -type f -exec sh -c 'i=$1; echo "${i}" && rpm -qip --dump "${i}" && echo' shell {} \;
```
pmtk, line 75: Can you provide a log of how this looks like in practice?

dhellmann: The output will look something like https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/44677/rehearse-44677-pull-ci-openshift-microshift-release-4.14-test-rpm/1716393559587819520/artifacts/test/build-log.txt

The idea is to verify that `make rpm` actually builds something.

Can we do something to report an error if no RPM files are found? Ideally `make rpm` would have already failed, but just in case.

pacevedom: Keep in mind that this is part of the same script that builds the RPMs, its just a post step running only if an env var is defined.
Do we foresee any situation in which the [rpmbuild](https://github.com/openshift/microshift/blob/main/packaging/rpm/make-rpm.sh#L68) will build nothing and not fail?

As for the output, it looks just like what Doug linked but with an additional title.

packaging/rpm/make-rpm.sh
```diff
@@ -76,10 +101,24 @@ usage() {
 [ $# -lt 2 ] && usage
 
 case $1 in
-  all)  RPMBUILD_OPT=-ba ;;
-  rpm)  RPMBUILD_OPT=-bb ;;
-  srpm) RPMBUILD_OPT=-bs ;;
-  *)    usage
+  all)
+    RPMBUILD_OPT=-ba
+    RPM_INFO_DIRS="RPMS SRPMS"
```
ggiguash, line 106: `RPM_INFO_DIRS` does not seem to be used.

packaging/rpm/make-rpm.sh
```diff
@@ -24,6 +24,9 @@ GITHUB_SHA="${GIT_SHA:0:7}"
 TARBALL_FILE="microshift-${GITHUB_SHA}.tar.gz"
 DEFAULT_RPMBUILD_DIR="$(git rev-parse --show-toplevel)/_output/rpmbuild/"
 RPMBUILD_DIR="${RPMBUILD_DIR:-${DEFAULT_RPMBUILD_DIR}}"
+RPM_INFO_DIRS=""
```
ggiguash, line 27: Not used, should remove?

packaging/rpm/make-rpm.sh
pacevedom: It is used in print_info function, but it should be an argument instead.

ggiguash: OK, let's fix it for consistency with other funcs.

ggiguash:
Looks like verify is failing

PR #2503, opened 2023-10-23
USHIFT-1405: Standard suite on RHEL-9.3

pmtk:


scripts/image-builder/configure.sh
```diff
@@ -18,6 +18,8 @@ sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
     createrepo yum-utils selinux-policy-devel jq wget lorax rpm-build \
     containernetworking-plugins expect
 
+sudo sed -i 's,dist/rhel9/9.3,beta/rhel9/9,g' /usr/share/osbuild-composer/repositories/rhel-93.json
```
ggiguash, line 21: We probably need to run this command only if the file exists and we're working on a beta OS, right?

test/image-blueprints/group1/rhel93.toml
```diff
@@ -0,0 +1,19 @@
+name = "rhel-9.3"
+description = "Basic RHEL 9.2 edge image without MicroShift."
```
ggiguash, line 2:
```suggestion
description = "Basic RHEL 9.3 edge image without MicroShift."
```

test/scenarios/el93-src@standard-suite.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Sourced from cleanup_scenario.sh and uses functions defined there.
```
ggiguash, line 3:
```suggestion
# Sourced from scenario.sh and uses functions defined there.
```

test/image-blueprints/group1/rhel93.toml
pmtk: I was editing and testing on 3 different hosts and lost track of these :laughing: Thx

test/scenarios/el93-src@standard-suite.sh
pmtk: Well, I just copied the el92 file but I'll update that ;)

scripts/image-builder/configure.sh
pmtk: Yup, it's just draft PR now, needed to check how things will work in CI compared to local env

test/resources/selinux.py
```diff
@@ -74,6 +78,10 @@
 
 
 def get_expected_ocp_microshift_fcontext_list() -> List[str]:
+    stdout = remote_sudo("bash -c 'source /etc/os-release && echo $VERSION_ID'")
+    if stdout == "9.3":
```
ggiguash, line 82: Is this specific to 9.3 version, or it's 9.3 and above?

pmtk:
Waiting for https://github.com/openshift/release/pull/44785 before further testing

test/resources/selinux.py
pmtk: I don't know, 9.4 might have even more (or different) entries

ggiguash: Should we better add `stdout >= 9.3` to make it potentially future proof?

scripts/image-builder/configure.sh
```diff
@@ -18,6 +18,12 @@ sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
     createrepo yum-utils selinux-policy-devel jq wget lorax rpm-build \
     containernetworking-plugins expect
 
+RHEL93_REPOS=/usr/share/osbuild-composer/repositories/rhel-93.json
+if [[ ! -f "${RHEL93_REPOS}" ]]; then
+    sudo cp -v /usr/share/osbuild-composer/repositories/rhel-9.json "${RHEL93_REPOS}"
+    sudo sed -i "s,content/dist/rhel9,content/beta/rhel9,g" "${RHEL93_REPOS}"
+fi
```
ggiguash, line 25: We may be forced to move to RHEL 9.3 before it's in GA and then we'd need to update this code.
Should we consider a full condition (i.e. beta + 9.3) ?

pmtk: Actually, we might want to stay on 9.2 EUS version and only update to next EUS (9.4), to keep our CI stable

pmtk:
/test metal-periodic-test microshift-metal-tests

Retesting metal-tests to get some sanity checks about RPM signature verification compared to osbuild-composer v88

scripts/image-builder/configure.sh
pmtk: PTAL

test/scenarios/el93-src@standard-suite.sh
ggiguash, line 1: Should we put this in periodics? 
We're already running pre-submit on RHEL 9.2, and much different with RHEL 9.3 is not likely.

pmtk:
/hold

I'm seeing issues with mirror's RPM signature verification

PR #2507, opened 2023-10-24
OCPQE-17114: Stress Testing

agullon:
Stress Testing script to enable/disable conditions (limiting resources) on local or remote host:
- New `stress_testing.sh` script added.
- `run.sh` add a new `-s` option to set the Stress Condition
- `scenario.sh` add new `stress_testing` function to enable/disable Stress condition before/after `run_tests`
- Add a new package (iproute-tc) to the blueprint to run the conditions for Red Hat Device Edge image where these tests will run


test/bin/stress_testing.sh
```diff
@@ -0,0 +1,221 @@
...
+# must run as root
+if [[ $(id -u) -ne 0 ]]; then 
+    echo "ERROR: It must be run as root"
+    exit 1
+fi
+
+# main loop
+TOTAL_PARAMS=$#
+arg_precheck "$@"
+while [ $# -gt 0 ]; do
```
dhensel-rh, line 190: Would it be easier to use getopts with the while function instead of creating your own functions (arg_precheck and arg_postcheck) ?

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,221 @@
+#!/bin/bash
+
+set -uo pipefail
+
+SCRIPT_NAME="stress_testing.sh"
+VERSION="0.2.0"
+
+INTERFACE="enp1s0" # interface in the MicroShift Jenkins VMs
```
dhensel-rh, line 8: You might consider deriving the interface. This example matches on the route with a metric of 100 and the default GW
route | awk '/100/ && /default/{print $8}'

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,221 @@
...
+  sudo tc qdisc replace dev "${INTERFACE}" root netem delay "${VALUE}"
+}
+
+function disable_latency {
+  echo -e "Latency scenario disabled"
+  sudo tc qdisc delete dev "${INTERFACE}" root
+}
+
+function get_latency {
+  AVG_LATENCY=$(ping -c 3 8.8.8.8 | tail -1 | awk '{print $4}' | cut -d '/' -f 2 | sed 's/\..*//')
```
dhensel-rh, line 177: You could simplify the AVG_LATENCY down to a couple of awk commands and two pipes

ping -c 3 8.8.8.8 | awk '/avg/{print $4}' | awk -F"/" '{printf "%d\n",$2}'

agullon: thanks, I change the command

test/bin/stress_testing.sh
agullon: thanks, I added a command to get default interface

test/bin/stress_testing.sh
agullon: yep, I just find this approach easier for this particular case

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,221 @@
...
+  bandwidth_scenario_ready
+  if [ "${ACTION}" == "enable" ]; then
+    enable_bandwidth
+  elif [ "${ACTION}" == "disable" ]; then
+    disable_bandwidth
+  elif [ "${ACTION}" == "get" ]; then
+    get_bandwidth
+  fi
+}
+
```
jogeo, line 104: dnf can not be used to install packages on a RHEL for Edge os-tree based system.
These tools need to exist on the MicroShift host, and would therefore need to be added from a [[packages]] stanza in the toml blueprint files. This script could check that the dependencies are available but can't assume it's able to install them.

The tests/image-blueprints directory has the toml files used to build images (See the "Image Blueprints" section of the README).

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,221 @@
...
+  echo -e "USAGE: "
+  echo -e "       ./${SCRIPT_NAME} ACTION SCENARIO [VALUE]"
+  echo -e "       ./${SCRIPT_NAME} -h | --help"
+  echo -e "       ./${SCRIPT_NAME} -V | --version\n"
+  echo -e "ACTION:"
+  echo -e "  -e, --enable SCENARIO VALUE     Enable a scenario"
+  echo -e "  -d, --disable SCENARIO          Disable a scenario\n"
+  echo -e "  -g, --get SCENARIO              Returns the current real value for the scenario\n"
+  echo -e "SCENARIO:"
+  echo -e "  -s, --scenario { cpu | memory | disk | bandwith | packet_loss | latency } [ VALUE ]"
```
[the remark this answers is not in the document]
agullon: thanks! fixed

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,178 @@
...
+
+function check_command {
+  if ( ! $1 &> /dev/null ) ; then
+    echo -e "ERROR: $1: command not found"
+    exit 1
+  fi
+}
+
+# script info
+function version {
```
jogeo, line 104: I don't see the version function being used?

jogeo (commented):
Thanks for adding this and for the demo earlier. I like how this adds a framework for introducing adverse conditions, to the running MicroShift host, without the need to modify existing test cases.

test/bin/stress_testing.sh
agullon: yes, just removed it

scripts/image-builder/config/blueprint_v0.0.1.toml
```diff
@@ -53,6 +53,10 @@ version = "*"
 name = "strace"
 version = "*"
 
+[[packages]]
+name = "iproute-tc"
+version = "*"
```
ggiguash, line 58: Why do we need it here? I'd add it to base blueprints under test.

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,173 @@
+#!/bin/bash
+#
+# This script will enable stress scenarios to run MicroShift tests
+# with constricted resources
+
+set -uo pipefail
+
+SCRIPT_NAME=$(basename "$0")
+IP=$(nslookup "$(hostname)" | grep -v "#" | awk '/Address:/{print $2}')
```
ggiguash, line 9: Use `hostname -I | awk '{print $1}'` to get the first IP address of the local host.

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,173 @@
...
+INTERFACE=$(ip route | grep default | awk -v ip="${IP}" '$0~ip{print $5}')
+
+WONDERSHAPER="/usr/local/bin/wondershaper"
+SPEEDTEST_CLI="/usr/local/bin/speedtest-cli"
```
ggiguash, line 13: Where are these coming from? We need to add support for those in fetch_tools.sh rather than hard-coding their paths.

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,173 @@
...
+    echo -e "ERROR: $1: command not found"
+    exit 1
+  fi
+}
+
+# usage
+function help {
+  echo -e "USAGE: "
+  echo -e "       ./${SCRIPT_NAME} ACTION SCENARIO [VALUE]"
+  echo -e "       ./${SCRIPT_NAME} -h | --help\n"
```
ggiguash, line 107: Please, remove `./` from the script name. Paths are usually avoided in the usage message

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,173 @@
...
+  echo -e "        memory: limit the available memory to the value"
+  echo -e "        disk: limit the free space to the value"
+  echo -e "        bandwidth: limit the max download and upload bandwidth to the value in Kbps"
+  echo -e "        packet_loss: set packet loss % to the value"
+  echo -e "        latency: limit minimun latency to the value in ms\n"
+  echo -e "VALUE:  The target value, integer type\n"
+  echo -e "EXAMPLE: "
+  echo -e "        ./${SCRIPT_NAME} --help"
+  echo -e "        ./${SCRIPT_NAME} --enable latency 100"
+  echo -e "        ./${SCRIPT_NAME} --disable latency"
```
ggiguash, line 124: Same here, please remove `./`

test/run.sh
```diff
@@ -10,24 +10,26 @@ RF_VENV="${ROOTDIR}/_output/robotenv"
...
-  -n       Dry-run, do not run the tests.
-  -o DIR   The output directory. (${OUTDIR})
-  -v DIR   The venv directory. (${RF_VENV})
-  -i PATH  The variables file. (${RF_VARIABLES})
+  -h                 Print this help text.
+  -n                 Dry-run, do not run the tests.
+  -o DIR             The output directory. (${OUTDIR})
+  -v DIR             The venv directory. (${RF_VENV})
+  -i PATH            The variables file. (${RF_VARIABLES})
+  -s NAME-VALUE      To enable an stress scenario.
```
ggiguash, line 28: This is a generic script for running any suites. I do not think we should add options that are specific to stress-testing here. Instead, we just need to document how to run it with all the options.

test/suites/stress-testing/stress-testing-disable-scenario.robot
ggiguash, line 1: The .robot file name repeats the directory in its prefix. Let's shorten it to `disable-scenario.robot`

test/suites/stress-testing/stress-testing-enable-scenario.robot
ggiguash, line 1: The .robot file name repeats the directory in its prefix. Let's shorten it to `enable-scenario.robot`

test/run.sh
```diff
@@ -71,20 +76,28 @@ if [ -z "${TESTS}" ]; then
     TESTS="./suites/standard"
 fi
 
+# if stress testing is set
```
dhellmann, line 79: So when a stress test scenario is active, a separate RF file is run before the regular tests to enable it? Is that always going to make it possible to configure the host? I would think we would need to create the VM differently for different RAM settings, for example.

scripts/image-builder/config/blueprint_v0.0.1.toml
agullon: What are the blueprints under test? As I can see on Jenkins jobs, this is the blueprint used to generate the image that is deployed to run RF and ginkgo tests. Am I missing something?

test/bin/stress_testing.sh
agullon: thanks! changed

test/bin/stress_testing.sh
agullon: These are the local paths to 2 scripts that will be downloaded if they don't exist when they are going to run. No installation is needed, only download them, so I don't think there's need to add it to `fetch_tools.sh` 

I'm fine to add them into `fetch_tools.sh`, but I think it would be better if this `stress_testing.sh` script does not depend on any other one.

test/bin/stress_testing.sh
agullon: fixed

test/bin/stress_testing.sh
agullon: fixed

test/run.sh
agullon: My idea is to add an opt-in option to enable and disable any Stress condition before running any test.

test/suites/stress-testing/stress-testing-disable-scenario.robot
agullon: done, also rename `scenario` to `condition`

test/suites/stress-testing/stress-testing-enable-scenario.robot
agullon: done, also rename `scenario` to `condition`

test/run.sh
agullon: Hi, this Stress Testing approach is independent of the requirements of the VM. You can enable/disable the Stress condition at any given moment and at any given time before running any test suite.

The separate RF file will enable the Stress condition before running any set of tests. Then, when tests are finished, the Stress condition will be disabled.

My idea is that will be always possible to configure the host, no need to change the way the VM is created at all. Any Stress condition can be enabled and disabled at any time.

agullon:
hi @dhellmann can you point me what are all the blueprints where we need to add the `iproute-tc` package? Thanks

PR #2522, opened 2023-10-29
USHIFT-1547: Implement build ISO artifact caching in AWS S3 buckets

ggiguash:
**Bottom line**: the current build times of `group1` and `group2` are ~40m. When testing cached builds on AWS metal host, the build times of `group1` and `group2` go down to ~25m.

> There are opportunities to also cache ostree commits in addition to ISO images, which will result in more savings
> of the overall build time.

The following changes were implemented in the test system:
- `scripts/fetch_tools.sh`
  - Installation of awscli for manipulating the buckets
-  `test/bin/common.sh`
   - Definition of the SCENARIO_BUILD_* denoting branch and current/previous tag
-  `test/bin/manage_build_cache.sh`
   - Cache manipulation script working with AWS S3 bucket

The main cache update and usage logic was implemented in `test/bin/ci_phase_iso_build.sh`. 
If access to a AWS S3 bucket is not configured, the builds should continue working normally, building everything locally.
- See [build_cache_update](https://github.com/openshift/microshift/pull/2522/files#diff-8852daecec3ead41a3def231ae3a3376cbae01858790680d6004dc1df2829b89R29) on how the cache is created
- See [build_with_cache](https://github.com/openshift/microshift/pull/2522/files#diff-8852daecec3ead41a3def231ae3a3376cbae01858790680d6004dc1df2829b89R43) on how the cache is used, optionally, when S3 bucket access is configured.

The cache creation should be scheduled as a separate periodic job using `test/bin/ci_phase_iso_build.sh -update_cache` command. This is to be implemented as a separate PR.

### DIFF
--- a/test/bin/ci_phase_iso_build.sh
+++ b/test/bin/ci_phase_iso_build.sh
@@ -1,13 +1,12 @@
 #!/bin/bash
 #
-# This script runs on the hypervisor, from the iso-build step.
+# This script runs on the build host to create all test artifacts.
 
 set -xeuo pipefail
 
-SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-# Cannot use common.sh because virsh is not installed, but we only
-# need ROOTDIR to set up logging in this script.
-ROOTDIR="$(cd "${SCRIPTDIR}/../.." && pwd)"
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# shellcheck source=test/bin/common.sh
+source "${SCRIPTDIR}/common.sh"
 
 # Log output automatically
 LOGDIR="${ROOTDIR}/_output/ci-logs"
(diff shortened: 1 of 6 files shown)


scripts/image-builder/configure.sh
```diff
@@ -27,6 +27,14 @@ sudo dnf install -y "https://dl.fedoraproject.org/pub/epel/epel-release-latest-$
 sudo dnf install -y mock nginx tomcli parallel
 sudo usermod -a -G mock "$(whoami)"
 
+# Install AWS CLI
+AWSCLI_DIR=$(mktemp -d /tmp/awscli.XXXXXXXX)
+pushd "${AWSCLI_DIR}" &>/dev/null
+curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
+    unzip -q awscliv2.zip && sudo ./aws/install --update
```
eslutsky, line 34: will this  update the aws binary to latest (in case its used in a local dev-env) ?

test/bin/manage_build_cache.sh
```diff
@@ -0,0 +1,169 @@
...
+            in the AWS S3 bucket to a copy of the specified build ID.
+
+  keep:     Delete all data from the '${AWS_BUCKET_NAME}/<release>'
+            AWS S3 bucket, only keeping the '<buildid>' sub-directory.
+
+Options:
+
+  -r <release>  The release sub-directory in the cache.
+
+  -b <buildid>  The build ID sub-directory in the cache.
```
pmtk, line 47: What is `buildid`?

test/bin/manage_build_cache.sh
```diff
@@ -0,0 +1,169 @@
...
+
+EOF
+}
+
+action_upload() {
+    local -r src_dir="${IMAGEDIR}/${VM_POOL_BASENAME}"
```
pmtk, line 53: `VM_POOL_BASENAME` suggests we only cache iso?

test/bin/manage_build_cache.sh
```diff
@@ -0,0 +1,169 @@
...
+# shellcheck source=test/bin/common.sh
+source "${SCRIPTDIR}/common.sh"
+
+AWS_BUCKET_NAME="${AWS_BUCKET_NAME:-microshift-build-cache}"
+RELEASE_SUBDIR=
+BUILDID_SUBDIR=
+
+usage() {
+    cat - <<EOF
+${BASH_SOURCE[0]} (upload|download|exist|latest|keep) [options]
```
pmtk, line 16: Can you show examples of actual invocation of the script?

scripts/image-builder/configure.sh
```diff
@@ -27,6 +27,14 @@ sudo dnf install -y "https://dl.fedoraproject.org/pub/epel/epel-release-latest-$
 sudo dnf install -y mock nginx tomcli parallel
 sudo usermod -a -G mock "$(whoami)"
 
+# Install AWS CLI
+AWSCLI_DIR=$(mktemp -d /tmp/awscli.XXXXXXXX)
+pushd "${AWSCLI_DIR}" &>/dev/null
+curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
```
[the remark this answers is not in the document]
ggiguash: It simplifies things if we always install it on devenv. It's a good-to-have tool, so...

test/bin/manage_build_cache.sh
ggiguash: I'll change this to `tag`. Is it better?

test/bin/manage_build_cache.sh
ggiguash: At the moment, yes. I thougth about this - if we need to expand to commits, we can add third mandatory option which will switch between pool and ostree repo dirs.

test/bin/manage_build_cache.sh
ggiguash:
```
manage_build_cache.sh upload -r 4.15 -b 231030
manage_build_cache.sh download -r 4.15 -b latest

if manage_build_cache.sh exist -r 4.15 -b latest &>/dev/null; then
   build_opts=-I # disable ISO builds
fi 

manage_build_cache.sh latest -r 4.15 -b 231031
```

scripts/image-builder/configure.sh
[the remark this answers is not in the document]
ggiguash: Good catch. Will fix asap

scripts/image-builder/configure.sh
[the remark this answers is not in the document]
ggiguash: We can move it to _output directory, but it's a good-to-have tool on the devenv machine, so I decided to put it in the system path.

ggiguash: Yes,  this is what the `--update` option is for. 
It does not do anything if the version did not change

test/bin/manage_build_cache.sh
pmtk: Still, what is `tag`? to what it corresponds from local dir structure?

test/bin/manage_build_cache.sh
```diff
@@ -0,0 +1,169 @@
...
+
+  latest:   Update the '${AWS_BUCKET_NAME}/<release>/latest' directory
+            in the AWS S3 bucket to a copy of the specified '<tag>'.
+
+  keep:     Delete all data from the '${AWS_BUCKET_NAME}/<release>'
+            AWS S3 bucket, only keeping the '<tag>' sub-directory.
+
+Options:
+
+  -r <release>  The release sub-directory in the AWS S3 bucket.
```
pmtk, line 45: would it be easier to use branch? `main, 4.14, 4.15` vs `main, release-4.14, release-4.15`? less parsing, just getting from CI ref inputs

test/bin/manage_build_cache.sh
pmtk: What ISOs would we cache right now? rhel92.image-installer and what else?

I think it would be nice to cache commits like current (4.15-ec) or previous (4.14-rc) release, they're pretty much static (i.e. updating nightly should be good enough).

pmtk: Also, we could have commits based of `main` cache using post-submit job

test/bin/manage_build_cache.sh
pmtk: how will presubmit know which tag to use?
could it pick partial image? e.g. when one job starts pushing tagX/a.iso, could another job see tagX as existing?

ggiguash: OK, there are good questions:
- I think we pre-submit always need to use `latest` tag
- Since the bucket operations are atomic, it cannot pick a partial image. A file becomes available only after its upload is finished.
- The question is how to implement an atomic rename/copy to latest.

Since we would have a check for the cache validity, maybe it's not a big deal is a job falls back to building its ISOs once in a while.

test/bin/manage_build_cache.sh
ggiguash: Yes, it's a good comment. Let me change the name of `release` to `branch` to make it more explicit.

test/bin/manage_build_cache.sh
ggiguash: I'm planning to cache group1 and group2, so we'll get 2 ISO's.
Let's see how this works and proceed to commits later?

scripts/image-builder/configure.sh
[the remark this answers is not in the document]
dhellmann: Should this be in `fetch_tools.sh`, or is it here because it's not being installed to the private location?

test/bin/manage_build_cache.sh
ggiguash: We can further discuss this issue in the next PR where this script will be used.

test/bin/manage_build_cache.sh
dhellmann: Why do we have any ISOs in group2?

Can we put the cacheable images into their own group(s) so it's clear what items can/should be cached? Maybe use `cachedN` as the pattern for directory names?

test/bin/manage_build_cache.sh
```diff
@@ -0,0 +1,169 @@
...
+#
+# This script should be run on the build host to manage cache of build artifacts
+set -euo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# shellcheck source=test/bin/common.sh
+source "${SCRIPTDIR}/common.sh"
+
+AWS_BUCKET_NAME="${AWS_BUCKET_NAME:-microshift-build-cache}"
+BCH_SUBDIR=
```
dhellmann, line 11: What is BCH?

test/bin/manage_build_cache.sh
```diff
@@ -0,0 +1,169 @@
...
+
+  verify:   Exit with 0 status if the specified
+            '${AWS_BUCKET_NAME}/<branch>/<tag>' sub-directory exists
+            and contains files, 1 otherwise.
+
+  latest:   Update the '${AWS_BUCKET_NAME}/<branch>/latest' directory
+            in the AWS S3 bucket to a copy of the specified '<tag>'.
+
+  keep:     Delete all data from the '${AWS_BUCKET_NAME}/<branch>'
+            AWS S3 bucket, only keeping the '<tag>' sub-directory.
```
dhellmann, line 41: Why is an action that deletes something called "keep"?

Why keep the tag subdirectory?

test/bin/manage_build_cache.sh
```diff
@@ -0,0 +1,169 @@
...
+  download: Download build artifacts from the specified
+            '${AWS_BUCKET_NAME}/<branch>/<tag>' AWS S3 bucket to
+            the local disk.
+
+  verify:   Exit with 0 status if the specified
+            '${AWS_BUCKET_NAME}/<branch>/<tag>' sub-directory exists
+            and contains files, 1 otherwise.
+
+  latest:   Update the '${AWS_BUCKET_NAME}/<branch>/latest' directory
+            in the AWS S3 bucket to a copy of the specified '<tag>'.
```
dhellmann, line 38: Instead of "latest" being a copy of the content, could it be a file that tells the download script which tag is the latest? Then it could be updated atomically and wouldn't cause issues by deleting content that other jobs (or users) might be in the middle of downloading.

scripts/image-builder/configure.sh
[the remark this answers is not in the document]
ggiguash: Yes, fetch_tools.sh installs packages to _output dir. We can always reconsider, but I think this is a useful utility to have on a devenv machine

test/bin/manage_build_cache.sh
ggiguash: Probably, I was planning to work out the details when implementing the usage of this script. We can adopt it as we go, according to our needs.

test/bin/manage_build_cache.sh
ggiguash: That's a cleanup function inverted (name suggested by ChatGPT :), similar to -k options that are common). 
The idea is to specify what to keep and delete everything else in a given release sub-dir.
Depending on what we decide to implement for the cache update algo, we'd want to change this further.

test/bin/manage_build_cache.sh
ggiguash: Yes, we can certainly reorg the structure further. At this time, however, we can start by implementing some form of caching that will evolve in the future and dictate new structure of dirs, etc.

scripts/image-builder/configure.sh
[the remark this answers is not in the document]
pmtk: I don't disagree about usefulness, but the same could be said about at least yq and shellcheck.
You can provide `DEST_DIR` to `fetch_tools.sh` to install tools somewhere else.
What if, someone would want it on their machine, would run fetch_tools.sh with custom DEST_DIR?

test/bin/manage_build_cache.sh
ggiguash: I reimplemented the usage of `last` tag according to Doug's suggestion.
Let's review and repost comment if any.

test/bin/manage_build_cache.sh
ggiguash: I implemented the use of `last` tag according to your suggestion. 
It seems to be much better

test/bin/ci_phase_iso_build.sh
```diff
@@ -21,49 +20,111 @@ exec &> >(tee >(awk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0; fflush() }' >"${
...
+        exit 1
     fi
+    build_cache_update
 else
-    # Fall back to full build when not running in CI
-    bash -x ./bin/build_images.sh
+    if ! ${HAS_BUILD_CACHE} ; then
+        echo "Build cache is not available, rebuilding all the artifacts"
+    fi
+    build_with_cache
```
pmtk, line 127: Nit: slightly confusing naming: `not $HAS_BUILD_CACHE` and then `build_with_cache`. Maybe that `echo` could be an `else` to the `if ${HAS_BUILD_CACHE}` inside `build_with_cache`?

test/bin/ci_phase_iso_build.sh
```diff
@@ -21,49 +20,111 @@ exec &> >(tee >(awk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0; fflush() }' >"${
...
+    # Upload the images and update the 'last' setting
+    ./bin/manage_build_cache.sh upload  -b "${SCENARIO_BUILD_BRANCH}" -t "${SCENARIO_BUILD_TAG}"
+    ./bin/manage_build_cache.sh setlast -b "${SCENARIO_BUILD_BRANCH}" -t "${SCENARIO_BUILD_TAG}"
+
+    # Cleanup older images in the cache, preserving the previous cache if any
+    # The 'last' cache is preserved by default
+    ./bin/manage_build_cache.sh keep -b "${SCENARIO_BUILD_BRANCH}" -t "${SCENARIO_BUILD_TAG_PREV}"
+}
+
+build_with_cache() {
```
pmtk, line 43: I think naming if a bit off: `_with_cache` but inside there's `if $HAS_BUILD_CACHE` so it can be "without cache" after all.

Maybe we could rename these two new functions so they hint which one we expect in presubmits and which one in special periodic?
Or do you think it would be beneficial for readability to have separate files for "testing flow" and "cache building flow"? I.e. instead of `if [ $# -gt 0 ] && [ "$1" = "-update_cache" ]`  we'd just have another script?

ggiguash: I will fix the naming, no problem.
About spliting the files, I thought about it, but there will be 90% duplication of code in them, so I thought to keep it in the same script.

ggiguash:
The metal-periodic-test is failing due to an unrelated problem to be fixed in another PR. See #2565 
/unhold

PR #2523, opened 2023-10-29
USHIFT-1801: change build_rpms.sh to create rpm with podman in CI

eslutsky:
this change will  use openshift's  go container toolchain in CI  ,make the resulting binary/rpm FIPs compliant and consistent with ART .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,8 @@ GO_PACKAGES=$(go list ./cmd/... ./pkg/...)
 # Build to a place we can ignore
 GO_BUILD_BINDIR :=$(OUTPUT_DIR)/bin
 
+GO_CACHE :=$(shell go env GOCACHE)
+
 ifeq ($(DEBUG),true)
 	# throw all the debug info in!
 	LD_FLAGS =
@@ -293,15 +295,15 @@ commit: image-build-configure image-build-commit
 .PHONY: commit
 
 rpm-podman:
-	RPM_BUILDER_IMAGE_TAG="rhel-8-release-golang-1.19-openshift-4.13"; \
+	RPM_BUILDER_IMAGE_TAG="rhel-8-release-golang-1.20-openshift-4.14"; \
 	podman build \
 		--volume /etc/pki/entitlement/:/etc/pki/entitlement \
 		--build-arg TAG=$$RPM_BUILDER_IMAGE_TAG \
 		--tag microshift-builder:$$RPM_BUILDER_IMAGE_TAG - < ./packaging/images/Containerfile.rpm-builder ; \
 	podman run \
 		--rm -ti \
-		--volume $$(pwd):/opt/microshift \
-		--volume $$(go env GOCACHE):/go/.cache \
+		--volume $$(pwd):/opt/microshift:z \
+		--volume $(GO_CACHE):/go/.cache:z \
 		--env TARGET_ARCH=$(TARGET_ARCH) \
 		microshift-builder:$$RPM_BUILDER_IMAGE_TAG \
 		bash -ilc 'cd /opt/microshift && make rpm & pid=$$! ; trap "pkill $${pid}" INT ; wait $${pid}'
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -293,15 +293,14 @@ commit: image-build-configure image-build-commit
...
-	RPM_BUILDER_IMAGE_TAG="rhel-8-release-golang-1.19-openshift-4.13"; \
+	RPM_BUILDER_IMAGE_TAG="rhel-8-release-golang-1.20-openshift-4.14"; \
 	podman build \
 		--volume /etc/pki/entitlement/:/etc/pki/entitlement \
 		--build-arg TAG=$$RPM_BUILDER_IMAGE_TAG \
 		--tag microshift-builder:$$RPM_BUILDER_IMAGE_TAG - < ./packaging/images/Containerfile.rpm-builder ; \
 	podman run \
 		--rm -ti \
-		--volume $$(pwd):/opt/microshift \
-		--volume $$(go env GOCACHE):/go/.cache \
```
pmtk, line 304: Have you checked how removing cache mount influences rebuild times?

eslutsky: actually I've received `permission error` with that cache, did it work for you ?

test/bin/build_rpms.sh
```diff
@@ -14,7 +14,7 @@ rm -rf _output/rpmbuild*
 
 BUILD_CMDS=( \
     # Normal build of current branch from source
-    'make rpm' \
+    'make rpm-podman' \
```
pmtk, line 17: This add new requirement to test harness. I think we should mention it in the docs

Makefile
eslutsky: looks like does improves the runtime in couple of minutes , i'll try adding `:z`  for selinux relabeling , like i did with the first volume.

test/bin/build_rpms.sh
eslutsky: yes actually its requires:
1. pull-secret for the registry.ci.openshift.org 
2.  build has to run from a RHEL  RHSM registered machine

pmtk: FYI you could also copy entitlement from RHEL to Fedora and it should work. But for docs, let's keep it simple

Makefile
```diff
@@ -57,6 +57,9 @@ GO_PACKAGES=$(go list ./cmd/... ./pkg/...)
 # Build to a place we can ignore
 GO_BUILD_BINDIR :=$(OUTPUT_DIR)/bin
 
+# Go Cache
```
pmtk, line 60: Nit: I'm not sure we need this comment ;)

test/bin/build_rpms.sh
ggiguash: Can/should we make this optional, only if registry.ci.openshift.org pull secret is present?

eslutsky: @ggiguash , i can do something different:
if the command `make rpm-podman` then fallback to `make rpm`

eslutsky: @pmtk  , now this change affect CI  only its not really affects the dev-env requirement. do we still need to change the Docs?

pmtk: Nah, I think this is fine

test/bin/build_rpms.sh
```diff
@@ -12,9 +12,15 @@ source "${SCRIPTDIR}/common.sh"
 cd "${ROOTDIR}"
 rm -rf _output/rpmbuild*
 
-BUILD_CMDS=( \
-    # Normal build of current branch from source
-    'make rpm' \
+# Normal build of current branch from source
+BUILD_CMDS=('make rpm')
+
+#in CI build current branch from source using podman
```
ggiguash, line 18:
```suggestion
# In CI, build the current branch from source with the build tools used by OCP.
# This is necessary for enabling FIPS support.
```

PR #2538, opened 2023-10-31
merge greenboot RPM content into main package

dhellmann:
Eliminate the need for a separate greenboot RPM.

/hold


dhellmann:
No longer needed since CWFCONF-7884 is resolved.

/close

PR #2542, opened 2023-11-01
USHIFT-1816: use the package as source of truth for version of microshift in tests

dhellmann:
Instead of reading the local source tree to determine what version
MicroShift should report, look at the dnf info for the package.

/assign @ggiguash @jogeo

### DIFF
--- a/test/suites/standard/version.robot
+++ b/test/suites/standard/version.robot
@@ -4,6 +4,7 @@ Documentation       Tests related to the version of MicroShift
 Resource            ../../resources/common.resource
 Resource            ../../resources/oc.resource
 Resource            ../../resources/microshift-process.resource
+Resource            ../../resources/microshift-rpm.resource
 Library             Collections
 Library             ../../resources/DataFormats.py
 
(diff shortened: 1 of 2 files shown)


dhellmann:
/hold

This needs some rework.

test/suites/standard/version.robot
```diff
@@ -74,20 +75,27 @@ Teardown
...
-
-    # 4.14.0-0.nightly-arm64-2023-05-04-012046
-    ${version_full_raw}=    Fetch From Right    ${unparsed}    :=
-    ${version_full}=    Strip String    ${version_full_raw}
+    IF    ${rc}==0
+        # This returns a string like 4.14.0-0.nightly-arm64-2023-05-04-012046
+        ${unparsed}=    Get Version Of MicroShift RPM From DNF
+    ELSE
+        ${unparsed}=    Get Version Of MicroShift RPM From Ostree
+    END
```
ggiguash, line 95: @dhellmann , since we are checking the version of an installed package, can we use `rpm -q` which works on both system types?

dhellmann: Nice, I've updated it to work that way.

test/suites/standard/version.robot
```diff
@@ -74,20 +75,22 @@ Teardown
...
     ...    the content.
 
-    # The file contains content in this format:
-    #
-    #    OCP_VERSION := 4.14.0-0.nightly-arm64-2023-05-04-012046
+    ${rc}=    SSHLibrary.Execute Command
+    ...    test -f /usr/bin/dnf
+    ...    return_rc=True
+    ...    return_stdout=False
+    ...    return_stderr=False
```
ggiguash, line 88: A nit: looks like we no longer need this code.
LGTM otherwise.

dhellmann: thanks, fixed

PR #2543, opened 2023-11-01
OCPBUGS-22809: Fix selinux labeling for microshift executable files

ggiguash:


dhellmann:
I verified manually that I no longer see the warnings when installing packages built from this branch.

PR #2545, opened 2023-11-01
USHIFT-1818: fix argument processing for login command

dhellmann:
The script requires at least 2 arguments, but the login command
accepts an extra one.

/assign @ggiguash

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -611,11 +611,11 @@ Settings
 
 Login
 
-  scenario.sh login <scenario-script> <host>
+  scenario.sh login <scenario-script> [<host>]
 EOF
 }
 
-if [ $# -ne 2 ]; then
+if [ $# -lt 2 ]; then
     usage
     exit 1
 fi


test/bin/scenario.sh
```diff
@@ -611,11 +611,11 @@ Settings
 
 Login
 
-  scenario.sh login <scenario-script> <host>
+  scenario.sh login <scenario-script> [<host>]
```
ggiguash, line 614: A nit: we should probable remove `login` and `[args]` from the top of the usage message because all-but-login actions expect 2 arguments, right?

ggiguash: I'd suggest something like this:

```
scenario.sh (create|boot|run|cleanup|rerun) <scenario-script>
scenario.sh login <scenario-script> [host]

  create|boot -- Set up the infrastructure for the test, such as VMs.

  run -- Run the scenario.

  rerun -- cleanup, create, run for the same scenario.

  cleanup -- Remove the VMs created for the scenario.

  login -- Login to a host for a scenario (use 'host1' host name by default)

Settings

  The script looks for /home/microshift/microshift/test/scenario_settings.sh for some global settings.
```

test/bin/scenario.sh
```diff
@@ -611,11 +611,11 @@ Settings
...
 EOF
 }
 
-if [ $# -ne 2 ]; then
+if [ $# -lt 2 ]; then
```
ggiguash, line 618:
```suggestion
if [ $# -lt 2 ] || [ $# -gt 3 ] ; then
```

test/bin/scenario.sh
dhellmann: Line 596 already indicates that commands may take optional arguments after the scenario script. The change on line line is showing that host is actually optional, after the fix.

PR #2548, opened 2023-11-02
USHIFT-1716: Report microshift service status in the microshift-cleanup-data script

ggiguash:
Note the `MicroShift service...` messages before the exit.

When run with `--all` option, the output is:
```
Stopping MicroShift services
Disabling MicroShift services
Removing MicroShift pods
Removing crio image storage
Killing conmon, pause and OVN processes
Removing MicroShift configuration
Removing OVN configuration
MicroShift service was stopped
MicroShift service was disabled
Cleanup succeeded
```

When run with `--ovn` option, the output is:
```
Stopping MicroShift services
Removing MicroShift pods
Killing conmon, pause and OVN processes
Removing OVN configuration
MicroShift service was stopped
Cleanup succeeded
```

### DIFF
--- a/scripts/microshift-cleanup-data.sh
+++ b/scripts/microshift-cleanup-data.sh
@@ -9,6 +9,9 @@ PODS_NS_LIST=(openshift-service-ca openshift-ingress openshift-dns openshift-sto
 FULL_CLEAN=false
 KEEP_IMAGE=false
 OVN_CLEAN=false
+# Flags used for generating a user summary message
+SERVICE_STOPPED=false
+SERVICE_DISABLED=false
 
 function usage() {
     if [ $# -gt 0 ] ; then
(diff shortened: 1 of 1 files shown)


scripts/microshift-cleanup-data.sh
```diff
@@ -29,11 +32,13 @@ function stop_disable_services() {
     for service in microshift microshift-etcd ; do
         systemctl stop --now   ${service} 2>/dev/null || true
         systemctl reset-failed ${service} 2>/dev/null || true
+        SERVICE_STOPPED=true
```
dhellmann, line 35: Why not just echo here? That way if something goes wrong and the script exits there is still information about what progress it did make before the error.

ggiguash: Well, we already have a message about stopping MicroShift. I think the ask was to have it towards the end where users can see it.

ggiguash: @dhensel-rh , what's your preference? I can add the message right on the spot vs in the end.

dhensel-rh: @ggiguash I guess my request was alerting the user that MicroShift would need to be restarted at the bottom of the script. I guess it is implied at the top when the first line says Stopping MicroShift services. Sorry if it feels I am being pedantic

ggiguash: Does the current messaging address this request? We're printing the following in the end of the output.
```
...
...
MicroShift service was stopped
MicroShift service was disabled
Cleanup succeeded
```

pmtk: I kind of agree with @dhensel-rh. We cannot assume that someone using that script lives and breathes systemd, so adding some friendly UX might be good

dhensel-rh: @ggiguash Yes, your message addresses the request.

PR #2553, opened 2023-11-03
USHIFT-1812: `make rpm-podman` only for x86_64

pmtk:
Because FIPS is not yet supported on ARM and there's no build image available for ARM, let's use `make rpm` to build RPMs for ARM.


pacevedom (approved):
/lgtm

PR #2554, opened 2023-11-03
OCPBUGS-22322: Don't create backups for current boot

pmtk:
Adds a check that prevents backup creation for current boot. Existence of such backup would prevent a proper backup (i.e. created on boot, with data compatible with previous boot) from being created (because microshift does not update backups - only creates new ones while deleting previous).

### DIFF
--- a/pkg/admin/prerun/prerun.go
+++ b/pkg/admin/prerun/prerun.go
@@ -86,6 +86,20 @@ func (dm *dataManagement) backup() error {
 	}
 	klog.InfoS("Contents of version file", "contents", versionFile)
 
+	currentBootID, err := getCurrentBootID()
+	if err != nil {
+		return err
+	}
+	if currentBootID == versionFile.BootID {
+		// We don't want to create a backup if versionFile has current boot ID:
+		// backup would be created for current boot - in the middle of it.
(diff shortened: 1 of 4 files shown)


pkg/admin/prerun/prerun.go
```diff
@@ -86,6 +86,20 @@ func (dm *dataManagement) backup() error {
 	}
 	klog.InfoS("Contents of version file", "contents", versionFile)
 
+	currentBootID, err := getCurrentBootID()
```
dhellmann, line 89: If we're no longer doing a backup on every startup, do we still need the boot ID in the backup name, or could we just use the ostree commit ID?

Other than that question, this looks good.

pmtk: I guess we could get rid of it. Backup's boot ID should be in version file, so it's not like we'd lose any info.
We'd have to change backing up: `create new, delete old` -> `create tmp, rename into place`.
Though I don't think we necessarily need to, there's no penalty for having this, right?

ggiguash: If it complicates the backup procedure, I would not touch it.

PR #2518, opened 2023-10-26
USHIFT-1804: Implement VM creation procedure for no-NIC scenario

ggiguash:
Implementation based on #2513 by @copejon.

Depends on the following PRs to optimize ISO image builds for pre-submits:
* https://github.com/openshift/release/pull/44963 
* https://github.com/openshift/release/pull/44973

Co-authored with @copejon

PR #2555, opened 2023-11-03
USHIFT-1804: Report OK status for IP and SSH in offline scenarios

ggiguash:
A follow-up fix on #2518

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -385,8 +385,9 @@ launch_vm() {
         touch "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/ip"
 
         echo "VM ${full_vmname} has no NICs, skipping IP assignment and ssh polling"
-        record_junit "${vmname}" "ip-assignment" "SKIPPED"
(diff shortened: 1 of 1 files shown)


pmtk:
/lgtm
/hold
Unhold when ready

ggiguash:
/unhold 
The remaining problem in the periodic test is about FDO sos-report failure.
It will be fixed in another PR.

test/bin/scenario.sh
```diff
@@ -385,8 +385,9 @@ launch_vm() {
         touch "${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/ip"
 
         echo "VM ${full_vmname} has no NICs, skipping IP assignment and ssh polling"
-        record_junit "${vmname}" "ip-assignment" "SKIPPED"
-        record_junit "${vmname}" "ssh-access" "SKIPPED"
+        # Anything other than "OK" status is reported as an error
```
dhellmann, line 388: That could be fixed by fixing the logic in record_junit.

dhellmann: See #2556

PR #2558, opened 2023-11-03
USHIFT-1824: add ssh wrapper functions to scenario.sh

dhellmann:
This allows scenario setup steps to do more with the VM before starting the
test.

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -47,6 +47,28 @@ set_vm_property() {
     echo "${value}" > "${property_file}"
 }
 
+run_command_on_vm() {
+    local -r vmname="$1"
+    shift
+    local -r command="$*"
+
+    local -r ip=$(get_vm_property "${vmname}" ip)
+    local -r ssh_port=$(get_vm_property "${vmname}" ssh_port)
+
+    ssh "redhat@${ip}" -p "${ssh_port}" -t "${command}"
+}
+
+copy_file_to_vm() {
+    local -r vmname="$1"
+    local -r local_filename="$2"
+    local -r remote_filename="$3"
+
+    local -r ip=$(get_vm_property "${vmname}" ip)
+    local -r ssh_port=$(get_vm_property "${vmname}" ssh_port)
+
+    scp -P "${ssh_port}" "${local_filename}" "redhat@${ip}:${remote_filename}"
+}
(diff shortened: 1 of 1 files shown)


test/bin/scenario.sh
```diff
@@ -35,16 +80,14 @@ sos_report() {
...
+        if [ -z "${ip}" ]; then
             # skip hosts without NICs
             # FIXME: use virsh to copy sos report files
             continue
         fi
-        # Copy the sos helper for compatibility, it is only available in 4.14 RPMs
-        scp "${ROOTDIR}/scripts/microshift-sos-report.sh" "redhat@${ip}":/tmp
-        ssh "redhat@${ip}" \
-            "[ -f /usr/bin/microshift-sos-report ] && sudo /usr/bin/microshift-sos-report || sudo /tmp/microshift-sos-report.sh ; sudo chmod +r /tmp/sosreport*"
+        run_command_on_vm "${vmname}" "sudo sos report --quiet --batch --all-logs --tmp-dir /tmp && sudo chmod +r /tmp/sosreport*"
```
ggiguash, line 90: I'm not sure we need to go back to collecting all the logs. They slow down the process and they contain too much unnecessary information. I'd rather continue using the `microshift-sos-report` script.

ggiguash: Note: I just approved a fix in #2565 which fixes the logic of sos collection when MicroShift is not installed.

test/bin/scenario.sh
```diff
@@ -58,16 +80,31 @@ sos_report() {
...
+        cat - >/tmp/sos-wrapper.sh <<EOF
+#!/usr/bin/env bash
+if [ -f /usr/bin/microshift-sos-report ]; then
+    sudo /usr/bin/microshift-sos-report
+else
+    sudo chmod +x /tmp/microshift-sos-report.sh
+    sudo PROFILES=network,security /tmp/microshift-sos-report.sh
+fi
+sudo chmod +r /tmp/sosreport*
+EOF
```
ggiguash, line 104: The /tmp/sos-wrapper.sh should be executable, so chmod +x seems to be missing either on the source or dest host.

ggiguash:
This job was marked as required by mistake. The current change is not related to the job failure.

PR #2564, opened 2023-11-03
[USHIFT-2378] CI: Build group3 images unconditionally

copejon:

This is a temporary workaround to allow #2378 to pass pre-submits.  The PR's jobs cannot pass because the scenario image doesn't exist.  After merging 2378, we should revert this PR (or maybe enhance the logic to permit presubmit builds when changes are made to group3 files).


dhellmann (commented):
/lgtm

PR #2565, opened 2023-11-03
USHIFT-1820: Skip sos microshift profile if its disabled

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -44,7 +44,7 @@ sos_report() {
         # Copy the sos helper for compatibility, it is only available in 4.14 RPMs
         scp "${ROOTDIR}/scripts/microshift-sos-report.sh" "redhat@${ip}":/tmp
         ssh "redhat@${ip}" \
-            "[ -f /usr/bin/microshift-sos-report ] && sudo /usr/bin/microshift-sos-report || sudo /tmp/microshift-sos-report.sh ; sudo chmod +r /tmp/sosreport*"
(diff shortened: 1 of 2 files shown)


test/bin/scenario.sh
```diff
@@ -44,7 +44,7 @@ sos_report() {
         # Copy the sos helper for compatibility, it is only available in 4.14 RPMs
         scp "${ROOTDIR}/scripts/microshift-sos-report.sh" "redhat@${ip}":/tmp
         ssh "redhat@${ip}" \
-            "[ -f /usr/bin/microshift-sos-report ] && sudo /usr/bin/microshift-sos-report || sudo /tmp/microshift-sos-report.sh ; sudo chmod +r /tmp/sosreport*"
+            "[ -f /usr/bin/microshift-sos-report ] && sudo /usr/bin/microshift-sos-report || PROFILES="network,security" sudo /tmp/microshift-sos-report.sh; sudo chmod +r /tmp/sosreport*"
```
dhellmann, line 47: I think we need a change in microshift-sos-report.sh to allow PROFILES to be overridden from the outside. 

It might be more effective to just run a basic sos command directly, since that removes any concern of dependencies on the script for scenarios where microshift isn't even installed.

ggiguash: We were running a basic sos report, but it generates too large output and it's slow. I'm not sure we'd want to get back to that situation.

pacevedom: Done with the first approach, somehow forgot to turn those variables into overrideable.

As for creating a direct command, I wanted to have everything centralized in the script since it simplifies logic around it (like permissions, dirs, copying, etc.).

PR #2567, opened 2023-11-03
OCPBUGS-22936: don't fail to clean up if crio is not running

dhellmann:
In some situations, especially in tests but not necessarily limited to
those cases, crio is not running when the microshift cleanup script is
run.  Detect the missing socket and skip the crio steps to avoid failures.

### DIFF
--- a/scripts/microshift-cleanup-data.sh
+++ b/scripts/microshift-cleanup-data.sh
@@ -46,6 +46,10 @@ function stop_clean_pods() {
     # It is necessary to remove the pods (OVN-related last) to allow for further termination
     # of processes (i.e. conmon, etc.) that use the files under /var/run/ovn.
     # The cleanup of OVN data only works if the files under /var/run/ovn are not in use.
+    if [ ! -e /var/run/crio/crio.sock ]; then
+        echo "crio.sock is not present, not attempting to clean up pods"
+        return 0
+    fi
     echo Removing MicroShift pods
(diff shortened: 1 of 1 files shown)


pmtk:
I don't think we need all the commits here, just the last one, right?

scripts/microshift-cleanup-data.sh
```diff
@@ -46,6 +46,10 @@ function stop_clean_pods() {
     # It is necessary to remove the pods (OVN-related last) to allow for further termination
     # of processes (i.e. conmon, etc.) that use the files under /var/run/ovn.
     # The cleanup of OVN data only works if the files under /var/run/ovn are not in use.
+    if [ ! -e /var/run/crio/crio.sock ]; then
+        echo "crio.sock is not present, not attempting to clean up pods"
+        return 0
```
pmtk, line 51: I guess missing `crio.sock` means its not running (or not installed but I think microshift pulls it as a dependency so we shouldn't end up in such situation)?
Pods might not be running, but should we attempt to remove image storage?

dhellmann: That also uses crictl, which won't be able to talk to the backend if the socket is not there.

This case came up in the course of testing #2568, which uninstalls MicroShift and its dependencies.

ggiguash: Another case is when microshift RPM is just installed, but not started. So, crio is not running at all, but the script can be run.

PR #2568, opened 2023-11-03
USHIFT-1822: scenario and suite for testing rpm-based installation and upgrades

dhellmann:
Merge the existing upgrade test into a new suite that includes installing
packages from scratch, rather than relying on any packages being in the
base image.

During installation, watch for errors and warnings in the output of dnf and
fail if any are present. To make this work reliably, NetworkManager-ovs is
installed separately instead of as a dependency, because installing it
frequently causes a warning about the systemd unit changing on disk.

Rename the existing installation keyword to be specific that it needs a 
repo url as arg. Add a new keyword for installing from the repos known to
the system.


dhellmann:
/hold

Some of the dependency PRs need to merge first then I will rebase this one

dhellmann:
It looks like the last rebase was bad and I missed some of the new functions that went in via a separate PR. I've rebased again to pick those up.

test/bin/scenario.sh
```diff
@@ -620,6 +670,12 @@ action_create() {
     fi
     record_junit "setup" "load_global_settings" "OK"
 
+    if ! load_subscription_manager_plugin; then
+        record_junit "setup" "load_subscription_manager_plugin" "FAILED"
+        return 1
+    fi
+    record_junit "setup" "load_subscription_manager_plugin" "OK"
```
ggiguash, line 677: Looks like we're unconditionally calling the registration procedure on the VM creation?

pmtk: I think this just sources the functions, but they're called in scenario script

test/bin/subscription_manager_register.sh
```diff
@@ -0,0 +1,27 @@
...
+        --activationkey="\$(cat /tmp/subscription-manager-act-key)"
+fi
+EOF
+        copy_file_to_vm "${vmname}" /tmp/sub.sh /tmp/sub.sh
+        copy_file_to_vm "${vmname}" /tmp/subscription-manager-org /tmp/subscription-manager-org
+        copy_file_to_vm "${vmname}" /tmp/subscription-manager-act-key /tmp/subscription-manager-act-key
+        run_command_on_vm "${vmname}" "chmod +x /tmp/sub.sh && sudo /tmp/sub.sh"
+    else
+        # Local developer workflow
+        run_command_on_vm "${vmname}" "sudo subscription-manager register"
```
pmtk, line 25: Will this pause the execution and present a prompt? Could we automatically use `/etc/pki/entitlement` for RHEL hosts?

pmtk (commented):
couple questions, but overall looks good to me

test/bin/scenario.sh
dhellmann: Right, this loads the script. The script defines a function. The function is then available to scenarios that need it. The new scenario in this PR uses it.

test/bin/subscription_manager_register.sh
dhellmann: It does pause, yes. I'm not sure what the files in that directory are. Can those just be copied to a new host as they are?

test/resources/microshift-rpm.resource
```diff
@@ -10,10 +10,38 @@ ${REPO_FILE_NAME}       "/etc/yum.repos.d/microshift-local.repo"
...
     [Documentation]    Installs MicroShift RPM packages from the specified URL
     [Arguments]    ${repo_url}
 
+    Configure MicroShift Repository    ${repo_url}
+
+    ${stdout}    ${stderr}    ${rc}=    SSHLibrary.Execute Command
+    ...    dnf install -y microshift
+    ...    sudo=True    return_rc=True    return_stderr=True    return_stdout=True
+    Log    ${stderr}
+    Should Be Equal As Integers    0    ${rc}
```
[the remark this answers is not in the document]
dhellmann: Good point, I think we can reuse the keyword. Or maybe just call "Command Should Work"?

test/suites/rpm/install-and-upgrade-successful.robot
```diff
@@ -0,0 +1,113 @@
+*** Comments ***
+# NOTE: These tests rely on being run in order.
```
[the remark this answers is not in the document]
dhellmann: Yeah, I can expand the comment to talk about that.

test/bin/subscription_manager_register.sh
dhellmann: I have a few different RHEL systems, and the contents of those files are all different. I think that means the content is host specific.

test/suites/rpm/install-and-upgrade-successful.robot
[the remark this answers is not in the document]
dhellmann: done

dhellmann:
The 4.14.1 RPM installation is reporting a warning, which is no surprise because that's why I'm writing this test. The work-around is to ignore the warning in that step of the upgrade test, so I've updated the PR to do that.

test/bin/subscription_manager_register.sh
pmtk: You're right. Entitlement can be used to pass-through the subscription to a container, but in case of OS on VM it's probably not the right way.
I guess the subscription plugin system is flexible enough we can do something about these prompts if they get annoying

dhellmann:
@pmtk I rebased this to see if that would pick up any of the CI fixes you all made last week. It needs a lgtm again.

dhellmann:
Rebased for all the latest CI stability goodness.

pmtk:
Ah, 4.15 is not GA'd yet, so the repo is not available.
Until then `Enable MicroShift Dependency Repositories` should enable `rhocp-4.14` repositories...

```
"Errors during downloading metadata for repository 'rhocp-4.15-for-rhel-9-x86_64-rpms':\n  - Status code: 403 for https://cdn.redhat.com/content/dist/layered/rhel9/x86_64/rhocp/4.15/os/repodata/repomd.xml (IP: 23.198.106.83)\nError: Failed to download metadata for repo 'rhocp-4.15-for-rhel-9-x86_64-rpms': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

pmtk:
with https://github.com/openshift/microshift/pull/2760 it should be possible to dynamically enable right rhocp-4.Y repository in the test

dhellmann:

The failures in this job are known flakes not related to the new test suite.

/hold cancel

test/bin/subscription_manager_register.sh
dhellmann: I'm not sure it's possible for individual users to get the kinds of keys that we use in CI. Maybe? I think I saw something that said the account administrator had to create them.

dhellmann:
The test failures are all known flakes and not related to this change. We can save the override instruction for when the code review is done.

test/scenarios/el92-src@rpm-install.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+# shellcheck source=test/bin/get_rel_version_repo.sh
+source "${SCRIPTDIR}/get_rel_version_repo.sh"
+
+# Sourced from scenario.sh and uses functions defined there.
+
+# The RPM-based system won't run greenboot automatically on startup,
+# so tell the framework not to wait for that.
+export SKIP_GREENBOOT=true
```
ggiguash, line 10: Greenboot is installed as a dependency of MicroShift, so it's present.
The defaul preset for greenboot-healthcheck service seems to be `enabled`

```
$ systemctl status greenboot-healthcheck.service 
● greenboot-healthcheck.service - greenboot Health Checks Runner
     Loaded: loaded (/usr/lib/systemd/system/greenboot-healthcheck.service; enabled; preset: enabled)
    Drop-In: /etc/systemd/system/greenboot-healthcheck.service.d
             └─overrides.conf
     Active: active (exited) since Tue 2024-01-02 14:06:44 UTC; 2 days ago
   Main PID: 823 (code=exited, status=0/SUCCESS)
        CPU: 77ms
```

scripts/verify/verify-rf.sh
```diff
@@ -16,6 +16,7 @@ set -x
 "${RF_VENV}/bin/robocop" \
     --exclude 1015 \
     --configure 0504:max_len:40 \
-    --configure 0505:max_calls:20
+    --configure 0505:max_calls:20 \
+    --configure 0508:line_length:200
```
ggiguash, line 20: I've been trying to keep the line length on 80-100 chars, and most tests are like this.
Maybe, we can introduce a more conservative limit to keep tests easily readable?

test/bin/subscription_manager_register.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+# Enable subscription in the VM
+subscription_manager_register() {
+    local -r vmname="$1"
+
+    if [ -f /tmp/subscription-manager-org ]; then
+        # CI workflow
+        cat <<EOF > /tmp/sub.sh
```
ggiguash, line 9: Should we use `mktemp` for all the temporary files?

test/scenarios/el92-src@rpm-install.sh
```diff
@@ -0,0 +1,79 @@
...
+    fi
+    local -r release_info_rpm_base=$(find "${BASE_REPO}" -name 'microshift-release-info-*.rpm' | sort | tail -n 1)
+    if [ -z "${release_info_rpm_base}" ]; then
+        error "Failed to find microshift-release-info RPM in ${BASE_REPO}"
+        return 1
+    fi
+    local -r SOURCE_VERSION=$(rpm -q --queryformat '%{version}' "${release_info_rpm}")
+    local -r CURRENT_VERSION=$(echo "${SOURCE_VERSION}" | cut -f1-2 -d.)
+    local -r MINOR_VERSION=$(echo "${CURRENT_VERSION}" | cut -f2 -d.)
+    local -r PREVIOUS_MINOR_VERSION=$(( "${MINOR_VERSION}" - 1 ))
```
ggiguash, line 61: This code already exists in `build_images.sh` and partially in `get_latest_rpm_version.sh`.
It's an implementation detail. 
Should we abstract it from the test by putting it in a separate script and calling it in those scripts and also in the test?

test/bin/subscription_manager_register.sh
dhellmann: Good catch

test/scenarios/el92-src@rpm-install.sh
dhellmann: The VM boots into an image without either package installed, then the tests install MicroShift. I'll clarify the comment.

scripts/verify/verify-rf.sh
dhellmann: The command to run subscription manager to enable a repo is longer than 80 characters and trying to split it across lines as a shell command in RF was causing issues.

dhellmann:
New but unrelated infrastructure flake in the periodic test:

```
Building image-installer from rhel-9.2-microshift-source-isolated
+ 16:35:13.315092171 ./bin/build_images.sh:332 	build_cmd='sudo composer-cli compose start rhel-9.2-microshift-source-isolated image-installer'
++ 16:35:13.316922694 ./bin/build_images.sh:333 	sudo composer-cli compose start rhel-9.2-microshift-source-isolated image-installer
++ 16:35:13.317185196 ./bin/build_images.sh:333 	awk '{print $2}'
ERROR: ContainerResolveError: failed to resolve container: '': error getting manifest: too many requests to registry
```

The ARM job failed with known flakes on the networking tests.

test/scenarios/el92-src@rpm-install.sh
dhellmann: I cleaned this up.

PR #2574, opened 2023-11-06
More el93 scenarios

pmtk:


scripts/image-builder/configure.sh
```diff
@@ -13,11 +13,28 @@ auth_file_path = "/etc/osbuild-worker/pull-secret.json"
...
 
-sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
+# osbuild from COPR to install version that:
+# 1. can build rhel-93 images
+# 2. doesn't have issues with unexpected mirror's RPM verification
+#    (like the one in 9.3 beta at the moment of writing this comment)
+sudo dnf copr enable -y @osbuild/osbuild
+sudo dnf copr enable -y @osbuild/osbuild-composer rhel-9-x86_64
+
+sudo dnf install -y \
```
dhellmann, line 23: Can we do something to say "at least" this version, so that when it makes it into RHEL we get newer versions without having to change this code? I don't even know if that's possible with dnf.

test/scenarios-periodics/el93-crel@upgrade-from-el92-prel.sh
```diff
@@ -0,0 +1,35 @@
...
+# Scenario tests upgrade from RHEL9.2 to RHEL9.3 (including MicroShift).
+
+start_commit="rhel-9.2-microshift-4.$(previous_minor_version)"
+target_commit=rhel-9.3-microshift-crel
+
+scenario_create_vms() {
+    if ! does_commit_exist "${target_commit}"; then
+        echo "Commit '${target_commit}' not found in ostree repo - skipping test"
+        return 0
+    fi
```
dhellmann, line 14: I see this pattern repeated a few times. Maybe we should add an interface for the scenarios to report the images they require? I wonder if that could be optional.

test/scenarios-periodics/el93-crel@upgrade-from-el92-prel.sh
```diff
@@ -0,0 +1,35 @@
...
+    prepare_kickstart host1 kickstart.ks.template "${start_commit}"
+    launch_vm host1
+}
+
+scenario_remove_vms() {
+    if ! does_commit_exist "${target_commit}"; then
```
dhellmann, line 20: Cleanup should be safe even if the VM was not created, so I don't think this is needed.

test/scenarios-periodics/el93-crel@upgrade-from-el92-prel.sh
pmtk: I would love that. I was actually thinking about this and it could be long-term solution for our image blueprints manual grouping

scripts/image-builder/configure.sh
pmtk: That change was part of https://github.com/openshift/microshift/pull/2503.
I think if we'd remove explicit, we'd always get newest one?
Just FYI: these are main branch builds, so bleeding edge ;)

ggiguash: Ideally, we should stop using copr as soon as the functionality is integrated to RHEL. So, I would add a comment about this here, i.e. FIXME, etc.
For packages from copr, I recommended using something specific to avoid breaking our CI, etc.

pmtk:
Split into:
- https://github.com/openshift/microshift/pull/2583
- https://github.com/openshift/microshift/pull/2582
- https://github.com/openshift/microshift/pull/2581

PR #2576, opened 2023-11-06
Test for OCP-63099

dhensel-rh:

Closes #<Issue Number>

### DIFF
--- a/test/suites/standard/logging.robot
+++ b/test/suites/standard/logging.robot
@@ -16,39 +16,39 @@ Test Tags           configuration    etcd    restart    slow
 ${LEVELNORMAL}          SEPARATOR=\n
 ...                     ---
 ...                     debugging:
-...                     \ \ logLevel: NORMAL
+...                     \ \ LogLevel: NORMAL
 
 ${LEVELdebug}          SEPARATOR=\n
 ...                     ---
 ...                     debugging:
-...                     \ \ logLevel: debug
+...                     \ \ LogLevel: debug
 
 ${LEVELUNKNOWN}          SEPARATOR=\n
 ...                     ---
 ...                     debugging:
-...                     \ \ logLevel: Normal
+...                     \ \ LogLevel: Normal
 *** Test Cases ***
 Test LogLevel Normal
-    [Documentation]    Set logLevel to NORMAL
+    [Documentation]    Set LogLevel to NORMAL
     ...
-    ...    Test various spellings of the logLevelkeyword
+    ...    Test various spellings of the LogLevelkeyword
     ...    uppercase, lowercase, Capitol, and camelcase are supported
     [Setup]    Setup With Custom Config    ${LEVELNORMAL}
-    Expect loglevel   NORMAL
+    Expect LogLevel   NORMAL
 Test LogLevel Debug
-    [Documentation]    Set logLevel to debug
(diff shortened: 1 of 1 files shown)


test/suites/standard/logging.robot
```diff
@@ -0,0 +1,89 @@
...
+    ${merged}=    Extend MicroShift Config    ${config_content}
+    Upload MicroShift Config    ${merged}
+    Restart MicroShift
+
+Expect logLevel
+    [Documentation]    Verify that the logLevel matches the expected value
+    [Arguments]    ${expected}
+
+    ${stdout}=    Execute Command
+    ...    microshift show-config | awk '/logLevel/{print $2}'
```
dhellmann, line 86: You can use `Yaml Parse` to avoid needing to call awk like this. https://github.com/openshift/microshift/blob/main/test/suites/standard/show-config.robot#L35-L36 is an example of getting config output and checking the results directly.

dhellmann (commented):
Good idea for a test. One suggestion for making it easier to write inline.

dhellmann:
It looks like this PR is changing files that would be updated by running the microshift rebase script. Is that intentional, or a side-effect of something in your git repo?

dhensel-rh:
> It looks like this PR is changing files that would be updated by running the microshift rebase script. Is that intentional, or a side-effect of something in your git repo?

I am trying to add one file. This must be something in my git repo. I attempted a rebase yesterday. I wonder if I did something incorrectly.

test/suites/standard/logging.robot
```diff
@@ -0,0 +1,93 @@
...
+    ${merged}=    Extend MicroShift Config    ${config_content}
+    Upload MicroShift Config    ${merged}
+    Restart MicroShift
+
+Expect LogLevel
+    [Documentation]    Verify that the logLevel matches the expected value
+    [Arguments]    ${expected}
+
+    ${output}    ${stdout}=    Execute Command
+    ...    microshift show-config
```
dhellmann, line 90: The show-config command reads the file and reports the contents, so it's not strictly necessary to restart microshift as part of these tests. If you wanted to ensure that the microshift service picked up the settings correctly after a restart, you could use some the keywords we have for watching the logs and picking out patterns. test/suites/standard/log-scan.robot has some examples and the keywords are defined in test/resources/journalctl.py

dhellmann:
This looks good. There are some similar tests in https://github.com/openshift/microshift/blob/main/test/suites/standard/configuration.robot so maybe the 2 suites can be combined here into this new file?

I'll approve the PR as something we want. We should give another member of the team a chance to review it and lgtm it.

/approve

dhellmann:
> This looks good. There are some similar tests in https://github.com/openshift/microshift/blob/main/test/suites/standard/configuration.robot so maybe the 2 suites can be combined here into this new file?

Now that https://github.com/openshift/microshift/pull/2832 has merged, you probably want to rebase this before considering how to combine the 2 suites.

dhensel-rh:
> > This looks good. There are some similar tests in https://github.com/openshift/microshift/blob/main/test/suites/standard/configuration.robot so maybe the 2 suites can be combined here into this new file?
> 
> Now that #2832 has merged, you probably want to rebase this before considering how to combine the 2 suites.

Thanks. I'll rebase and have a look at combining the 2 suites.

PR #2577, opened 2023-11-06
USHIFT-1848: add periodic test scenario for the standard tests on rpm-based system

dhellmann:


dhellmann:
/hold

The dependencies should land first

test/scenarios/el92-src@rpm-install.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+
+export SKIP_GREENBOOT=true
+export TEST_RANDOMIZATION=none
+
+scenario_create_vms() {
+    prepare_kickstart host1 kickstart-liveimg.ks.template ""
+    launch_vm host1 "rhel-9.2"
```
eslutsky, line 10: "rhel-9.2" is the [default](https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L12) blueprint name, do we need to be explicit about it ?

dhellmann: I like being explicit, but it's not necessary.

test/bin/scenario.sh
```diff
@@ -497,6 +497,61 @@ remove_vm() {
...
+fi
+EOF
+        copy_file_to_vm "${vmname}" /tmp/sub.sh /tmp/sub.sh
+        copy_file_to_vm "${vmname}" /tmp/subscription-manager-org /tmp/subscription-manager-org
+        copy_file_to_vm "${vmname}" /tmp/subscription-manager-act-key /tmp/subscription-manager-act-key
+        run_command_on_vm "${vmname}" "chmod +x /tmp/sub.sh && sudo /tmp/sub.sh"
+    else
+        # Local developer workflow
+        run_command_on_vm "${vmname}" "sudo subscription-manager register"
+    fi
```
ggiguash, line 552: I would like to suggest a few alternatives for the current implementation. The goal is not to include CI-specific code in the `scenario.sh` generic script and avoid prompt when running tests in devenv:

- Introduce a scenario setting parameter pointing to subscription information
- Move org and activationKey detection code to the scenario and pass the values to the function.
- Can we use the subscription technique used by containers?

dhellmann: This commit is really part of #2568 so I made the change there. I will rebase this PR when that one lands.

dhellmann:
Rebasing this is awkward, so I'm going to submit a new PR with a new branch.

/close

PR #2581, opened 2023-11-07
USHIFT-1492: Test upgrade el92 to el93 without microshift upgrade

pmtk:


ggiguash:
@pmtk , Looks like we have RPM instalation problems that prevent the tests from running.

pmtk:
/test metal-periodic-test

booting cncf failed

PR #2585, opened 2023-11-07
USHIFT-1846: Reorganize image blueprints into layers of job types

ggiguash:
The restructuring of the image blueprint directories introduces a notion of "layers" to make it more explicit that base and presubmit blueprints are prerequisites to the periodics. It also helps to keep the necessary abc order of execution.


test/bin/ci_phase_iso_build.sh
```diff
@@ -75,13 +75,13 @@ run_image_build() {
...
     if [ -v CI_JOB_NAME ] ; then
-        $(dry_run) bash -x ./bin/build_images.sh ${build_opts} -g ./image-blueprints/group1
-        $(dry_run) bash -x ./bin/build_images.sh ${build_opts} -g ./image-blueprints/group2
-        # Groups 3 and 4 contains images exclusively used in periodic CI jobs
+        $(dry_run) bash -x ./bin/build_images.sh ${build_opts} -l ./image-blueprints/layer1-base
+        $(dry_run) bash -x ./bin/build_images.sh ${build_opts} -l ./image-blueprints/layer2-presubmit
         if [[ "${CI_JOB_NAME}" =~ .*periodic.* ]]; then
-            $(dry_run) bash -x ./bin/build_images.sh ${build_opts} -g ./image-blueprints/group3
-            $(dry_run) bash -x ./bin/build_images.sh ${build_opts} -g ./image-blueprints/group4
+            $(dry_run) bash -x ./bin/build_images.sh ${build_opts} -l ./image-blueprints/layer3-periodic
```
pmtk, line 84: Correct me if I'm wrong, but seems we'd have the same order we already have?

I thought we'd build `rhel93.image-installer` and `rhel93.toml` along `rhel92.toml` in periodics. Then we could get rid of `layer3-periodic/group2` by moving `rhel93-source.toml` to `l3-periodic/group1`

ggiguash: That's an optimization we can apply after the rename. To make it straighforward, I did not want to touch the toml structure.

ggiguash: The good thing is that restructuring no longer afftects any scripts.

test/bin/build_images.sh
```diff
@@ -478,10 +492,14 @@ configure_package_sources
 
 trap 'osbuild_logs' EXIT
 
-if [ -n "${GROUP}" ]; then
+if [ -n "${LAYER}" ]; then
+    for group in "${TESTDIR}"/image-blueprints/"${LAYER}"/group*; do
+        do_group "${group}" ""
```
dhellmann, line 497: Does do_group not need to change to understand that the groups are in the layer directory now?

test/bin/build_images.sh
```diff
@@ -424,14 +435,17 @@ while getopts "iIg:st:h" opt; do
...
-            echo
-            usage
+            usage "ERROR: Unknown option ${opt}"
             exit 1
             ;;
     esac
 done
 
+if [ -n "${LAYER}" ] && [ -n "${GROUP}" ] ; then
+    usage "The layer and group options are mutually exclusive"
```
dhellmann, line 445: Do they have to be? Maybe we can undo that requirement later?

test/bin/build_images.sh
ggiguash: Layer is just a logical grouping, so I do not think we need any changes. I have not run the CI yet :). Let me try it just in case.

test/bin/build_images.sh
ggiguash: It's more of a logical problem. If I specify a group, a layer is already implied. It's similar for the template, which is on the 3rd level. 
Maybe, we should have all the 3 of layer, group and template mutually exclusive?

ggiguash:
The periodic test failed with a registry DoS. Restarting.
```
ERROR: ContainerResolveError: failed to resolve container: '': error getting manifest: too many requests to registry
```
/test metal-periodic-test

pmtk:
/lgtm
/hold 
unhold when ready

PR #2589, opened 2023-11-08
USHIFT-1868: Remove readonly flag from vars in loop

pmtk:
Local variables are function scoped, not block (loop) scoped, so on subsequent executions of loop-body they cannot be reassigned resulting in error.


pmtk:
/hold

Let's wait until we're sure periodic/cncf works (periodics not required)

PR #2590, opened 2023-11-08
USHIFT-1870: Record sos-report in junit

pmtk:


test/bin/scenario.sh
```diff
@@ -72,31 +72,62 @@ copy_file_to_vm() {
...
-        # can't rely on the wrapper being there or working if it
-        # is. Copy the script to the host, just in case, along with a
-        # wrapper that knows how to execute it or the installed
-        # version.
-        cat - >/tmp/sos-wrapper.sh <<EOF
+
+        if ! sos_report_for_vm "${vmdir}" "${vmname}" "${ip}"; then
+            scenario_result=1
+            if "${junit}"; then
+                record_junit "${vmname}" "sos-report" "FAILED"
```
dhellmann, line 109: I wonder if we really want to fail a job if only the SOS gather fails. Maybe? Let's keep it and see how it goes.

dhellmann (commented):
/lgtm

test/bin/scenario.sh
pmtk: Actually that would be useful. For example, if nightlies wouldn't stop after iso-boot because of this ([example - 4.14 metal nightly](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-microshift-release-4.14-ocp-metal-nightly/1722955034682986496/artifacts/ocp-metal-nightly/openshift-microshift-infra-iso-boot/artifacts/scenario-info/el92-src@fdo/boot.log)) then at least we would have some test runs.
However, to ignore them I guess we would have to add two junit entries: fail and success, so prow interprets them as flaky?
And also make sure that at the end of the final trap, we return the status of the overall procedure, not sos-report. On the other hand, when we have some syntax errors, we want it to fail...

PR #2599, opened 2023-11-11
USHIFT-1845 USHIFT-1846: Update testing framework documentation with the new layering and caching techniques

ggiguash:


test/README.md
```diff
@@ -158,26 +158,26 @@ rhsm = true
...
+images can be reused.
+
+Images are organized into "layers" that contain "groups".
+
+Each layer has its designation:
+- `base` layer is a prerequisite for all the subsequent ones
+- `presubmit` layer contains images used in presubmit CI jobs
+- `periodic` layer contains images used in periodic CI jobs
+
+Each group in a given layer is built in order and all of the images in
```
dhellmann, line 170: Does this mean it builds each group of each layer in order as a separate step

base group 1
base group 2
base group 3
presubmit group 1
presubmit group 2

or does it build all of the same group across all layers at the same time

group 1 from base and presubmit
group 2 from base and presubmit

?

ggiguash: I have not implemented parallelization on the level of job type yet, so it's the former case. We can apply this optimization in the future if we see it's necessary. 
I changed the wording in the doc to make it more explicit. Is this better now?

pmtk:
/lgtm
/hold
unhold when ready

ggiguash:
Documentation change. Metal tests irrelevant.
/override ci/prow/microshift-metal-tests ci/prow/metal-periodic-test ci/prow/images

PR #2603, opened 2023-11-13
USHIFT-1949: Require openvswitch3.1 or openvswitch to allow installation on Fedora

aleskandro:
The version number is not part of the openvswitch package name in Fedora. This PR proposes changing the networking rpmspec requirement to either openvswitch3.1 (RHEL or CS9) or openvswitch (Fedora).

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -109,7 +109,7 @@ The microshift-selinux package provides the SELinux policy modules required by M
 %package networking
 Summary: Networking components for MicroShift
 Requires: microshift = %{version}
-Requires: openvswitch3.1
+Requires: (openvswitch3.1 or openvswitch >= 3.1)
 Requires: NetworkManager
 Requires: NetworkManager-ovs
 Requires: jq


packaging/rpm/microshift.spec
```diff
@@ -109,7 +109,7 @@ The microshift-selinux package provides the SELinux policy modules required by M
 %package networking
 Summary: Networking components for MicroShift
 Requires: microshift = %{version}
-Requires: openvswitch3.1
+Requires: (openvswitch3.1 or openvswitch)
```
ggiguash, line 112: I'm a bit uneasy with this change because we're basically allowing unversioned package installation for `openvswitch`. We do not support Fedora, so I would vote not to make this change.

aleskandro: Hi @ggiguash , I could even change this to 
```
(openvswitch3.1 or openvswitch >= 3.1)
```

but that's still nice to have if that can just help to let the installation go, although in a non-officially-supported scenario.

ggiguash: That code would certainly be better. 
@dhellmann / @pacevedom , do you have any objections from merging this code?
```
(openvswitch3.1 or openvswitch >= 3.1)
```

aleskandro: Thanks @ggiguash, let me push it first and run some testing even on Fedora.

aleskandro: confirmed as working good in fc38 too.

ggiguash:
/retitle USHIFT-1949: Require openvswitch3.1 or openvswitch to allow installation on Fedora

ggiguash:
Failures have nothing to do with the change.

PR #2604, opened 2023-11-13
OCPBUGS-23220: Remove keyfile nm plugin force when installing

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/packaging/network-manager-conf/microshift-nm.conf
+++ b/packaging/network-manager-conf/microshift-nm.conf
@@ -1,6 +1,0 @@
-# This file configures NetworkManager to use the keyfile plugin
-# that will make OVS initialization work with IP static addressing
-
-[main]
-plugins = keyfile
-
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -188,9 +188,6 @@ install -p -m644 packaging/crio.conf.d/microshift_amd64.conf %{buildroot}%{_sysc
 
 install -p -m644 packaging/crio.conf.d/microshift-ovn.conf %{buildroot}%{_sysconfdir}/crio/crio.conf.d/microshift-ovn.conf
 
-install -d -m755  %{buildroot}%{_sysconfdir}/NetworkManager/conf.d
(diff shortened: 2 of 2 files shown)


packaging/network-manager-conf/microshift-nm.conf
```diff
@@ -1,6 +0,0 @@
-# This file configures NetworkManager to use the keyfile plugin
-# that will make OVS initialization work with IP static addressing
```
dhellmann, line 2: This comment makes me think this is required for some configurations. Is that not actually the case?

pacevedom: If you configure network manager to already have these plugins then there is no need to override anything.
This is the default config for nm:
```
# Configuration file for NetworkManager.
#
# See "man 5 NetworkManager.conf" for details.
#
# The directories /usr/lib/NetworkManager/conf.d/ and /run/NetworkManager/conf.d/
# can contain additional .conf snippets installed by packages. These files are
# read before NetworkManager.conf and have thus lowest priority.
# The directory /etc/NetworkManager/conf.d/ can contain additional .conf
# snippets. Those snippets are merged last and overwrite the settings from this main
# file.
#
# The files within one conf.d/ directory are read in asciibetical order.
#
# You can prevent loading a file /usr/lib/NetworkManager/conf.d/NAME.conf
# by having a file NAME.conf in either /run/NetworkManager/conf.d/ or /etc/NetworkManager/conf.d/.
# Likewise, snippets from /run can be prevented from loading by placing
# a file with the same name in /etc/NetworkManager/conf.d/.
#
# If two files define the same key, the one that is read afterwards will overwrite
# the previous one.
[main]
#plugins=keyfile,ifcfg-rh
```

Also note this line:
```
# If two files define the same key, the one that is read afterwards will overwrite
# the previous one.
```
So it might even override what it shouldnt. This config file was introduced in july 2022, so the content might be dated?

pliurh: I think this file was added to support the microshift-ovs-init.service. That service used to require NM to use keyfile plugin.  But now as we decoupled the br-ex bridge from the physical interface. I think we no longer need this file.

PR #2617, opened 2023-11-16
upload CI built RPMs to s3 Cache

eslutsky:


test/bin/manage_build_cache.sh
```diff
@@ -100,6 +100,19 @@ action_upload() {
...
+    # Upload RPMs
+    local -r rpm_src="${src_base}/rpms.tar"
+    local -r rpm_dst="${dst_base}/rpms.tar"
+
+    # Archive the rpms files before the upload
+    rm -f "${rpm_src}"
+    tar cf "${rpm_src}" -C "${RPM_SOURCE}" RPMS
+
+    local -r rpm_size="$(du -csh "${rpm_src}" | awk 'END{print $1}')"
+    echo "Uploading ${rpm_size} of rpms to '${rpm_dst}'"
```
ggiguash, line 113:
```suggestion
    echo "Uploading ${rpm_size} of RPMs to '${rpm_dst}'"
```

test/bin/manage_build_cache.sh
ggiguash, line 1: I've a bit of a problem with the non-symmetric implementation of RPM upload, while no download / verify operations were implemented.

We cannot implement download the way it is now because that would override the CI RPM artfifacts. So, we should think about how to implement this properly.

Can I suggest uploading and downloading into a special rpm-source-cache directory? This way, the CI will simply ignore these files and we should be able to use them explicitly when necessary.

ggiguash (changes requested):

test/bin/manage_build_cache.sh
eslutsky: for download case it does make sense  to save them  outside "${RPM_SOURCE}" directory, and yes,  its only downloaded  on special  cases when explicitly required.
for upload either way will work  for example: ${dst_base}/rpm-source-cache/ ?

ggiguash: So, to be explicit:
- When uploading RPMs, we should tar them up and save them at `${AWS_BUCKET_NAME}/<branch>/<arch>/<tag>/rpm-source-cache.tar`
- When downloading RPMs, we should unpack them at `${IMAGEDIR}/rpm-source-cache.tar`
- The `valid` operation should check for presence of `${AWS_BUCKET_NAME}/<branch>/<arch>/<tag>/rpm-source-cache.tar` in the bucket.

Does it make sense?

eslutsky:
closing for now

PR #2619, opened 2023-11-16
USHIFT-1968: SELinux syntax cleanup

zpytela:
There is no real problem, just some enhancements.

What would help in the future is if there was the /var/lib/microshift directory and it was a part of the main rpm with possible subdirectories like backups or microshift-backups, saved etc. Packaged files will get the correct label and files inside it will inherit the type from their parent, subdirectories do not need to be packaged if they are not always used. It would then be possible to drop the filetrans calls, so effectively have the .te file empty, also not declaring the microshift_t type which is not used anywhere.


ggiguash:
@zpytela , the current proposal seems to break CI tests [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_microshift/2619/pull-ci-openshift-microshift-main-microshift-metal-tests/1725397165527797760/artifacts/microshift-metal-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el92-src@standard-suite/log.html).
Please, discuss with @eggfoobar on how to move ahead with this.

eggfoobar:
Hey @zpytela Thanks for opening this up, the failure in the test is due to this [line](https://github.com/openshift/microshift/blob/main/test/resources/selinux.py#L61) from `/var/lib/microshift.saved(/.*)?` to `/var/lib/microshift\.saved(/.*)? ` That should resolve the test failure

dhellmann:
> Hey @zpytela Thanks for opening this up, the failure in the test is due to this [line](https://github.com/openshift/microshift/blob/main/test/resources/selinux.py#L61) from `/var/lib/microshift.saved(/.*)?` to `/var/lib/microshift\.saved(/.*)? ` That should resolve the test failure

Several tests are still failing for these reasons.

zpytela:
The only actual change is escaping the dot in file context specification as it accepts regexps. Anyway, for now I'd rather stick to the idea of having just /var/lib/microshift in the package and subsequently in configuration.

dhellmann:
> The only actual change is escaping the dot in file context specification as it accepts regexps. Anyway, for now I'd rather stick to the idea of having just /var/lib/microshift in the package and subsequently in configuration.

We did consider the change as part of the release plan for 4.14. Unfortunately rearranging the files will be a breaking change for existing deployments, and we had users with supported systems on earlier versions. We need to ensure we have good upgrade and downgrade plans in place before making the change.

eggfoobar:
/close

Closing now that #2746

PR #2636, opened 2023-11-21
USHIFT-1553: Use right OCPVERSION when enabling repos

pmtk:
Upgrades should be tested from 4.14 GA, not RC.


pmtk:
Obsoleted by https://github.com/openshift/microshift/pull/2640

PR #2640, opened 2023-11-22
USHIFT-1928: Switch to OCP 4.14 dependencies for building ISOs and ostree commits

ggiguash:
Also includes [USHIFT-1553](https://issues.redhat.com//browse/USHIFT-1553)  fix.
As a byproduct of this change, we need to upgrade our devenv to RHEL 9.3 because latest osbuild dependencies are no longer available on RHEL 9.2.


ggiguash:
We need this PR merged to unblock CI tests in a different region. Otherwise, no jobs can pass without osbuild package update.

PR #2647, opened 2023-11-24
OCPQE-17114: Stress Testing

agullon:
Stress Testing script to enable/disable conditions (limiting resources) on local or remote host:
- New `stress_testing.sh` script added.
- `run.sh` add a new `-s` option to set the Stress Condition
- `scenario.sh` add new `stress_testing` function to enable/disable Stress condition before/after `run_tests`
- Add a new package `iproute-tc` to blueprints
    - This package is needed to run `tc qdisc` commands for networking Stress conditions.

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -557,7 +557,7 @@ stress_testing() {
     local -r condition="${3}"
     local -r value="${4}"
 
-    local -r ssh_host="${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/public_ip"
+    local -r ssh_host="$(get_vm_property "${vmname}" public_ip)"
     local -r ssh_user=redhat
     local -r ssh_port="${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/ssh_port"
     local -r ssh_pkey="${SSH_PRIVATE_KEY:-}"
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -550,6 +550,31 @@ next_minor_version() {
     echo $(( $(current_minor_version) + 1 ))
 }
 
+# Enable/Disable Stress Condition
+stress_testing() {
+    local -r vmname="${1}"
+    local -r action="${2}"
+    local -r condition="${3}"
+    local -r value="${4}"
+
+    local -r ssh_host="${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/public_ip"
+    local -r ssh_user=redhat
+    local -r ssh_port="${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${vmname}/ssh_port"
+    local -r ssh_pkey="${SSH_PRIVATE_KEY:-}"
+
+
+	if [ "${action}" == "enable" ]; then
+        echo "${action}d stress condition: ${condition} ${value}"
+        "${SCRIPTDIR}/stress_testing.sh" -e "${condition}" -v "${value}" -h "${ssh_host}" -u "${ssh_user}" -p "${ssh_port}" -k "${ssh_pkey}"
+    elif [ "${action}" == "disable" ]; then
+        echo "${action}d stress condition: ${condition}"
+        "${SCRIPTDIR}/stress_testing.sh" -d "${condition}" -h "${ssh_host}" -u "${ssh_user}" -p "${ssh_port}" -k "${ssh_pkey}"
+    else
+        error "Invalid Stress Testing action"
+        exit 1
+    fi
+}
+
 # Run the tests for the current scenario
 run_tests() {
     local -r vmname="${1}"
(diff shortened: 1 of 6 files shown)


dhellmann (commented):
This looks good. I have a few small comments inline, but nothing that I think should hold you up.

What's the plan for running the tests? Do you want to set up a new periodic CI job, for example, or are you just planning to run them in the QE environment for now?

test/run.sh
```diff
@@ -11,23 +11,23 @@ RF_VARIABLES="${SCRIPTDIR}/variables.yaml"
 DRYRUN=false
 OUTDIR="${ROOTDIR}/_output/e2e-$(date +%Y%m%d-%H%M%S)"
 
-
 function usage {
     local -r script_name=$(basename "$0")
     cat - <<EOF
-${script_name} [-h] [-n] [-o output_dir] [-v venv_dir] [-i var_file] [test suite files]
+${script_name} [-h] [-n] [-o output_dir] [-v venv_dir] [-i var_file] [-s name-value] [test suite files]
```
ggiguash, line 17: Can we use `name=value` format?

test/bin/scenario.sh
```diff
@@ -550,6 +550,30 @@ next_minor_version() {
     echo $(( $(current_minor_version) + 1 ))
 }
 
+# Enable/Disable Stress Condition
+stress_testing() {
```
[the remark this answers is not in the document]
ggiguash: We need to document its arguments in the comment please.

test/bin/stress_testing.sh
```diff
@@ -0,0 +1,229 @@
...
+    run sudo chmod +x ${WONDERSHAPER} &> /dev/null
+    check_command "sudo ${WONDERSHAPER} -v"
+  fi
+
+  # speedtest-cli
+  if ( ! ${SPEEDTEST_CLI} --version &> /dev/null);  then
+    run sudo curl -Lo ${SPEEDTEST_CLI} https://raw.githubusercontent.com/sivel/speedtest-cli/master/speedtest.py &> /dev/null
+    run sudo chmod +x ${SPEEDTEST_CLI} &> /dev/null
+    check_command "${SPEEDTEST_CLI} --version"
+  fi
```
ggiguash, line 62: This code is not going to work on ostree systems because /usr/local/bin is read-only.
Should we put these utilitties in the home directory instead?

ggiguash:
/approve
Need to fix a few things according to the comments.

test/bin/stress_testing.sh
agullon: No need to change the directory. There's a symlink from `/usr/local` to `/var/usrlocal` on ostree systems. I tested it and it works.

ggiguash: Hah, nice. I did not know that.

agullon:
> This looks good. I have a few small comments inline, but nothing that I think should hold you up.
> 
> What's the plan for running the tests? Do you want to set up a new periodic CI job, for example, or are you just planning to run them in the QE environment for now?

Thanks! There's not a well-defined plan yet where to use these Stress Conditions. The goal of this PR is to set a Stress Testing framework to be used by any existing Test Case or Test Suite.

dhellmann (commented):
/retest required

test/bin/scenario.sh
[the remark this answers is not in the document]
agullon: I added it with the idea to explicitly invoke this function from `scenarios/*.sh` scripts. But I'm happy with both approaches. Also, we need to discuss if it makes sense to apply Stress Conditions to all the scenarios or create new scenarios for that.

agullon:
> I think we have 2 approaches we can take in scenario.sh. Either scenarios explicitly invoke this function, or we add new actions like stress-enable and stress-disable that can be used to apply the settings after running create. The second approach has the benefit of allowing us to reuse existing scenarios, so unless there's a strong reason for the first approach (or some better third approach?) I think we should plan on that.

> If we're going to add a new action, this function can eventually evolve into the implementation of that action function.

hi @dhellmann my idea for this PR is to create a Stress Testing framework where we can limit resources at OS level for any Test or Test Suite. I worked on the assumption to explicitly call this function before any Test or Test Suite, to apply the Stress Condition before running the Tests. So we may need to duplicate existing Scenario tests if we want to rerun them under Stress Conditions.

It's also fine for me if we evolve this function to stress-enable and stress-disable actions after running create. As I said, this is just a framework, we all need to agree on the better way to integrate it into our Testing workflow.

ggiguash:
We are currently having "known" problems with the CI and the tests that failed in this PR do not seem to have anything to do with the changes.
Overriding those tests to save on CI cycles.

/ovveride ci/prow/metal-periodic-test ci/prow/metal-periodic-test-arm ci/prow/microshift-metal-tests-arm

ggiguash:
/override ci/prow/metal-periodic-test-arm  ci/prow/microshift-metal-tests-arm

PR #2648, opened 2023-11-24
USHIFT-1769: Abort rebase if manifests change

pmtk:


scripts/auto-rebase/rebase.sh
```diff
@@ -1052,6 +1075,9 @@ rebase_to() {
     copy_manifests
     update_openshift_manifests
     if [[ -n "$(git status -s assets)" ]]; then
+        if [[ -n "${FAIL_ON_MANIFEST_CHANGE+x}" ]] && [[ "${FAIL_ON_MANIFEST_CHANGE}" == "1" ]]; then
```
dhellmann, line 1078: What is the `+x` part in the variable expansion?

pmtk: If the `FAIL_ON_MANIFEST_CHANGE` is set, then it'll be substituted with `x` and `-n` will return true as the string is not empty. If `FAIL_ON_MANIFEST_CHANGE` is unset, it'll be just an empty string and `-n` will fail.

Yet another way of checking if the variables is set/is not null so it can be used without risking `unbound variable` error.

PR #2652, opened 2023-11-25
USHIFT-1941: Fix osconfig tests failing due to unplanned reboots and service failures

ggiguash:
Besides adding `Wait Until Greenboot Health Check Exited` calls following the `Reboot MicroShift Host` function, the following fixes were implemented:
- Make the `systemctl` command retry less frequently (once in 15s instead of 10s) not to hit the 10s start limit
- Fixed the `Systemctl enable --now` call, which was wrong because `Systemctl` does not expect such usage
- Limited the journalctl output to 32K characters instead of 100 lines. The 32K characters is the max size of RF buffer anyway.

### DIFF
--- a/test/resources/systemd.resource
+++ b/test/resources/systemd.resource
@@ -52,7 +52,7 @@ Systemctl    # robocop: disable=too-long-keyword
         ...    sudo=True
         ...    return_stdout=True
         ${log_text}=    Execute Command
-        ...    journalctl -u ${unit_name} -o short | tail -n 100
+        ...    journalctl -u ${unit_name} -o short | tail -c 32000
         ...    sudo=True
         ...    return_stdout=True
     END
@@ -74,7 +74,9 @@ Systemctl    # robocop: disable=too-long-keyword
 Systemctl With Retry
     [Documentation]    Run Systemctl keyword but retry 10 times
     [Arguments]    ${verb}    ${unit_name}
-    Wait Until Keyword Succeeds    10x    10s
+    # Need to wait for more than 10s between the attempts not to
+    # trigger the service start limit condition (10s by default)
+    Wait Until Keyword Succeeds    6x    15s
     ...    Systemctl    ${verb}    ${unit_name}
 
 Systemctl Daemon Reload
--- a/test/suites/osconfig/systemd-resolved.robot
+++ b/test/suites/osconfig/systemd-resolved.robot
@@ -10,6 +10,8 @@ Resource            ../../resources/systemd.resource
 Suite Setup         Setup
 Suite Teardown      Teardown
 
+Test Tags           slow
+
 
 *** Variables ***
 ${KUBELET_CONFIG_FILE}      /var/lib/microshift/resources/kubelet/config/config.yaml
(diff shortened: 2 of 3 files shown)


test/resources/systemd.resource
```diff
@@ -52,7 +52,7 @@ Systemctl    # robocop: disable=too-long-keyword
         ...    sudo=True
         ...    return_stdout=True
         ${log_text}=    Execute Command
-        ...    journalctl -u ${unit_name} -o short | tail -n 100
+        ...    journalctl -u ${unit_name} -o short | tail -c 32000
```
dhellmann, line 55: Maybe we could write the output to a file instead of truncating it?

test/suites/osconfig/systemd-resolved.robot
```diff
@@ -129,6 +131,7 @@ Stop Systemd-Resolved
 
     Systemctl    stop    systemd-resolved
     Reboot MicroShift Host
+    Wait Until Greenboot Health Check Exited
```
dhellmann, line 134: Should we just do this in "Reboot MicroShift Host"? That would mean we don't have to remember to do this 2 step process every time we want to reboot.

ggiguash: We are calling reboot in a few other places. Going through those places, we do not always need to wait for greenboot. I think it's the problem in a specific test rather than a global one.

test/resources/systemd.resource
ggiguash: This command includes the output in the RF log. The idea was to limit the size of the buffer to 32K, which is anyway the max size displayed in the RF log. If we redirect it to a file, it won't be in the log.
Maybe, I'm missing something.

test/suites/osconfig/systemd-resolved.robot
dhellmann: That's surprising. What's an example of a case where we do want to reboot but do not need to wait for the system to come all the way up?

ggiguash: There are different ways for the system to come up. For example, in this case we only need MicroShift service to start.
https://github.com/openshift/microshift/blob/main/test/suites/backup/backup-restore-on-reboot.robot#L27

ggiguash: I think this is a bigger change that might be better handled in a separate task is we decide to implement it. 
From looking at the sources, I found a couple of places where tests need to be updated if we make such a  change.

For example:
- This case might be broken because we need to switch SSH to another IP before retrying greenboot checks
https://github.com/openshift/microshift/blob/main/test/suites/network/multi-nic.robot#L217
- This case expects a failing greenboot https://github.com/openshift/microshift/blob/main/test/resources/ostree.resource#L49

ggiguash: Should I create a separate task for this work?

dhellmann: OK, it sounds like we do have some cases where we would not want to wait automatically. Maybe we want 2 separate keywords, where the most obvious one does wait and another one does not? Let's leave it for now and see how things go. I don't think we need a ticket for it.

PR #2660, opened 2023-11-28
WIP: Fail fast for the greenboot checks in schenario.sh

ggiguash:


test/bin/scenario.sh
```diff
@@ -247,37 +247,77 @@ wait_for_greenboot() {
...
+        local svc_state
+
+        svc_state="$(${ssh_cmd} systemctl show --property=SubState --value greenboot-healthcheck || true)"
+        if [ "${svc_state}" = "exited" ] ; then
+            record_junit "${vmname}" "greenboot-check" "OK"
             return 0
         fi
+
+        # Print the last log and check for terminal failure
+        ${ssh_cmd} "sudo journalctl -n 10 -u greenboot-healthcheck" || true
```
pmtk, line 275: This looks like will print the logs on each loop-body execution when svc isn't neither `exited` nor `failed`. Should this be inside `if` below?

ggiguash: In the original code, it was printed unconditionally on every iteration to show greenboot progress. I guess, I can change this, but it's not how it was

pmtk: Ah, right, ack. I though this is something extra

test/bin/scenario.sh
```diff
@@ -247,37 +247,77 @@ wait_for_greenboot() {
...
+    record_junit "${vmname}" "greenboot-check" "FAILED"
     return 1
 }
 
 start_junit() {
     local outputfile="${SCENARIO_INFO_DIR}/${SCENARIO}/vms/junit.xml"
     mkdir -p "$(dirname "${outputfile}")"
 
-    echo "Creating ${outputfile}"
+    if [ -e "${outputfile}" ] ; then
```
dhellmann, line 309: I fixed this a different way in https://github.com/openshift/microshift/pull/2664

dhellmann (commented):
The retry logic looks good. I've done some different work that will conflict with the junit changes but solves the problem in a more robust way, so maybe we can agree on the order to merge.

test/bin/scenario.sh
ggiguash: OK, let me approve that PR and I'll rebase mine on top of it.

PR #2663, opened 2023-11-28
USHIFT-1861 Router/Ingress Smoke tests should wait for env cleanup

copejon:


test/suites/standard/networking-smoke.robot
```diff
@@ -63,6 +65,20 @@ Delete Hello MicroShift Route
     [Documentation]    Delete route for cleanup.
     Oc Delete    route/hello-microshift -n ${NAMESPACE}
 
+Wait For Service Deletion With Timeout
```
pacevedom, line 68: Is `--wait=true` not waiting for complete deletion in the delete keyword above?

dhellmann: Do you mean on line 66? I would expect a wait there to actually block, and the controllers to delete the service and endpoint before deleting the route.

Maybe the `Oc Delete` keyword should always add `--wait`?

copejon: I'm going to 2x this. Before I went this route, I'd tried `--wait` and examined to logs to see what the deletion sequence looked like.  AFAICT though, the deletions are asynchronous.  This makes sense as the service does not have a finalizer placed on it to pause it's deletion.

copejon: Examining the logs + referring to upstream docs, the EP deletion is asynchronous.  The endpoint is not reconciled until the service disappears:
```
Dec 04 12:10:38 el92-src-standard-suite-host1 microshift[13935]: microshift-loadbalancer-service-controller I1204 12:10:38.429501   13935 controller.go:150] Service default/hello-microshift does not exist anymore
Dec 04 12:10:38 el92-src-standard-suite-host1 microshift[13935]: kube-controller-manager I1204 12:10:38.435518   13935 garbagecollector.go:672] "Deleting item" item="[discovery.k8s.io/v1/EndpointSlice, namespace: default, name: hello-microshift-b7jvm, uid: ce614dbc-e818-47cd-a880-62b8778a5b25]" propagationPolicy="Background"
```

The delta between the events on the backend is on the order of hundredths of a second, but that does not include the latency in updating the cluster's object store to the latest cluster state.

test/suites/standard/networking-smoke.robot
```diff
@@ -63,6 +65,20 @@ Delete Hello MicroShift Route
...
+    [Documentation]    Polls for service and endpoint by "app=hello-microshift" label for 1 minute.    Fails if timeout
+    ...    expires. This check is unique to this test suite because each test here reuses the same namespace. Since
+    ...    the tests reuse the service name, a small race window exists between the teardown of one test and the setup
+    ...    of the next.    This produces flakey failures when the service or endpoint names collide.
+    ${match}=    Catenate    No resources found in    ${NAMESPACE}    namespace.
+    ${match}=    Remove String    ${match}    "
+    FOR    ${i}    IN RANGE    60
```
dhellmann, line 75: Instead of writing out this loop, could we use `Wait Until Keyword Succeeds` here?

test/suites/standard/networking-smoke.robot
dhellmann: OK, it sounds like we need the test to be a little smart about what it waits for.

test/suites/standard/networking-smoke.robot
copejon: fixed!

copejon:
Used the more sane and readable "Wait Until Keyword Succeeds" convention used throughout the test.  Wish I could say why I didn't do that in the beginning 😅

test/suites/standard/networking-smoke.robot
```diff
@@ -63,6 +65,22 @@ Delete Hello MicroShift Route
     [Documentation]    Delete route for cleanup.
     Oc Delete    route/hello-microshift -n ${NAMESPACE}
 
+Wait For Service Deletion With Timeout
```
dhellmann, line 68: I probably would have called this "The Service Should Be Deleted" but it's not worth holding up this PR to change it.

dhellmann (commented):
/lgtm

Nice work figuring out what was causing this flakiness!

PR #2667, opened 2023-11-29
USHIFT-1954: Fix the check for resolvConf option in the kubelet configuration file

ggiguash:
Looks like the latest systemd-resolved package added quotes on the file name.
```
# grep "^resolvConf" /var/lib/microshift/resources/kubelet/config/config.yaml
resolvConf: "/run/systemd/resolve/resolv.conf" 
```

### DIFF
--- a/test/suites/osconfig/systemd-resolved.robot
+++ b/test/suites/osconfig/systemd-resolved.robot
@@ -30,9 +30,7 @@ Verify Kubelet Config With Systemd-Resolved Running
     Restart MicroShift
 
     # Verify the presence of the kubelet option
-    ${rc}=    Execute Command
-    ...    grep -q "resolvConf: ${RESOLVE_CONF_FILE}" ${KUBELET_CONFIG_FILE}
-    ...    sudo=True    return_rc=True    return_stdout=False    return_stderr=False
+    ${rc}=    Check ResolveConf Option Presence
     Should Be Equal As Integers    0    ${rc}
 
     [Teardown]    Run Keywords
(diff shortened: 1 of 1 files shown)


test/suites/osconfig/systemd-resolved.robot
```diff
@@ -170,3 +164,12 @@ Restore Systemd-Resolved
...
     END
+
+Check ResolveConf Option Presence
+    [Documentation]    Check if the 'resolvConf' option is present in the kubelet
+    ...    configuration file. Return a none-zero code if not present.
+
+    ${rc}=    Execute Command
+    ...    grep -qE "^resolvConf:.*${RESOLVE_CONF_FILE}" ${KUBELET_CONFIG_FILE}
+    ...    sudo=True    return_rc=True    return_stdout=False    return_stderr=False
+    RETURN    ${rc}
```
pmtk, line 175: Can we include here
```
Should Not Be Equal As Integers 0 ${rc}
```

All usages of these keyword do that.

+ rename to "ResolvConf Should Be Present" or smth like this

ggiguash: I think this one does not, right?
https://github.com/openshift/microshift/pull/2667/files#diff-ec352c7ad6712e0a73a467e27ee66dc0b1ae368db3453136387b8da7a5406266R33

pmtk: Ah, obviously... We could then have an argument to expect existence or lack of existence, but that's your choice then

ggiguash:
Need this fix for nightlies to work
/override microshift-metal-tests

PR #2669, opened 2023-11-29
USHIFT-1860: introduce certificate rotation tests

eslutsky:
this test will change the system date then expects Microshift service to be restarted and cert rotation to occur.

### DIFF
--- a/test/suites/standard/validate-certificate-rotation.robot
+++ b/test/suites/standard/validate-certificate-rotation.robot
@@ -1,0 +1,73 @@
+*** Settings ***
+Documentation       Tests related to MicroShift automated certificate rotation.
+
+Resource            ../../resources/common.resource
+Resource            ../../resources/microshift-process.resource
+Resource            ../../resources/microshift-host.resource
+Library             DateTime
+Library             Collections
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+
+*** Variables ***
+${KUBE_SCHEDULER_CLIENT_CERT}       /var/lib/microshift/certs/kube-control-plane-signer/kube-scheduler/client.crt
+${OSSL_CMD}                         openssl x509 -noout -dates -in
+${OSSL_DATE_FORMAT}                 %b %d %Y
+${TIMEDATECTL_DATE_FORMAT}          %Y-%m-%d %H:%M:%S
+${FUTURE_DAYS}                      150
+
+
+*** Test Cases ***
+Certificate Rotation
+    [Documentation]    Performs Certificate Expiration Rotation test
+    ${first_cert_date}=    Compute Date After Days    365    ${OSSL_DATE_FORMAT}
+    Certs Should Expire On    ${KUBE_SCHEDULER_CLIENT_CERT}    ${first_cert_date}
+    ${cert_should_expire_in_days}=    Evaluate    365+${FUTURE_DAYS}
+    ${cert_expiry_date}=    Compute Date After Days    ${cert_should_expire_in_days}    ${OSSL_DATE_FORMAT}
+    ${future_date}=    Compute Date After Days    ${FUTURE_DAYS}    ${TIMEDATECTL_DATE_FORMAT}
+    Change System Date To    ${future_date}
+    Certs Should Expire On    ${KUBE_SCHEDULER_CLIENT_CERT}    ${cert_expiry_date}
+
+
+*** Keywords ***
+Setup
+    [Documentation]    Test suite setup
+    Check Required Env Variables
+    Login MicroShift Host
+    Setup Kubeconfig    # for readiness checks
+
+Teardown
+    [Documentation]    Test suite teardown
+    Restore System Date
+    Logout MicroShift Host
+
+Restore System Date
+    [Documentation]    Reset Microshift date to current date
+    ${ushift_pid}=    MicroShift Process ID
+    Systemctl    start    chronyd
+    Wait Until MicroShift Process ID Changes    ${ushift_pid}
+    Wait For MicroShift
+
+Change System Date To
+    [Documentation]    Move the system to a future date.
+    [Arguments]    ${future_date}
+    ${ushift_pid}=    MicroShift Process ID
+    Systemctl    stop    chronyd
+    Command Should Work    timedatectl set-time "${future_date}"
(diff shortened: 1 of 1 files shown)


test/resources/microshift-process.resource
```diff
@@ -76,6 +76,12 @@ Wait For MicroShift Service
     Wait Until Keyword Succeeds    50x    10s
     ...    MicroShift Service Is Active
 
+Wait For MicroShift Service Restart
```
dhellmann, line 79: This may be too-specific to the test you're writing to be reusable. It assumes something else is causing MicroShift to restart, right?

test/suites/standard/validate-certificate-rotation.robot
```diff
@@ -0,0 +1,73 @@
...
+${USHIFT_USER}                      ${EMPTY}
+${KUBE_SCHEDULER_CLIENT_CERT}       /var/lib/microshift/certs/kube-control-plane-signer/kube-scheduler/client.crt
+${OSSL_CMD}                         openssl x509 -noout -dates -in
+
+
+*** Test Cases ***
+Certificate Rotation
+    [Documentation]    Performs Certificate Expiration Rotation test
+    Check Cert Expiration Date    ${KUBE_SCHEDULER_CLIENT_CERT}    365
+    Go To Future Date    150
```
dhellmann, line 23: Does this cause the certs to be regenerated? The way the test is written doesn't make me think it does, but maybe I'm missing something?

test/resources/microshift-process.resource
eslutsky: microshift service  is restarted after time change [detected](https://github.com/openshift/microshift/blob/main/pkg/sysconfwatch/sysconfwatch_linux.go#L139),
i was assuming we will need this keyword for diferent cases (IP change?)

test/suites/standard/validate-certificate-rotation.robot
```diff
@@ -0,0 +1,73 @@
...
+    Check Cert Expiration Date    ${KUBE_SCHEDULER_CLIENT_CERT}    365
+    Reset To Current Date
```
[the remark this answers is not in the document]
eslutsky: after the clock is restored to the current date,certificate will be rotated again, i guess we need to test this also.

eslutsky:
@dhellmann , i've refactored the keywords names , let me know if its become more clear ..

test/resources/microshift-process.resource
dhellmann: OK, reusing it for IP change tests makes sense.

test/suites/standard/validate-certificate-rotation.robot
dhellmann: OK, when I read the test, I interpreted it as "certs should live 365 days; move to the future; certs should live 365 days". It would be clearer if the argument to "Check Cert Expiration Date" was a date, instead of a duration.

test/suites/standard/validate-certificate-rotation.robot
```diff
@@ -0,0 +1,75 @@
...
+${USHIFT_HOST}                      ${EMPTY}
+${USHIFT_USER}                      ${EMPTY}
+${KUBE_SCHEDULER_CLIENT_CERT}       /var/lib/microshift/certs/kube-control-plane-signer/kube-scheduler/client.crt
+${OSSL_CMD}                         openssl x509 -noout -dates -in
+
+
+*** Test Cases ***
+Certificate Rotation
+    [Documentation]    Performs Certificate Expiration Rotation test
+    Cert Expiration Date Should Match    ${KUBE_SCHEDULER_CLIENT_CERT}    365
```
dhellmann, line 22: This would be clearer if the "365" was a date instead of a duration. Let the test compute the expected values, then check, change the clock, check again (with a new expected value).

eslutsky: how does this sounds:
1. `Cert Expiration Date Should Match` rename  "Certifcate Will Valid for One Year" , and remove the passed duration paramter
1.A  calculate  the date a year from now 
1.B - check if the passed  cert file matches  the outcome of 1.A.

2. change the clock - 
2.A - expect microshift reboot and cert regenerate
2.B - repeat 1.

test/suites/standard/validate-certificate-rotation.robot
```diff
@@ -0,0 +1,79 @@
...
+    ...    sudo=True    return_stdout=True    return_stderr=True    return_rc=True
+    Should Be Equal As Integers    ${rc}    0
+    Wait Until MicroShift Process ID Changes    ${ushift_pid}
+    Wait For MicroShift
+
+Compute Date After Days
+    [Documentation]    return system date after number of days elapsed
+    [Arguments]    ${number_of_days}    ${date_format}
+    ${future_date}    ${stderr}    ${rc}=    Execute Command
+    ...    date "+${date_format}" -d "$(date) + ${number_of_days} day"
```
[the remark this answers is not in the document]
eslutsky: done

ggiguash:
@eslutsky , should we allow for 1d diff instead of the exact comparison? The test failed
```
{May 12 2025 != May 13 2025 AssertionError }
```

/lgtm cancel

pmtk:
> @eslutsky , should we allow for 1d diff instead of the exact comparison? The test failed https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/2669/pull-ci-openshift-microshift-main-microshift-metal-tests-arm/1735408190490677248
> 
> ```
> {May 12 2025 != May 13 2025 AssertionError }
> ```

I wonder what's this about... is this leap year, of by one error, or some architecture difference.

eslutsky:
> > @eslutsky , should we allow for 1d diff instead of the exact comparison? The test failed https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/2669/pull-ci-openshift-microshift-main-microshift-metal-tests-arm/1735408190490677248
> > ```
> > {May 12 2025 != May 13 2025 AssertionError }
> > ```
> 
> I wonder what's this about... is this leap year, of by one error, or some architecture difference.

the issue here is that we are using python to calculate the current date (which by default using UTC ), 
and the certs are rotated using the local machine timezone.

PR #2670, opened 2023-11-29
USHIFT-1926: Implement certificate cleanup option in microshift-cleanup-data script

ggiguash:
The new script usage is as follows.
```
Stop all MicroShift services, also cleaning their data

Usage: microshift-cleanup-data.sh <--all [--keep-images] | --ovn | --cert>
   --all         Clean all MicroShift and OVN data
   --keep-images Keep container images when cleaning all data
   --ovn         Clean OVN data only
   --cert        Clean certificates only
```

Note: the changes in `systemd.resource` file come to make the logs more verbose and explicit, which assists with troubleshooting. The previous implementation did not show the returned values.

### DIFF
--- a/scripts/microshift-cleanup-data.sh
+++ b/scripts/microshift-cleanup-data.sh
@@ -9,6 +9,7 @@ PODS_NS_LIST=(openshift-operator-lifecycle-manager openshift-service-ca openshif
 FULL_CLEAN=false
 KEEP_IMAGE=false
 OVN_CLEAN=false
+CERT_CLEAN=false
 # Flags used for generating a user summary message
 SERVICE_STOPPED=false
 SERVICE_DISABLED=false
@@ -20,10 +21,11 @@ function usage() {
         echo "Stop all MicroShift services, also cleaning their data"
     fi
     echo ""
-    echo "Usage: ${SCRIPT_NAME} <--all [--keep-images] | --ovn>"
+    echo "Usage: ${SCRIPT_NAME} <--all [--keep-images] | --ovn | --cert>"
     echo "   --all         Clean all MicroShift and OVN data"
     echo "   --keep-images Keep container images when cleaning all data"
     echo "   --ovn         Clean OVN data only"
+    echo "   --cert        Clean certificates only"
     exit 1
 }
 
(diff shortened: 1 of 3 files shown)


scripts/microshift-cleanup-data.sh
```diff
@@ -145,9 +157,11 @@ while [ $# -gt 0 ] ; do
 done
 
 # Verify valid option combination
-! ${FULL_CLEAN} && ! ${OVN_CLEAN}  && usage "Either --all or --ovn option must be specified"
-! ${FULL_CLEAN} &&   ${KEEP_IMAGE} && usage "The --keep-images option can only be used with --all"
-  ${FULL_CLEAN} &&   ${OVN_CLEAN}  && usage "The --all and --ovn options are mutually exclusive"
+! ${FULL_CLEAN} && ! ${OVN_CLEAN}  && ! ${CERT_CLEAN} && usage "Either --all, --ovn or --cert option must be specified"
```
dhellmann, line 160: Some of these combinations seem redundant, but not "wrong". For example, combining `--all` with anything else or `--keep-images` and `--cert`. Can we be more permissive and allow those and only fail when the combinations are not valid?

ggiguash: Maybe, there is a way to simplify the code by making positional assumptions about the arguments and checking their amount. Is this where the question is coming from? Should we try simplifying the code?

dhellmann: It's more a usability question. If I say "delete everything and delete the certs" that doesn't feel like it should be an error. Think about someone trying to write an ansible role to run this command, and automatically build up the argument list based on variables set in the playbook. The template for that command line could be simple if they check individual flags that map to the command line options. If when the "delete all" flag is set they must not pass any other arguments then the template becomes more complicated.

For our testing, if we know that some of the options are redundant and supersede each other, I don't think we need additional test scenarios.

ggiguash: What you're suggesting is a valid approach, but it's a change in the current behavior. I did not change the usage, which states that the options are mutually exclusive

```
Usage: microshift-cleanup-data.sh <--all [--keep-images] | --ovn | --cert>
```

If we want to change this behavior, I prefer to do it in a separate task. Should I create one and follow-up on this?

dhellmann: Yes, let's leave that for another PR

dhellmann (commented):
/test microshift-metal-tests

ggiguash:
Wait until the https://github.com/openshift/microshift/pull/2695/files is merged. This is necessary to fix frequent failures when restarting microshift and greenboot services.

/hold

ggiguash:
ARM failing due to lack of AWS resources.

ggiguash:
The failure is for the known network test.
/unhold

PR #2677, opened 2023-12-02
USHIFT-1952: Retry after microshift-cleanup-data when greenboot reports a failure in scenario.sh

ggiguash:
The deletion of pods using `kubectl` command gets stuck and returns with timeout.
Trying a stronger technique by using `crictl` commands.


dhellmann (commented):
/lgtm

PR #2678, opened 2023-12-03
USHIFT-1961: make sure FIPS compliance test runs on a stopped system

eslutsky:
due to a race condition  Microshift service  must be stopped before running the test.

### DIFF
--- a/test/suites/fips/validate-fips.robot
+++ b/test/suites/fips/validate-fips.robot
@@ -4,6 +4,7 @@ Documentation       Tests related to FIPS Validation
 Resource            ../../resources/ostree-health.resource
 Resource            ../../resources/common.resource
 Resource            ../../resources/selinux.resource
+Resource            ../../resources/microshift-process.resource
 Library             Collections
 
 Suite Setup         Setup
@@ -11,14 +12,15 @@ Suite Teardown      Teardown
 
 
 *** Variables ***
-${USHIFT_HOST}      ${EMPTY}
-${USHIFT_USER}      ${EMPTY}
+${USHIFT_HOST}              ${EMPTY}
+${USHIFT_USER}              ${EMPTY}
+${USHIFT_LIBS_DUMP_FILE}    /tmp/microshift-libs
+${FIPS_PATTERN}             ossl-modules/fips.so$
 
 
 *** Test Cases ***
 Verify Host Is FIPS Enabled
     [Documentation]    Performs a FIPS validation against the host
-    Wait Until Greenboot Health Check Exited
     Fips Should Be Enabled
 
 Verify Binary Is FIPS Compliant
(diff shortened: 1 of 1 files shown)


test/suites/fips/validate-fips.robot
```diff
@@ -38,8 +41,15 @@ Teardown
 
 Microshift Binary Should Dynamically Link FIPS Ossl Module
     [Documentation]    Check if Microshift binary is FIPS compliant.
+
+    ${stdout_file}=/tmp/microshift-fips.run.log
```
ggiguash, line 45: Can we generate a random file name?
```
    ${rand}=    Generate Random String
    ${local_tmp}=    Join Path    /tmp    ${rand}
```

test/suites/fips/validate-fips.robot
```diff
@@ -38,8 +41,15 @@ Teardown
...
+    ${stdout}    ${rc}=    Execute Command
+    ...    LD_DEBUG=symbols microshift run 2>${stdout_file}
```
ggiguash, line 47: The file name seems to be "wrong". Did you mean `strerr_file`?

test/suites/fips/validate-fips.robot
```diff
@@ -38,8 +41,15 @@ Teardown
...
+    ...    sudo=False    return_rc=True
+    Should Be Equal As Integers    1    ${rc}
+
     ${stdout}    ${rc}=    Execute Command
-    ...    LD_DEBUG=symbols microshift run 2>&1 | grep ossl-modules/fips.so$
+    ...    cat ${stdout_file} && cat ${stdout_file} | grep oassl-modules/fips.so$
```
ggiguash, line 52: I'm not sure we should dump all the output into stdout. It's longer than 32K limit of the RF buffer. 
Let's just run `grep  oassl-modules/fips.so$ ${stdout_file}`?

eslutsky: i'd like to see also  the content of  LD_DEBUG_OUTPUT ,can we store LD_DEBUG_OUTPUT file  as job artifacts? (using RF keyword)
this needed to detect cases when the binary is build incorrectly.

test/suites/fips/validate-fips.robot
eslutsky: Actually  it contains both stdout and stderr.

test/suites/fips/validate-fips.robot
eslutsky: in theory we can, but if we are going to store this log file as CI artifact  better to have constant name.

eslutsky: not according to the [man](https://man7.org/linux/man-pages/man8/ld.so.8.html):
```
       LD_DEBUG_OUTPUT (since glibc 2.1)
              By default, LD_DEBUG output is written to standard error.
              If LD_DEBUG_OUTPUT is defined, then output is written to
              the pathname specified by its value, with the suffix "."
              (dot) followed by the process ID appended to the pathname.

              LD_DEBUG_OUTPUT is ignored in secure-execution mode.
```

test/suites/fips/validate-fips.robot
```diff
@@ -31,17 +31,29 @@ Setup
...
+    Wait For MicroShift
     Logout MicroShift Host
 
 Microshift Binary Should Dynamically Link FIPS Ossl Module
     [Documentation]    Check if Microshift binary is FIPS compliant.
+    ${output_libs_file}=    Set Variable    /tmp/microshift-libs
+    ${look_for_libs}=    Set Variable    ossl-modules/fips.so$
+    ${stdout}    ${rc}=    Execute Command
+    ...    LD_DEBUG_OUTPUT=${output_libs_file} LD_DEBUG=libs microshift run
+    ...    sudo=False    return_rc=True
```
ggiguash, line 50:
```suggestion
    ...    sudo=False    return_rc=True    return_stdout=True
```

test/suites/fips/validate-fips.robot
```diff
@@ -31,17 +31,29 @@ Setup
     [Documentation]    Test suite setup
     Check Required Env Variables
     Login MicroShift Host
+    Setup Kubeconfig
+    Wait Until Greenboot Health Check Exited
+    Stop MicroShift
 
 Teardown
     [Documentation]    Test suite teardown
+    Start MicroShift
```
ggiguash, line 40: This is a bit assymetric because we're stopping MicroShift in a test, while starting it in the suite Teardown.
Do you think we should shut down the service in Setup?

test/suites/fips/validate-fips.robot
```diff
@@ -31,17 +31,29 @@ Setup
...
+    ${stdout}    ${rc}=    Execute Command
+    ...    LD_DEBUG_OUTPUT=${output_libs_file} LD_DEBUG=libs microshift run
+    ...    sudo=False    return_rc=True
+    Should Be Equal As Integers    1    ${rc}
     ${stdout}    ${rc}=    Execute Command
-    ...    LD_DEBUG=symbols microshift run 2>&1 | grep ossl-modules/fips.so$
+    ...    grep ${look_for_libs} ${output_libs_file}*
     ...    sudo=False    return_rc=True
     Should Be Equal As Integers    0    ${rc}
+    SSHLibrary.Get File    ${output_libs_file}*    ${OUTPUTDIR}/
```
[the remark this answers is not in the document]
ggiguash: Also, let's make sure the command does not fail when / if the files do not exist.

PR #2687, opened 2023-12-05
USHIFT-1961: add stderr to FIPS  test

eslutsky:


ggiguash:

The failure is not related to the current fix.

PR #2688, opened 2023-12-05
USHIFT-1814: Run cmd to stop etcd in its own process group

pmtk:
Goal is to avoid getting terminated before stopping etcd.


pmtk:
/hold

Change doesn't fix (log from non-CI testing):
```
etcd E1205 09:07:13.082335   26246 etcd.go:132] "failed to stop microshift-etcd" err="signal: terminated" output=""
```

PR #2699, opened 2023-12-07
USHIFT-1977: add strictfipsruntime tag to microshift binary build

eslutsky:


Makefile
```diff
@@ -102,7 +102,7 @@ debug:
 	@echo PATCH:"$(PATCH)"
 
 # These tags make sure we can statically link and avoid shared dependencies
```
dhellmann, line 104: This comment is no longer accurate, we should have removed it before when we enabled cgo.

dhellmann (commented):
/lgtm
/hold

One nit, but nothing worth holding up the PR. I'll leave a hold in place in case there are test issues. Feel free to remove it when you're ready to merge.

eslutsky:
it appears that those build flags only works with podman build  (doesnt work with upstream golang builds)

Makefile
```diff
@@ -228,6 +228,15 @@ validate-cluster:
 OS := $(shell go env GOOS)
 ARCH := $(shell go env GOARCH)
 
+
+# check if FIPS supported
```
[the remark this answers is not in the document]
eslutsky: the only place that its officially documented its here https://docs.google.com/presentation/d/1o3IowxHX6BsnxGkIInaQ0lBgnn_K5Ex8jxwCYCeNsqs/edit#slide=id.g2679cb578c3_0_17

Makefile
eslutsky: we should address this in separate PR

dhellmann:

The failing test (`el92-src@storage.Reboot: Rebooting Healthy System Should Keep Functional PVC `) is not related to this change.

PR #2700, opened 2023-12-07
USHIFT-1955: Implement microshift-olm plugin altering greenboot functionality

ggiguash:
* Implement a new greenboot check script for the `microshift-olm` RPM, based on deployment readiness.
* Refactor the existing scripts to share some common functionality

### DIFF
--- a/packaging/greenboot/functions.sh
+++ b/packaging/greenboot/functions.sh
@@ -9,6 +9,7 @@ SCRIPT_PID=$$
 OCCONFIG_OPT="--kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig"
 OCGET_OPT="--no-headers"
 OCGET_CMD="oc get ${OCCONFIG_OPT}"
+OCROLLOUT_CMD="oc rollout ${OCCONFIG_OPT}"
 
 # Note about the output
 # This file runs as part of a systemd unit, greenboot-healthcheck. All of the
@@ -142,6 +143,22 @@ function namespace_images_downloaded() {
     return 0
 }
 
+# Check the deployment rollout status in a given namespace is successful.
+#
+# args: None
+# env: 'CHECK_DEPLOY_NS' environment variable for the namespace to check
+# return: 0 if deployment rollout is successful, or 1 otherwise
+function namespace_deployment_ready() {
+    local -r ns="${CHECK_DEPLOY_NS}"
+
+    if ${OCROLLOUT_CMD} status deployment -n "${ns}" \
+            --watch=false --timeout=1s &>/dev/null ; then
+        return 0
+    fi
+
+    return 1
+}
+
 # Check if a given number of pods in a given namespace are in the 'Ready' status,
 # terminating the script with the SIGTERM signal if more pods are ready than expected.
 #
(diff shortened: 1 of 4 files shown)


dhellmann:
/hold

Instead of adding another layer of plugins, we should make the greenboot health check for OLM its own executable that is configured to run after MicroShift's script runs.

packaging/greenboot/functions.sh
```diff
@@ -142,6 +142,38 @@ function namespace_images_downloaded() {
...
+    if [ "${lcount}" -ne "${ct}" ] ; then
+        return 1
+    fi
+
+    # Count the number of available replicas that is not equivalent
+    # to the number of ready replicas
+    local -r rcount=$(echo -e "${output}" | awk '$2 != $3 {print $1}' | grep -vc '^$')
+    if [ "${rcount}" -ne 0 ] ; then
+        return 1
+    fi
```
pmtk, line 172: Instead of manually comparing ready and expected replicas we could use on of these:
```
$ oc rollout status deployment -n openshift-operator-lifecycle-manager        
Waiting for deployment "catalog-operator" rollout to finish: 0 of 1 updated replicas are available...
deployment "catalog-operator" successfully rolled out
deployment "olm-operator" successfully rolled out
```

```
$ oc wait deployment -n openshift-operator-lifecycle-manager catalog-operator --for=condition=Available --timeout=300s
deployment.apps/catalog-operator condition met
```

packaging/greenboot/functions.sh
```diff
@@ -142,6 +142,38 @@ function namespace_images_downloaded() {
...
 
+# Check if all the deployments in a given namespace are ready.
+#
+# args: None
+# env1: 'CHECK_DEPLOY_NS' environment variable for the namespace to check
+# env2: 'CHECK_DEPLOY_CT' environment variable for the deployment count to check
+# return: 0 if deployment is ready, or 1 otherwise
+function namespace_deployment_ready() {
+    local -r ns="${CHECK_DEPLOY_NS}"
+    local -r ct="${CHECK_DEPLOY_CT}"
```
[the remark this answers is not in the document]
ggiguash: It's the pattern in the functions.sh because those functions are called in `wait_for`, which does not support other ways of argument passing.

packaging/greenboot/functions.sh
ggiguash: Fixed.

packaging/greenboot/functions.sh
```diff
@@ -221,3 +237,43 @@ function print_failure_logs() {
...
+# variable is set, also log the list of pods and their events on failure.
+#
+# args: None
+# env: 'LOG_POD_EVENTS' environment variable to enable pod-specific logging
+# return: None
+function log_script_exit() {
+    if [ "$?" -ne 0 ] ; then
+        if ${LOG_POD_EVENTS}; then
+            log_failure_cmd "pod-list" "${OCGET_CMD} pods -A -o wide"
+            log_failure_cmd "pod-events" "${OCGET_CMD} events -A"
```
pmtk, line 272: This is outside of scope of this PR, but what do you think about sorting events by time? Currently it's sorted by namespace and it's not helping with quickly seeing what happened last.
```--sort-by='.metadata.creationTimestamp'`

File: packaging/greenboot/microshift-running-check-olm.sh [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+# Set the wait timeout for the current check based on the boot counter
+WAIT_TIMEOUT_SECS=$(get_wait_timeout)
+
+# Starting pod-specific checks
+# Log list of pods and their events on failure
+LOG_POD_EVENTS=true
+
+# Wait for the deployments to be ready
+echo "Waiting ${WAIT_TIMEOUT_SECS}s for '${CHECK_DEPLOY_NS}' deployments to be ready"
+if ! wait_for "${WAIT_TIMEOUT_SECS}" namespace_deployment_ready ; then
```
pmtk, line 63: I don't think this is right usage of `kubectl rollout status`. The command WAITS for the deployment, there's no need for additional layer of `wait_for`.
More so, it doesn't even work as expected right now - because it blocks, loop in `wait_for` will only happen once. It's also possible that without `--timeout` it will block indefinitely and that `wait_for` might never return

pmtk: Unless we add `--watch=false`, then we'll have behavior which we expect in other functions

```
    -w, --watch=true:
	Watch the status of the rollout until it's done.
```

ggiguash:
The test is failing on the "known" network problem

PR #2704, opened 2023-12-08
USHIFT-1971: More MicroShift start up hardening 

pmtk:


pkg/controllers/etcd.go
```diff
@@ -104,27 +107,44 @@ func (s *EtcdService) Run(ctx context.Context, ready chan<- struct{}, stopped ch
...
 	defer func() {
-		klog.Info("stopping microshift-etcd")
-		cmd := exec.Command("systemctl", "stop", "microshift-etcd.scope", "--no-block")
+		stopWatchingForUnexpectedShutdown <- struct{}{}
 
-		if out, err := cmd.CombinedOutput(); err != nil {
-			klog.ErrorS(err, "failed to stop microshift-etcd", "output", string(out))
-			return
+		klog.Info("Waiting for kube-apiserver shutdown before terminating microshift-etcd")
+		<-s.kasShutdownSignal
```
pmtk, line 135: Okay, this fixed one thing and introduced another problem...
Scenario:
```
- etcd starts
- shutdown request comes
- etcd waits for kube-apiserver to shutdown, but KAS didn't even attempted to start, so this blocks until timeout (15s)
- MicroShift just exits, etcd lives on preventing subsequent restarts from starting healthy
```

PR #2726, opened 2023-12-13
USHIFT-2046: Source common.sh after deps are installed

pmtk:


pmtk:
/override ci/prow/metal-periodic-test  ci/prow/microshift-metal-tests-arm

PR #2737, opened 2023-12-14
USHIFT-2051: always build images not available in the cache

dhellmann:
Also adds a -f flag to force rebuilding locally, and introduces force
logic for source image rebuilds and when building a specific image from
its template.

/assign @ggiguash @pmtk

### DIFF
--- a/test/bin/ci_phase_iso_build.sh
+++ b/test/bin/ci_phase_iso_build.sh
@@ -76,16 +76,9 @@ update_build_cache() {
 # - Always build the 'presubmit' layer.
 # - Only build the 'periodic' layer when 'CI_JOB_NAME' contains 'periodic' token.
 run_image_build() {
-    local -r with_cached_data=$1
-
     if [ -v CI_JOB_NAME ] ; then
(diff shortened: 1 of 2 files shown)


test/bin/ci_phase_iso_build.sh
```diff
@@ -76,16 +76,9 @@ update_build_cache() {
...
-
     if [ -v CI_JOB_NAME ] ; then
         # Conditional per-layer builds when running in CI
-        if ${with_cached_data} ; then
-            $(dry_run) bash -x ./bin/build_images.sh -d -l ./image-blueprints/layer1-base
-        else
-            $(dry_run) bash -x ./bin/build_images.sh    -l ./image-blueprints/layer1-base
-        fi
-
+        $(dry_run) bash -x ./bin/build_images.sh -l ./image-blueprints/layer1-base
```
ggiguash, line 81: Can we add a comment here explaining how / when the cache is used?

dhellmann: done

PR #2739, opened 2023-12-14
USHIFT-2053: only install osbuild from COPR on RHEL 9.3

dhellmann:
The version on RHEL 9.2 works OK for building only 9.2 images, so
allow it to be used.

/assign @ggiguash @pmtk

### DIFF
--- a/scripts/image-builder/configure.sh
+++ b/scripts/image-builder/configure.sh
@@ -13,20 +13,31 @@ auth_file_path = "/etc/osbuild-worker/pull-secret.json"
 EOF
 fi
 
-# osbuild from COPR to install version that:
-# 1. can build rhel-93 images
-# 2. doesn't have issues with unexpected mirror's RPM verification
-#    (like the one in 9.3 beta at the moment of writing this comment)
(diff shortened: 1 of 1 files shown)


scripts/image-builder/configure.sh
```diff
@@ -13,20 +13,31 @@ auth_file_path = "/etc/osbuild-worker/pull-secret.json"
...
+    sudo dnf copr enable -y @osbuild/osbuild-composer rhel-9-"$(uname -m)"
+    composer_packages="
+    osbuild-composer-96-1.20231213092324044071.main.10.g991293a89.el9
+    osbuild-101-1.20231213161521815551.main.24.g05c0fd31.el9
+    "
+else
+    # osbuild from the standard RPM repository, limits about which os
+    # versions can be built apply
+    composer_packages="osbuild-composer osbuild"
+fi
```
ggiguash, line 32: Why is this necessary? We're supposed to be using RHEL 9.3 for devenv now.

pmtk: Yes, we should makes sure that all 4.14, 4.15, and main branches use rhel93 as hypervisor

PR #2745, opened 2023-12-15
USHIFT-2063: Display manage_build_cache.sh output for getlast command during CI builds

ggiguash:
This is necessary to see the errors occasionally appearing when running the cache update procedure in CI


pacevedom (approved):
/lgtm

PR #2746, opened 2023-12-15
USHIFT-1968: SELinux syntax cleanup

eggfoobar:
Incorporates the fixes from @zpytela with updates to the unit tests.

/assign @ggiguash


eggfoobar:
/override microshift-metal-tests-arm

Overriding due to known failure on arm platform

test/resources/selinux.py
```diff
@@ -58,7 +58,7 @@
     "/var/lib/lxd(/.*)?",
     "/var/lib/microshift(/.*)?",
     "/var/lib/microshift-backups(/.*)?",
-    "/var/lib/microshift.saved(/.*)?",
+    r"/var/lib/microshift\.saved(/.*)?",
```
zpytela, line 61: There seems to be an unintentional leading "r"

eggfoobar: Good eye, that's actually intentional in this instance, because we have an escape character in the `\.saved` we want python to treat it as a raw string so it doesn't interpret the escape character, you do that by prepending the `r` in front of a string to denote that.

ggiguash:
The failure has nothing to do with the current change. Saving CI cycles.

PR #2756, opened 2023-12-18
USHIFT-1961: remove golang cache when building from ART container

eslutsky:
since we are running concurrent  builds with different golang compilers and shared cache,  it may produce  FIPS incompateble RPMS,   which may lead  to FIPS test failures.


eslutsky:
this failure not related to this PR

PR #2760, opened 2023-12-18
USHIFT-2078: Dynamically use latest available RHOCP repo

pmtk:


scripts/get-latest-rhocp-repo.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+cacert="/etc/rhsm/ca/redhat-uep.pem"
+cert=$(find /etc/pki/entitlement -iname '*.pem' -not -iname '*-key.pem')
+key=$(find /etc/pki/entitlement -iname '*-key.pem')
+
+current_minor=$(make -C "${REPOROOT}" debug | grep "MINOR:" | cut -d':' -f2)
+stop=$(( current_minor - 3 ))
+
+for ver in $(seq "${current_minor}" -1 "${stop}"); do
```
ggiguash, line 30: Can we add some comments here explaining what's going on?

scripts/get-latest-rhocp-repo.sh
```diff
@@ -0,0 +1,49 @@
...
+if ! sudo subscription-manager status >&/dev/null; then
+    >&2 echo "System must be subscribed"
+    exit 1
+fi
+
+cacert="/etc/rhsm/ca/redhat-uep.pem"
+cert=$(find /etc/pki/entitlement -iname '*.pem' -not -iname '*-key.pem')
+key=$(find /etc/pki/entitlement -iname '*-key.pem')
+
+current_minor=$(make -C "${REPOROOT}" debug | grep "MINOR:" | cut -d':' -f2)
```
ggiguash, line 27: Can we comment / explain?

scripts/get-latest-rhocp-repo.sh
```diff
@@ -0,0 +1,49 @@
...
+        --location \
+        --output /dev/null \
+        --write-out "%{http_code}" \
+        --cacert "${cacert}" \
+        --cert "${cert}" \
+        --key "${key}" \
+        "${repository}/os/repodata/repomd.xml")
+
+    if [[ "${exit_code}" == "200" ]]; then 
+        echo "${ver}"
```
ggiguash, line 43: I wonder if we need to just output the whole version, i.e. `4.14`? 
It looks like we're appending `4.` everywhere it's used.

pmtk: By printing only the minor, we can easily manipulate it further. For example, on `release-4.14` branch we're enabling RHOCP 4.13 and 4.14. So if we have similar situation in future, we can easily get minor-1 without too much problem.
On `r-4.15` it will happen when `rhocp-4.15` will be published and we'll switch to it to test upgrade from released RPMs, but without `rhocp-4.14` will only use RCs and they will be long outdated by then.

scripts/get-latest-rhocp-repo.sh
pmtk: Umm, it's this: https://github.com/openshift/microshift/blob/main/Makefile#L101

pmtk: added comment

scripts/get-latest-rhocp-repo.sh
pmtk: added comment

scripts/get-latest-rhocp-repo.sh
pmtk: This happens because https://github.com/openshift/microshift/blob/main/test/bin/get_rel_version_repo.sh, for released RPMs, only checks system repositories - if rhocp-4.14 is not enabled, it will take version from openshift mirror

ggiguash:
The failure is not related to the current changes.

PR #2764, opened 2023-12-18
USHIFT-2085: RPM produced when building base branch should not use the default name

ggiguash:


test/Makefile
```diff
@@ -31,7 +31,8 @@ build-base-branch: TMP_REPO !=mktemp -d
 build-base-branch:
 	if ! [ -d "$(REPO)/_output" ]; then mkdir $(REPO)/_output; fi
 	git clone --single-branch --branch $(BASE_BRANCH) https://github.com/openshift/microshift.git $(TMP_REPO)
-	cd $(TMP_REPO) && $(MAKE) rpm
+	MICROSHIFT_VERSION="$(MAJOR).$(MINOR).$(PATCH)_base_branch" \
+		$(MAKE) -C $(TMP_REPO) rpm
```
pmtk, line 35: Can you provide example values? How about using $(BASE_BRANCH)?

ggiguash:
The failure is not related to the chage

PR #2766, opened 2023-12-18
USHIFT-2067: Use registry mirror in all architectures

pacevedom:
- Enable registry mirror in x86 and aarch64.
- Use docker distribution instead of quay.

Closes #<Issue Number>

### DIFF
--- a/test/bin/mirror_registry.sh
+++ b/test/bin/mirror_registry.sh
@@ -5,8 +5,8 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=test/bin/common.sh
 source "${SCRIPTDIR}/common.sh"
 
-QUAY_VERSION=1.3.9
-REGISTRY_HOST=${REGISTRY_HOST:-$(hostname):8443}
+DISTRIBUTION_VERSION=2.8.3
+REGISTRY_HOST=${REGISTRY_HOST:-$(hostname):5000}
 REGISTRY_ROOT=${REGISTRY_ROOT:-${HOME}/mirror-registry}
 REGISTRY_CONTAINER_DIR=${REGISTRY_CONTAINER_DIR:-${REGISTRY_ROOT}/containers}
 REGISTRY_CONTAINER_LIST=${REGISTRY_CONTAINER_LIST:-${REGISTRY_ROOT}/mirror-list.txt}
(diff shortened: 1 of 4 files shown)


test/bin/mirror_registry.sh
```diff
@@ -29,23 +29,19 @@ get_container_images() {
 prereqs() {
     mkdir -p "${REGISTRY_ROOT}"
     mkdir -p "${REGISTRY_CONTAINER_DIR}"
-    curl -L https://github.com/quay/mirror-registry/releases/download/v${QUAY_VERSION}/mirror-registry-offline.tar.gz -o /tmp/mirror-registry-offline.tar.gz
-    tar xf /tmp/mirror-registry-offline.tar.gz -C "${REGISTRY_ROOT}"
-    rm -f /tmp/mirror-registry-offline.tar.gz
     sudo dnf install -y podman skopeo jq
+    podman run -d -p 5000:5000 --restart always --name registry "docker.io/distribution/distribution:${DISTRIBUTION_VERSION}"
```
pmtk, line 33: We might want to mirror it to our quay, docker.io has pull limits

pacevedom: Done

pacevedom:
metal-periodic-test-arm/test

pacevedom:
/hold
Need to solve the multiarch image first.

pacevedom:
Failed because of a problem in the registry when test was running. Metal tests pulled the images correctly.
Failed because of a greenboot issue in one of the tests. Unrelated to this change.

test/bin/mirror_registry.sh
```diff
@@ -29,23 +29,19 @@ get_container_images() {
 prereqs() {
     mkdir -p "${REGISTRY_ROOT}"
     mkdir -p "${REGISTRY_CONTAINER_DIR}"
-    curl -L https://github.com/quay/mirror-registry/releases/download/v${QUAY_VERSION}/mirror-registry-offline.tar.gz -o /tmp/mirror-registry-offline.tar.gz
-    tar xf /tmp/mirror-registry-offline.tar.gz -C "${REGISTRY_ROOT}"
-    rm -f /tmp/mirror-registry-offline.tar.gz
     sudo dnf install -y podman skopeo jq
+    podman run -d -p 5000:5000 --restart always --name registry "quay.io/microshift/distribution:${DISTRIBUTION_VERSION}"
```
ggiguash, line 33: This command fails if run for the second time. Can we add something like the following to allow running the script more than once without manual intervention?
```suggestion
    podman stop microshift-local-registry || true
    podman run -d -p 5000:5000 --restart always --name microshift-local-registry "quay.io/microshift/distribution:${DISTRIBUTION_VERSION}"
```

test/bin/mirror_registry.sh
```diff
@@ -29,23 +29,19 @@ get_container_images() {
...
-    jq -s '.[0] * .[1]' "${PULL_SECRET}" "${REGISTRY_ROOT}/local-auth.json" > "${PULL_SECRET}.tmp"
-    mv "${PULL_SECRET}.tmp" "${PULL_SECRET}"
-    popd &>/dev/null
+    # Docker distribution does not support TLS authentication. The mirror-images.sh helper uses skopeo without tls options
+    # and it defaults to https. Since this is not supported we need to configure registries.conf so that skopeo tries http instead.
+    sudo bash -c 'cat >> /etc/containers/registries.conf' << EOF
+[[registry]]
+location = "$(hostname)"
+insecure = true
+EOF
```
ggiguash, line 43: This one also needs to be fixed not to append to the file when run a few times.
May I suggest just creating a file in `/etc/containers/registries.d`,  so it an be overwritten when run a few times.

test/bin/mirror_registry.sh
pacevedom: Added a `rm` too, or else it wont run the second time. Thanks!

pmtk: We might consider mounting path where images are stored on the host, so after kill and run we don't need to recache the images

pacevedom: Keep in mind that this procedure happens without knowledge of the RPMs contents. You are going to download and re-push everything everytime because you dont know if something changed.
The container is named to keep the volumes for the next `start` too.

Thinking about this I would be more inclined to include commands to this script. Something along the lines of: setup, start, stop. Only the setup part would try to download and push containers, while the others would be basic operations.

test/bin/mirror_registry.sh
pacevedom: Done

test/bin/mirror_registry.sh
pmtk: I want to believe that `scopeo copy` is smart enough to not do unnecessary work when image exists in dest registry (especially we're using digests, not tags). Also, volumes for `start` don't matter if you `rm` the container.

I'm inclined to tackle this in another PR - for now it's intended for CI, if we want to leverage this on local envs, we should do it some other time

pacevedom: > Also, volumes for start don't matter if you rm the container.

In case you want to start the registry because for some reason it was stopped, you wouldnt use this script and would need to do it manually. `mirror_registry.sh` should only be used when needing to mirror all containers. If that was already done, then you should handle it with `podman start/stop`. Thats why I am inclined to include commands.

All this should be included in a separate PR.

ggiguash:
/override ci/prow/metal-periodic-test  ci/prow/microshift-metal-tests-arm
The failures have nothing to do with the current changes

ggiguash:
/override ci/prow/microshift-metal-tests ci/prow/metal-periodic-test-arm

PR #2770, opened 2023-12-19
OCPBUGS-24444: Check topolvm only when it should be deployed

pmtk:


packaging/greenboot/microshift-running-check.sh
```diff
@@ -25,6 +25,36 @@ function forced_termination() {
...
+    if [ -f /etc/microshift/lvmd.yaml ]; then
+        return 0
+    fi
+
+    local -r volume_groups=$(vgs --readonly --options=name --noheadings)
+    if [ -z "${volume_groups}" ]; then
+        return 1
+    fi
+
+    local -r volume_groups_count=$(echo "${volume_groups}" | wc -l)
```
ggiguash, line 48: I think we should count words to avoid formatting problems.
```suggestion
    local -r volume_groups_count=$(echo "${volume_groups}" | wc -w)
```

packaging/greenboot/microshift-running-check.sh
```diff
@@ -25,6 +25,36 @@ function forced_termination() {
...
+# Adapted from MicroShift code.
+#
+# args: None
+# return: 0 if topolvm readiness should be checked, 1 otherwise
+function topolvmShouldBeDeployed() {
+    if ! hash vgs 2>/dev/null; then
+        return 1
+    fi
+
+    # TODO: If lvmd.yaml exists, should we expect topolvm even if the OS doesn't has lvms2 installed?
```
ggiguash, line 38: I'm not sure how to intepret this comment. Can you explain more please?

packaging/greenboot/microshift-running-check.sh
```diff
@@ -25,6 +25,36 @@ function forced_termination() {
...
+
+    local -r volume_groups=$(vgs --readonly --options=name --noheadings)
+    if [ -z "${volume_groups}" ]; then
+        return 1
+    fi
+
+    local -r volume_groups_count=$(echo "${volume_groups}" | wc -l)
+    if [ "${volume_groups_count}" -eq 1 ]; then
+        return 0
+    elif echo "${volume_groups}" | grep -q "microshift"; then
```
ggiguash, line 51: Do we expect the volume to be called `microshift` exactly, or container `microshift` in the name?

packaging/greenboot/microshift-running-check.sh
pmtk: If `lvmd.yaml` exists, but `lvms2` tools are not installed - which is stronger signal that user wants topolvm?
What if user created `lvmd.yaml` but forgot to install lvms2?

packaging/greenboot/microshift-running-check.sh
pmtk: We expect volume group to be named `microshift`
https://github.com/openshift/microshift/blob/682d0d580b6d568915baa67188d918428837ad65/pkg/config/lvmd/lvmd.go#L103
https://github.com/openshift/microshift/blob/682d0d580b6d568915baa67188d918428837ad65/pkg/config/lvmd/lvmd.go#L57

packaging/greenboot/microshift-running-check.sh
pmtk: Yes! `wc -l` was giving me 1 even on empty output, this helps, thx

packaging/greenboot/microshift-running-check.sh
ggiguash: So, in this case, would microshift decide to start topolvm and fail?

packaging/greenboot/microshift-running-check.sh
ggiguash: In this case, let's perform an exact comparison.
```suggestion
    elif echo "${volume_groups}" | grep -qw "microshift"; then
```

pmtk: Done

packaging/greenboot/microshift-running-check.sh
pmtk: from impl perspective it's just like currently implemented here: if `vgs` is absent, existence of `lvmd.yaml` doesn't have any impact

pmtk: and what I'm wondering is: should we flip it - if `lvms.yaml` is present, but `vgs` is not, then should we fail (and where - here or microshift itself)

pmtk: The check stays as it is - question was about flipping this one and first one (`hash vgs`)

packaging/greenboot/microshift-running-check.sh
```diff
@@ -25,6 +25,33 @@ function forced_termination() {
...
+# args: None
+# return: 0 if topolvm readiness should be checked, 1 otherwise
+function topolvmShouldBeDeployed() {
+    if ! hash vgs 2>/dev/null; then
+        return 1
+    fi
+
+    if [ -f /etc/microshift/lvmd.yaml ]; then
+        return 0
+    fi
```
pmtk, line 40: @ggiguash 
This has one caveat: in the code, after loading the file, we also check if `len(DeviceClass) > 0`. Should we skip this level of detail? Or do you have idea how to easily get this info with using standard linux tools without `yq`

ggiguash: I think we can skip this detail. It's too much for this script.

pmtk:
/retest

```
ci/prow/microshift-metal-tests-arm
Error: Timed out waiting for pod image(s) from the 'openshift-storage' namespace to be downloaded
```

PR #2780, opened 2023-12-20
OCPBUGS-24689: The microshift-cleanup-data script should delete pods from all namespaces

ggiguash:


scripts/microshift-cleanup-data.sh
```diff
@@ -57,15 +54,23 @@ function stop_clean_pods() {
...
     echo Removing MicroShift pods
-    for i in "${!PODS_NS_LIST[@]}"; do
-        local ns
-        ns=${PODS_NS_LIST[${i}]}
+    # The 'openshift-ovn-kubernetes' namespace must be in the end of the list
+    # to allow for pod deletion when MicroShift is stopped
+    local namespaces
+    namespaces=$(crictl pods --output json | jq -r '.items[].metadata.namespace' | sort -u)
+    if [ -n "${namespaces}" ] ; then
+        namespaces=$(grep -vw openshift-ovn-kubernetes <<<"${namespaces}" || true)
```
eslutsky, line 64: can we do it in one liner in 62:
namespaces=$(crictl pods --output json | jq -r '.items[].metadata.namespace' | sort -u | grep -v openshift-ovn-kubernetes)

ggiguash: The problem with one liner is that there is no way of knowing if there were any namespaces to start with. 
That way, we would always need to add `openshift-ovn-kubernetes` because it's filtered out.

ggiguash: So, we'll attempt to clear a namespace which does not exist.

PR #2781, opened 2023-12-20
[WIP] OCPBUGS-22216: version subcmd displays lvms version

copejon:
The LVMS version isn't easy to obtain from microshift itself, and usually requires inspecting the container topolvm contain image.  This is complicated by the fact that microshift lumps the lvms images in with core openshift images and discards the lvms version.

The changes here do a couple things:
- UX: changes the output of `microshift version` to append the lvms version. For example:

```
❯ _output/bin/microshift version
MicroShift Version: 4.16.0-0.nightly-2023-12-15-211129-20231220224009-6684e03d9-dirty
Base OCP Version: 4.16.0-0.nightly-2023-12-15-211129
Base LVMS Version: v4.14.1-4
```
> LVMS versions are tracked only by the image tag and not by specific git commit.

- Isolate lvms pull-specs and version data in a separate release file.
- Inject lvms pull-spects into the release-images data struct to keep internal release-image handling as is
- Update the lvms rebase logic to handle version and pull-spect parsing and storing

TODO:
- Remove old lvms rebase logic


dhellmann:
If we make the version command show the LVMS version we have to keep doing that and that will make it hard to decouple the CSI later.

We should be able to get the version from the input to the rebase command, can't we?

Can we save the version to a new config map in the same namespace as the controller? That should meet the requirements without adding another point of tight coupling. Later, from the command line, someone could look at the rpm version.

copejon:
Closing as this has crept out of scope

PR #2793, opened 2023-12-22
USHIFT-2090: MicroShift Cluster ID generation and display

ggiguash:


packaging/rpm/microshift.spec
```diff
@@ -298,6 +299,9 @@ usermod -a -G hugetlbfs openvswitch
 
 %post
 
+# Create the cluster ID file
+uuidgen > %{_datadir}/microshift/cluster-id
```
pmtk, line 303: So this creates ID during installation. Should we consider MicroShift after `microshift-cleanup` to be the same cluster?

ggiguash: Looking at the https://issues.redhat.com/browse/OCPSTRAT-1073 feature, the requirement is to create an ID during the installation. 
For the overwrite, it's a good point, I'll add a code preventing that if the file already exists.

ggiguash: Otherwise, the cluster ID generation goes to /var/lib/microshift and the task is somewhat simplified.

pmtk: It says "during installation/startup". I was expecting this to be created during first microshift start up. I'm not fond of this logic being in the RPM.

pmtk: Also, does `/usr/share/` persists through ostree upgrades/rollbacks? I remember `/etc/` and `/var/lib`. If it doesn't, then this will re-create the file on each upgrade/rollback.
BTW. When creating ostree commit, does %post run when the image is created? I doubt that it works when booted. If that's the case, we cannot have this logic in the RPM

ggiguash: All these are good points. I will incorporate them into the design doc.

ggiguash: The open question about whether to keep the cluster ID after clean can be resolved by the location of the `cluster-id` file. It can either be inside or outside of `/var/lib/microshift`

ggiguash:
Closing this PR after an agreement for an alternative implementation.
/close

PR #2796, opened 2023-12-24
USHIFT-2110: certificate rotation test revert to date command

eslutsky:


ggiguash:
The failure is not related to the current changes. Saving on CI cycles.

PR #2814, opened 2023-12-28
NO-ISSUE: Upgrade osbuild copr version to 97 and 102

ggiguash:


ggiguash:
The failure is not related to the change.

PR #2832, opened 2024-01-02
USHIFT-2142: Split standard suite into 2 parts to parallelize its execution

ggiguash:
This change is about reducing the runtime of **presubmit** jobs to improve the overall CI responsiveness. 
The goal is to reduce the time under 1h, and this change brings us closer, but not quite there yet.

Note: The original "standard" presubmit test takes a bit over 20m. It is currently the longest test and all the others are waiting for it to complete.


ggiguash:
Looking at the following runs, we're getting very close to 1h mark for pre-submit tests. Note that while the test step took ~15m, the longest test ran ~11m.
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/2832/pull-ci-openshift-microshift-main-microshift-metal-tests/1742235011181973504
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_microshift/2832/pull-ci-openshift-microshift-main-microshift-metal-tests-arm/1742235011253276672

PR #2850, opened 2024-01-05
USHIFT-1848: add periodic test scenario for the standard tests on rpm-based system

dhellmann:
Depends on #2568

/hold

### DIFF
--- a/test/resources/selinux.py
+++ b/test/resources/selinux.py
@@ -1,8 +1,9 @@
-from robot.libraries.BuiltIn import BuiltIn
-from libostree import remote_sudo_rc, remote_sudo
-from typing import List
-from packaging.version import Version
 import re
+from typing import List
+
+from robot.libraries.BuiltIn import BuiltIn
+
+from libostree import remote_sudo, remote_sudo_rc
 
 ACCESS_CHECK_MAP = {
     "/var/lib/microshift/version": ["cat"],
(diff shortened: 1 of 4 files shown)


dhellmann:
Rebased to see if that helps with the flakes. The SELinux failure looks like it might be real because I can reproduce it locally.

test/resources/selinux.py
ggiguash, line 1: These changes do not seem to be related to the main task, or are they?

test/scenarios-periodics/el92-src@rpm-standard1.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+
+scenario_create_vms() {
+    prepare_kickstart host1 kickstart-liveimg.ks.template ""
+    launch_vm host1 "rhel-9.2-microshift-source"
+    configure_vm_firewall host1
```
ggiguash, line 8: Why do we need this here? What's different from other scenarios?

test/scenarios-periodics/el92-src@rpm-standard1.sh
```diff
@@ -0,0 +1,18 @@
...
+    run_command_on_vm host1 "sudo systemctl enable --now microshift"
```
ggiguash, line 9: I thought we were enabling microshift service in the source base template. Are we losing this setting somehow?
If so, should we also enable the `microshift-test-agent` service here?

test/resources/selinux.py
dhellmann: The commit message explains. https://github.com/openshift/microshift/pull/2850/commits/d1396b335ff61d4c865abb911b069ad289a51900

I get different results from the fcontext list on RPM and ostree systems. The test was overly rigid, testing for an exact match. The important thing to do is ensure that the rules that are specific to MicroShift are installed properly, so I rewrote the test to do that so it would work on both deployment types and be more resilient to changes in the future.

test/scenarios-periodics/el92-src@rpm-standard1.sh
dhellmann: The RPM-based system doesn't have the firewall configured by the blueprint. This is the same function that's used in the RPM installation test scenario.

test/scenarios-periodics/el92-src@rpm-standard1.sh
dhellmann: The blueprint item to enable the service does not apply to the RPM-based system.

The tests do not use the agent service, so it's not being enabled.

test/scenarios-periodics/el92-src@rpm-standard1.sh
ggiguash: We have other scenarios i.e. el92-src@fips.sh using liveimg, but they do not configure firewall explicitly. 
What is different in the current test?

test/image-blueprints/layer3-periodic/group3/rhel92-microshift-source.image-installer
```diff
@@ -0,0 +1 @@
+rhel-9.2-microshift-source
```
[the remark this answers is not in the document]
dhellmann: Done

PR #2858, opened 2024-01-08
OCPBUGS-21901: increase core dns pod buffersize

eslutsky:
increase the buffer-size to be in-sync with the upsteam fixed configuration [value](https://github.com/openshift/cluster-dns-operator/pull/266) .

### DIFF
--- a/assets/components/openshift-dns/dns/configmap.yaml
+++ b/assets/components/openshift-dns/dns/configmap.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 data:
   Corefile: |
     .:5353 {
-        bufsize 512
+        bufsize 1232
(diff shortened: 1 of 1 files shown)


pmtk:
Worth noting that they were going back and forth with 512 and 1232:
- Revert to 512: https://github.com/openshift/cluster-dns-operator/pull/276
- Bumped to 1232: https://github.com/openshift/cluster-dns-operator/pull/370

[And it's `1232` as of now](https://github.com/openshift/cluster-dns-operator/blob/75c5dc17111dc498efd849f5a905b006681d8ddf/pkg/operator/controller/controller_dns_configmap.go#L67)

/lgtm

PR #2866, opened 2024-01-10
USHIFT-2169: -f option to build_images.sh does not require a parameter

dhellmann:


dhellmann:
The failures are all known issues that occur after the image build phase.

PR #2867, opened 2024-01-10
USHIFT-2105: Recreate core components on microshift restarts

pacevedom:

Closes #<Issue Number>


test/kickstart-templates/kickstart-liveimg.ks.template
```diff
@@ -109,4 +109,8 @@ ln -sf /dev/null /etc/systemd/user/grub-boot-success.timer
 # configure FIPs
 REPLACE_FIPS_COMMAND
 
+echo "fs.inotify.max_user_watches = 65536" >> /etc/sysctl.conf
+echo "fs.inotify.max_user_instances = 8192" >> /etc/sysctl.conf
+sysctl --system
```
ggiguash, line 114: We need a comment here please, explaining why this is necessary

test/kickstart-templates/kickstart-offline.ks.template
```diff
@@ -66,4 +66,8 @@ ln -sf /dev/null /etc/systemd/user/grub-boot-success.timer
 # Disable SELinux enforcement for the qemu-guest-agent
 sudo semanage permissive -a virt_qemu_ga_t
 
+echo "fs.inotify.max_user_watches = 65536" >> /etc/sysctl.conf
+echo "fs.inotify.max_user_instances = 8192" >> /etc/sysctl.conf
+sysctl --system
```
ggiguash, line 71: We need a comment here please, explaining why this is necessary

test/kickstart-templates/kickstart.ks.template
```diff
@@ -159,4 +159,8 @@ done
 # and system needing much more time to roll back.
 ln -sf /dev/null /etc/systemd/user/grub-boot-success.timer
 
+echo "fs.inotify.max_user_watches = 65536" >> /etc/sysctl.conf
+echo "fs.inotify.max_user_instances = 8192" >> /etc/sysctl.conf
+sysctl --system
```
ggiguash, line 164: We need a comment here please, explaining why this is necessary

ggiguash (changes requested):

pacevedom:
Found other pods with similar issues, working on them.

test/resources/microshift-process.resource
```diff
@@ -80,6 +77,16 @@ Restart MicroShift
...
+    # Force delete of core pods to avoid extended downtime after a restart.
+    # Whenever there is apiserver unavailability some of these pods enter a
+    # crash loop because they can not update/get resources they need from the
+    # cluster. This backoff may cap at 5min and trigger greenboot failures.
+    # Manifests must be deleted or they may be lingering pods trying to query
+    # apiserver when its up again, triggering transient errors regarding access
+    # to certain resources. The safest option is to remove them before restart
+    # and start clean.
+    Setup Kubeconfig
+    Delete Core Components
```
ggiguash, line 89: Are we not creating a race condition by doing this before the restart? I mean, part of those pods may be recreated before the actual shutdown takes place, right?

test/resources/microshift-process.resource
```diff
@@ -112,3 +119,12 @@ Cleanup MicroShift
...
     Enable MicroShift
+
+Delete Core Components
+    [Documentation]    Delete core components pods from MicroShift's control
+    ...    plane.
+    Run With Kubeconfig    oc delete deployment topolvm-controller -n openshift-storage
+    Run With Kubeconfig    oc delete daemonset topolvm-node -n openshift-storage
+    Run With Kubeconfig    oc delete deployment service-ca -n openshift-service-ca
+    Run With Kubeconfig    oc wait --for delete pod --selector=app.kubernetes.io/name=topolvm-csi-driver -n openshift-storage --timeout=120s
+    Run With Kubeconfig    oc wait --for delete pod --selector=app=service-ca -n openshift-service-ca --timeout=120s
```
ggiguash, line 130: What heppens if pods are deleted before we get into the way command?

test/resources/microshift-process.resource
ggiguash: Would moving it after `Wait For MicroShift` have the same effect?

dhellmann (commented):
I don't think this is the approach we want to take to fix this. See comment inline.

test/kickstart-templates/kickstart-liveimg.ks.template
```diff
@@ -109,4 +109,10 @@ ln -sf /dev/null /etc/systemd/user/grub-boot-success.timer
 # configure FIPs
 REPLACE_FIPS_COMMAND
 
+# Force configuration for inotify watches, which are used when interacting with the
+# apiserver, as in getting logs, for example. These are the same defaults as OCP uses.
+echo "fs.inotify.max_user_watches = 65536" >> /etc/sysctl.conf
+echo "fs.inotify.max_user_instances = 8192" >> /etc/sysctl.conf
+sysctl --system
```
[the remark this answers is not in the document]
pacevedom: This is only an "issue" when the cluster has a high number of pods and/or heavy interaction with oc/networking. Should we make a reference in docs?

test/resources/microshift-process.resource
```diff
@@ -80,6 +77,16 @@ Restart MicroShift
     [Documentation]    Restart the MicroShift service
     ${ushift_pid}=    MicroShift Process ID
     ${ushift_etcd_pid}=    MicroShift Etcd Process ID
+    # Force delete of core pods to avoid extended downtime after a restart.
```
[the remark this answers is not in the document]
pacevedom: While you are right, there are some details that are specific to the tests, like performing many restarts in a short period of time. While this is also possible in a real deployment, its unlikely to happen as it may cause disruptions to applications, not just the core components. If you wait just a bit more it will heal by itself, but tests dont allow that much waiting.
Anyhow, what I am after now is the 'forbidden' logs from a different test, which is unexpected. I will try again with the TIMESTAMP env var.
If this doesnt fix it then I think we are better off with skipping these tests until its ok again.

test/resources/microshift-process.resource
pacevedom: The pod deletion is triggered by deleting the top level resource, which is the daemonset/deployment, so they can not be recreated.

test/resources/microshift-process.resource
pacevedom: The delete commands for daemonset/deployment return in milliseconds, plus the wait command returns immediately if there is no match for the selector, meaning the pods have already been deleted.

test/kickstart-templates/kickstart-liveimg.ks.template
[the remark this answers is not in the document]
ggiguash: We should probably create a doc issue as our policy has been not to have MicroShift configure the OS.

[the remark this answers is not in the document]
dhellmann: It does sound like maybe we want documentation instead of a production code change. Keeping the settings here in the kickstart is OK, too.

test/resources/microshift-process.resource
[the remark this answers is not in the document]
dhellmann: > While you are right, there are some details that are specific to the tests, like performing many restarts in a short period of time. While this is also possible in a real deployment, it's unlikely to happen as it may cause disruptions to applications, not just the core components. If you wait just a bit more it will heal by itself, but tests dont allow that much waiting.

Should the tests wait for longer?

> Anyhow, what I am after now is the 'forbidden' logs from a different test, which is unexpected. 

What is being forbidden?

> I will try again with the TIMESTAMP env var. If this doesnt fix it then I think we are better off with skipping these tests until we solve it again.

pacevedom:
Will split the log scan test to its own scenario

pkg/components/controllers.go
```diff
@@ -175,7 +177,8 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		return err
 	}
 
-	if err := assets.ApplyDeployments(ctx, apps, renderTemplate, renderParamsFromConfig(cfg, nil), kubeconfigPath); err != nil {
+	now := time.Now().Unix()
```
dhellmann, line 180: Instead of repeating this pattern everywhere, maybe we want a factory for RenderParams to add the timestamp? If we already use renderParamsFromConfig everywhere we could just add it there.

If ApplyDeployments() added the annotation, we wouldn't even need to update the assets to include the template parameter.

pacevedom: Hm, you are right, didnt realize that was a function we owned. Actually this is a pattern in library-go. Will do that and also rebase the PR to the latest changes in tests.

pkg/assets/apps.go
```diff
@@ -55,6 +57,10 @@ func (d *dpApplier) Reader(objBytes []byte, render RenderFunc, params RenderPara
 		panic(err)
 	}
 	d.dp = obj.(*appsv1.Deployment)
+	if d.dp.Spec.Template.ObjectMeta.Annotations == nil {
+		d.dp.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+	d.dp.Spec.Template.ObjectMeta.Annotations["operator.openshift.io/force"] = uuid.NewString()
```
pmtk, line 63: Do we want this approach for all Deployments/DaemonSets we have or only the selected ones? ovn-k for example

pmtk, line 63: I wonder if this is an annotation string we should use. Function in library-go is deprecated so I wouldn't consider this to an established convention and without `microshift` in there it could send someone on a trip in wrong direction.

dhellmann: Maybe we want a flag to this function? Which is the more common case?

dhellmann: Good point about the name.

pacevedom: Turns out this also flakes. I have seen an execution where the old pod had a deletion timestamp (but was not deleted yet and was signaling readiness) and the new one was ready as well. This failed greenboot because it does not expect 2 pods from different deployment/daemonset versions.
Going to try the old way again about restarting the pod from the test until we figure out how to do this gracefully from within MicroShift (or its components).

test/scenarios/el92-src@log-scan.sh
```diff
@@ -0,0 +1,16 @@
...
+    prepare_kickstart host1 kickstart.ks.template rhel-9.2-microshift-source
+    launch_vm host1
+}
+
+scenario_remove_vms() {
+    remove_vm host1
+}
+
+scenario_run_tests() {
+    run_tests host1 suites/logscan/
```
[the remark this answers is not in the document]
pacevedom: These tests look specifically for graceful start of MicroShift. When doing restarts you disrupt the pods and the apiserver in such a way that errors will happen, but shouldnt on a fresh reboot.
These tests also look at the journal without knowing where to stop since the test execution order is randomized. If there have been no restarts it works, if there were any it may fail.

pacevedom:
/test microshift-metal-tests
/test microshift-metal-tests-arm

`metal-tests-arm` failed because of ingress taking too much time to be ready after a reboot. This is unrelated to the PR, but worth looking.

pacevedom:
Log scan tests are flaking for the same reason as networking smoke. In order to stabilize them we will need to change the way in which a restart is made in them. We need a full wipe (which is disruptive) to avoid having old containers while restarting the apiserver, as there are some error messages we expect to happen in that time frame.
Will do this in a separate PR.

test/resources/microshift-process.resource
```diff
@@ -112,3 +113,17 @@ Cleanup MicroShift
...
+    ...    connection with apiserver.
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    crictl pods --namespace openshift-service-ca | grep -v "POD ID" | awk '{print $1}' | xargs crictl stopp
+    ...    return_stdout=True    return_stderr=True    return_rc=True
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    crictl pods --namespace openshift-storage | grep -v "POD ID" | awk '{print $1}' | xargs crictl stopp
+    ...    return_stdout=True    return_stderr=True    return_rc=True
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    crictl pods --namespace openshift-ingress | grep -v "POD ID" | awk '{print $1}' | xargs crictl stopp
+    ...    return_stdout=True    return_stderr=True    return_rc=True
```
ggiguash, line 129: You can use a shorter expression, i.e.
```
crictl pods -q --namespace openshift-storage | xargs crictl stopp
```

pacevedom: Done, thanks!

test/resources/microshift-process.resource
```diff
@@ -112,3 +113,22 @@ Cleanup MicroShift
...
+
+    Crictl Stop Pods    openshift-service-ca
+    Crictl Stop Pods    openshift-storage
+    Crictl Stop Pods    openshift-ingress
+
+Crictl Stop Pods
+    [Documentation]    Stops CRI pods using crictl in the given namespace.
+    [Arguments]    ${namespace}
+
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
```
ggiguash, line 130: Don't we need SUDO here?

test/kickstart-templates/kickstart.ks.template
```diff
@@ -9,4 +9,10 @@
 %include /post-system.cfg
 %include /post-network.cfg
 
+# Force configuration for inotify watches, which are used when interacting with the
+# apiserver, as in getting logs, for example. These are the same defaults as OCP uses.
+echo "fs.inotify.max_user_watches = 65536" >> /etc/sysctl.conf
+echo "fs.inotify.max_user_instances = 8192" >> /etc/sysctl.conf
+sysctl --system
```
ggiguash, line 16: Should we add this code to post-system.cfg?

pacevedom:
Ready for review again, last couple of runs didnt have any more flakes.

test/kickstart-templates/includes/post-system.cfg
```diff
@@ -20,3 +20,9 @@ EOF
 # user login, this results in more than requested amount of red-boot-induced reboots
 # and system needing much more time to roll back.
 ln -sf /dev/null /etc/systemd/user/grub-boot-success.timer
+
+# Force configuration for inotify watches, which are used when interacting with the
+# apiserver, as in getting logs, for example. These are the same defaults as OCP uses.
+echo "fs.inotify.max_user_watches = 65536" >> /etc/sysctl.conf
+echo "fs.inotify.max_user_instances = 8192" >> /etc/sysctl.conf
+sysctl --system
```
dhellmann, line 28: What did we decide to do about these settings? Should we document them?

dhellmann (commented):
I'm happy with this version. I had one question about future work, and I'll leave the final review to someone else on the team.

You may want to remove the hold before that review.

test/kickstart-templates/includes/post-system.cfg
pacevedom: Yes, I will open a ticket.

test/suites/standard1/networking-smoke.robot
```diff
@@ -77,6 +74,11 @@ Wait For Service Deletion With Timeout
     Wait Until Keyword Succeeds    30s    1s
     ...    Network APIs With Test Label Are Gone
 
+Restart Router
+    [Documentation]    Testing
```
ggiguash, line 78: Docs should be updated.

test/suites/standard1/networking-smoke.robot
```diff
@@ -20,13 +20,11 @@ ${HOSTNAME}                 hello-microshift.cluster.local
 *** Test Cases ***
 Router Smoke Test
     [Documentation]    Run a router smoke test
-    [Tags]    robot:exclude
     [Setup]    Run Keywords
-    ...    Restart Greenboot And Wait For Success
```
pmtk, line 25: So this assumes the VM is already healthy when the test starts?

test/scenarios/el92-src@log-scan.sh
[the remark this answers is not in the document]
ggiguash: Can we avoid adding another VM? 
I mean, if we're looking for clean logs, can we deleted them?

P.S. It's getting very tight with the number of VMs

test/suites/standard1/networking-smoke.robot
ggiguash: Can we add an explanation in a comment (one that cannot be missed easily) on the problem and solution?

test/scenarios/el92-src@log-scan.sh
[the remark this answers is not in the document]
pacevedom: I am thinking of doing something different, which is to wipe the whole microshift deployment as part of this test. This is disruptive with what was running before, but getting a clean log start from an already running system is somehow complex/flaky by nature. Would that be ok? I would do it as part of a different PR though.

test/suites/standard1/networking-smoke.robot
pacevedom: Instead of waiting for the whole system it just waits for the components it needs to test, that is apiserver and router.

test/suites/standard1/networking-smoke.robot
pacevedom: Done.

PR #2873, opened 2024-01-14
USHIFT-2183: Make sos report collection more robust in scenario.sh

ggiguash:


test/bin/scenario.sh
```diff
@@ -137,19 +143,21 @@ sos_report_for_vm() {
...
-    sudo /usr/bin/microshift-sos-report
+if ! hash sos ; then
+    touch /tmp/sosreport-command-does-not-exist
+elif [ -f /usr/bin/microshift-sos-report ]; then
+    /usr/bin/microshift-sos-report || touch /tmp/sosreport-command-failed
 else
-    sudo chmod +x /tmp/microshift-sos-report.sh
-    sudo PROFILES=network,security /tmp/microshift-sos-report.sh
+    chmod +x /tmp/microshift-sos-report.sh
+    PROFILES=network,security /tmp/microshift-sos-report.sh || touch /tmp/sosreport-command-failed
```
copejon, line 152: Is `touch /tmp/sosreport-command-failed` checked in another script?

ggiguash: It's copied to the CI logs using the same `/tmp/sosreport-*` pattern.
So, if sos collection fails, it does not fail the CI job, but we'll see a special file created in the logs

ggiguash: The ideas is - never fails CI job because of sos collection errors.

dhellmann: We're collecting the logs from this script already. Maybe instead of creating separate files we could just log clear messages?

test/bin/scenario.sh
```diff
@@ -137,19 +143,21 @@ sos_report_for_vm() {
...
+elif [ -f /usr/bin/microshift-sos-report ]; then
+    /usr/bin/microshift-sos-report || echo "ERROR: The /usr/bin/microshift-sos-report script failed"
 else
-    sudo chmod +x /tmp/microshift-sos-report.sh
-    sudo PROFILES=network,security /tmp/microshift-sos-report.sh
+    chmod +x /tmp/microshift-sos-report.sh
+    PROFILES=network,security /tmp/microshift-sos-report.sh || echo "ERROR: The /tmp/microshift-sos-report.sh script failed"
 fi
-sudo chmod +r /tmp/sosreport*
+chmod +r /tmp/sosreport-* || echo "ERROR: The sos report files do not exist in /tmp"
```
dhellmann, line 154: Why is the sudo command removed from this line? I think we need it to change the permissions on files owned by root, don't we?

test/bin/scenario.sh
```diff
@@ -137,19 +143,21 @@ sos_report_for_vm() {
...
 EOF
     copy_file_to_vm "${vmname}" "/tmp/sos-wrapper.sh" "/tmp/sos-wrapper.sh"
     copy_file_to_vm "${vmname}" "${ROOTDIR}/scripts/microshift-sos-report.sh" "/tmp/microshift-sos-report.sh"
     run_command_on_vm "${vmname}" "sudo bash -x /tmp/sos-wrapper.sh"
     mkdir -p "${vmdir}/sos"
-    scp "redhat@${ip}:/tmp/sosreport*.tar.xz" "${vmdir}/sos/"
+    scp "redhat@${ip}:/tmp/sosreport-*" "${vmdir}/sos/" || true
```
dhellmann, line 160:
```suggestion
    scp "redhat@${ip}:/tmp/sosreport-*" "${vmdir}/sos/" || echo "Ignoring error copying sos report files"
```

test/bin/scenario.sh
ggiguash: Because the script is called with sudo already and it all runs elevated. So, all sudo in the script were removed for better readability.

test/bin/scenario.sh
```diff
@@ -65,7 +65,13 @@ run_command_on_vm() {
...
     local -r ssh_port=$(get_vm_property "${vmname}" ssh_port)
 
-    ssh "redhat@${ip}" -p "${ssh_port}" -t "${command}"
+    local term_opt=""
+    if [ -t 0 ] ; then
+        # Allocate pseudo-terminal for SSH commands when stdin is a terminal
+        # Necessary in devenv for entering input i.e. system registration, etc.
+        term_opt="-t"
+    fi
+    ssh "redhat@${ip}" -p "${ssh_port}" "${term_opt}" "${command}"
```
dhellmann, line 74: What happens here if term_opt is not set and ssh gets an empty argument? I'd expect to not quote the variable here so that the empty variable would be transparently ignored.

ggiguash: Empty arguments are usually ignored by commands, it's like a no-op
```
$ args=""
$ ssh microshift@microshift-dev-rhel92 -p 22 "${args}" "hostname"
microshift-dev-rhel92

$ args="-t"
$ ssh microshift@microshift-dev-rhel92 -p 22 "${args}" "hostname"
microshift-dev-rhel92
Connection to microshift-dev-rhel92 closed.
```

ggiguash: In any case, I will fix it not to raise questions.

ggiguash: It does not even need disabling the warning because we're manipulating constants.
```
    # shellcheck disable=SC2086
```

PR #2885, opened 2024-01-16
OCPBUGS-23336: Check AdvertiseAddress configured in NICs

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -279,6 +279,13 @@ func (c *Config) validate() error {
 		)
 	}
 
+	if c.ApiServer.SkipInterface {
+		err := checkAdvertiseAddressConfigured(c.ApiServer.AdvertiseAddress)
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -316,3 +323,21 @@ func getAllHostnames() ([]string, error) {
...
+	}
+	for _, addr := range addrs {
+		// interface addresses come with the mask at the end.
+		addrStr := addr.String()
+		if idx := strings.Index(addrStr, "/"); idx != -1 {
+			addrStr = addrStr[:idx]
+		}
+		if addrStr == advertiseAddress {
+			return nil
+		}
```
pmtk, line 340: What do you think about using `strings.Contains()`?

pacevedom: If I understood you correctly, that could make false matches.
For example, if the advertiseAddress is configured to `10.44.0.1` and one interface has `10.44.0.11/24` then the contains would return true.

pkg/config/config_test.go
```diff
@@ -281,7 +281,7 @@ func TestGetActiveConfigFromYAML(t *testing.T) {
 func TestValidate(t *testing.T) {
 	mkDefaultConfig := func() *Config {
 		c := NewDefault()
-		c.ApiServer.SkipInterface = true
+		c.ApiServer.SkipInterface = false
```
dhellmann, line 284: Why was the original test written to set this to true?

pacevedom: This variable went through at least one major change, but in the context of these tests there was no difference. It only marks whether a later stage should configure the `advertiseAddress` in a local interfacce or not.
In the config update/validation phase this is irrelevant, as there are only sets to this variable. With this change the variable becomes relevant for validations.
If its set to true it means a user explicitly introduced the advertiseAddress, regardless of the value, even if its the same as the default. This triggers validations to check whether the IP is already configured.
If its set to false it will just calculate a default by itself and skip the validations.

pkg/config/config.go
pmtk: You're right, thx for clarifying

PR #2902, opened 2024-01-19
USHIFT-2200: Fix configure-vm.sh to work in standalone mode and on CentOS

ggiguash:


scripts/devenv-builder/configure-vm.sh
```diff
@@ -59,14 +61,24 @@ if [ $# -ne 1 ]; then
     usage "Wrong number of arguments"
 fi
 
-# The conditional check for dnf_retry.sh script presence is required 
-# because configure-vm.sh can be used as a standalone script when
-# bootstrapping the development environment
+# The conditional check for presence of the used scripts and files
+# is required because configure-vm.sh can be run as a standalone
+# script when bootstrapping the development environment
```
dhellmann, line 66: Could we say that to install this script you clone the git repo? That way as we add new scripts over time we don't have to keep updating this section of the code.

ggiguash: We can do this, but it requires changes in the script, docs and boostrap procedure. 
Currently the script checks out the repo itself. I think it might be a good improvement, but it's a new "feature".

If you think we should try it, I'd be happy to implement a prototype - can you open an issue if you do?

dhellmann: It would require updates to the docs, but it's how I already use the script so it already works that way.

ggiguash: It's a bit more than docs because to get the sources, we need git and this comes with problems:
* It is not always installed on the system, so we need to register it first. 
* dnf_retry must be used or it's failing occasionally

The configure-vm.sh script can be optimized, but it's a different task. Currently, it simply does not work from scratch - it's a bug.

pmtk:
/lgtm

If we want to change behavior of the script (cloning repo instead of curl'ing individual files) let's do it in separate PR.

PR #2919, opened 2024-01-23
USHIFT-2213: wrong namespace in docs

copejon:


dhellmann:
/lgtm
/hold

The change looks good. I suggested moving the ticket to USHIFT, so the hold is to give you time to do that and update the title here before the PR merges.

PR #2923, opened 2024-01-24
OCPBUGS-27849: Create microshift-olm-release-info RPM

pmtk:
Creates new dedicated RPM for OLM's release info that depends on `microshift-release-info`

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -137,6 +137,17 @@ Requires: microshift = %{version}
 %description olm
 The microshift-olm package provides the required manifests for the Operator Lifecycle Manager to be installed on MicroShift.
 
+%package olm-release-info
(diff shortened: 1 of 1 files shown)


packaging/rpm/microshift.spec
```diff
@@ -137,6 +137,17 @@ Requires: microshift = %{version}
 %description olm
 The microshift-olm package provides the required manifests for the Operator Lifecycle Manager to be installed on MicroShift.
 
+%package olm-release-info
+Summary: Release information for Operator Lifecycle Manager components for MicroShift
+BuildArch: noarch
+Requires: microshift-release-info = %{version}
```
ggiguash, line 143: This does not correspond to the description of the PR. I guess, we need to fix the description because the code seems to be OK.

pmtk: fixed

PR #2940, opened 2024-01-28
USHIFT-2090: Introduce cluster-id file creation on MicroShift startup

ggiguash:


pkg/controllers/clusterid.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (s *ClusterID) Run(ctx context.Context, ready chan<- struct{}, stopped chan<- struct{}) error {
+	defer close(stopped)
+	defer close(ready)
+
+	// Read the 'kube-system' namespace attributes
+	restConfig, err := clientcmd.BuildConfigFromFlags("", s.cfg.KubeConfigPath(config.KubeAdmin))
+	if err != nil {
+		panic(err)
```
pmtk, line 54: I'd prefer not to panic. Returning error should be equally effective

pkg/controllers/clusterid.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	coreClient := clientv1.NewForConfigOrDie(rest.AddUserAgent(restConfig, "core-agent"))
+	namespace, err := coreClient.Namespaces().Get(ctx, "kube-system", metav1.GetOptions{})
+	if err != nil {
+		panic(err)
```
pmtk, line 59: I'd prefer not to panic. Returning error should be equally effective

pkg/controllers/clusterid.go
```diff
@@ -0,0 +1,88 @@
...
+	_, err := os.Stat(fileName)
+	if !os.IsNotExist(err) {
+		return
+	}
+
+	// Write the cluster ID to a new file
+	klog.Infof("Writing MicroShift Cluster ID '%v' to '%v'", clusterID, fileName)
+	err = os.WriteFile(fileName, []byte(clusterID), 0400)
+	if err != nil {
+		panic(err)
```
pmtk, line 86: I'd prefer not to panic. Let's return error from this func and propagate it to ServiceManager

pkg/controllers/clusterid.go
```diff
@@ -0,0 +1,88 @@
...
+	klog.Infof("MicroShift Cluster ID: %v", clusterID)
+
+	return ctx.Err()
+}
+
+func initClusterIDFile(clusterID string) {
+	// The location of the cluster ID file
+	fileName := filepath.Join(config.DataDir, "cluster-id")
+
+	// Do not create the cluster ID file if it already exists
```
pmtk, line 76: Have we decided that getting this out of sync is non-issue? For example, what if user writes something to that file.

test/suites/osconfig/clusterid.robot
```diff
@@ -0,0 +1,80 @@
...
+Test Tags           restart    slow
+
+
+*** Variables ***
+${CLUSTERID_FILE}       /var/lib/microshift/cluster-id
+${CLUSTERID_NS}         kube-system
+
+
+*** Test Cases ***
+Compare Cluster ID From Namespace And File
```
pmtk, line 21: Do we need this test? The other one does that and more.
Regardless if this test runs as first or second, it duplicates checks from the other.

pkg/controllers/clusterid.go
ggiguash: I think it's not an issue. The file is created with `0400` permissions. Users can tamper with any database file potentially and get wierd results.

Do you think I should add a check between the namespace value and the file value? It's easy to detect such an inconsistency and abort.

dhellmann: I thought we would overwrite the file on every startup to ensure the value is as close to up to date as possible.

ggiguash: I would rather create the file once and not touch it. If it's tampered by the user, we could log an error message and exit. 
Otherwise, it may be a bit of a problem to explain this file purpose in the user docs, i.e. we have this read-only file here, but it is overwritten on every startup...

ggiguash: So, I'm leaning to the stricter implementation, but I'm open to suggestions either way. 
WDYT?

dhellmann: The only reason we're writing the file is to have a way to get the cluster ID when MicroShift is down and it can't be read from the database. I don't think it's good practice to refuse to start because the user has taken an action we can easily circumvent. This is a good example of how we can make the software more resilient by making it flexible.

dhellmann: You could draw a parallel to a status field in a kube API. If this value was in the database and we wanted to prevent it from being overwritten we would have a controller constantly watch it and reset the value if someone changed it.

ggiguash: Added a warning and rewrite in case of tampering with the cluster ID.
Also added a test checking this scenario.

pkg/controllers/clusterid.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func initClusterIDFile(clusterID string) error {
+	// The location of the cluster ID file
+	fileName := filepath.Join(config.DataDir, "cluster-id")
+
+	// Read and verify the cluster ID file if it already exists,
+	// logging a warning if the cluster ID is inconsistent
+	_, err := os.Stat(fileName)
+	if !os.IsNotExist(err) {
```
dhellmann, line 84: This check will pass if the error is that we can't read the file for some reason.

If the logic is reversed, so that on line 83 we try to read the contents of the file and then if the file does not exist we skip the comparison, that should be shorter and clearer. Something like

```
data, err := os.ReadFile(fileName)
if err && !os.IsNotExist(err) {
    return err
}
if len(data) && string(data) == clusterID {
    return nil
}
// log warning
// write new file contents
```

ggiguash: Good point. Fixing.

pmtk:
`osconfig` scenario is part of periodics
/test metal-periodic-test metal-periodic-test-arm

test/suites/osconfig/clusterid.robot
```diff
@@ -0,0 +1,92 @@
...
+    [Documentation]    Read and return the cluster ID from the kube-system namespace.
+    ${clusterid}=    Oc Get Jsonpath    namespaces    ${CLUSTERID_NS}    ${CLUSTERID_NS}    .metadata.uid
+
+    Should Not Be Empty    ${clusterid}
+    RETURN    ${clusterid}
+
+Tamper With Cluster ID File
+    [Documentation]    Append invalid characters to the cluster ID file.
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    echo -n 123 >> ${CLUSTERID_FILE}
```
pmtk, line 89:
```
bash: line 1: /var/lib/microshift/cluster-id: Permission denied
```

I think it's failing because `sudo=True` might only prepend sudo to the provided command (rather than wrap it with `sudo bash -c` or whatever), so `echo` runs with sudo, but >> does not.

ggiguash: Exactly, thanks for pointing out that we need to run periodics.

PR #2941, opened 2024-01-28
USHIFT-881: Add support for CentOS 9 OS in test harness framework

ggiguash:
- Add support for new blueprint `.image-fetcher` type to download arbitrary ISO image or file
- Add CentOS-specific kickstart installing from CDROM and using microshift-local repo on the hypervisor
- Mimic the code in configure-vm.sh script for bootstrapping CentOS configuration for MicroShift
- Update test framework docs to include `.image-fetcher` description
- Fix rpm install test that warned on the conflict with the existing /etc/containers/policy.json file


test/bin/build_images.sh
```diff
@@ -376,12 +377,47 @@ do_group() {
...
+            echo "Starting image-fetch for ${download_file}"
+            wget -nv -O "${expected_iso_file}" "${download_url}" &
+            download_pids+=($!)
+        done
+        # Make sure all image-fetch processes are terminated on exit
+        if [ ${#download_pids[@]} -ne 0 ]; then
+            # Expand the values of download PIDs here not to reference
+            # the local variable on script exit
+            # shellcheck disable=SC2064
+            trap "kill ${download_pids[*]}" EXIT
```
dhellmann, line 403: Does this override any other exit trap, like the one to collect logs?

test/bin/build_images.sh
```diff
@@ -376,12 +377,47 @@ do_group() {
...
         echo "Waiting for builds to complete..."
         # wait_images.py returns possibly updated list of builds that must be handled
         # "update" means replacing initial build ID with retry build ID
         builds_to_get=$(time "${SCRIPTDIR}/wait_images.py" "${buildid_list[@]}")
     fi
+    if [ ${#download_pids[@]} -ne 0 ]; then
+        # Make sure all potentially running wget processes are complete
+        echo "Waiting for image-fetch to complete..."
+        # shellcheck disable=SC2068
+        wait ${download_pids[@]}
```
dhellmann, line 417: In other places in the test framework we use `parallel` to manage concurrency? Could we do that here? It's far less error prone.

test/bin/build_images.sh
```diff
@@ -376,12 +377,47 @@ do_group() {
         done
     fi
 
-    if (( ${#buildid_list[@]} )); then
+    if ${BUILD_INSTALLER} && ! ${COMPOSER_DRY_RUN}; then
+        for download_file in "${groupdir}"/*.image-fetch; do
```
dhellmann, line 381: Since there is no build step for these, would it make sense to put them in their own group or process them separately instead of mingling them in with the groups of image build instructions?

ggiguash: The pro for putting them together with everything else is that images are fetched in parallel together with osbuilder running. It can take up to 5-10m to download an .iso, which I thought would be a waste.

test/bin/build_images.sh
ggiguash: If I use parallel, this one should not be necessary, I think. 
I'll upload the new code asap.

test/bin/build_images.sh
```diff
@@ -381,7 +382,31 @@ do_group() {
...
+
+            echo "Adding image-fetcher for ${download_file}"
+            download_opts+=("${expected_iso_file} ${download_url}")
+        done
+    fi
+
+    # Run image-fetcher while osbuilder is running in background
+    if [ ${#download_opts[@]} -ne 0 ]; then
+        echo "Waiting for image-fetcher to complete..."
+        parallel --colsep ' ' wget -nv -O "{1}" "{2}" ::: "${download_opts[@]}"
```
dhellmann, line 407: The build_rpms.sh script uses some additional options to provide progress information when running on the console and to save log output to files for debugging.

test/bin/build_images.sh
```diff
@@ -381,7 +382,31 @@ do_group() {
...
+            if [ -f "${expected_iso_file}" ]; then
+                echo "${expected_iso_file} already exists"
+                if should_skip "${blueprint}"; then
+                    record_junit "${groupdir}" "${download_file}" "download" "SKIPPED"
+                    continue
+                fi
+            fi
+
+            echo "Adding image-fetcher for ${download_file}"
+            download_opts+=("${expected_iso_file} ${download_url}")
```
dhellmann, line 400: Downloading the file directly to its final location means that cancelling the download leaves a file in place that prevents the operation from being repeated until the user cleans up that file. It would be more resilient to download to a temporary name/location and then rename the file after it is fully downloaded. I don't know if that's something wget or curl will do automatically or if we would have to wrap the command.

test/kickstart-templates/includes/post-firewall.cfg
```diff
@@ -0,0 +1,10 @@
+dnf install -y firewalld
+firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16
+firewall-offline-cmd --zone=trusted --add-source=169.254.169.1
+firewall-offline-cmd --zone=public --add-port=80/tcp
+firewall-offline-cmd --zone=public --add-port=443/tcp
+firewall-offline-cmd --zone=public --add-port=5353/udp
+firewall-offline-cmd --zone=public --add-port=6443/tcp
+firewall-offline-cmd --zone=public --add-port=30000-32767/tcp
+firewall-offline-cmd --zone=public --add-port=30000-32767/udp
+firewall-offline-cmd --zone=public --add-service=mdns
```
dhellmann, line 10: There's a function in scenario.sh to set up the firewall on a running host. We also have these settings in blueprints, which of course aren't used for CentOS, but I would at least like to avoid adding a 3rd location with the same settings.

test/scenarios/cos9-src@rpm-standard1.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
pmtk, line 1: Do you think we need 2 COS9 scenarios in presubmit? We don't even support COS9. I'd move one of them to periodics and remove other one, so that we have some coverage of this new functionality.

test/bin/scenario.sh
```diff
@@ -139,17 +139,23 @@ sos_report_for_vm() {
...
     # can't rely on the wrapper being there or working if it
     # is. Copy the script to the host, just in case, along with a
-    # wrapper that knows how to execute it or the installed
-    # version.
+    # wrapper that knows how to execute it or the installed version.
+    # Add anaconda to the list of the plugins to collect information
+    # about failed image installations.
     cat - >/tmp/sos-wrapper.sh <<EOF
 #!/usr/bin/env bash
+PLUGINS=container_log,crio,logs,rpmostree,rpm,anaconda
```
pmtk, line 147: Let's think if including `anaconda` in current state it the best action

ggiguash: We need /var/log/anaconda/ directory contents for debugging the installation.
Should we just copy those explicitly?

test/scenarios/cos9-src@rpm-standard1.sh
ggiguash: I agree that we should be flexible with this. 
It's mostly about testing that MicroShift install and startup work. I'll implement the suggestion.

test/bin/scenario.sh
ggiguash: The problem seems to be in the /root/anaconda-ks.cfg file.
Let me try to copy the contents of /var/log/anaconda explicitly, without using the plugin.

ggiguash:
FIPS failure, not related to CentOS changes

pmtk:
nice!
/lgtm

PR #2949, opened 2024-01-30
USHIFT-2239: Destroy VM when virt-install exits with errors

ggiguash:


ggiguash:
Ready for review. The failure is not related to the change.

test/bin/scenario.sh
```diff
@@ -486,6 +486,10 @@ launch_vm() {
     if ${vm_created} ; then
         record_junit "${vmname}" "install_vm" "OK"
     else
+        # Make sure to stop the VM on error before the control is returned.
+        # This is necessary not to leave running qemu child processes so that
+        # the caller considers the script fully complete.
+        sudo virsh destroy "${full_vmname}" || true
```
dhellmann, line 492: This is going to make it more difficult to debug failures when running scenarios locally. Could we have a flag to control the behavior? Leaving the default set to cleanup automatically would make sense.

ggiguash: I only enable this option when stdin is not available (non-interactive session).
Is this better now?

dhellmann: That works, thanks!

pacevedom (commented):
/lgtm

ggiguash:
FIPS failures are not related to the current change

PR #2950, opened 2024-01-30
USHIFT-2241: use unique repo name when extracting container images

dhellmann:
Reusing the same temporary repo name for different repository locations
can introduce issues if the dnf cache is not updated. Use the base name
of the repository URL as a unique name.

/assign @pacevedom

### DIFF
--- a/test/bin/build_images.sh
+++ b/test/bin/build_images.sh
@@ -29,13 +29,13 @@ extract_container_images() {
     local -r version=$1
     local -r repo_url=$2
     local -r outfile=$3
-    local -r repo_name="microshift-extract-images"
 
     echo "Extracting images from ${version}"
     mkdir -p "${IMAGEDIR}/release-info-rpms"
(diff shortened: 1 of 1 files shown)


test/bin/build_images.sh
```diff
@@ -29,7 +29,7 @@ extract_container_images() {
     local -r version=$1
     local -r repo_url=$2
     local -r outfile=$3
-    local -r repo_name="microshift-extract-images"
+    local -r repo_name="$(basename "${repo_url}")"
```
pacevedom, line 32: The repo_url var may be the empty string at times. `basename` will fail on that and the script runs with `set -e`, shouldnt that trigger a failure? Maybe we could just get rid of the variable since it is only used in line 39?

dhellmann: I can compute the name later where it's used when we know the URL has a value.

pacevedom (approved):
/lgtm

PR #2951, opened 2024-01-30
USHIFT-2242: add option to build_images.sh to skip log collection step

dhellmann:
Clean up help text and add new option to skip logs.

/assign @pacevedom


dhellmann:
Jobs are failing with an error about the pull secret being missing from the pod so it can't be copied into the test host.

pacevedom (approved):
/lgtm

PR #2956, opened 2024-01-31
USHIFT-2220: Initial Multus manifests

pmtk:


assets/optional/multus/06-daemonset.yaml
```diff
@@ -0,0 +1,113 @@
...
+        - mountPath: /entrypoint
+          name: cni-binary-copy
+        - mountPath: /host/etc/os-release
+          name: os-release
+        - name: system-cni-dir
+          mountPath: /host/etc/cni/net.d
+        - name: cnibin
+          mountPath: /host/opt/cni/bin
+        env:
+        - name: RHEL8_SOURCE_DIRECTORY
```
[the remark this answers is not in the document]
pmtk: No, but we don't have the microshift specific Multus image yet and I plan to change these `env`s and the ConfigMap that copies binaries later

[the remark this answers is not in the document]
pmtk: Changed, thx

assets/optional/multus/06-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+        - name: cnibin
+          mountPath: /host/opt/cni/bin
+        - mountPath: /entrypoint
+          name: cni-binary-copy
+        env:
+        - name: DEFAULT_SOURCE_DIRECTORY
+          # TODO: Change when microshift-specific multus image is ready
+          value: "/usr/src/plugins/rhel9/bin/"
+        - name: BINARIES_TO_COPY
+          value: "{bridge,ipvlan,macvlan,static,dhcp,host-local}"
```
s1061123, line 51: Q: `dhcp` will be supported in MicroShift (if so, you need to create DHCP plugin server pod as https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/multus/003-dhcp-daemon.yaml)

assets/optional/multus/06-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+        image: multus-cni
+        imagePullPolicy: IfNotPresent
+        command: [ "/bin/bash", "-ec", "--" ]
+        args:
+          - >
+            /entrypoint/cnibincopy.sh;
+            exec /usr/src/multus-cni/bin/thin_entrypoint
+            --multus-conf-file=auto
+            --multus-autoconfig-dir=/host/etc/cni/net.d
+            --readiness-indicator-file=/etc/cni/net.d/10-ovn-kubernetes.conf
```
s1061123, line 63: `--readiness-indicator-file=/etc/cni/net.d/10-ovn-kubernetes.conf` expects that ovn-k config will be created under '/etc/cni/net.d' of multus pod. Is that actually created by ovn-k?

pmtk: Is it multus pod or multus cni (/opt/cni/bin/multus)? But yeah, we have that file on the host ([ref](https://github.com/openshift/microshift/blob/main/assets/components/ovn/single-node/master/daemonset.yaml#L376))

PR #2977, opened 2024-02-06
USHIFT-2262: Cleanup before running log scan tests

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/test/suites/logscan/log-scan.robot
+++ b/test/suites/logscan/log-scan.robot
@@ -40,21 +40,25 @@ Should Find MicroShift Is Ready
     [Tags]    robot:exclude
     Pattern Should Appear In Log Output    ${CURSOR}    MICROSHIFT READY
 
-
 *** Keywords ***
 Setup
     [Documentation]    Test suite setup
     Check Required Env Variables
     Login MicroShift Host
-    Setup Kubeconfig    # for readiness checks
+    Stop MicroShift
+    Cleanup MicroShift    --all    --keep-images
     # Save the journal cursor then restart MicroShift so we capture
     # the shutdown messages and startup messages.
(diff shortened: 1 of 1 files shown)


test/suites/logscan/log-scan.robot
```diff
@@ -40,21 +40,25 @@ Should Find MicroShift Is Ready
...
 *** Keywords ***
 Setup
     [Documentation]    Test suite setup
     Check Required Env Variables
     Login MicroShift Host
-    Setup Kubeconfig    # for readiness checks
+    Stop MicroShift
+    Cleanup MicroShift    --all    --keep-images
     # Save the journal cursor then restart MicroShift so we capture
     # the shutdown messages and startup messages.
```
dhellmann, line 51: This comment could do with updating now that the logic of Setup is changing. Explaining why we stop the service, do a full clean up, then do the cursor stuff and start & stop it would be good.

test/suites/logscan/log-scan.robot
```diff
@@ -40,21 +40,25 @@ Should Find MicroShift Is Ready
...
     ${cursor}=    Get Journal Cursor
     Set Suite Variable    \${CURSOR}    ${cursor}
-    Restart MicroShift
+    Setup Kubeconfig    # for readiness checks
+    Start MicroShift
+    Wait For MicroShift Service
     Restart Greenboot And Wait For Success
+    Stop MicroShift
```
dhellmann, line 58: The previous implementation left the service running after Setup. This doesn't. We've seen issues with restarting the service, so maybe instead of stopping here it should restart so we continue to monitor that case?

pacevedom:
/close
Will change the approach

PR #2982, opened 2024-02-06
USHIFT-2266: have the release note script handle the tag commands

dhellmann:
The earlier version of the release note script let the user run the
commands to create the tag. This update applies the tags automatically
to make the script easier to use.

/assign @pacevedom


pacevedom (approved):
/lgtm

PR #2986, opened 2024-02-08
Add layer4 blueprints to support building from ART rpms

jogeo:
This PR addresses [USHIFT-2344](https://issues.redhat.com//browse/USHIFT-2344) and provides a layer of blueprints intended to be used with RPMs build by ART.

### DIFF
--- a/test/bin/build_images.sh
+++ b/test/bin/build_images.sh
@@ -545,6 +545,10 @@ build_images.sh [-iIsdf] [-l layer-dir | -g group-dir] [-t template]
 
   -d      Dry run by skipping the composer start commands.
 
+  -e      Extract container images.
+
+  -E      Do not extract container images.
+
   -f      Force rebuilding images that already exist.
 
   -g DIR  Build only one group (cannot be used with -l or -t).
@@ -580,12 +584,18 @@ GROUP=""
 TEMPLATE=""
 FORCE_REBUILD=false
 FORCE_SOURCE=false
+EXTRACT_CONTAINER_IMAGES=true
 
 selCount=0
-while getopts "dfg:hiIl:sSt:" opt; do
+while getopts "deEfg:hiIl:sSt:" opt; do
     case "${opt}" in
         d)
             COMPOSER_DRY_RUN=true
+            ;;
+        e)
+            EXTRACT_CONTAINER_IMAGES=true
+            ;;
+        E)  EXTRACT_CONTAINER_IMAGES=false
             ;;
         f)
             FORCE_REBUILD=true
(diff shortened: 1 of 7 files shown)


test/bin/build_images.sh
```diff
@@ -545,6 +545,10 @@ build_images.sh [-iIsdf] [-l layer-dir | -g group-dir] [-t template]
 
   -d      Dry run by skipping the composer start commands.
 
+  -e      Extract container images.
+
+  -E      Do not extract container images.
+
```
ggiguash, line 551: We should not need both options. Let's keep the extraction ON by default and add "do not extract" flag.

test/image-blueprints/layer4-art/group2/rhel-92_microshift-release.toml
ggiguash, line 1: Is `microshift-release` a name we should use if the packages are coming from ART?

test/image-blueprints/layer4-art/group1/rhel.toml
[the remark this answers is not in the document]
jogeo: layer1 requires the microshift-test-agent package. That rpm package is built from source.
Layer4 will be run without building the source but rather after downloading rpms from brew or the mirror.

test/image-blueprints/layer4-art/group2/rhel-92_microshift-release.toml
```diff
@@ -0,0 +1,39 @@
+name = "microshift_release"
+description = "A RHEL 9.2 image with the RPMs built from ART."
+version = "0.0.1"
+modules = []
+groups = []
+distro = "rhel-92"
+
+[[packages]]
+name = "microshift"
+version = "{{ .Env.SOURCE_VERSION }}"
```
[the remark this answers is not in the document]
jogeo: Right, the RPMs are first downloaded, then just run build_images.sh.
Build_images.sh already extracts the version from the microshift-release-info rpm.

SOURCE_VERSION=$(rpm -q --queryformat '%{version}' "${release_info_rpm}")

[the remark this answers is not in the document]
dhellmann: I guess if you download the RPMs and put them in the same location as the ones built from source, they work with all of the other existing images from other layers, too? That seems like a reasonable benefit to trade off against the fact that we couldn't test source RPMs and ART built RPMs together in that setup.

test/image-blueprints/layer4-art/group2/rhel-92_microshift-release.toml
jogeo: I can pick a different name. The main goal here is to create a series of commits that have a common parent and ref so we can upgrade through the entire lineage; base-no-microshift --> rhel9.2 --> rhel9.3. The way the build_images.sh script works today is that it sets the ref to the value of name defined in the toml file. The result is that when you run 'rpm-ostree upgrade' matching refs are not found and it fails. The RF CI tests are not using 'rpm-ostree upgrade' but rather 'rpm-ostree rebase' so they don't hit this issue.

jogeo: Setting an alias might also work as well, so we can keep the names unique.

PR #2993, opened 2024-02-08
OCPBUGS-29271: Rebase lvms registry.access.redhat.com lvms4 lvms operator bundle v4.15.0

jakobmoellerdev:

Update to LVMS 4.15, also solves https://issues.redhat.com/browse/USHIFT-1874 by moving to an embedded lvmd implementation which has feature parity but uses less containers and slightly fewer resources. The lvmd socket is not necessary.

Closes #<Issue Number>

### DIFF
--- a/assets/components/lvms/topolvm-node_daemonset.yaml
+++ b/assets/components/lvms/topolvm-node_daemonset.yaml
@@ -30,26 +30,8 @@ spec:
                 - linux
       containers:
       - command:
-        - /lvmd
-        - --config=/etc/topolvm/lvmd.yaml
-        - --container=true
-        image: '{{ .ReleaseImage.topolvm_csi }}'
-        name: lvmd
-        resources:
-          requests:
-            cpu: 1m
-            memory: 22Mi
-        securityContext:
-          privileged: true
-          runAsUser: 0
-        volumeMounts:
-        - mountPath: '{{ Dir .SocketName  }}'
-          name: lvmd-socket-dir
-        - mountPath: /etc/topolvm
-          name: lvmd-config-dir
-      - command:
         - /topolvm-node
-        - --lvmd-socket={{ .SocketName }}
+        - --embed-lvmd
         env:
         - name: NODE_NAME
           valueFrom:
@@ -78,16 +60,16 @@ spec:
           protocol: TCP
         resources:
           requests:
-            cpu: 1m
-            memory: 16Mi
+            cpu: 2m
+            memory: 35Mi
         securityContext:
           privileged: true
           runAsUser: 0
         volumeMounts:
+        - mountPath: /etc/topolvm
+          name: lvmd-config-dir
         - mountPath: /run/topolvm
           name: node-plugin-dir
-        - mountPath: '{{ Dir .SocketName  }}'
-          name: lvmd-socket-dir
         - mountPath: /var/lib/kubelet/pods
           mountPropagation: Bidirectional
           name: pod-volumes-dir
@@ -163,8 +145,5 @@ spec:
             path: lvmd.yaml
           name: lvmd
         name: lvmd-config-dir
-      - emptyDir:
-          medium: Memory
-        name: lvmd-socket-dir
   updateStrategy:
     type: RollingUpdate
--- a/scripts/auto-rebase/lvms_assets.yaml
+++ b/scripts/auto-rebase/lvms_assets.yaml
@@ -4,6 +4,7 @@ assets:
     src: lvms/amd64/
     no_clean: True
     files:
+      - file: topolvm-configmap_lvms-version.yaml
       - file: topolvm-controller_rbac.authorization.k8s.io_v1_clusterrole.yaml
       - file: topolvm-controller_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
       - file: topolvm-controller_rbac.authorization.k8s.io_v1_role.yaml
@@ -30,6 +31,4 @@ assets:
       - file: topolvm-openshift-storage_namespace.yaml
         ignore: "provided by MicroShift"
       - file: topolvm_default-storage-class.yaml
-        ignore: "provided by MicroShift"
-      - file: topolvm-configmap_lvms-version.yaml
         ignore: "provided by MicroShift"
--- a/scripts/auto-rebase/rebase-lvms.sh
+++ b/scripts/auto-rebase/rebase-lvms.sh
@@ -31,16 +31,6 @@ load_version(){
     fi
 }
 
-parse_version(){
-    local image_tag="$1"
-    local v
-    v="${image_tag#*:}"
-    if [ -z "${v}" ]; then
-        >&2 echo "error: version not found in image tag ${image_tag}"
-        return 1
-    fi
-}
-
 title() {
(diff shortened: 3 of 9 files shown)


assets/components/lvms/topolvm-node_daemonset.yaml
```diff
@@ -78,16 +60,16 @@ spec:
           protocol: TCP
         resources:
           requests:
-            cpu: 1m
-            memory: 16Mi
+            cpu: 2m
+            memory: 35Mi
```
dhellmann, line 64: Those are still small numbers, but they're doubling. Is that absolutely necessary?

scripts/auto-rebase/lvms_assets.yaml
```diff
@@ -31,5 +32,3 @@ assets:
         ignore: "provided by MicroShift"
       - file: topolvm_default-storage-class.yaml
         ignore: "provided by MicroShift"
-      - file: topolvm-configmap_lvms-version.yaml
-        ignore: "provided by MicroShift"
```
dhellmann, line 35: This change is surprising. Why don't we need this any more?

scripts/auto-rebase/rebase-lvms.sh
```diff
@@ -181,6 +169,10 @@ download_lvms_operator_bundle_manifest(){
         local namespace="openshift-storage"
         extract_lvms_rbac_from_cluster_service_version "${PWD}" "${csv}" "${namespace}"
 
+        # Push the configMap to the kube-public namespace so that it is available to all users/apps
+        generate_version_config_map "${version}" "lvms-version" "kube-public"\
```
dhellmann, line 173: Do we get one of these per ${ARCHS} now? Are the contents the same?

scripts/auto-rebase/lvms_assets.yaml
jakobmoellerdev: this was introduced earlier and is a config map generated in the script. Its still created and maintained, but since the manifest extraction does it it should not be ignored. (otherwise the file would not be copied from the staging repository).

scripts/auto-rebase/rebase-lvms.sh
jakobmoellerdev: The contents are identical but you use one staging environment per arch (not sure when or why this was introduced), so if i want to generate files I have to do this per arch.

assets/components/lvms/topolvm-node_daemonset.yaml
jakobmoellerdev: I am removing an entire container whose work is now done in this one. Total usage is still down, but we can of course always reduce further if the requests are continously lower than 70% requests

scripts/auto-rebase/lvms_assets.yaml
jakobmoellerdev: Here it is, now generated: https://github.com/openshift/microshift/pull/2993/files#diff-645a8be2d813fd3ee4d2d90e0377d488623dd8c77948101fe58a42a0af563bf0R7

scripts/auto-rebase/rebase-lvms.sh
jakobmoellerdev: If you look at release-aarch64 / x86_64, the docker manifests are exactly the same because we build multi-arch containers, so Im not sure why one per arch is necessary

dhellmann: Maybe that's a legacy from when we were using images that were not multi-arch. @pmtk would likely know.

assets/components/lvms/topolvm-node_daemonset.yaml
dhellmann: Nice, that sounds like a good improvement overall.

jakobmoellerdev:
@copejon would be great if you could take a look today, ideally want to open up a rebase with the same functionality for 4.15 so that we finally get LVMS into the correct merge window for microshift-4.15

scripts/auto-rebase/rebase-lvms.sh
pmtk: I recall something about osbuild not handling multiarch images correctly? @ggiguash is that correct?

jakobmoellerdev: Im wondering how that fits together. LVMS has multiarch images since forever (4.12 I believe already had them), so im wondering how they wouldnt work with osbuild? Is that even relevant?

scripts/auto-rebase/rebase-lvms.sh
```diff
@@ -149,12 +139,10 @@ download_lvms_operator_bundle_manifest(){
 
     # Persist the version of the LVMS operator bundle for use in manifest steps
     local version
-    version="$(parse_version "${bundle_manifest}")"
-    dump_version "${version}"
```
pmtk, line 153: Seems this is not used anymore. Should we remove it?

scripts/auto-rebase/rebase-lvms.sh
pmtk: Wait, I may have misread things.

Previously the file was created at `"${REPOROOT}/assets/components/lvms/topolvm-configmap_lvms-version.yaml"` and now it ends up there because you removed "ignore" in lvms-assets.yaml.

I don't think we should make that change - entries without `ignore` are the ones we obtain from lvms' CSV, this is not the case for the file - we generate it during rebase, in microshift repo.

jakobmoellerdev:
@pmtk The file ends up here because it is the canonical way to generate files during rebases. Before the change it was actually run twice.

I changed this logic because it simply does not do what its supposed to do and the lvms assets.yaml was generated directly into the assets directory and was not generated in the staging directory that is usually used in the rebase process.

To replicate, simply run the rebase from current main with the command from the last lvms rebase here without adjusting the script.

We do not obtain any of the RBACs from the CSV, I generate them in the script because you have no other way. Generating the config map with a version imo is the same functionality. Generating content based on the pulled image (in this case the tag instead of the CSV file inside the bundle image).

scripts/auto-rebase/rebase-lvms.sh
pmtk: I discussed with @jakobmoellerdev what's going on here and now I know the story: version configmap was not generated correctly.

What Jakob proposed here is not strictly adhering to the flow we have in `rebase.sh` (that is: any file created during rebase, is created directly in assets, not in staging and then copied; like kustomizes per $ARCH for OLM).

However, I think proposed solution is more elegant - we don't need to use intermediate file where version gets dumped and then read during manifest update - it's generated during "download" which also extracts RBACs from CSV, then just copied with the power of `lvms_assets.yaml` and `handle_assets.py`. LGTM

jakobmoellerdev:
@copejon Ready for Dev Test / Confirmation

scripts/auto-rebase/rebase-lvms.sh
copejon: I agree. It's still reusing the existing flow, but in a less round-about way.

pmtk:
deployed using clusterbot, `lvms-version` CM and image refs in topolvm Pods look good
/lgtm

PR #3011, opened 2024-02-14
USHIFT-2372: Correct the GROUP path assignment when calling build_images.sh with -t.

jogeo:
Corrects the GROUP variable value when build_images.sh is called with the -t option.

### DIFF
--- a/test/bin/build_images.sh
+++ b/test/bin/build_images.sh
@@ -362,6 +362,15 @@ do_group() {
 
     if ${BUILD_INSTALLER} && ! ${COMPOSER_DRY_RUN}; then
         for image_installer in "${groupdir}"/*.image-installer; do
+            # If a template arg was given, only build the image installer for the
+            # matching template.
+            if [ -n "${template_arg}" ]; then
+                installer_file=$(basename "${image_installer}")
+                template_file=$(basename "${template_arg}")
+                if [ "${installer_file%.image-installer}" != "${template_file%.toml}" ]; then
+                    continue
+                fi
+            fi
             blueprint=$("${GOMPLATE}" --file "${image_installer}")
             local expected_iso_file="${VM_DISK_BASEDIR}/${blueprint}.iso"
             if [ -f "${expected_iso_file}" ]; then
@@ -623,7 +632,7 @@ while getopts "dEfg:hiIl:sSt:" opt; do
             ;;
         t)
             TEMPLATE="${OPTARG}"
-            GROUP="$(basename "$(dirname "$(realpath "${OPTARG}")")")"
+            GROUP="$(dirname "$(realpath "${OPTARG}")")"
             selCount=$((selCount+1))
             FORCE_REBUILD=true
             ;;


test/bin/build_images.sh
```diff
@@ -618,7 +618,7 @@ while getopts "dfg:hiIl:sSt:" opt; do
             ;;
         t)
             TEMPLATE="${OPTARG}"
-            GROUP="$(basename "$(dirname "$(realpath "${OPTARG}")")")"
+            GROUP="$(dirname "$(realpath "${OPTARG}")")"
```
ggiguash, line 621: Yes, group needs to point to a path, not a name.

ggiguash:
/lgmt
/hold

@jogeo , please attach this PR to a JIRA issue you're working on, or create a separate bug for tracking. 
You're welcome to unhold the PR once it's linked to the right JIRA.

dhellmann (commented):
/lgtm

jogeo:
There is a secondary issue once the group path is set correctly. When -t is passed only one template is built but the code to build the image installer loops over all the file in the directory that have the .image-installer suffix.

I'll keep the /hold until a fix for this is added as well.

test/bin/build_images.sh
```diff
@@ -361,7 +361,13 @@ do_group() {
...
 
     if ${BUILD_INSTALLER} && ! ${COMPOSER_DRY_RUN}; then
-        for image_installer in "${groupdir}"/*.image-installer; do
+        image_installer_list="${groupdir}"/*.image-installer
+        # If a template arg was given, only build the image installer for the
+        # matching template.
+        if [ -n "${template_arg}" ]; then
+            image_installer_list="${template_arg%.toml}".image-installer
+        fi
+        for image_installer in "${image_installer_list}"; do
```
ggiguash, line 370: Can we not alter the `for` loop pattern? There are edge cases on how wildcards are interpreted and expanded when passed through variables. 

As an alternative, inside the `for` loop, please just check if a `template_arg` was given and compare the file names. If names do not match, use `continue` to skip to the next loop iteration.

test/bin/build_images.sh
```diff
@@ -362,6 +362,15 @@ do_group() {
 
     if ${BUILD_INSTALLER} && ! ${COMPOSER_DRY_RUN}; then
         for image_installer in "${groupdir}"/*.image-installer; do
+            # If a template arg was given, only build the image installer for the
```
pmtk, line 365: I'm curious - can you give me a command that has wrong behavior without the fix? I think I noticed some unnecessary installer builds, but I can't recall the exact scenario.

jogeo: test/bin/build_images.sh -E -i -f -t test/image-blueprints/layer1-base/group1/rhel93.toml

pmtk: Argh, I get it now - this one doesn't show up until you fix the primary one...

PR #3012, opened 2024-02-14
USHIFT-2344: Add layer4 image-blueprints to support creating images with RPMs buillt elsewhere

jogeo:
This PR addresses [USHIFT-2344](https://issues.redhat.com//browse/USHIFT-2344) and provides a layer of blueprints intended to be used with RPMs build by ART.

At this time I do not expect the images built from the layer4 directory to be run in CI. Rather they will be used in the QE lab. Creating a layer4 directory also sets up a storage location for additional blueprint templates with variations the QE team would like to use in their lab.

### DIFF
--- a/test/bin/build_images.sh
+++ b/test/bin/build_images.sh
@@ -545,6 +545,8 @@ build_images.sh [-iIsdf] [-l layer-dir | -g group-dir] [-t template]
 
   -d      Dry run by skipping the composer start commands.
 
+  -E      Do not extract container images.
+
   -f      Force rebuilding images that already exist.
 
   -g DIR  Build only one group (cannot be used with -l or -t).
@@ -580,12 +582,15 @@ GROUP=""
 TEMPLATE=""
 FORCE_REBUILD=false
 FORCE_SOURCE=false
+EXTRACT_CONTAINER_IMAGES=true
 
 selCount=0
-while getopts "dfg:hiIl:sSt:" opt; do
+while getopts "dEfg:hiIl:sSt:" opt; do
     case "${opt}" in
         d)
             COMPOSER_DRY_RUN=true
+            ;;
+        E)  EXTRACT_CONTAINER_IMAGES=false
             ;;
         f)
             FORCE_REBUILD=true
@@ -682,10 +687,13 @@ configure_package_sources
 
 # Prepare container lists for mirroring registries.
 rm -f "${CONTAINER_LIST}"
-extract_container_images "${SOURCE_VERSION}" "${LOCAL_REPO}" "${CONTAINER_LIST}"
-extract_container_images "4.${FAKE_NEXT_MINOR_VERSION}.*" "${NEXT_REPO}" "${CONTAINER_LIST}"
-extract_container_images "4.${FAKE_YPLUS2_MINOR_VERSION}.*" "${YPLUS2_REPO}" "${CONTAINER_LIST}"
-extract_container_images "${PREVIOUS_RELEASE_VERSION}" "${PREVIOUS_RELEASE_REPO}" "${CONTAINER_LIST}"
+if ${EXTRACT_CONTAINER_IMAGES}; then
+    extract_container_images "${SOURCE_VERSION}" "${LOCAL_REPO}" "${CONTAINER_LIST}"
+    # The following images are specific to layers that use fake rpms built from source.
+    extract_container_images "4.${FAKE_NEXT_MINOR_VERSION}.*" "${NEXT_REPO}" "${CONTAINER_LIST}"
+    extract_container_images "4.${FAKE_YPLUS2_MINOR_VERSION}.*" "${YPLUS2_REPO}" "${CONTAINER_LIST}"
+    extract_container_images "${PREVIOUS_RELEASE_VERSION}" "${PREVIOUS_RELEASE_REPO}" "${CONTAINER_LIST}"
+fi
 
 trap 'osbuild_logs' EXIT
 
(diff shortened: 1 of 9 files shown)


ggiguash:
/retitle USHIFT-2344: Add layer4 image-blueprints to support creating images with RPMs buillt elsewhere

test/bin/build_images.sh
```diff
@@ -545,6 +545,10 @@ build_images.sh [-iIsdf] [-l layer-dir | -g group-dir] [-t template]
 
   -d      Dry run by skipping the composer start commands.
 
+  -e      Extract container images.
```
ggiguash, line 548: Remove this option from help

test/image-blueprints/layer4-art/group1/rhel92_microshift-art.image-installer
ggiguash, line 1: We should be able to use the existing `rhel92.iso` and `rhel93.iso` images built in layer1 and cached.
Why do we need another ISO?

test/image-blueprints/layer4-art/group1/rhel92_microshift-art.toml
ggiguash, line 1: The only difference of this toml with the other source commits is the missing `microshift-test-agent`.
We should be able to build it on top of the existing rhel92.iso image.

Did you try naming the file `rhel92-microshift-art.toml`? It would then using the existing build as a base.

test/image-blueprints/layer4-art/group1/rhel92_microshift-art.toml
```diff
@@ -0,0 +1,39 @@
+name = "rhel-92-microshift-art-base"
+description = "A RHEL 9.2 image with the RPMs built from ART."
+version = "0.0.1"
+modules = []
+groups = []
+distro = "rhel-92"
+
+[[packages]]
+name = "microshift"
+version = "{{ .Env.SOURCE_VERSION }}"
```
ggiguash, line 10: I'm not sure we should be reusing the same `SOURCE_VERSION` setting for something else. 
May I suggest using the notion of external rpms for what you're trying to implement. 

We could have a special `EXTERNAL_SOURCE_VERSION` setting that would be valid during the build only if a dedicated directory is populated with RPMs manually.

ggiguash: The build should fail if such RPMs do not exist.

test/image-blueprints/layer4-art/group1/rhel92_microshift-art.toml
ggiguash: I would not use `-art` suffix here, because practically, one can drop anything in place of those RPMs. 
What about using `-ext` or `-external` in accordance with my comment below.

test/image-blueprints/layer4-art/group1/rhel92_microshift-art.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[[packages]]
+name = "microshift-selinux"
+version = "{{ .Env.SOURCE_VERSION }}"
+
+[[packages]]
+name = "systemd-resolved"
+version = "*"
+
```
agullon, line 27: Can we add here these packages? They may be useful for troubleshooting. Also, `iproute-tc` is needed for Stress Testing.
I copy them from [blueprint_v0.0.1.toml](https://github.com/openshift/microshift/blob/main/scripts/image-builder/config/blueprint_v0.0.1.toml)
```
[[packages]]
name = "iputils"
version = "*"

[[packages]]
name = "bind-utils"
version = "*"

[[packages]]
name = "net-tools"
version = "*"

[[packages]]
name = "iotop"
version = "*"

[[packages]]
name = "strace"
version = "*"

[[packages]]
name = "iproute-tc"
version = "*"
```

test/image-blueprints/layer4-art/group1/rhel93_microshift-art.toml
```diff
@@ -0,0 +1,39 @@
...
+version = "{{ .Env.SOURCE_VERSION }}"
+
+[[packages]]
+name = "microshift-selinux"
+version = "{{ .Env.SOURCE_VERSION }}"
+
+[[packages]]
+name = "systemd-resolved"
+version = "*"
+
```
agullon, line 27: Can we add here the same extra packages [I commented in the rhel92 toml file](https://github.com/openshift/microshift/pull/3012/files#r1489492489)?

test/image-blueprints/layer4-art/group1/rhel92_microshift-art.image-installer
dhellmann: QE deploys directly from the ISO in some cases (without a kickstart telling the system to override the ostree image in the ISO with one from a web server). To do that, they need an ISO with a full image with MicroShift in it.

dhellmann (commented):
This looks mostly OK to me. There's a question about the use of `-art` in the image names and one bit of cleanup to do in the help message, but otherwise I think the choices all make sense.

test/image-blueprints/layer4-art/group1/rhel92_microshift-art.toml
jogeo: I did try using the rhel92 image from from layer1 but since it includes packages that are not provided by RHEL (i.e. microshift-test-agent) it requires building the additional RPM from source. The goal of layer4 is to use packaging from the official build tooling and avoid building locally.

dhellmann (commented):
/lgtm

PR #3017, opened 2024-02-14
USHIFT-2262: Fix log scan test to check for errors on startup and after restart

ggiguash:


pacevedom:
Retesting as sometimes the "should not find forbidden" flakes. Pods remain running when MicroShift is down, and when it comes back up they might start sending requests before the apiserver is ready, triggering some of the unwanted logs. It is a known/controlled/transient situation, but it happens.
If retests show no trouble I am good to go with the change.

test/suites/logscan/log-scan.robot
```diff
@@ -27,34 +81,12 @@ Should Not Find Cannot Patch Resource
 
 Services Should Not Timeout When Stopping
     [Documentation]    Logs should not say "Timed out waiting for services to stop"
-    [Tags]    robot:exclude
     Pattern Should Not Appear In Log Output    ${CURSOR}    Timed out waiting for services to stop
```
pmtk, line 84: We no longer have this exact string, it was changed to `MICROSHIFT STOP TIMED OUT`. Shall we change it in this PR?

PR #3041, opened 2024-02-19
OCPBUGS-24442: Add network checks on startup

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/config/network.go
+++ b/pkg/config/network.go
@@ -3,8 +3,11 @@ package config
 import (
 	"fmt"
 	"net"
+	"slices"
 
 	"github.com/apparentlymart/go-cidr/cidr"
+	"github.com/openshift/microshift/pkg/config/ovn"
+	"github.com/vishvananda/netlink"
 )
 
 type Network struct {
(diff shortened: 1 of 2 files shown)


pkg/config/network.go
```diff
@@ -58,3 +61,28 @@ func getClusterDNS(serviceCIDR string) (string, error) {
 
 	return dnsClusterIP.String(), nil
 }
+
+func (c *Config) EnsureNetworksDontOverlap() error {
```
dhellmann, line 65: This could use a more descriptive name. From this name, I can't tell which networks should not overlap with which other networks.

pkg/config/network.go
```diff
@@ -58,3 +61,28 @@ func getClusterDNS(serviceCIDR string) (string, error) {
...
+	linkList, err := netlink.LinkList()
+	if err != nil {
+		return fmt.Errorf("unable to load NIC list: %v", err)
+	}
+	for _, link := range linkList {
```
dhellmann, line 70: This loop could use some logging calls to report what it's doing, which interfaces are OK, etc.

pkg/config/network.go
```diff
@@ -58,3 +61,28 @@ func getClusterDNS(serviceCIDR string) (string, error) {
 
 	return dnsClusterIP.String(), nil
 }
+
+func (c *Config) EnsureNetworksDontOverlap() error {
+	linkList, err := netlink.LinkList()
```
dhellmann, line 66: It should be possible to test the logic of this function without spinning up different VMs with "bad" network settings. If the public interface took no arguments, as this function does, and called a separate private function that took as arguments a function to return the list of links and a function to return the routes for a given link, then a unit test could provide fake implementations of those functions.

PR #3051, opened 2024-02-21
USHIFT-2260: Add support to run Stress Testing conditions from Jenkins

agullon:

Closes Jira: https://issues.redhat.com/browse/USHIFT-2260

### DIFF
--- a/test/bin/stress_testing.sh
+++ b/test/bin/stress_testing.sh
@@ -35,7 +35,7 @@ function reboot_and_wait {
   local -r PREV_BOOT_ID=$(run cat /proc/sys/kernel/random/boot_id)
   run sudo reboot &
   sleep 10
-  TIMEOUT_LIMIT=$(( SECONDS + 300 ))
+  local -r TIMEOUT_LIMIT=$(( SECONDS + 300 ))
   while true; do
     NEW_BOOT_ID=$(run cat /proc/sys/kernel/random/boot_id)
     if [ "${NEW_BOOT_ID:-}" ] && [ "${PREV_BOOT_ID}" != "${NEW_BOOT_ID}" ]; then
--- a/test/bin/stress_testing.sh
+++ b/test/bin/stress_testing.sh
@@ -31,18 +31,30 @@ function run {
   fi
 }
 
+function reboot_and_wait {
+  PREV_BOOT_ID=$(run cat /proc/sys/kernel/random/boot_id)
+  run sudo reboot &
+  sleep 10
+  TIMEOUT_LIMIT=$(( SECONDS + 300 ))
+  while true; do
+    NEW_BOOT_ID=$(run cat /proc/sys/kernel/random/boot_id)
+    if [ "${NEW_BOOT_ID:-}" ] && [ "${PREV_BOOT_ID}" != "${NEW_BOOT_ID}" ]; then
+      break
+    fi
+    if (( SECONDS > TIMEOUT_LIMIT )); then
+      echo "ERROR: Timed out waiting for the VM to boot"
+      exit 1
+    fi
+    sleep 10
+  done
+}
+
 # install dependecies
 function install_kernel-modules-extra {
   if ! run rpm-ostree status &> /dev/null ; then
     if ! run rpm -q kernel-modules-extra &> /dev/null ; then
       run sudo dnf -y install kernel-modules-extra &> /dev/null
-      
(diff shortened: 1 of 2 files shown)


test/bin/stress_testing.sh
```diff
@@ -25,12 +25,46 @@ function run {
...
-    ssh -o LogLevel=ERROR ${SSH_PORT_PARAM:-} ${SSH_PKEY_PARAM:-} ${SSH_USER_PARAM:-}${SSH_HOSTNAME} "$@"
+    ssh -o StrictHostKeyChecking=no -o LogLevel=ERROR ${SSH_PORT_PARAM:-} ${SSH_PKEY_PARAM:-} ${SSH_USER_PARAM:-}${SSH_HOSTNAME} "$@"
   else
     "$@"
   fi
 }
 
+# install dependecies
+function install_kernel-modules-extra {
+  if ( ! run rpm-ostree status &> /dev/null); then
```
ggiguash, line 36: Nit: The () brackets are not usually used by us when checking command statuses in `if`.
Can we remove those in this script as well to adhere to the common style?

test/bin/stress_testing.sh
```diff
@@ -25,12 +25,46 @@ function run {
...
+    if ( ! run rpm -q kernel-modules-extra &> /dev/null); then
+      run sudo dnf -y install kernel-modules-extra &> /dev/null
+      
+      run sudo reboot &
+      sleep 10
+      TIMEOUT_LIMIT=$(( SECONDS + 300 ))
+      until run true || (( SECONDS > TIMEOUT_LIMIT )); do
+          sleep 10
+      done
```
ggiguash, line 45: There is a robust way to check if a host was rebooted. 
Please, see how it's implemented in the test and let's do it a in similar way here.
https://github.com/openshift/microshift/blob/main/test/resources/microshift-host.resource#L49

ggiguash (changes requested):

test/bin/stress_testing.sh
agullon: thanks, round brackets removed

test/bin/stress_testing.sh
agullon: Thanks, but I don't think I need to add a robust validation here. I just need to wait until the VM is up again.

ggiguash: The current code has a high chance of detecting a `false` reboot. The thing is that it's hard to predict how much time it will take a host to go down. While it's going down, ssh still works, so your `run true` after a 10s sleep has a high chance of succeeding on a host before the reboot.

test/bin/stress_testing.sh
```diff
@@ -25,12 +25,58 @@ function run {
...
+    ssh -o StrictHostKeyChecking=no -o LogLevel=ERROR ${SSH_PORT_PARAM:-} ${SSH_PKEY_PARAM:-} ${SSH_USER_PARAM:-}${SSH_HOSTNAME} "$@"
   else
     "$@"
   fi
 }
 
+function reboot_and_wait {
+  PREV_BOOT_ID=$(run cat /proc/sys/kernel/random/boot_id)
+  run sudo reboot &
+  sleep 10
```
ggiguash, line 37: We probably do not need this sleep anymore. 
Anyway, the host rerely shuts down in 10s after calling a reboot command.

test/bin/stress_testing.sh
```diff
@@ -25,12 +25,58 @@ function run {
...
+  TIMEOUT_LIMIT=$(( SECONDS + 300 ))
```
ggiguash, line 38:
```suggestion
  local -r TIMEOUT_LIMIT=$(( SECONDS + 300 ))
```

test/bin/stress_testing.sh
```diff
@@ -25,12 +25,58 @@ function run {
...
+  while true; do
+    NEW_BOOT_ID=$(run cat /proc/sys/kernel/random/boot_id)
```
ggiguash, line 40: A variable in the loop cannot be read-only
```suggestion
    local NEW_BOOT_ID
    NEW_BOOT_ID=$(run cat /proc/sys/kernel/random/boot_id)
```

test/bin/stress_testing.sh
agullon: I agree, removed

PR #3064, opened 2024-02-26
USHIFT-2225: Create CRI-O cfg to use Multus over ovn-k

pmtk:


packaging/crio.conf.d/12-microshift-multus.conf
```diff
@@ -0,0 +1,10 @@
+[crio.network]
+# Override default network so CRI-O waits for and calls Multus CNI instead of ovn-kubernetes.
+cni_default_network = "multus-cni-network"
+
+# Change the order, so the CNIs from the container-networking-plugins (copied to host's /opt/cni/bin)
+# are picked over CNIs from containernetworking-plugins RPM (which a dependency of CRI-O).
+plugin_dirs = [
+        "/opt/cni/bin",
+        "/usr/libexec/cni"
+]
```
ggiguash, line 10: Looking at the ovn config file, we also have `crio.runtime.workloads.management` to pin the CPU 0.
Should we duplicate it here as well so that the config is self-contained?

pmtk: It's not needed, These are overrides over what's present in the other file - we'll install both files. The lines you commented are in different order.

pmtk: I don't think this works like that - it seems to be using annotations to target specific Pods, `target.workload.openshift.io/management` in this case and that annotation is also present in manifests of csi-snapshotter, dns, node-resolver, router, service-ca, and OLM. If anything, that configuration should be in 10-microshift.conf and if other CNIs should make use of it, they should include that annotation in the manifests.

ggiguash: I see. Thank you for explaining. 
We should better refactor it in a different PR then.

PR #3071, opened 2024-02-27
USHIFT-2460: Fix Stress Testing script

agullon:

Closes #<Issue Number>

### DIFF
--- a/test/bin/stress_testing.sh
+++ b/test/bin/stress_testing.sh
@@ -32,35 +32,6 @@ function run {
 }
 
 # reboot and wait
-function wait_for() {
-    local timeout=$1
-    shift 1
-    local -r start=$(date +%s)
-    until ("$@"); do
-        sleep 1
-        echo -n " ."
-        local now
-        now=$(date +%s)
-        if [ $(( now - start )) -ge "${timeout}" ]; then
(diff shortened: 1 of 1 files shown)


test/bin/stress_testing.sh
```diff
@@ -47,13 +48,15 @@ function reboot_and_wait {
     fi
     sleep 10
   done
+  sleep 10
```
pacevedom, line 51: I guess this is allowing some time for microshift to come up after the boot. The RF tests also need to wait for this to happen, but they do it in a different way. There is a [keyword](https://github.com/openshift/microshift/blob/main/test/resources/microshift-process.resource#L26-L29) to check the `/readyz` endpoint in api server, [another](https://github.com/openshift/microshift/blob/main/test/resources/microshift-process.resource#L31-L34) one to check `/livez`, and this is [tried](https://github.com/openshift/microshift/blob/main/test/resources/microshift-process.resource#L39-L42) a few times.
This might be safer than a sleep, which could take more than 10s and then fail the oc wait command anyway.

agullon: thanks, this is a great idea. I just added a check for `/livez` endpoint.

PR #3075, opened 2024-02-28
USHIFT-2261 USHIFT-2459: Add namespace ownership support for router

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/cmd/generate-config/config/config-openapi-spec.json
+++ b/cmd/generate-config/config/config-openapi-spec.json
@@ -5,6 +5,7 @@
     "debugging",
     "dns",
     "etcd",
+    "ingress",
     "manifests",
     "network",
     "node"
@@ -66,6 +67,26 @@
           "description": "Set a memory limit on the etcd process; etcd will begin paging memory when it gets to this value. 0 means no limit.",
           "type": "integer",
           "format": "int64"
+        }
+      }
+    },
+    "ingress": {
+      "type": "object",
+      "required": [
+        "routeAdmissionPolicy"
+      ],
+      "properties": {
+        "routeAdmissionPolicy": {
+          "type": "object",
+          "required": [
+            "namespaceOwnership"
+          ],
+          "properties": {
+            "namespaceOwnership": {
+              "type": "string",
+              "default": "InterNamespaceAllowed"
+            }
+          }
         }
       }
     },
--- a/assets/components/openshift-router/deployment.yaml
+++ b/assets/components/openshift-router/deployment.yaml
@@ -42,7 +42,7 @@ spec:
             - name: ROUTER_DISABLE_HTTP2
               value: "true"
             - name: ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK
-              value: "false"
+              value: '{{.RouterNamespaceOwnership}}'
             - name: ROUTER_LOAD_BALANCE_ALGORITHM
               value: random
             - name: ROUTER_METRICS_TYPE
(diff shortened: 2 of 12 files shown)


pacevedom:
wait until https://github.com/openshift/microshift/pull/3066 is merged

dhellmann:
It looks like this PR includes the contents of #3066, is that right? If so, that's good, I'd like to land the code change and tests at the same time. Maybe we can close the other PR?

pacevedom:
> It looks like this PR includes the contents of #3066, is that right? If so, that's good, I'd like to land the code change and tests at the same time. Maybe we can close the other PR?

Yes, I was hoping for a quick review of the implementation (as its simple) and then land the tests later (which are more complex). To have them pass the tests I included all the other commits here too and planned to remove them once #3066 merged.

ShudiLi:
The function worked well: By default, the ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK in the router deployment was "true", two route with same host names but with different path were admitted and could curl the routes successfully.
Changed it to "false", one route was  admitted, the other wasn't.

1.
% oc -n openshift-ingress get deployment -oyaml | grep  -A1 ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK 
          - name: ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK
            value: "true"
2.
% oc -n test2 get route
NAME           HOST                                  ADMITTED   SERVICE        TLS
route1         test-route1.apps.example.com          True       unsec-apach2   


3.
% oc -n test get route 
NAME           HOST                                 ADMITTED   SERVICE        TLS  
route1         test-route1.apps.example.com         True       unsec-apach2   
% oc -n test get route route1  -oyaml | grep -A2 spec:
spec:
  host: test-route1.apps.example.com
  path: /srv/docs

4.  curl the two routes
sh-4.4# curl http://test-route1.apps.example.com --resolve test-route1.apps.example.com:80:10.42.0.16
this a test!
sh-4.4# curl http://test-route1.apps.example.com/srv/docs/book.txt --resolve test-route1.apps.example.com:80:10.42.0.16
11111ss
sh-4.4#

5. also checked for edge route and reencrypt route

ShudiLi:
/label qe-approved
thanks

cmd/generate-config/config/config-openapi-spec.json
```diff
@@ -69,6 +70,26 @@
         }
       }
     },
+    "ingress": {
```
jerpeter1, line 73: In your EP, you said you intended to expose this via:

```yaml
router:
  routerAdmissionPolicy:
    namespaceOwnership: <Strict|InterNamespaceAllowed> # Defaults to InterNamespaceAllowed.
```

here, you're exposing it via ingress.routerAdmissionPolicy.namespaceOwnership - if the implementation is what we want, you should probably go back and fixup the EP.

jerpeter1:
You might consider adding a test case which checks that the default (unconfigured) code path works as expected. Maybe another one for invalid configuration, but I wouldn't consider either test a blocker for merging this.

/lgtm

cmd/generate-config/config/config-openapi-spec.json
pacevedom: Good catch, will update it right away.

pacevedom:
> You might consider adding a test case which checks that the default (unconfigured) code path works as expected. Maybe another one for invalid configuration, but I wouldn't consider either test a blocker for merging this.
> 
> /lgtm

Will add them in a separate PR. Thanks!

PR #3080, opened 2024-02-28
[release-4.15] USHIFT-2467: Hardcode version values to ensure right repositories

pmtk:


pmtk:
`el92-src@rpm-install: Upgrade From Previous Version` fails because test harness only uses repositories enabled on the hypervisor to obtain latest "previous minor version". After GA of 4.15, it stopped enabling rhocp-4.14, so the most recent 4.14 `get_rel_version_repo()` detects is 4.14~rc.7.
We might enable Y and Y-1 rhocp on the hypervisor or change the logic of the `get_rel_version_repo()` so it searches also not enabled rhocp repositories.

test/suites/rpm/install-and-upgrade-successful.robot
```diff
@@ -54,9 +54,9 @@ Upgrade From Previous Version
     # Depending on where we are in the current release cycle, we might
     # need to install from an unofficial repository if the previous
     # release only has EC and RC builds.
-    IF    '${PREVIOUS_VERSION_REPO_URL}' != '${EMPTY}'
+    IF    '${PREVIOUS_VERSION_REPO}' != '${EMPTY}'
         Install MicroShift RPM Packages From Repo
-        ...    ${PREVIOUS_VERSION_REPO_URL}
+        ...    ${PREVIOUS_VERSION_REPO}
         ...    4.${PREVIOUS_MINOR_VERSION}.*
     ELSE
```
pmtk, line 61: That `ELSE` branch is no longer necessary - ${PREVIOUS_VERSION_REPO} will always have a value - sometimes it's rhocp, sometimes it's openshift mirror, sometimes it's http repo from hypervisor

test/resources/microshift-rpm.resource
```diff
@@ -27,12 +27,19 @@ Install MicroShift RPM Packages From System Repo
...
 
-Configure MicroShift Repository
+    IF    "http" in "${repo}"
+        Configure MicroShift File Repository    ${repo}
+        Install MicroShift RPM Packages From System Repo    ${version}
+        Unconfigure MicroShift File Repository
+    ELSE
+        Command Should Work    subscription-manager repos --enable ${repo}
+        Install MicroShift RPM Packages From System Repo    ${version}
+        Command Should Work    subscription-manager repos --disable ${repo}
```
pmtk, line 39: There's a risk that this would disable repo with dependencies (e.g. 4.15 at 4.16/main)

test/bin/build_images.sh
```diff
@@ -534,15 +555,26 @@ mkdir -p "${IMAGEDIR}/blueprints"
...
-configure_package_sources
-
 # Prepare container lists for mirroring registries.
 rm -f "${CONTAINER_LIST}"
 extract_container_images "${SOURCE_VERSION}" "${LOCAL_REPO}" "${CONTAINER_LIST}"
 extract_container_images "4.${FAKE_NEXT_MINOR_VERSION}.*" "${NEXT_REPO}" "${CONTAINER_LIST}"
 extract_container_images "4.${FAKE_YPLUS2_MINOR_VERSION}.*" "${YPLUS2_REPO}" "${CONTAINER_LIST}"
 extract_container_images "${PREVIOUS_RELEASE_VERSION}" "${PREVIOUS_RELEASE_REPO}" "${CONTAINER_LIST}"
 
+# Original values of CURRENT_RELEASE_REPO and PREVIOUS_RELEASE_REPO are needed for proper
```
dhellmann, line 565: This is OK for now. When we have the current issue resolved, we may want to consider using separate variables for the "real" repos vs. the pre-release content from the mirror. It means more variables, but the logic using those variables can be more explicit.

test/resources/microshift-rpm.resource
```diff
@@ -27,12 +27,26 @@ Install MicroShift RPM Packages From System Repo
...
 Install MicroShift RPM Packages From Repo
     [Documentation]    Installs MicroShift RPM packages from the specified URL
-    [Arguments]    ${repo_url}    ${version}
-    Configure MicroShift Repository    ${repo_url}
-    Install MicroShift RPM Packages From System Repo    ${version}
-    Unconfigure MicroShift Repository
+    [Arguments]    ${repo}    ${version}
 
-Configure MicroShift Repository
+    IF    "http" in "${repo}"
```
dhellmann, line 32: Quoting can be weird in RF. You might not need quotes around the variable here.

The other way to solve this might be to do more of the work to set up the repos from the scenario script, which might be easier because it's straight bash. For example, if that script logged in and enabled both repos, and then passed very detailed version info into the test, the test itself would be "install old version; verify it works; install new version; verify it works" without all of the repo manipulation.

dhellmann: If you want to take that approach, we could just disable this scenario to remove some of the time pressure to get it working.

pmtk: I think I had issues when missed the quotes - I think python didn't interpret that as a string type.

As for the idea - sounds good, but isn't it more like refactor than fix? We have all the tools already in the RF, so to me it would be a candidate for separate PR.

dhellmann: I didn't notice this was the 4.15 branch. We could also just hard-code repo names here, since they aren't going to change. Then we could work on making the main branch smarter so we don't have to keep updating those settings over time.

And yes, what I proposed would have been a refactor/rewrite, that's why I suggested temporarily disabling the scenario in 1 PR and doing the other work in another.

test/bin/build_images.sh
pmtk: PTAL at most recent force push and see if it's better

test/resources/microshift-rpm.resource
pmtk: On the second thought - that might be a better idea. I think this problem shows that we should keep the repos of the hypervisor and test harness separate, so they are explicitly enabled and used rather than having luck

test/package-sources/rhocp-y.toml
```diff
@@ -0,0 +1,10 @@
+{{- if env.Getenv "RHOCP_MINOR_Y" "" -}}
```
dhellmann, line 1: Is this saying "if RHOCP_MINOR_Y is empty" or is the second argument to Getenv the default value to use if the variable is not set?

test/scenarios/el92-src@rpm-install.sh
```diff
@@ -41,20 +41,16 @@ scenario_run_tests() {
...
-    local -r previous_version_repo_url=$(echo "${previous_version_repo}" | cut -d, -f2)
 
-    # Enable the repositories with the dependencies using the most
-    # recent GA OCP version. This value is used to build the repo
-    # name, so should include the major and minor version number.
-    local -r dependency_version="4.$("${ROOTDIR}/scripts/get-latest-rhocp-repo.sh")"
+    run_command_on_vm host1 "sudo subscription-manager repos \
+        --enable rhocp-4.14-for-rhel-9-${UNAME_M}-rpms \
+        --enable rhocp-4.15-for-rhel-9-${UNAME_M}-rpms \
+        --enable fast-datapath-for-rhel-9-${UNAME_M}-rpms"
```
dhellmann, line 48: We may want to bring this up to the main branch, too.

dhellmann (commented):
This is definitely easier to follow. Let's see what CI thinks.

test/scenarios/el92-src@rpm-install.sh
pmtk: Yes, it's easier when they hardcoded, but seems like better approach overall

test/bin/build_images.sh
```diff
@@ -519,13 +522,13 @@ SOURCE_VERSION_BASE=$(rpm -q --queryformat '%{version}' "${release_info_rpm_base
...
 current_version_repo=$(get_rel_version_repo "${MINOR_VERSION}")
 CURRENT_RELEASE_VERSION=$(echo "${current_version_repo}" | cut -d, -f1)
-CURRENT_RELEASE_REPO=$(echo "${current_version_repo}" | cut -d, -f2)
 
 previous_version_repo=$(get_rel_version_repo "${PREVIOUS_MINOR_VERSION}")
 PREVIOUS_RELEASE_VERSION=$(echo "${previous_version_repo}" | cut -d, -f1)
 PREVIOUS_RELEASE_REPO=$(echo "${previous_version_repo}" | cut -d, -f2)
 
-LATEST_RHOCP_MINOR="$("${SCRIPTDIR}/../../scripts/get-latest-rhocp-repo.sh")"
+RHOCP_MINOR_Y="15"
```
pmtk, line 530: Need to delete this

PR #3084, opened 2024-02-28
USHIFT-2222: RPM packages for Multus CNI

pmtk:


assets/optional/multus/06-daemonset.yaml
```diff
@@ -57,6 +57,7 @@ spec:
         args:
           - >
             /entrypoint/cnibincopy.sh;
+            while [ ! -f /host/etc/cni/net.d/10-ovn-kubernetes.conf ]; do sleep 2; done;
```
ggiguash, line 60: Should we delete this file before running the script then?

ggiguash, line 60: This file is created when OVN starts up. Is it only on the first startup, or everytime?
If OVN startup fails, are we going to remain in this loop "forever"?

pmtk: I think this should be created each time ovn-k daemon starts, so the contents are accurate. If that file is not created then that loop will continue indefinitely. Alternatively we could exit, but it won't help because after Pod restart it'll happen again. And we cannot proceed because multus will use wrong primary CNI (like it did with the ocp-conformance tests).
We could add more logs here to help with the debugging, wdyt?

pmtk: Done
```
2024-03-11T10:38:42+00:00 Waiting for /etc/cni/net.d/10-ovn-kubernetes.conf
2024-03-11T10:43:44+00:00 Timed out waiting for /etc/cni/net.d/10-ovn-kubernetes.conf
```

PR #3088, opened 2024-02-29
USHIFT-2286: Use timeout command for interrupting failed VM installations

ggiguash:
The fix comes to work around stuck VM installations when the ostree payload cannot be pulled. 
The `--wait <time>` option of `virt-install` does not terminate the `tclsh` process used by the `unbuffer` command.

When `timeout` is used, all the process chain of `unbuffer->sudo->virt-install` is terminated.

Note: this problem is reproducing itself on ARM platform (not on x86) as noted in the JIRA issue comments.

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -403,7 +403,6 @@ launch_vm() {
     local -r full_vmname="$(full_vm_name "${vmname}")"
     local -r kickstart_url="${WEB_SERVER_URL}/scenario-info/${SCENARIO}/vms/${vmname}/kickstart.ks"
 
-    local -r vm_wait_timeout=$(( VM_BOOT_TIMEOUT / 60 ))
     local -r vm_pool_name="${VM_POOL_BASENAME}-${SCENARIO}"
     local -r vm_pool_dir="${VM_DISK_BASEDIR}/${vm_pool_name}"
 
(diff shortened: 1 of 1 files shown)


test/bin/scenario.sh
```diff
@@ -479,12 +478,18 @@ launch_vm() {
...
+        # Note: Using the '--wait <time>' virt-install options may not work for
+        # failed installations when 'unbuffer' command is used.
+        local timeout_install="timeout -v --kill-after=1m ${VM_BOOT_TIMEOUT}s"
         # When bash creates a background job (using `&`),
         # the bg job does not get its own TTY.
         # If the TTY is not provided, virt-install refuses
         # to attach to the console. `unbuffer` provides the TTY.
         # shellcheck disable=SC2086
-        if ! unbuffer sudo virt-install \
+        if ! ${timeout_install} unbuffer sudo virt-install \
```
dhellmann, line 492: How many different places are we applying timeouts in the call stack by the time we reach here? Does the call to parallel in the phase script also have a timeout, for example?

ggiguash: We have taken a similar approach when running tests. It allows for per-test override of timeouts - so, if necessary, we can also allow per-test VM boot timeouts here in the future.

dhellmann: OK, as long as we're consistent in only applying timeouts at the "bottom" of the call stack, this change makes sense.

dhellmann (commented):
/lgtm

/hold

This looks good, but I don't want to merge a CI change on Friday evening. Feel free to remove the hold when you're back online next week.

PR #3105, opened 2024-03-04
USHIFT-2454: audit log configuration

copejon:


pkg/controllers/kube-apiserver.go
```diff
@@ -255,45 +261,19 @@ func (s *KubeAPIServer) configure(cfg *config.Config) error {
 	return nil
 }
 
-func (s *KubeAPIServer) configureAuditPolicy() error {
-	data := []byte(`
-apiVersion: audit.k8s.io/v1
-kind: Policy
-metadata:
-  name: Default
```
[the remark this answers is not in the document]
copejon: The generated default policy (source from OCP) includes a couple extra rules which are moot on MicroShift. Including the rules as their defined for an OCP cluster doesn't interfere with the kube-apiserver.

1. This one would be applied to the `apiserver.openshift.io` group, which isn't installed on MicroShift.
```yaml
  - level: None
    namespaces:
      - ""
    resources:
      - group: apiserver.openshift.io
        resources:
          - apirequestcounts
          - apirequestcounts/*
    users:
      - system:apiserver
```

2. Same for this rule. The `user` and `oauth` APIs aren't deployed on MicroShift.
```yaml
  - level: RequestResponse
    resources:
      - group: user.openshift.io
        resources:
          - identities
      - group: oauth.openshift.io
        resources:
          - oauthaccesstokens
          - oauthauthorizetokens
    verbs:
      - create
      - update
      - patch
      - delete
```

pkg/config/apiserver.go
```diff
@@ -13,7 +13,20 @@ type ApiServer struct {
...
+
 	// The URL and Port of the API server cannot be changed by the user.
 	URL  string `json:"-"`
 	Port int    `json:"-"`
 }
+
+type AuditLog struct {
+	// maxFileAge is the maximum number of days to retain old audit log files
+	MaxFileAge  int    `json:"maxFileAge"`
+	// maxFiles is the maximum number of audit log files to retain
```
pmtk, line 26:
```suggestion
	// maxFiles is the maximum number of rotated audit log files to retain
```

pkg/config/apiserver/audit.go
```diff
@@ -0,0 +1,21 @@
...
+	ac := configV1.Audit{
+		Profile:    configV1.AuditProfileType(forProfile),
+		CustomRules: nil,
+	}
+
+	ap, err := audit.GetAuditPolicy(ac)
+	if err != nil {
+		return nil, err
+	}
+	return ap.DeepCopy(), nil
```
pmtk, line 20: Is `DeepCopy` necessary? Looks like `GetAuditPolicy` already copies base policy before editing just to return it.

pkg/config/apiserver.go
```diff
@@ -13,7 +13,20 @@ type ApiServer struct {
...
+	MaxFiles    int    `json:"maxFiles"`
+	// maxFileSize is the maximum size in megabytes of the audit log file before it gets rotated
+	MaxFileSize int    `json:"maxFileSize"`
+	// profile is the OpenShift profile specifying a specific logging policy
+	Profile     string `json:"profile"`
```
pmtk, line 31: Take a look at `dns.go`, it's using special comments to provide value for the example in `howto_config.md`:
```
	// +kubebuilder:default=example.com
	// +kubebuilder:example=microshift.example.com
```

```suggestion
	// profile is the OpenShift profile specifying a specific logging policy
	// +kubebuilder:example=Default
	Profile     string `json:"profile"`
```

pkg/config/apiserver.go
```diff
@@ -13,7 +13,20 @@ type ApiServer struct {
...
+	AuditLog AuditLog `json:"auditLog"`
+
 	// The URL and Port of the API server cannot be changed by the user.
 	URL  string `json:"-"`
 	Port int    `json:"-"`
 }
+
+type AuditLog struct {
+	// maxFileAge is the maximum number of days to retain old audit log files
+	MaxFileAge  int    `json:"maxFileAge"`
```
pmtk, line 25:
```suggestion
	// maxFileAge is the maximum number of days to retain old audit log files
    // +kubebuilder:example=10
	MaxFileAge  int    `json:"maxFileAge"`
```

pkg/config/apiserver.go
```diff
@@ -13,7 +13,20 @@ type ApiServer struct {
...
+	// maxFiles is the maximum number of audit log files to retain
+	MaxFiles    int    `json:"maxFiles"`
+	// maxFileSize is the maximum size in megabytes of the audit log file before it gets rotated
+	MaxFileSize int    `json:"maxFileSize"`
```
pmtk, line 29:
```suggestion
	// maxFileSize is the maximum size in megabytes of the audit log file before it gets rotated
    // +kubebuilder:example=200
	MaxFileSize int    `json:"maxFileSize"`
```

pkg/controllers/kube-apiserver.go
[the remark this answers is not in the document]
pmtk: So the new, generated Policy is superset (with the same values) of this hardcoded Police we're deleting?

pkg/config/apiserver/audit.go
copejon: The way I understood the comment on audit.GetAuditPolicy, the prudent thing to do is deepcopy the returned value:

`// Note: the returned policy must not be modifed by the caller prior to a deepcopy.`

pkg/config/apiserver.go
copejon: Ah, thanks!

pkg/controllers/kube-apiserver.go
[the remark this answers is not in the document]
copejon: Right. The policies are sourced from embedded manifests in the vendored package.  It's somewhat simple in that it has a `base` manifest to which it appends different rules depending on the profile.

pkg/config/apiserver.go
```diff
@@ -13,7 +13,23 @@ type ApiServer struct {
...
 	SkipInterface bool `json:"-"`
 
+	AuditLog AuditLog `json:"auditLog"`
+
 	// The URL and Port of the API server cannot be changed by the user.
 	URL  string `json:"-"`
 	Port int    `json:"-"`
 }
+
+type AuditLog struct {
```
pmtk, line 23: I would expect adding this would update the default config (packaging/microshift/config.yaml) and configs in docs (docs/user/howto_config.md)

vendor/github.com/openshift/library-go/pkg/crypto/crypto.go
```diff
@@ -371,7 +371,7 @@ func (c *TLSCertificateConfig) GetPEMBytes() ([]byte, []byte, error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	keyBytes, err := EncodeKey(c.Key)
+	keyBytes, err := encodeKey(c.Key)
```
pmtk, line 374: How this change happened without updating go.mod?

test/suites/standard1/configuration.robot
```diff
@@ -70,3 +108,22 @@ Setup With Debug Log Level
...
+    ${merged}=    Extend MicroShift Config    ${BAD_AUDIT_PROFILE}
+    Upload MicroShift Config    ${merged}
+    Stop MicroShift
+    Start MicroShift
+
+Setup Known Audit Log Profile
+    [Documentation]    Setup audit
+    ${merged}=    Extend MicroShift Config    ${AUDIT_PROFILE}
+    Upload MicroShift Config    ${merged}
+    Restart MicroShift
```
pmtk, line 123: I'm wondering if we should refresh the ${CURSOR} before restarting so that we look at more specific log span? Otherwise it seems to use CURSOR set up at the beginning of this suite

test/suites/standard1/configuration.robot
```diff
@@ -70,3 +108,22 @@ Setup With Debug Log Level
     ${merged}=    Extend MicroShift Config    ${BAD_LOG_LEVEL}
     Upload MicroShift Config    ${merged}
     Restart MicroShift
+
+Setup With Unknown Audit Log Profile
+    [Documentation]    Set audit log profile to unknown value
+    ${merged}=    Extend MicroShift Config    ${BAD_AUDIT_PROFILE}
+    Upload MicroShift Config    ${merged}
+    Stop MicroShift
+    Start MicroShift
```
pmtk, line 117: Any reason here it's Stop + Start but in other places it's Restart?

vendor/github.com/openshift/library-go/pkg/crypto/crypto.go
copejon: good question, let me see

test/suites/standard1/configuration.robot
copejon: This was a pain to realize, but the Restart MicroShift keyword fails the test if MicroShift doesn't restart, which is the expected outcome here.

test/suites/standard1/configuration.robot
```diff
@@ -35,6 +49,30 @@ Debug Log Level Produces No Warning
     Setup With Debug Log Level
     Pattern Should Not Appear In Log Output    ${CURSOR}    Unrecognized log level "debug", defaulting to "Normal"
 
+Unknown Audit Log Profile Should Fail
```
pacevedom, line 52: Is this covered with https://github.com/openshift/microshift/pull/3105/files#diff-8f642c21bb21f507564f8c19ec55c9c83fa24f6578fd9dd26dcf12bf49aabee0R27 ?
Saying it because of time savings for including some e2e tests as unit tests, if possible.

pkg/controllers/kube-apiserver.go
```diff
@@ -255,45 +261,19 @@ func (s *KubeAPIServer) configure(cfg *config.Config) error {
...
-  - "/readyz"
-# A catch-all rule to log all other requests at the Metadata level.
-- level: Metadata
-  # Long-running requests like watches that fall under this rule will not
-  # generate an audit event in RequestReceived.
-  omitStages:
-  - "RequestReceived"`)
-
+func (s *KubeAPIServer) configureAuditPolicy(cfg *config.Config) error {
+	p, err := apiserver.GetPolicy(cfg.ApiServer.AuditLog.Profile)
```
pacevedom, line 265: Would it be possible to validate this in the config package?

pkg/config/config.go
```diff
@@ -84,6 +84,7 @@ func (c *Config) fillDefaults() error {
 		URL:             "https://localhost:6443",
 		Port:            6443,
 	}
+	c.ApiServer.AuditLog.Profile = "Default"
```
pacevedom, line 87: Reference https://github.com/openshift/api/blob/master/config/v1/types_apiserver.go#L81 ?

copejon: Ah, good catch. It does effectively duplicate the test. I'll remove the e2e test and leave the unit.

pkg/controllers/kube-apiserver.go
copejon: Yes, but I'm doubtful of the value there. Is the idea to catch breaking changes to the OpenShift policies?

docs/user/howto_config.md
```diff
@@ -48,6 +53,11 @@ In case `config.yaml` is not provided, the following default settings will be us
 ```yaml
 apiServer:
     advertiseAddress: ""
+    auditLog:
+        maxFileAge: 0
+        maxFileSize: 0
+        maxFiles: 0
+        profile: ""
```
pmtk, line 60: This doesn't look like the default values we use. Have you tried with `// +kubebuilder:default=VALUE`?

pmtk:
One q about the generated default config in the docs, but otherwise looks good to me

vendor/github.com/openshift/library-go/pkg/operator/resource/resourceread/config.go
```diff
@@ -1,58 +0,0 @@
-package resourceread
```
pmtk, line 1: Why this file got deleted? There are no changes to go.mod and I couldn't find any refs in this PR to `{Read,Write}FeatureGate*` functions, so if we don't use these functions why this file was not cleaned rebase or other changes?

pkg/config/apiserver/testdata/AllRequestBodies
```diff
@@ -0,0 +1,105 @@
+typemeta:
```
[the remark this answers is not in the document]
copejon: You're right, this is an artifact of early unit test iterations. Will remove.

vendor/github.com/openshift/library-go/pkg/operator/resource/resourceread/config.go
copejon: This is indeed weird. A `make vendor` restored the file.

pkg/config/config.go
pacevedom: I think this is the answer to the unit test vs e2e test comment? :)

pkg/controllers/kube-apiserver.go
pacevedom: You would get the validation also in `show-config` command, so that any invalid values also signal the error and you dont need to wait until starting microshift.

copejon: Ah my bad, I see what you mean (thought you were referring to testing).

pmtk: I think this is not addressed yet? Otherwise the PR looks pretty good

pacevedom (commented):
This is lgtm for me, waiting on others reviews. When done I will label it if its not already there.

PR #3108, opened 2024-03-05
USHIFT-2431: Add junit integration to CNCF tests

ggiguash:


test/scenarios-periodics/el92-src@cncf-conformance.sh
```diff
@@ -106,6 +162,10 @@ scenario_remove_vms() {
 }
 
 scenario_run_tests() {
-    prepare_hosts
-    run_sonobuoy
-}
\ No newline at end of file
+    start_junit
+    trap "close_junit" EXIT
```
dhellmann, line 166: The framework creates and closes the file, so you don't want these 2 lines. Calling start_junit again will overwrite any existing content and having the trap will cause the file to be closed twice so the syntax is not correct.

test/scenarios-periodics/el92-src@cncf-conformance.sh
```diff
@@ -14,83 +14,139 @@ prepare_hosts() {
...
+    record_junit "run_sonobuoy" "wait_e2e_running" "OK"
 
     # Use 1h timeout. A normal run on 2 CPUs takes 40-45min.
     start=$(date +%s)
     while [ "$(~/go/bin/sonobuoy status --json | jq -r '.status')" = "running" ] ; do
         now=$(date +%s)
         if [ $(( now - start )) -ge ${TIMEOUT_TEST} ]; then
+            rc=1
             echo "Tests running for 1h. Timing out"
+            record_junit "run_sonobuoy" "wait_e2e_finished" "FAILED"
```
dhellmann, line 104: Does the full status output include any information useful to debug the tests? Should we copy it to the artifacts directory?

ggiguash: Good idea. Will do.

test/scenarios-periodics/el92-src@cncf-conformance.sh
ggiguash: @dhellmann , it looks like the tests do not work without this change. 
The reason is because the junit is initialized in `run_tests` function, which is never called for CNCF. 
I have to return the code back

dhellmann: This function is the plugin interface used by `run_tests`. If `run_tests` is not called, then these lines wouldn't do anything, right?

ggiguash: Looking at `scenario.sh`, running tests is a separate action and it calls `action_run` [function](https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L911). This function is different from `action_create` because it does not directly call `start_junit`.

The latter is only called inside `run_tests` function (RF invocation), which is never used in the CNCF test because it's not written using RF.

One possible solution to this problem would be to move the following block to `action_run` instead of `run_tests`.
```
    start_junit
    trap "close_junit; sos_report" EXIT
```

Then, it will always be called regardless of what's executed in the scenario run.

ggiguash: I uploaded this implementation - let me know what you think.

dhellmann:
/lgtm

I like the change to move the junit setup into the action function. It's much more consistent.

PR #3110, opened 2024-03-05
USHIFT-2464: scan FIPS on container images

eslutsky:


test/suites/fips/validate-fips.robot
```diff
@@ -32,6 +32,9 @@ Verify Node RPMs FIPS Compliant
     [Documentation]    Performs a FIPS validation against the Installed RPMs
     Check Payload Tool Must Pass
 
+Verify Payload FIPS Compliant
```
ggiguash, line 35: Nit: would using `Container Images` instead of `Payload` be more explicit?
The latter can be anything, so it was a bit confusing to read the code,

test/suites/fips/validate-fips.robot
```diff
@@ -86,3 +106,13 @@ Fips Should Be Enabled
     ...    sudo=True    return_rc=True    return_stdout=True    return_stderr=True
     Should Be Equal As Integers    0    ${rc}
     Should Match    ${stdout}    FIPS mode is enabled.
+
+Get Images From Release
```
ggiguash, line 110:
```suggestion
Get Container Images From Release File
```

ggiguash:
FIPS job runs in periodic tests.

/test metal-periodic-test
/test metal-periodic-test-arm

test/suites/fips/validate-fips.robot
```diff
@@ -68,6 +74,22 @@ Check Payload Tool Must Pass
     ...    sudo=True    return_rc=True    return_stdout=False    return_stderr=False
     Should Be Equal As Integers    0    ${rc}
 
+Check Container Images In Release Must Pass
+    [Documentation]    Run check-paylod Tool for Release containers
+    ${podman_pull_secret}=    Set Variable    /root/.config/containers/auth.json
+    ${podman_args}=    Set Variable
+    ...    --authfile ${PULL_SECRET_PATH} --rmi --privileged -v ${PULL_SECRET_PATH}:${podman_pull_secret}
```
ggiguash, line 81: Should we use `--rm` here to clean up the container per-each run?

eslutsky: yes, each iteration check-payload is pulling the image into its `/var/lib/containers/storage/` , so we need to remove the `anonymous unnamed volume`  with `-rm`

if we have those images embedded  we can try mounting  them into `/var/lib/containers/storage/`, so it could  save some downloading time for each image..

PR #3116, opened 2024-03-06
USHIFT-2543: add overrides for CI scripts

dhellmann:
Add the ability to override settings used by the CI scripts for local
developer workflows.

/assign @jogeo

### DIFF
--- a/test/bin/common.sh
+++ b/test/bin/common.sh
@@ -207,3 +207,11 @@ SCENARIO_BUILD_TAG_PREV="$(date -d "yesterday" '+%y%m%d')"
 # The location of the awscli binary.
 # shellcheck disable=SC2034  # used elsewhere
 AWSCLI="${OUTPUTDIR}/bin/aws"
+
+# The location of developer overrides files.
+DEV_OVERRIDES="${TESTDIR}/dev_overrides.sh"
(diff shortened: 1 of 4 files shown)


test/bin/common.sh
```diff
@@ -207,3 +207,11 @@ SCENARIO_BUILD_TAG_PREV="$(date -d "yesterday" '+%y%m%d')"
 # The location of the awscli binary.
 # shellcheck disable=SC2034  # used elsewhere
 AWSCLI="${OUTPUTDIR}/bin/aws"
+
+# The location of developer overrides files.
+DEV_OVERRIDES="${TESTDIR}/dev_overrides.sh"
```
ggiguash, line 212: Should we keep this file in the _output directory instead?

Otherwise, we have similar scripts in the `test` directory, i.e. scenario_settings.sh.example and variables.yaml.example. Should we follow that pattern if we want to keep the file in the test directory?

dhellmann: This location follows the pattern set by `scenario_settings.sh`, which is a similar file meant only for the variables in `scenario.sh`.

ggiguash: OK. In this case, should we follow the existing pattern and create `dev_overrides.sh.example` file to be checked into the repo with comments / examples of what it's for?

PR #3125, opened 2024-03-08
USHIFT-2550: Change configure-vm.sh to not install optional RPMs by default

pmtk:
Adds a --optional-rpms flag to install all RPMs.

### DIFF
--- a/scripts/devenv-builder/configure-vm.sh
+++ b/scripts/devenv-builder/configure-vm.sh
@@ -10,6 +10,7 @@ RHEL_BETA_VERSION=false
 SET_RHEL_RELEASE=true
 DNF_UPDATE=true
 PULL_IMAGES=false
+OPTIONAL_RPMS=false
 DNF_RETRY="${SCRIPTDIR}/../dnf_retry.sh"
 PULL_RETRY="${SCRIPTDIR}/../pull_retry.sh"
 RHOCP_REPO="${SCRIPTDIR}/../get-latest-rhocp-repo.sh"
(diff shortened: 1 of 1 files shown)


scripts/devenv-builder/configure-vm.sh
```diff
@@ -216,7 +222,11 @@ else
 fi
 
 if ${BUILD_AND_RUN}; then
-    "${DNF_RETRY}" "localinstall" "$(ls -1 ~/microshift/_output/rpmbuild/RPMS/*/*.rpm)"
+    rpms="$(ls -1 ~/microshift/_output/rpmbuild/RPMS/*/*.rpm)"
+    if ! ${OPTIONAL_RPMS}; then
+        rpms="$(echo "${rpms}" | grep -E 'microshift-(4|greenboot|selinux|networking|release-info)')"
+    fi
+    "${DNF_RETRY}" "localinstall" "${rpms}"
```
dhellmann, line 229: Could we just install the "microshift" RPM and let its dependencies bring in the others? Or do we have to specify them all on the command line because we're installing from local files?

pmtk: It's the latter - if we'd do `dnf install ./some/path/microshift-4.16.rpm` then it wouldn't know about other RPMs unless we'd put them into a repository

dhellmann: Makes sense. It would be cool if we could derive this info from the RPM or its spec file instead of hard-coding it. For example, if the RPM is built already, this command gives you a list of the dependencies:

```
$ rpm -q _output/rpmbuild/RPMS/aarch64/microshift-4.16.0_0.nightly_arm64_2024_01_15_061927_20240124205332_e27825c8b-1.el9.aarch64.rpm --requires
```

and that can be filtered to the packages we build

```
$ rpm -q _output/rpmbuild/RPMS/aarch64/microshift-4.16.0_0.nightly_arm64_2024_01_15_061927_20240124205332_e27825c8b-1.el9.aarch64.rpm --requires | grep '^microshift' | cut -f1 -d=
microshift-greenboot
microshift-networking
microshift-selinux
```

ggiguash: I think the best way may be to let `dnf` calculate the dependencies itself. 

```
createrepo _output/rpmbuild
sudo dnf install -y _output/rpmbuild/*/*/microshift-4.*.rpm --repofrompath=microshift-local,./_output/rpmbuild/ 
```

PR #3128, opened 2024-03-08
USHIFT-2555: detect existing VM in scenario.sh create

dhellmann:
virt-install reports errors, but with the retry loop it can take a
while before the script exits when the VM already exists. Add a check
for that case and return early. Treat it as not an error, since the
instruction is to ensure the VM exists and it does.

/assign @ggiguash


dhellmann:
/override ci/prow-metal-periodic-test

Overriding test for non-production code change.

PR #3137, opened 2024-03-11
USHIFT-2226: Test Multus - Bridge CNI

pmtk:


dhellmann (commented):
This looks good. One suggestion about a potential race condition, but I see you're trying to work around that so I'll leave it up to you and other reviewers to decide how risky it is.

test/suites/optional/multus.robot
```diff
@@ -0,0 +1,139 @@
...
+    Curl Pod Over Bridge Interface    ${POD_NAME}    ${POD_NS}    ${BRIDGE_INTERFACE}
+
+    [Teardown]    Remove NAD And Pod
+
+No Pre-Existing Bridge Interface
+    [Documentation]    Test verifies if Bridge CNI plugin will
+    ...    work correctly if there is no pre-existing bridge
+    ...    interface - it needs to be created.
+    [Setup]    Run Keywords
+    ...    Make Sure Interface Does Not Exist    ${BRIDGE_INTERFACE}
```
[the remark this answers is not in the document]
pmtk: Yes, though it'll need two (or single templated) NetworkAttachmentDefinition and Pod manifests.
We'll have similar problem with ipvlan / macvlan tests, because:
> A single master interface can not be enslaved by both macvlan and ipvlan. ([src](https://www.cni.dev/plugins/current/main/ipvlan/#notes))

Perhaps we want to add more interface to the VM to avoid that?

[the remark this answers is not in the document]
dhellmann: Would we need more than 2? The bridge test doesn't need a new one, right?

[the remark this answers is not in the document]
pmtk: No, bridge plugin doesn't have this exclusivity rule. Other solution is to (probably) reboot the host after the test, so the other *vlan can enslave main interface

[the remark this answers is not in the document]
dhellmann: Which approach seems like it will be most reliable?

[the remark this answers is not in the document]
pmtk: I don't have data on that, but adding two nics will be faster (if we keep random order of the tests, then we'll have two reboots).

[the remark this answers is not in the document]
dhellmann: Adding the extra NICs sounds like a good place to start, then.

test/suites/optional/multus.robot
```diff
@@ -0,0 +1,151 @@
...
+${PREEXISTING_BRIDGE_POD_NAME}      test-bridge-preexisting
+${PREEXISTING_BRIDGE_IP}            10.10.1.10/24
+
+${NS}                               default
+
+
+*** Test Cases ***
+Pre-Existing Bridge Interface
+    [Documentation]    Test verifies if Bridge CNI plugin will work correctly with pre-existing interface.
+    [Setup]    Run Keywords
```
dhellmann, line 30: You don't need to change it, but I'm curious why there are so many individual keywords in the setup for this case. Why not include them in the test body, or write a separate keyword?

dhellmann (commented):
You've addressed my comments, so I'm happy with this version. I had one philosophical question inline, but it shouldn't block merging this.

You have listed other reviewers, so I'll wait to lgtm to give them a chance to review it.

test/suites/optional/multus.robot
pmtk: > Why not include them in the test body

I was being literal - test setup in `[Setup]`

> or write a separate keyword?

I didn't like how this requires a jump to different location in the file - this way everything is single place.

So I think I just felt like it :D It's not perfect though, this keyword (in this or future PR) is maxing out the length, but separate keyword was "context jump" and writing general function for both Bridge tests was quite ugly and required some mental processing on what happens with which arguments.

dhellmann: It can definitely be hard to find a balance between inline and separate code blocks. That max keyword aspect to the linter is a nudge towards what the linter author thought we should be doing, but there have definitely been cases where we overrode the rule. Maybe we should have written more of those cases in Python? I don't know. As I said, there's no need to change anything here, I was just curious about the evolution of the code.

pmtk: > Maybe we should have written more of those cases in Python?

Perhaps, but python function don't support so great support for log.html. We'll see in the future

s1061123:
Regarding of CNI config in net-attach-def, it looks good to me.

pliurh:
LGTM

test/assets/multus/bridge-preexisting-pod.yaml
```diff
@@ -0,0 +1,28 @@
...
+    k8s.v1.cni.cncf.io/networks: bridge-preexisting-conf
+  labels:
+    app: test-bridge-preexisting
+spec:
+  terminationGracePeriodSeconds: 0
+  containers:
+  - name: hello-microshift
+    image: quay.io/microshift/busybox:1.36
+    command: ["/bin/sh"]
+    args: ["-c", "while true; do echo -ne \"HTTP/1.0 200 OK\r\nContent-Length: 16\r\n\r\nHello MicroShift\" | nc -l -p 8080 ; done"]
```
ggiguash, line 15: Do we want to add "sleep" here not to burn CPU?

test/assets/multus/bridge-pod.yaml
```diff
@@ -0,0 +1,28 @@
...
+    k8s.v1.cni.cncf.io/networks: bridge-conf
+  labels:
+    app: test-bridge
+spec:
+  terminationGracePeriodSeconds: 0
+  containers:
+  - name: hello-microshift
+    image: quay.io/microshift/busybox:1.36
+    command: ["/bin/sh"]
+    args: ["-c", "while true; do echo -ne \"HTTP/1.0 200 OK\r\nContent-Length: 16\r\n\r\nHello MicroShift\" | nc -l -p 8080 ; done"]
```
ggiguash, line 15: Do we want to add "sleep" here not to burn CPU?

test/suites/optional/multus.robot
```diff
@@ -0,0 +1,151 @@
...
+${BRIDGE_POD_NAME}                  test-bridge
+${BRIDGE_IP}                        10.10.0.10/24
+
+${PREEXISTING_BRIDGE_INTERFACE}     br-preexisting
+${PREEXISTING_BRIDGE_NAD_YAML}      ./assets/multus/bridge-preexisting-nad.yaml
+${PREEXISTING_BRIDGE_POD_YAML}      ./assets/multus/bridge-preexisting-pod.yaml
+${PREEXISTING_BRIDGE_POD_NAME}      test-bridge-preexisting
+${PREEXISTING_BRIDGE_IP}            10.10.1.10/24
+
+${NS}                               default
```
ggiguash, line 24: Is there a special reason we're not using a dedicated namespace for this test?

test/suites/optional/multus.robot
```diff
@@ -0,0 +1,151 @@
...
+Cleanup Bridge Test
+    [Documentation]    Removes provided NetworkAttachmentDefinition, Pod and network interface to allow for test rerun.
+    [Arguments]    ${nad}    ${pod}    ${if}
+    Run Keyword And Continue On Failure
+    ...    Oc Delete    -f ${pod}
+    Run Keyword And Continue On Failure
+    ...    Oc Delete    -f ${nad}
+    Command Should Work    ip link delete ${if}
+
+Curl Pod Over Local Interface
```
ggiguash, line 100: Nit: Curl is an implementation detail. Should we call it `Connect Pod` or something?

test/suites/optional/multus.robot
```diff
@@ -0,0 +1,151 @@
...
+    ${stdout}=    Command Should Work    curl -v --interface ${if} ${extra_ip}:8080
+    Should Contain    ${stdout}    Hello MicroShift
+
+Interface Should Not Exist
+    [Documentation]    Verifies that network interface does not exist.
+    [Arguments]    ${if}
+    ${stdout}    ${rc}=    Execute Command
+    ...    ip link show ${if}
+    ...    sudo=True    return_rc=True
+    Should Be Equal As Integers    ${rc}    1
```
ggiguash, line 116: Feels as if we would benefit from `Command Should Not Work` keyword.

test/assets/multus/bridge-preexisting-pod.yaml
pmtk: This doesn't spin furiously - `nc` blocks until a client arrives, so the loop has a many iterations as many GETs.

test/assets/multus/bridge-pod.yaml
pmtk: This doesn't spin furiously - `nc` blocks until a client arrives, so the loop has a many iterations as many GETs.

test/suites/optional/multus.robot
pmtk: Not really, let me have a look

test/suites/optional/multus.robot
pmtk: `Connect To Pod Over Local Interface`? Because in future PRs there's `Curl Pod From The Hypervisor` so we need to be more specific

test/suites/optional/multus.robot
pmtk: Done

test/suites/optional/multus.robot
pmtk: done

PR #3142, opened 2024-03-12
OCPBUGS-30833: Added a check for loadBalancerClass for built-in load balancing service.

### LINKED ISSUE #3136: [BUG] Metallb conflict with built-in load balancing service
#### What happened?
After deploying metallb in microshift, it constantly conflicts with the built-in microshift load balancer service. This produces an extreme amount of error logs, even though it does appear that metallb does "win" the fight.

Microshift should not watch services that have the loadBalancerClass set to "metallb", presumably there should be some class that microshift does listen for, along with blank services.

1. Deploy Microshift 4.13
2. Deploy metallb via helm: https://metallb.universe.tf/installation/
3. Install a service that uses a loadBalancer set with loadBalancerClass as "metallb". An example would be ingress-nginx with the following helm chart values:
```
          controller:
            name: controller
            service:
              annotations:
                metallb.universe.tf/allow-shared-ip: "share-10.110.1.228"
              loadBalancerIP: 10.110.1.228
              loadBalancerClass: metallb
```

#### Environment
- MicroShift version 4.13.21
- Hardware configuration: aarch64 node in oracle cloud
```
VERSION="9.3 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.3"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.3 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.3
REDHAT_SUPPORT_PRODUCT_VERSION="9.3"
```
- Kernel (e.g. `uname -a`): `Linux ocp-micro 5.14.0-362.8.1.el9_3.aarch64 #1 SMP PREEMPT_DYNAMIC Tue Oct 3 11:57:53 EDT 2023 aarch64 aarch64 aarch64 GNU/Linux`
- Others: 

#### Relevant logs 
Microshift Logs:
```
Mar 11 09:07:11 ocp-micro microshift[2470]: microshift-loadbalancer-service-controller I0311 09:07:11.285555    2470 controller.go:154] Process service ingress-nginx/ingress-nginx-controller
```
Metallb logs:
```
{"caller":"service_controller.go:60","controller":"ServiceReconciler","level":"info","start reconcile":"ingress-nginx/ingress-nginx-controller","ts":"2024-03-11T12:14:16Z"}
{"caller":"service.go:98","error":"[\"10.110.1.5\"] is not allowed in config","event":"clearAssignment","level":"info","msg":"current IP not allowed by config, clearing","ts":"2024-03-11T12:14:16Z"}
{"caller":"service.go:122","event":"clearAssignment","level":"info","msg":"user requested a different IP than the one currently assigned","reason":"differentIPRequested","ts":"2024-03-11T12:14:16Z"}
{"caller":"service.go:139","event":"ipAllocated","ip":["10.110.1.228"],"level":"info","msg":"IP address assigned by controller","ts":"2024-03-11T12:14:16Z"}
{"caller":"main.go:101","event":"serviceUpdated","level":"info","msg":"updated service object","ts":"2024-03-11T12:14:16Z"}
```

jdhirst:
Fixes #3136

**Which issue(s) this PR addresses**: #3136 

Closes #3136

This is my idea for an option to implement the checking of loadBalancerClass in the internal load balancing service. It would avoid conflicts with metallb or other load balancers being used. 

I specified "microshift-internal" as the name of the internal balancer service, but that could be anything really, was just the first thing that I thought of. 

Let me know what you think 😄

### DIFF
--- a/pkg/loadbalancerservice/controller.go
+++ b/pkg/loadbalancerservice/controller.go
@@ -154,7 +154,7 @@ func (c *LoadbalancerServiceController) updateServiceStatus(key string) error {
 		klog.Infof("Service %s does not exist anymore", key)
 	} else {
 		svc := obj.(*corev1.Service)
-		if svc.Spec.Type != corev1.ServiceTypeLoadBalancer || (svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass != "microshift-internal") {
+		if svc.Spec.Type != corev1.ServiceTypeLoadBalancer || svc.Spec.LoadBalancerClass != nil {
 			return nil
 		}
 		klog.Infof("Process service %s/%s", svc.Namespace, svc.Name)
--- a/pkg/loadbalancerservice/controller.go
+++ b/pkg/loadbalancerservice/controller.go
@@ -154,7 +154,7 @@ func (c *LoadbalancerServiceController) updateServiceStatus(key string) error {
 		klog.Infof("Service %s does not exist anymore", key)
 	} else {
 		svc := obj.(*corev1.Service)
-		if svc.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		if svc.Spec.Type != corev1.ServiceTypeLoadBalancer || (svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass != "microshift-internal") {
 			return nil
 		}
 		klog.Infof("Process service %s/%s", svc.Namespace, svc.Name)


pkg/loadbalancerservice/controller.go
```diff
@@ -154,7 +154,7 @@ func (c *LoadbalancerServiceController) updateServiceStatus(key string) error {
 		klog.Infof("Service %s does not exist anymore", key)
 	} else {
 		svc := obj.(*corev1.Service)
-		if svc.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		if svc.Spec.Type != corev1.ServiceTypeLoadBalancer || (svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass != "microshift-internal") {
```
pacevedom, line 157: While the controller is too open now (thanks for catching that!), this might be a bit too restrictive. This controller acts not just on MicroShift-owned resources, but also on any `LoadBalancer` `service` owned by applications.
By requiring the `loadBalancerClass` to match an expected value, MicroShift would be imposing that need to already deployed services and they would fail to pick up IPs again. This would happen in an upgrade, for example. Also, requiring this value in application's manifests would break compatibility with OpenShift and would require specific configuration for apps to run on MicroShift.
Instead of checking a specific value of the `loadBalancerClass`, I would check for emptiness instead.
From the [docs](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class):
```
If you specify .spec.loadBalancerClass, it is assumed that a load balancer implementation that matches the specified class is watching for Services. Any default load balancer implementation (for example, the one provided by the cloud provider) will ignore Services that have this field set.
```

I would drop the second part of the new check and that should work in all cases.

pacevedom (commented):
Thanks for the PR! I left one small comment.

pkg/loadbalancerservice/controller.go
jdhirst: Thanks for reviewing my request! Yep, that makes perfect sense! I updated the PR with only the check for nil now 😄

dhellmann:
/retitle OCPBUGS-30833: Added a check for loadBalancerClass for built-in load balancing service.

pacevedom (commented):
Verified to be working by QE.
/approve
/lgtm

asood-rh:
Details of the pre-merge testing can be found in JIRA ticket for the bug.

PR #3143, opened 2024-03-12
USHIFT-2087: Add configurable router ports

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -177,8 +177,9 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		return err
 	}
 	extraParams := assets.RenderParams{
-		"RouterHttpPort":  cfg.Ingress.Ports.Http,
-		"RouterHttpsPort": cfg.Ingress.Ports.Https,
+		"RouterNamespaceOwnership": cfg.Ingress.AdmissionPolicy.NamespaceOwnership == config.NamespaceOwnershipAllowed,
+		"RouterHttpPort":           *cfg.Ingress.Ports.Http,
+		"RouterHttpsPort":          *cfg.Ingress.Ports.Https,
 	}
 	if err := assets.ApplyServices(ctx, svc, renderTemplate, renderParamsFromConfig(cfg, extraParams), kubeconfigPath); err != nil {
 		klog.Warningf("Failed to apply service %v %v", svc, err)
@@ -197,11 +198,6 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		return err
 	}
 
-	extraParams = assets.RenderParams{
-		"RouterNamespaceOwnership": cfg.Ingress.AdmissionPolicy.NamespaceOwnership == config.NamespaceOwnershipAllowed,
-		"RouterHttpPort":           cfg.Ingress.Ports.Http,
-		"RouterHttpsPort":          cfg.Ingress.Ports.Https,
-	}
 	if err := assets.ApplyDeployments(ctx, apps, renderTemplate, renderParamsFromConfig(cfg, extraParams), kubeconfigPath); err != nil {
 		klog.Warningf("Failed to apply apps %v: %v", apps, err)
 		return err
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -176,7 +176,11 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		klog.Warningf("Failed to apply configMap %v: %v", cm, err)
 		return err
 	}
-	if err := assets.ApplyServices(ctx, svc, nil, nil, kubeconfigPath); err != nil {
+	extraParams := assets.RenderParams{
+		"RouterHttpPort":  cfg.Ingress.Ports.Http,
+		"RouterHttpsPort": cfg.Ingress.Ports.Https,
+	}
+	if err := assets.ApplyServices(ctx, svc, renderTemplate, renderParamsFromConfig(cfg, extraParams), kubeconfigPath); err != nil {
 		klog.Warningf("Failed to apply service %v %v", svc, err)
 		return err
 	}
@@ -193,7 +197,7 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		return err
 	}
 
-	extraParams := assets.RenderParams{
+	extraParams = assets.RenderParams{
 		"RouterNamespaceOwnership": cfg.Ingress.AdmissionPolicy.NamespaceOwnership == config.NamespaceOwnershipAllowed,
 		"RouterHttpPort":           cfg.Ingress.Ports.Http,
 		"RouterHttpsPort":          cfg.Ingress.Ports.Https,
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -195,6 +195,8 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 
 	extraParams := assets.RenderParams{
 		"RouterNamespaceOwnership": cfg.Ingress.AdmissionPolicy.NamespaceOwnership == config.NamespaceOwnershipAllowed,
+		"RouterHttpPort":           cfg.Ingress.Ports.Http,
+		"RouterHttpsPort":          cfg.Ingress.Ports.Https,
 	}
 	if err := assets.ApplyDeployments(ctx, apps, renderTemplate, renderParamsFromConfig(cfg, extraParams), kubeconfigPath); err != nil {
 		klog.Warningf("Failed to apply apps %v: %v", apps, err)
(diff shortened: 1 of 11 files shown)


pkg/config/config.go
```diff
@@ -193,6 +197,14 @@ func (c *Config) incorporateUserSettings(u *Config) {
...
 		c.Ingress.AdmissionPolicy.NamespaceOwnership = u.Ingress.AdmissionPolicy.NamespaceOwnership
 	}
+
+	if u.Ingress.Ports.Http != 0 {
+		c.Ingress.Ports.Http = u.Ingress.Ports.Http
+	}
+
+	if u.Ingress.Ports.Https != 0 {
+		c.Ingress.Ports.Https = u.Ingress.Ports.Https
+	}
```
dhellmann, line 207: There are some unit tests for the config package. Those should be updated as part of this work.

pkg/config/ingress.go
```diff
@@ -17,3 +18,10 @@ type RouteAdmissionPolicy struct {
 	// +kubebuilder:default="InterNamespaceAllowed"
 	NamespaceOwnership NamespaceOwnershipEnum `json:"namespaceOwnership"`
 }
+
+type IngressPortsConfig struct {
+	// +kubebuilder:default=80
+	Http uint16 `json:"http"`
+	// +kubebuilder:default=443
+	Https uint16 `json:"https"`
+}
```
dhellmann, line 27: Do we support disabling individual protocols? How do you tell the difference between a configuration where HTTP is explicitly disabled vs. one where the config does not mention HTTP at all?

test/suites/standard1/router.robot
```diff
@@ -14,19 +14,26 @@ Test Tags           restart    slow
...
+...                         ---
+...                         ingress:
+...                         \ \ routeAdmissionPolicy:
+...                         \ \ \ \ namespaceOwnership: Strict
+${ROUTER_CUSTOM_PORT}       SEPARATOR=\n
+...                         ---
+...                         ingress:
+...                         \ \ status: Enabled
+...                         \ \ ports:
+...                         \ \ \ \ http: ${ALTERNATIVE_HTTP_PORT}
```
dhellmann, line 36: It would be good to have a test for an alternative HTTPS port, too.

pkg/config/ingress.go
pacevedom: OpenShift router does not support disabling ports through any Ingress/IngressController APIs and we aligned to that. Routes are also involved in how they are exposed (https or http), it depends on the configuration they have.

pkg/components/controllers.go
```diff
@@ -176,8 +180,10 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		return err
 	}
 
-	extraParams := assets.RenderParams{
+	extraParams = assets.RenderParams{
```
pmtk, line 183: Can we define it with these 3 fields straightaway on line 162?

pkg/config/ingress.go
dhellmann: OK, sounds good

pacevedom:
/hold
Need to do some changes

pkg/config/config.go
```diff
@@ -120,6 +121,10 @@ func (c *Config) fillDefaults() error {
 		AdmissionPolicy: RouteAdmissionPolicy{
 			NamespaceOwnership: NamespaceOwnershipAllowed,
 		},
+		Ports: IngressPortsConfig{
+			Http:  &(&struct{ x int }{80}).x,
```
dhellmann, line 125: Is this just to get a pointer to the integer? We already have a dependency on https://pkg.go.dev/k8s.io/utils/pointer and that might be easier to read.

pkg/config/config.go
```diff
@@ -301,6 +316,13 @@ func (c *Config) validate() error {
 		return fmt.Errorf("unsupported namespaceOwnership value %v", c.Ingress.AdmissionPolicy.NamespaceOwnership)
 	}
 
+	if c.Ingress.Ports.Http != nil && (*c.Ingress.Ports.Http < 1 || *c.Ingress.Ports.Http > math.MaxUint16) {
```
dhellmann, line 319: These conditions should be tested in the unit tests.

pkg/config/config.go
pacevedom: Turns out that package was deprecated and had to use k8s.io/utils/ptr which uses generics. Let me know what you think.

pkg/config/config.go
pacevedom: Done

pkg/components/controllers.go
pacevedom: Done

test/suites/standard1/router.robot
```diff
@@ -119,3 +151,46 @@ Restart Router
...
+
+Network APIs With Test Label Are Gone
+    [Documentation]    Check for service and endpoint by "app=hello-microshift" label. Succeeds if response matches
+    ...    "No resources found in <namespace> namespace." Fail if not.
+    ${match_string}=    Catenate    No resources found in    ${NAMESPACE}    namespace.
+    ${match_string}=    Remove String    ${match_string}    "
+    ${response}=    Run With Kubeconfig    oc get svc,ep -l app\=hello-microshift -n ${NAMESPACE}
+    Should Be Equal As Strings    ${match_string}    ${response}    strip_spaces=True
+
+Check Http Port Open
```
pmtk, line 180: Would `Http Port Should Be Open` be better fit here?

pacevedom: Done.

test/suites/standard1/router.robot
```diff
@@ -119,3 +151,46 @@ Restart Router
...
+    Wait Until Keyword Succeeds    30s    1s
+    ...    Network APIs With Test Label Are Gone
+
+Network APIs With Test Label Are Gone
+    [Documentation]    Check for service and endpoint by "app=hello-microshift" label. Succeeds if response matches
+    ...    "No resources found in <namespace> namespace." Fail if not.
+    ${match_string}=    Catenate    No resources found in    ${NAMESPACE}    namespace.
+    ${match_string}=    Remove String    ${match_string}    "
+    ${response}=    Run With Kubeconfig    oc get svc,ep -l app\=hello-microshift -n ${NAMESPACE}
+    Should Be Equal As Strings    ${match_string}    ${response}    strip_spaces=True
```
pmtk, line 178: I don't see any of these being used in a test

pacevedom: Right, these were leftovers from previous tests, now they are simpler and not needed. Thanks.

test/suites/standard1/router.robot
```diff
@@ -192,3 +223,21 @@ Setup Hello MicroShift Pods In Multiple Namespaces
...
+Https Port Should Be Open
+    [Documentation]    Try to curl the router, disregard the response because we only care about connectivity.
+    [Arguments]    ${port}
+    ${result}=    Access Hello MicroShift    ${port}    protocol=https
+    Should Match Regexp    ${result}    HTTP.*503
+
+Port Should Be Closed
+    [Documentation]    Try to curl the router and expect a failure when connecting.
+    [Arguments]    ${port}
+    Run Keyword And Expect Error    *
```
dhellmann, line 242: Can we write this in a way that asserts the type of error accessing the port receives? It might be easier to do from python than RF.

test/suites/standard1/router.robot
```diff
@@ -192,3 +223,21 @@ Setup Hello MicroShift Pods In Multiple Namespaces
     Expose Hello MicroShift    ${NS_OWNERSHIP_2}
     Oc Expose    svc hello-microshift --hostname ${HOSTNAME} --path /${NS_OWNERSHIP_1} -n ${NS_OWNERSHIP_1}
     Oc Expose    svc hello-microshift --hostname ${HOSTNAME} --path /${NS_OWNERSHIP_2} -n ${NS_OWNERSHIP_2}
+
+Http Port Should Be Open
+    [Documentation]    Try to curl the router, disregard the response because we only care about connectivity.
+    [Arguments]    ${port}
+    ${result}=    Access Hello MicroShift    ${port}
+    Should Match Regexp    ${result}    HTTP.*503
```
dhellmann, line 231: Why do we expect a 503?

test/suites/standard1/router.robot
pacevedom: Did it with curl in the end because it re-uses parts of the router connection keywords. Doing it in a separate library meant having part in RF and part in python (plus, to mimick the --connect-to tests you have to rewrite some connect functions as they are not part of the urllib API).

test/suites/standard1/router.robot
pacevedom: Added a comment. This is because there is no route in the router to handle this path. The test only checks connectivity and a router listening on the other side, not its configuration.

PR #3144, opened 2024-03-12
USHIFT-2542: Fix scenario removal

pacevedom:

Closes #<Issue Number>


pacevedom:
/hold
just in case it fails again.

PR #3148, opened 2024-03-13
USHIFT-2459: Add docs to namespaceOwnership

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/config/ingress.go
+++ b/pkg/config/ingress.go
@@ -14,6 +14,16 @@ type IngressConfig struct {
 }
 
 type RouteAdmissionPolicy struct {
+	// Describes how host name claims across namespaces should be handled.
+	//
+	// Value must be one of:
+	//
+	// - Strict: Do not allow routes in different namespaces to claim the same host.
+	//
+	// - InterNamespaceAllowed: Allow routes to claim different paths of the same
+	//   host name across namespaces.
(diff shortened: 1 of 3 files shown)


pkg/config/ingress.go
```diff
@@ -14,6 +14,17 @@ type IngressConfig struct {
 }
 
 type RouteAdmissionPolicy struct {
+	// namespaceOwnership describes how host name claims across namespaces should
```
pmtk, line 17:
```suggestion
	// NamespaceOwnership describes how host name claims across namespaces should
```

pacevedom: There is this discrepancy, but I saw it in other parameters. Let me rephrase it to avoid mentioning it (its redundant anyway), as this is also quite common in other parameters.

pkg/config/ingress.go
```diff
@@ -14,6 +14,16 @@ type IngressConfig struct {
 }
 
 type RouteAdmissionPolicy struct {
+	// Describe how host name claims across namespaces should be handled.
```
dhellmann, line 17:
```suggestion
	// Describes how host name claims across namespaces should be handled.
```

pacevedom: Done

PR #3155, opened 2024-03-14
USHIFT-2228: Test Multus - macvlan CNI

pmtk:


pmtk:
Depends on https://github.com/openshift/microshift/pull/3137

PR #3156, opened 2024-03-14
USHIFT-2229: Test Multus - ipvlan CNI

pmtk:


pmtk:
Depends on https://github.com/openshift/microshift/pull/3155

test/bin/manage_hypervisor_config.sh
```diff
@@ -67,6 +67,12 @@ action_create() {
         sudo virsh net-autostart "${VM_ISOLATED_NETWORK}"
     fi
 
+    # Reconfigure `default` network's DHCP range to accomodate for Multus-ipvlan test's static IPs
+    if sudo virsh net-dumpxml default | grep -q "start='192.168.122.2'"; then
+        sudo virsh net-update default delete ip-dhcp-range "<range start='192.168.122.2' end='192.168.122.254'/>" --config --live
+        sudo virsh net-update default add ip-dhcp-range "<range start='192.168.122.10' end='192.168.122.254'/>" --config --live
+    fi
+
```
ggiguash, line 75: I believe we need to make this code more generic. Here's what I have on my machine as a range.
```
$ sudo virsh net-dumpxml default | grep -A4 'ip address'
  <ip address='192.168.100.1' netmask='255.255.255.0' localPtr='yes'>
    <dhcp>
      <range start='192.168.100.100' end='192.168.100.254'/>
    </dhcp>
  </ip>
```

ggiguash: We already have an isolated-network.xml, which we define for a special case. 
I would add a special case new network for this test, rather than changing the default.

pmtk: Changed according to the suggestion

PR #3159, opened 2024-03-14
USHIFT-1349: use pygithub library to get application token

dhellmann:
Add a wrapper script around the Python script to manage a virtualenv
with the dependencies needed to convert a Github application ID and
certificate to an API token so we can use the script in CI as well as
running it by hand.

/assign @pmtk


scripts/release-notes/tag_stable_releases.py
pmtk, line 1: Empty file?

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -134,6 +148,8 @@ def main():
     )
     args = parser.parse_args()
 
+    if not GITHUB_TOKEN:
+        GITHUB_TOKEN = get_access_token_for_app()
```
pmtk, line 152: Missing arguments?

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -160,6 +176,25 @@ def main():
...
+    key_path = os.environ.get('CLIENT_KEY')
+    if not key_path:
+        raise RuntimeError('CLIENT_KEY is not set')
+    integration = github.GithubIntegration(
+        app_id,
+        pathlib.Path(key_path).read_text(encoding='utf-8'),
+    )
+    app_installation = integration.get_installation(GITHUB_ORG, GITHUB_REPO)
+    if app_installation is None:
+        sys.exit(f"Failed to get app_installation for {GITHUB_ORG}/{GITHUB_REPO}. " +
```
pmtk, line 193: Just curious: here it's `sys.exit` but in lines above it's `raise` - any difference? I think uncaught exception produces stacktrace?

scripts/release-notes/tag_stable_releases.py
dhellmann: Right, I use tab completion to run the script and now I need to run the shell script so I removed the permissions on the python script.

scripts/release-notes/gen_ec_release_notes.py
dhellmann: Bah, clearly I didn't test this part of the change. Fixed.

scripts/release-notes/gen_ec_release_notes.py
dhellmann: I left the sys.exit() in the original copy I made from your script. I've switched to RuntimeError for consistency. I prefer the exception because it lets me see where in the code the error is coming from.

PR #3163, opened 2024-03-15
USHIFT-1349: automatically track pre-release versions in release note script

dhellmann:
We build a default list of versions to scan using the current version
and the previous version. This assumes the script is run out of the
main branch where we can find the most current version under
development. During the period where 4.n is being developed, 4.n-1 may
still be producing only release candidates, so that scanning those 2
versions give us the 2 most recent candidates for having no
releases. During the period where the main branch has not landed a
rebase to update the version and both main and the pre-release branch
have the same version, we will end up scanning for EC or RC releases
of 4.n-2, but that's

/assign @pmtk @pacevedom

### DIFF
--- a/scripts/release-notes/gen_ec_release_notes.py
+++ b/scripts/release-notes/gen_ec_release_notes.py
@@ -90,11 +90,6 @@ VERSION_RE = re.compile(
 # Save our token for use later
 GITHUB_TOKEN = os.environ.get('GITHUB_TOKEN')
 
-# Include the major.minor version string in this list to ignore
-# processing very old versions for which we do not anticipate future
-# candidate builds.
-OLD_VERSIONS = ['4.12', '4.13', '4.14', '4.15']
-
 
 # Representation of one release
 Release = collections.namedtuple(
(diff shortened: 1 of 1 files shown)


scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -110,6 +110,35 @@ def main():
...
+    # .../microshift/scripts/release-notes/gen_ec_release_notes.py
+    # and we want the path to
+    # .../microshift
+    root_dir = pathlib.Path(__file__).parent.parent.parent
+    version_makefile = root_dir / 'Makefile.version.aarch64.var'
+    # Makefile contains something like
+    #   OCP_VERSION := 4.16.0-0.nightly-arm64-2024-03-13-041907
+    # and we want
+    #   4.16
+    version = version_makefile.read_text('utf8').split('=')[-1].strip().split('.')[:2]
```
pmtk, line 123: Is the `version` a list? Looks like it due to `[:2]` and usage in line 136, but comment `4.16` suggests it's a `str`.
Do you think we could make some changes to have less list operations? Could we add a `struct {major, minor}` or at least `major` `minor` integers? All that `[idx]` (together with line 136) is a bit difficult to read

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -154,13 +192,16 @@ def main():
     if not GITHUB_TOKEN:
         raise RuntimeError('GITHUB_TOKEN does not appear to be set')
 
+    if not args.versions_to_scan:
+        args.versions_to_scan = versions_to_scan
```
pmtk, line 196: What's this for? I don't see `args` passed to another fuction and I don't see usage of `args.versions_to_scan`

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -265,7 +306,7 @@ def find_new_releases(url_base, release_type):
         # Skip very old RCs, indicated by the first 2 parts of the
         # version string major.minor.
         version_prefix = '.'.join(version.split('.')[:2])
-        if version_prefix in OLD_VERSIONS:
+        if version_prefix not in versions_to_scan:
```
pmtk, line 309: `OLD_VERSIONS` looks unused after this change

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -110,6 +110,35 @@ def main():
     """
     global GITHUB_TOKEN
 
+    # The script runs as
+    # .../microshift/scripts/release-notes/gen_ec_release_notes.py
+    # and we want the path to
+    # .../microshift
+    root_dir = pathlib.Path(__file__).parent.parent.parent
+    version_makefile = root_dir / 'Makefile.version.aarch64.var'
```
pmtk, line 118: Can't we assume we're executing the script from the `microshift/` and just use current working dir?
That's how it runs in the https://github.com/openshift/release/pull/49830/files#diff-24ec5dd51fc1cbf37ac7d08f85d9f4d8abc188083abce1ccc7ace59d1eaf6377R16-R17

scripts/release-notes/gen_ec_release_notes.py
dhellmann: Yeah, this could be written clearer. Maybe even as a regex. :-)

The comment is wrong here. `version` ends up a list, and then `versions_to_scan` below gets 2 strings.

I'll fix this up.

scripts/release-notes/gen_ec_release_notes.py
dhellmann: I think this is a vestige of an earlier implementation plan that I didn't follow..

dhellmann: Actually, it was mistakenly trying to replace the defaults with the values the user gave us, but I either overwrote the wrong variable or passed the wrong variable to all of the functions that need it below.

scripts/release-notes/gen_ec_release_notes.py
dhellmann: Good catch, removed

PR #3173, opened 2024-03-21
USHIFT-2798: MicroShift bootc pilot 

ggiguash:
Shift Week project integrating MicroShift on CentOS 9 bootc image in CI.


test/bin/pyutils/common.py
```diff
@@ -0,0 +1,72 @@
...
+        env=os.environ.copy(), stdout=subprocess.PIPE)
+    return result.stdout.strip()
+
+
+def create_dir(dir: str, chdir: bool):
+    """Attempt recursive directory creation and optionally change current directory"""
+    path = pathlib.Path(dir)
+    path.mkdir(parents=True, exist_ok=True)
+    if chdir:
+        os.chdir(path)
```
pmtk, line 58: I'm not fan of this especially without accompanying "popd`. What do you think about using full absolute paths whenever possible - it's verbose, but there's no "context" around the command with relative path.

test/image-blueprints/layer5-bootc/group1/cos9-bootc-source.containerfile
```diff
@@ -0,0 +1,59 @@
+FROM quay.io/centos-bootc/centos-bootc:stream9
+
+# Build arguments
+ARG PREVIOUS_MINOR_VERSION=15
+ARG MICROSHIFT_REPO_NAME=microshift-local
```
pmtk, line 5: May I suggest including `RPM` in the name? I was really wondering why we want to copy `openshift/microshift` to the container :)

test/kickstart-templates/includes/main-ostreecontainer.cfg
```diff
@@ -0,0 +1,10 @@
+# Mount the bootc image directory from the hypervisor
+%pre-install --log=/dev/console --erroronfail
+
+mkdir -p /mnt/bootc-images
+mount -v -t virtiofs bootc-images /mnt/bootc-images
```
pmtk, line 5: Can you remind what was the problem with the registry? I think we should add registry pretty soon

test/scenarios-bootc/cos9-src@bootc-standard2.sh
```diff
@@ -0,0 +1,23 @@
...
+scenario_create_vms() {
+    prepare_kickstart host1 kickstart-bootc.ks.template cos9-bootc-source
+    launch_vm host1 centos9 "" "" "" "" "" "" "1"
+
+    # Open the firewall ports. Other scenarios get this behavior by
+    # embedding settings in the blueprint, but there is no blueprint
+    # for this scenario. We need do this step before running the RF
+    # suite so that suite can assume it can reach all of the same
+    # ports as for any other test.
+    configure_vm_firewall host1
```
pmtk, line 14: Can't this be done in the containerfile?
Found example where ansible is executed during build to configure firewall: https://gitlab.com/CentOS/cloud/centos-bootc-examples/-/tree/main/ansible-firewalld

test/bin/ci_phase_iso_build.sh
```diff
@@ -153,7 +166,13 @@ else
     if ! ${GOT_CACHED_DATA} ; then
         echo "WARNING: Build cache is not available, rebuilding all the artifacts"
     fi
-    run_image_build
+
+    # Optionally run bootc image builds
+    if ${COMPOSER_CLI_BUILDS} ; then
+        run_image_build
+    else
+        run_bootc_image_build
```
pmtk, line 174: Is this short-term plan until we port all images? Seems we'll be spinning metal-host for single test?

test/bin/pyutils/build_bootc_images.py
```diff
@@ -0,0 +1,212 @@
...
+        # Process a layer with group directories
+        for item in os.listdir(args.layer_dir):
+            item_path = os.path.join(args.layer_dir, item)
+            # Check if this item is a directory
+            if os.path.isdir(item_path):
+                process_containerfiles(item_path, args.dry_run)
+        common.print_msg("Build complete")
+    except Exception as e:
+        common.print_msg(f"An error occurred: {e}")
+        traceback.print_exc()
```
pmtk, line 207: I see no logs, is this sufficient information to find a problem in your opinion?

test/bin/pyutils/build_bootc_images.py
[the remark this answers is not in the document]
ggiguash: Yes, it is exactly what you're saying. I wanted to experiment with using Python for writing these scripts. 
Since we continue usage of the .sh wrapper as a public script interface, we can revert to using Bash if we decide that Python is not appropriate.

test/bin/pyutils/common.py
```diff
@@ -0,0 +1,72 @@
...
+    return subprocess.run(command, check=True)
+
+
+def run_command_in_shell(command: str):
+    """Run the command through shell and return its output"""
+    print_msg(f"[SHELL] {command}")
+    result = subprocess.run(
+        command,
+        check=True, shell=True, text=True,
+        env=os.environ.copy(), stdout=subprocess.PIPE)
```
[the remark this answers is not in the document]
ggiguash: Stderr remains the default behavior, which is to print it to the terminal. We do not want stderr in the output captured by the caller.

test/scenarios-bootc/cos9-src@bootc-standard2.sh
ggiguash: It can be done in the containerfile, but it will then be part of each containerfile because we do not have templating for them yet. 
I can move it there if you think it's critical for the first implementation.

test/bin/pyutils/build_bootc_images.py
```diff
@@ -0,0 +1,212 @@
...
+BASE_REPO = common.get_env_var('BASE_REPO')
+NEXT_REPO = common.get_env_var('NEXT_REPO')
+
+
+def find_latest_rpm(repo_path, version=""):
+    rpms = glob.glob(f"{repo_path}/**/microshift-release-info-{version}*.rpm", recursive=True)
+    if not rpms:
+        raise Exception(f"Failed to find 'microshift-release-info-{version}*' RPM in {repo_path}")
+    rpms.sort()
+    return rpms[-1]
```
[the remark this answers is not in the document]
pmtk: But not in scope of test harness as `build_rpms.sh` cleans them, right?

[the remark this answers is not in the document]
ggiguash: Correct. We want the script to work correctly also in the development flows. 
In fact, this is a port of the following line:
https://github.com/openshift/microshift/blob/main/test/bin/build_images.sh#L122

test/scenarios-bootc/cos9-src@bootc-standard2.sh
pmtk: Nah, we can leave it like this for now, but in future I think it'd be better to change it. Like in production I'd have this in the Containerfile so the image is complete (boots and runs).
To reduce repetition we can build an intermediate image (it can be build locally on each job without pushing to registry) and use it in FROM in other images.

test/bin/pyutils/build_bootc_images.py
pmtk: I'm not sure if I like the idea of porting `set -x` - it produces massive logs that are cumbersome to read.

test/image-blueprints/layer5-bootc/group1/cos9-bootc-source.containerfile
```diff
@@ -0,0 +1,59 @@
...
+#   symbolic link as a workaround.
+# - The /var/run directory must be a symbolic link to /run, which is not
+#   the case in bootc images. This cases problems with services like DBus
+#   and NetworkManager. DBus creates its socket in /run/dbus directory, which
+#   is accessed by NetworkManager at /var/run/dbus.
+RUN mkdir -p /opt/cni && mkdir -p /var/opt/cni/bin && ln -s /var/opt/cni/bin /opt/cni/ ; \
+    [ ! -L /var/run ] && rm -rf /var/run && ln -s /run /var/
+
+# Install MicroShift, few helper utilities and cleanup
+RUN dnf install -y vi firewalld microshift && \
```
dhellmann, line 42: If we separate the helpers from microshift, we'll be able to reuse more of the image layers in CI. We should probably set up the layers as

- OVN-K directory massaging pre-reqs
- crio configuration
- firewall installation
- rpm repo configuration
- microshift installation

The last 2 layers are going to change the most often across the images we build for the various test scenarios.

dhellmann (commented):
A couple of nits, but nothing especially serious. I didn't review the actual container build script because I assume we'll want to decide which language we want that in before taking this or the PR that rewrites it from bash to go.

test/bin/ci_phase_iso_build.sh
ggiguash: In the initial implementation stage, I thought, it would be best to separate bootc and composer-cli flows until we stabilize bootc. Then, we can go back and revisit how or if at all we should combine those.

test/image-blueprints/layer5-bootc/group1/cos9-bootc-source.containerfile
ggiguash: Yes, once we have more configurations in place, we will have an opportunity to create layering. 
At this time, we only have one configuration, so I'd postpone layering for later when we have more combinations.

test/image-blueprints/layer5-bootc/group1/cos9-bootc-source.containerfile
```diff
@@ -0,0 +1,59 @@
...
+ADD $USHIFT_RPM_REPO_NAME $USHIFT_RPM_REPO_PATH
+
+# MicroShift local RPM repository
+RUN printf "\
+[microshift-local]\n\
+name=MicroShift Local Repository\n\
+baseurl=$USHIFT_RPM_REPO_PATH\n\
+enabled=1\n\
+gpgcheck=0\n\
+skip_if_unavailable=0\n" > "/etc/yum.repos.d/microshift-local.repo"
```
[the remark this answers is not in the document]
ggiguash: Copying works fine if there are no variables. In this case, variables are not necessary, but I was preparing the grounds for it. Eventually, we will probably use Go templating in containerfiles as well.

pmtk:
I think this is good enough for now
/lgtm

PR #3211, opened 2024-04-03
USHIFT-2443: Introduce router IP address listening configuration

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/cmd/generate-config/config/config-openapi-spec.json
+++ b/cmd/generate-config/config/config-openapi-spec.json
@@ -101,11 +101,19 @@
     "ingress": {
       "type": "object",
       "required": [
+        "expose",
         "ports",
         "routeAdmissionPolicy",
         "status"
       ],
       "properties": {
+        "expose": {
+          "description": "List of NIC names and IP addresses where the router will be listening. Defaults to the IPs in the host.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
         "ports": {
           "type": "object",
           "required": [
--- a/cmd/generate-config/config/config-openapi-spec.json
+++ b/cmd/generate-config/config/config-openapi-spec.json
@@ -101,14 +101,14 @@
     "ingress": {
       "type": "object",
       "required": [
-        "expose",
+        "listenAddress",
         "ports",
         "routeAdmissionPolicy",
         "status"
       ],
       "properties": {
-        "expose": {
-          "description": "List of NIC names and IP addresses where the router will be listening. Defaults to the IPs in the host.",
+        "listenAddress": {
+          "description": "List of IP addresses and NIC names where the router will be listening. The NIC names get translated to all their configured IPs dynamically. Defaults to the configured IPs in the host at MicroShift start.",
           "type": "array",
           "items": {
             "type": "string"
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"slices"
 	"strings"
 	"time"
 
@@ -126,7 +127,6 @@ func (c *Config) fillDefaults() error {
 			defaultManifestDirEtcGlob,
 		},
 	}
-
 	c.Ingress = IngressConfig{
 		Status: StatusManaged,
 		AdmissionPolicy: RouteAdmissionPolicy{
@@ -365,6 +365,12 @@ func (c *Config) validate() error {
 		return fmt.Errorf("unsupported value %v for ingress.ports.https", *c.Ingress.Ports.Https)
 	}
 
+	if len(c.Ingress.Expose) != 0 {
+		if err := validateRouterExpose(c.Ingress.Expose, c.ApiServer.AdvertiseAddress, c.ApiServer.SkipInterface); err != nil {
+			return fmt.Errorf("error validating ingress.expose: %v", err)
+		}
+	}
+
 	return nil
 }
 
@@ -419,6 +425,32 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
 		}
 	}
 	return fmt.Errorf("Advertise address: %s not present in any interface", advertiseAddress)
+}
+
+func validateRouterExpose(entries []string, advertiseAddress string, skipInterface bool) error {
+	addresses, err := GetConfiguredAddresses()
+	if err != nil {
+		return err
+	}
+	names, err := GetHostNICNames()
+	if err != nil {
+		return err
+	}
+	for _, entry := range entries {
+		if net.ParseIP(entry) != nil {
+			if entry == advertiseAddress && !skipInterface {
+				continue
+			}
+			if !slices.Contains(addresses, entry) {
+				return fmt.Errorf("IP %v not present in the host", entry)
+			}
+		} else if slices.Contains(names, entry) {
+			continue
+		} else {
+			return fmt.Errorf("entry %v not present in the host", entry)
+		}
+	}
+	return nil
 }
 
 func getBannedIPs() ([]*net.IPNet, error) {
@@ -480,4 +512,20 @@ func GetConfiguredAddresses() ([]string, error) {
 		addressList = append(addressList, addr.String())
 	}
 	return addressList, nil
+}
+
+func GetHostNICNames() ([]string, error) {
+	handle, err := netlink.NewHandle()
+	if err != nil {
+		return nil, err
+	}
+	links, err := handle.LinkList()
+	if err != nil {
+		return nil, err
+	}
+	names := make([]string, 0, len(links))
+	for _, link := range links {
+		names = append(names, link.Attrs().Name)
+	}
+	return names, nil
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -273,7 +273,7 @@ func (c *Config) updateComputedValues() error {
 		// include it here.
 		addresses, err := AllowedListeningIPAddresses()
 		if err != nil {
-			return fmt.Errorf("unable to compute configured addresses: %v", err)
+			return fmt.Errorf("unable to compute default listening addresses: %v", err)
 		}
 		if !c.ApiServer.SkipInterface {
 			addresses = append(addresses, c.ApiServer.AdvertiseAddress)
@@ -367,7 +367,7 @@ func (c *Config) validate() error {
 
 	if len(c.Ingress.ListenAddress) != 0 {
 		if err := validateRouterListenAddress(c.Ingress.ListenAddress, c.ApiServer.AdvertiseAddress, c.ApiServer.SkipInterface); err != nil {
-			return fmt.Errorf("error validating ingress.listenAddress: %v", err)
+			return fmt.Errorf("error validating ingress.listenAddress: %w", err)
 		}
 	}
 
@@ -427,33 +427,33 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
 	return fmt.Errorf("Advertise address: %s not present in any interface", advertiseAddress)
 }
 
-func validateRouterListenAddress(entries []string, advertiseAddress string, skipInterface bool) error {
+func validateRouterListenAddress(ingressListenAddresses []string, advertiseAddress string, skipInterface bool) error {
 	addresses, err := AllowedListeningIPAddresses()
 	if err != nil {
 		return err
 	}
-	names, err := AllowedNICNames()
+	nicNames, err := AllowedNICNames()
 	if err != nil {
 		return err
 	}
-	for _, entry := range entries {
+	for _, entry := range ingressListenAddresses {
 		if net.ParseIP(entry) != nil {
 			if entry == advertiseAddress && !skipInterface {
 				continue
 			}
 			if !slices.Contains(addresses, entry) {
-				return fmt.Errorf("IP %v not present in the host", entry)
+				return fmt.Errorf("IP %v not present in any of the host's interfaces", entry)
 			}
-		} else if slices.Contains(names, entry) {
+		} else if slices.Contains(nicNames, entry) {
 			continue
 		} else {
-			return fmt.Errorf("entry %v not present in the host", entry)
+			return fmt.Errorf("interface %v not present in the host", entry)
 		}
 	}
 	return nil
 }
 
-func getBannedIPs() ([]*net.IPNet, error) {
+func getForbiddenIPs() ([]*net.IPNet, error) {
 	banned := make([]*net.IPNet, 0)
 	for _, entry := range defaultRouterForbiddenCIDRs {
 		_, netIP, err := net.ParseCIDR(entry)
@@ -464,6 +464,7 @@ func getBannedIPs() ([]*net.IPNet, error) {
 	}
 	return banned, nil
 }
+
 func getHostAddresses() ([]net.IP, error) {
 	handle, err := netlink.NewHandle()
 	if err != nil {
@@ -473,8 +474,9 @@ func getHostAddresses() ([]net.IP, error) {
 	if err != nil {
 		return nil, err
 	}
-	addresses := make([]net.IP, 0)
+	addresses := make([]net.IP, 0, len(links)*2)
 	for _, link := range links {
+		// Filter out slave NICs. These include ovs/ovn created interfaces, in case of a restart.
 		if link.Attrs().ParentIndex != 0 || link.Attrs().MasterIndex != 0 {
 			continue
 		}
@@ -490,7 +492,7 @@ func getHostAddresses() ([]net.IP, error) {
 }
 
 func AllowedListeningIPAddresses() ([]string, error) {
-	bannedAddresses, err := getBannedIPs()
+	bannedAddresses, err := getForbiddenIPs()
 	if err != nil {
 		return nil, err
 	}
@@ -498,7 +500,7 @@ func AllowedListeningIPAddresses() ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	addressList := make([]string, 0)
+	addressList := make([]string, 0, len(hostAddresses))
 	for _, addr := range hostAddresses {
 		skip := false
 		for _, banned := range bannedAddresses {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -222,8 +222,8 @@ func (c *Config) incorporateUserSettings(u *Config) {
 		c.Ingress.Ports.Https = ptr.To[int](*u.Ingress.Ports.Https)
 	}
 
-	if len(u.Ingress.Expose) != 0 {
-		c.Ingress.Expose = u.Ingress.Expose
+	if len(u.Ingress.ListenAddress) != 0 {
+		c.Ingress.ListenAddress = u.Ingress.ListenAddress
 	}
 
 	if len(u.ApiServer.NamedCertificates) != 0 {
@@ -266,8 +266,8 @@ func (c *Config) updateComputedValues() error {
 		c.ApiServer.AdvertiseAddress = firstValidIP.String()
 	}
 
-	if len(c.Ingress.Expose) == 0 {
-		// If the expose is not configured we need to include all of the host addresses
+	if len(c.Ingress.ListenAddress) == 0 {
+		// If the listenAddress is not configured we need to include all of the host addresses
 		// to preserve previous behavior. However, if the apiserver advertise address has
 		// not been configured, it will do so in a later stage and we also need to
 		// include it here.
@@ -278,7 +278,7 @@ func (c *Config) updateComputedValues() error {
 		if !c.ApiServer.SkipInterface {
 			addresses = append(addresses, c.ApiServer.AdvertiseAddress)
 		}
-		c.Ingress.Expose = addresses
+		c.Ingress.ListenAddress = addresses
 	}
 
 	c.computeLoggingSetting()
@@ -365,9 +365,9 @@ func (c *Config) validate() error {
 		return fmt.Errorf("unsupported value %v for ingress.ports.https", *c.Ingress.Ports.Https)
 	}
 
-	if len(c.Ingress.Expose) != 0 {
-		if err := validateRouterExpose(c.Ingress.Expose, c.ApiServer.AdvertiseAddress, c.ApiServer.SkipInterface); err != nil {
-			return fmt.Errorf("error validating ingress.expose: %v", err)
+	if len(c.Ingress.ListenAddress) != 0 {
+		if err := validateRouterListenAddress(c.Ingress.ListenAddress, c.ApiServer.AdvertiseAddress, c.ApiServer.SkipInterface); err != nil {
+			return fmt.Errorf("error validating ingress.listenAddress: %v", err)
 		}
 	}
 
@@ -427,7 +427,7 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
 	return fmt.Errorf("Advertise address: %s not present in any interface", advertiseAddress)
 }
 
-func validateRouterExpose(entries []string, advertiseAddress string, skipInterface bool) error {
+func validateRouterListenAddress(entries []string, advertiseAddress string, skipInterface bool) error {
 	addresses, err := GetConfiguredAddresses()
 	if err != nil {
 		return err
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -19,11 +19,19 @@ import (
 
 	"github.com/apparentlymart/go-cidr/cidr"
 	"github.com/openshift/microshift/pkg/util"
+	"github.com/vishvananda/netlink"
 )
 
 const (
 	// default DNS resolve file when systemd-resolved is used
 	DefaultSystemdResolvedFile = "/run/systemd/resolve/resolv.conf"
+)
+
+var (
+	forbiddenCIDRs = []string{
+		"127.0.0.0/8",
+		"169.254.0.0/16",
+	}
 )
 
 type Config struct {
@@ -118,6 +126,7 @@ func (c *Config) fillDefaults() error {
 			defaultManifestDirEtcGlob,
 		},
 	}
+
 	c.Ingress = IngressConfig{
 		Status: StatusManaged,
 		AdmissionPolicy: RouteAdmissionPolicy{
@@ -213,6 +222,10 @@ func (c *Config) incorporateUserSettings(u *Config) {
 		c.Ingress.Ports.Https = ptr.To[int](*u.Ingress.Ports.Https)
 	}
 
+	if len(u.Ingress.Expose) != 0 {
+		c.Ingress.Expose = u.Ingress.Expose
+	}
+
 	if len(u.ApiServer.NamedCertificates) != 0 {
 		c.ApiServer.NamedCertificates = u.ApiServer.NamedCertificates
 	}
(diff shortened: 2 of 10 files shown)


pacevedom:
/hold
Tests pending.

pkg/loadbalancerservice/controller.go
```diff
@@ -209,3 +238,59 @@ func (c *LoadbalancerServiceController) patchStatus(svc *corev1.Service, newStat
 
 	return err
 }
+
+func (c *LoadbalancerServiceController) getRouterIPAddressList() ([]string, error) {
+	configuredAddresses, err := config.GetConfiguredAddresses()
```
dhellmann, line 243: In what sense are these addresses configured? Are they part of the config data the user gave us, or are they present on the host?

pkg/loadbalancerservice/controller.go
```diff
@@ -209,3 +238,59 @@ func (c *LoadbalancerServiceController) patchStatus(svc *corev1.Service, newStat
...
+	configuredNicNames, err := config.GetHostNICNames()
+	if err != nil {
+		return nil, err
+	}
+
+	ipList := make([]string, 0)
+
+	for _, ip := range c.IPAddresses {
+		if !slices.Contains(configuredAddresses, ip) {
+			klog.Infof("IP address %v not found in the host. Removing it", ip)
```
dhellmann, line 257: The address is just being ignored, right, not removed? Will a reader of the log understand why we're reporting about that address at all?

cmd/generate-config/config/config-openapi-spec.json
```diff
@@ -73,9 +73,17 @@
     "ingress": {
       "type": "object",
       "required": [
+        "expose",
         "routeAdmissionPolicy"
       ],
       "properties": {
+        "expose": {
```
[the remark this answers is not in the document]
pacevedom: How about listenAddress? It also allows NIC names, but always translates to IP addresses.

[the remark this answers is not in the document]
dhellmann: What is the field called in the kube API where these values end up being copied?

pkg/config/config.go
```diff
@@ -17,13 +18,21 @@ import (
...
+	"github.com/vishvananda/netlink"
 )
 
 const (
 	// default DNS resolve file when systemd-resolved is used
 	DefaultSystemdResolvedFile = "/run/systemd/resolve/resolv.conf"
 )
 
+var (
+	forbiddenCIDRs = []string{
```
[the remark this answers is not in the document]
pacevedom: Done

pkg/config/config.go
```diff
@@ -356,3 +390,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+			return nil, err
+		}
+		for _, addr := range addressList {
+			addresses = append(addresses, addr.IP)
+		}
+	}
+	return addresses, nil
+}
+
+func GetConfiguredAddresses() ([]string, error) {
```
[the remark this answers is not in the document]
pacevedom: Renamed the function. This is used from both validations and the load balancer service controller to obtain all the allowed IP addresses in the host. They may change dynamically without restarting MicroShift.

pkg/config/config.go
```diff
@@ -356,3 +390,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+		}
+		if skip {
+			continue
+		}
+		addressList = append(addressList, addr.String())
+	}
+	return addressList, nil
+}
+
+func GetHostNICNames() ([]string, error) {
```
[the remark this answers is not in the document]
pacevedom: Renamed the function. This is used as the function getting all the allowed ip addresses, interfaces may also change dynamically.

pkg/loadbalancerservice/controller.go
pacevedom: These are the already present IPs in the host. Renamed both the function and the variable. I was thinking about a context where configuredXYZ means its in the host and anything coming from the user is in cfg. This can be misleading though.

pkg/loadbalancerservice/controller.go
pacevedom: Depends on the context. These are the addresses that a user configured, and they were in the service status field for some time.
If somehow the IP disappears from the host then it is effectively removed from the service status. This may qualify better as a warning instead, as it points to a misconfiguration of MicroShift/the host.
It is true though that the removal only happens once. The next execution loop will not remove it because it was not there (provided that the situation persists). Will reword to skip and raise the level to warning.

pkg/loadbalancerservice/controller.go
```diff
@@ -193,9 +209,22 @@ func (c *LoadbalancerServiceController) getNewStatus(svc *corev1.Service) (*core
 		}
 	}
 
-	newStatus.Ingress = append(newStatus.Ingress, corev1.LoadBalancerIngress{
-		IP: c.NodeIP,
-	})
+	//TODO use annotations instead.
+	if svc.Name == "router-default" {
```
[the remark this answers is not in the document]
pacevedom: Added.

cmd/generate-config/config/config-openapi-spec.json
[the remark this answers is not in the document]
dhellmann: If we call it `listenAddresses` how confusing will it be that it takes an interface, too? Does `listenInterfaces` have the connotation that it could be a NIC name or IP?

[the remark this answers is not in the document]
pacevedom: Suggested `listenaddress` because after all everything gets translated into IP addresses, and users can see it in the service itself. It is true that it does not accommodate NIC names very good though. I see `listenInterfaces` with a similar issue but on the opposite direction. What do you think of `endpoints`? That would serve for anything that accepts network connections.

pacevedom:
/hold
Seems like the tests are taking too long and it times out. Studying where to cut time now.

ShudiLi:
the ingress port including listenAddress and port were added to the config.yaml, then restarted the microshift service, but sometimes, couldn't connect to the cluster anymore(could curl the http route with the specified LB ip and http port 10080, failed to curl https route on the specified LB ip and https port 10443)

One testing result was below:
1. debug node and check the host ips:
sh-5.1# ip address | grep "inet 10."
    inet 10.192.10.65/24 brd 10.192.10.255 scope global dynamic noprefixroute eth0
    inet 10.44.0.0/32 scope global br-ex
    inet 10.42.0.2/24 brd 10.42.0.255 scope global ovn-k8s-mp0
sh-5.1# 
2. check the default load balancer ip, which having 10.192.10.65 and 10.44.0.0 as expected
% oc -n openshift-ingress get svc
NAME                      TYPE           CLUSTER-IP    EXTERNAL-IP              PORT(S)                      AGE
router-default            LoadBalancer   10.43.37.67   10.192.10.65,10.44.0.0   80:32373/TCP,443:32150/TCP   24m
router-internal-default   ClusterIP      10.43.40.42   <none>                   80/TCP,443/TCP,1936/TCP      24m

3. create two routes
 %oc get route
NAME            HOST                                     ADMITTED   SERVICE         TLS
edge1           edge1-default.apps.example.com           True       unsec-server3   
unsec-server3   unsec-server3-default.apps.example.com   True       unsec-server3   

4.  curl the routes with destination to the LB
sh-4.4# curl http://unsec-server3-default.apps.example.com --resolve unsec-server3-default.apps.example.com:80:10.192.10.65
this a test!
sh-4.4# curl http://unsec-server3-default.apps.example.com --resolve unsec-server3-default.apps.example.com:80:10.44.0.0
this a test!
sh-4.4#
sh-4.4# curl https://edge1-default.apps.example.com -k --resolve edge1-default.apps.example.com:443:10.192.10.65
this a test!
sh-4.4# curl https://edge1-default.apps.example.com -k --resolve edge1-default.apps.example.com:443:10.44.0.0
this a test!
sh-4.4#

5. debug node and modify config.yaml, and then restart the microshift service
sh-5.1# vi config.yaml
sh-5.1# sudo systemctl restart microshift

6. after about 1h passed, couldn't connect to the server
% oc get nodes
The connection to the server 35.94.60.42:6443 was refused - did you specify the right host or port?

ShudiLi:
Tested it with modifying the config.yaml file by adding the ingress part(specifying the listening address 10.44.0.0), after restarted the microshift service:
1.
% oc -n openshift-ingress get svc
NAME                      TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)                      AGE
router-default            LoadBalancer   10.43.21.230   10.44.0.0     80:30448/TCP,443:31863/TCP   69m
router-internal-default   ClusterIP      10.43.88.186   <none>        80/TCP,443/TCP,1936/TCP      69m

2.
 % oc get route
NAME            HOST                                     ADMITTED   SERVICE         TLS
edge1           edge1-default.apps.example.com           True       unsec-server3   
unsec-server3   unsec-server3-default.apps.example.com   True       unsec-server3   

3.
% oc rsh centos-pod2
sh-4.4# curl http://unsec-server3-default.apps.example.com --resolve unsec-server3-default.apps.example.com:80:10.44.0.0
this a test!
sh-4.4# 
sh-4.4# curl https://edge1-default.apps.example.com -k --resolve edge1-default.apps.example.com:443:10.44.0.0
this a test!
sh-4.4#

ShudiLi:
/label qe-approved
thanks

ShudiLi:
tests for the ports:
% oc -n openshift-ingress get svc
NAME                      TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)                           AGE
router-default            LoadBalancer   10.43.21.230   10.44.0.0     10080:30448/TCP,10443:31863/TCP   113m
router-internal-default   ClusterIP      10.43.88.186   <none>        80/TCP,443/TCP,1936/TCP           113m

sh-4.4# curl https://edge1-default.apps.example.com:10443 -k --resolve edge1-default.apps.example.com:10443:10.44.0.0
this a test!
sh-4.4# curl http://unsec-server3-default.apps.example.com:10080 --resolve unsec-server3-default.apps.example.com:10080:10.44.0.0
this a test!
sh-4.4#

pacevedom:
/hold cancel
Ready for final review

pkg/config/config.go
```diff
@@ -249,6 +262,21 @@ func (c *Config) updateComputedValues() error {
...
 	}
 
+	if len(c.Ingress.ListenAddress) == 0 {
+		// If the listenAddress is not configured we need to include all of the host addresses
+		// to preserve previous behavior. However, if the apiserver advertise address has
+		// not been configured, it will do so in a later stage and we also need to
+		// include it here.
+		addresses, err := AllowedListeningIPAddresses()
+		if err != nil {
+			return fmt.Errorf("unable to compute configured addresses: %v", err)
```
pmtk, line 272:
```suggestion
			return fmt.Errorf("unable to compute allowed addresses: %v", err)
```

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+			if entry == advertiseAddress && !skipInterface {
+				continue
+			}
+			if !slices.Contains(addresses, entry) {
+				return fmt.Errorf("IP %v not present in the host", entry)
+			}
+		} else if slices.Contains(names, entry) {
+			continue
+		} else {
+			return fmt.Errorf("entry %v not present in the host", entry)
```
pmtk, line 446:
```suggestion
			return fmt.Errorf("interface %q not present in the host", entry)
```

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+		}
+	}
+	return nil
+}
+
+func getBannedIPs() ([]*net.IPNet, error) {
```
pmtk, line 452: forbidden?

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+	for _, entry := range defaultRouterForbiddenCIDRs {
+		_, netIP, err := net.ParseCIDR(entry)
+		if err != nil {
+			return nil, err
+		}
+		banned = append(banned, netIP)
+	}
+	return banned, nil
+}
+func getHostAddresses() ([]net.IP, error) {
```
pmtk, line 463:
```suggestion
}

func getHostAddresses() ([]net.IP, error) {
```

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+	handle, err := netlink.NewHandle()
+	if err != nil {
+		return nil, err
+	}
+	links, err := handle.LinkList()
+	if err != nil {
+		return nil, err
+	}
+	addresses := make([]net.IP, 0)
```
pmtk, line 472: Is there a reasonable size that we could preallocate?
```suggestion
	addresses := make([]net.IP, 0, len(links) * 2)
```

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+	for _, link := range links {
+		if link.Attrs().ParentIndex != 0 || link.Attrs().MasterIndex != 0 {
```
pmtk, line 474: What do `ParentIndex` and `MasterIndex` mean? What does this line do?

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+func AllowedListeningIPAddresses() ([]string, error) {
+	bannedAddresses, err := getBannedIPs()
+	if err != nil {
+		return nil, err
+	}
+	hostAddresses, err := getHostAddresses()
+	if err != nil {
+		return nil, err
+	}
+	addressList := make([]string, 0)
```
pmtk, line 497: At worse it's slightly too big, but single allocation instead of several allocs
```suggestion
	addressList := make([]string, 0, len(hostAddresses))
```

pkg/loadbalancerservice/controller.go
```diff
@@ -20,23 +22,38 @@ import (
...
+	indexer     cache.Indexer
+	queue       workqueue.RateLimitingInterface
+	informer    cache.SharedIndexInformer
 }
 
 var _ servicemanager.Service = &LoadbalancerServiceController{}
 
 func NewLoadbalancerServiceController(cfg *config.Config) *LoadbalancerServiceController {
+	ipAddresses := make([]string, 0)
+	nicNames := make([]string, 0)
```
pmtk, line 44: It's no like leaving it at 0 will give use slices without any wasted space after all that appending so we could just do it ahead of time
```suggestion
	ipAddresses := make([]string, 0, len(cfg.Ingress.ListenAddress))
	nicNames := make([]string, 0, len(cfg.Ingress.ListenAddress))
```

pkg/loadbalancerservice/controller.go
```diff
@@ -209,3 +228,25 @@ func (c *LoadbalancerServiceController) patchStatus(svc *corev1.Service, newStat
...
 	return err
 }
+
+func (c *LoadbalancerServiceController) updateDefaultRouterServiceStatus(ips []string) error {
+	return wait.PollUntilContextTimeout(context.Background(), time.Second, time.Minute, true, func(ctx context.Context) (bool, error) {
+		svc, err := c.client.CoreV1().Services(defaultRouterServiceNamespace).Get(ctx, defaultRouterServiceName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		klog.Infof("updating default router service status: %v", ips)
```
pmtk, line 238:
```suggestion
		klog.Infof("Updating default router service status: %v", ips)
```

pkg/loadbalancerservice/controller.go
```diff
@@ -209,3 +228,25 @@ func (c *LoadbalancerServiceController) patchStatus(svc *corev1.Service, newStat
...
+		newStatus := &corev1.LoadBalancerStatus{}
+		for _, ip := range ips {
+			newStatus.Ingress = append(newStatus.Ingress, corev1.LoadBalancerIngress{
+				IP: ip,
+			})
+		}
+		err = c.patchStatus(svc, newStatus)
+		if err != nil {
+			klog.Warningf("unable to patch default router service: %v", err)
```
pmtk, line 247:
```suggestion
			klog.Warningf("Unable to patch default router service: %v", err)
```


```suggestion
			klog.ErrorS(err, "Unable to patch default router service")
```

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+	for {
+		select {
+		case <-updateChan:
+			ips, err := defaultRouterListenAddresses(ipAddresses, nicNames)
+			if err != nil {
+				klog.Errorf("unable to determine default router listening addresses: %v", err)
+				break
+			}
+			err = updateFunc(ips)
+			klog.Infof("P333. This is something I am waiting for: %v", err)
```
pmtk, line 52: Debug log?

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			klog.Errorf("unable to determine default router listening addresses: %v", err)
+			continue
+		}
+		if err := updateFunc(ips); err != nil {
+			klog.Errorf("unable to update default router service status: %v", err)
+			continue
+		}
+		break
+	}
```
pmtk, line 33: I understand that this is "retry until it works (no err)"? Should there be some sleep between iterations?

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+
+	updateChan := make(chan netlink.AddrUpdate)
+	doneChan := make(chan struct{})
+	err := netlink.AddrSubscribe(updateChan, doneChan)
+	if err != nil {
+		klog.Errorf("unable to subscribe to IP address changes: %v", err)
+		return
```
pmtk, line 40: Is this the best we can do? It's not progagated so MicroShift could continue running without this functionality and I'm not sure if that's a good thing

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+				break
+			}
+			err = updateFunc(ips)
+			klog.Infof("P333. This is something I am waiting for: %v", err)
+			if err != nil {
+				klog.Errorf("unable to update default router service status: %v", err)
+				break
+			}
+		case <-stopCh:
+			klog.Info("default router watcher stopping")
```
pmtk, line 58: Should we close `doneChan` here?

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+			ips, err := defaultRouterListenAddresses(ipAddresses, nicNames)
+			if err != nil {
+				klog.Errorf("unable to determine default router listening addresses: %v", err)
+				break
+			}
+			err = updateFunc(ips)
+			klog.Infof("P333. This is something I am waiting for: %v", err)
+			if err != nil {
+				klog.Errorf("unable to update default router service status: %v", err)
+				break
```
pmtk, line 55: Why not retry in some time? I guess this also leaves MicroShift without this "controller"?

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	allowedNicNames, err := config.AllowedNICNames()
+	if err != nil {
+		return nil, err
+	}
+
+	ipList := make([]string, 0)
```
pmtk, line 83: what do you think about this?
```suggestion
	ipList := make([]string, 0, len(allowedNicNames) + len(nicNames)*2)
```

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	for _, nicName := range nicNames {
+		if !slices.Contains(allowedNicNames, nicName) {
+			klog.Warningf("NIC %v not found in the host. Skipping", nicName)
+			continue
+		}
+		nicAddresses, err := ipAddressesFromNIC(nicName)
+		if err != nil {
+			return nil, err
```
pmtk, line 100: Above and below we're continuing on "errors" but here we're quiting. What do you think if we'd continue on this error as well (or ignore it in the `defaultRouterWatch`) so this "controller" keeps running?

pkg/loadbalancerservice/router.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	addrList, err := netlink.AddrList(link, netlink.FAMILY_ALL)
+	if err != nil {
+		return nil, err
+	}
+
+	ipList := make([]string, 0)
```
pmtk, line 128:
```suggestion
	ipList := make([]string, 0, len(addrList))
```

pkg/config/config.go
pacevedom: Done

pkg/config/config.go
```diff
@@ -333,6 +361,12 @@ func (c *Config) validate() error {
 		return fmt.Errorf("unsupported value %v for ingress.ports.https", *c.Ingress.Ports.Https)
 	}
 
+	if len(c.Ingress.ListenAddress) != 0 {
+		if err := validateRouterListenAddress(c.Ingress.ListenAddress, c.ApiServer.AdvertiseAddress, c.ApiServer.SkipInterface); err != nil {
+			return fmt.Errorf("error validating ingress.listenAddress: %v", err)
```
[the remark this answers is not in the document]
pacevedom: Done

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+	if err != nil {
+		return err
+	}
+	for _, entry := range entries {
+		if net.ParseIP(entry) != nil {
+			if entry == advertiseAddress && !skipInterface {
+				continue
+			}
+			if !slices.Contains(addresses, entry) {
+				return fmt.Errorf("IP %v not present in the host", entry)
```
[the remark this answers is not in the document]
pacevedom: Done

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
...
+			}
+		} else if slices.Contains(names, entry) {
```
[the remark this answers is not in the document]
pacevedom: Yes. That is the assumption, an interface should not have an ip address as name.

pkg/config/config.go
```diff
@@ -388,3 +422,106 @@ func checkAdvertiseAddressConfigured(advertiseAddress string) error {
 	}
 	return fmt.Errorf("Advertise address: %s not present in any interface", advertiseAddress)
 }
+
+func validateRouterListenAddress(entries []string, advertiseAddress string, skipInterface bool) error {
+	addresses, err := AllowedListeningIPAddresses()
+	if err != nil {
+		return err
+	}
+	names, err := AllowedNICNames()
```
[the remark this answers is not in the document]
pacevedom: Done with `nicNames`

pkg/config/config.go
pacevedom: Done

pkg/config/config.go
pacevedom: Done

pkg/config/config.go
pacevedom: Done

pkg/config/config.go
pacevedom: Those variables tell you if the link is a slave to another link (like veth's from pods). Will include a comment.

pkg/config/config.go
pacevedom: I see the point, but I dont think this will matter much as the code is just ran once at the start, and the amount of total IPs in the system wont get to the point where reallocations hit performance.
Plus, slices are contiguous memory so you might be robbing cache lines from other variables. But again this is negligible with these sizes.
Will do for correctness though.

pkg/config/config.go
pacevedom: Same comment as above

pkg/loadbalancerservice/controller.go
pacevedom: Same comment as above

pkg/loadbalancerservice/controller.go
pacevedom: Done

pkg/loadbalancerservice/router.go
pacevedom: oops yes

pkg/loadbalancerservice/router.go
pacevedom: The second one already has that implicit sleep by trying every second for a minute. It will just start over.

pkg/loadbalancerservice/router.go
pacevedom: Done.

pkg/loadbalancerservice/router.go
pacevedom: Will include this as part of the initial loop. However, falling into this error would mean no more dynamically changing the load balancer IP addresses. The last good known state would remain until an admin took action by restarting MicroShift. This may be considered as a degraded state, but not failed.

pkg/loadbalancerservice/router.go
pacevedom: When you say continuing on errors you mean skipping IPs which are not present anymore? This is not an abnormal situation in MicroShift, but in the configuration. If the values you configured are gone then it is the admin who should take action to fix it.
If, however, MicroShift is unable to retrieve host addresses or NICs then this is a different issue. Since its difficult to determine the cause from here, an error log will be written in the top handler and it will not change the last good known state. The next IP change (or an admin restarting something) will trigger this function again.
The key in this controller is avoiding changing the last good known state unless the new one is accurate.

pkg/loadbalancerservice/router.go
pacevedom: See next comment.

pkg/loadbalancerservice/router.go
pmtk: Exactly, but are we communicating this in any other way than logging it? How does the admin should recognize microshift needs restarting?

pkg/loadbalancerservice/router.go
pmtk: True, but it seems we're not communicating degraded state in any way (beside logs, but come on)? And the remediation is basically "restart microshift" - either manually or by reconfiguring some IPs?

pacevedom: This shall be part of the docs. We do not have a way of signaling this kind of condition in MicroShift. Killing the process could have other implications for an app if its unexpected.

pkg/loadbalancerservice/router.go
pacevedom: There is no other way right now but logs and docs. This situation shall be explicitly stated there as a degraded condition alongside a remediation procedure to solve it.

pkg/loadbalancerservice/router.go
pmtk: ack

pkg/loadbalancerservice/router.go
pmtk: ack

PR #3220, opened 2024-04-05
[WIP] USHIFT-2370: Changes to allow MicroShift 4.16 to build on RHEL 9.4 Beta

dhensel-rh:

Closes #USHIFT-2370

### DIFF
--- a/scripts/image-builder/configure.sh
+++ b/scripts/image-builder/configure.sh
@@ -23,6 +23,13 @@ fi
      createrepo yum-utils selinux-policy-devel jq wget lorax rpm-build \
      containernetworking-plugins expect"
 
+VID=$(source /etc/os-release && echo "${VERSION_ID}")
+if { [[ "${VID}" == "9.4" ]] && grep -qE 'Red Hat Enterprise Linux.*Beta' /etc/redhat-release; }; then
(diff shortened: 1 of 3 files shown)


scripts/image-builder/configure.sh
```diff
@@ -23,6 +23,13 @@ fi
      createrepo yum-utils selinux-policy-devel jq wget lorax rpm-build \
      containernetworking-plugins expect"
 
+VID=$(source /etc/os-release && echo "${VERSION_ID}")
+if { [[ "${VID}" == "9.4" ]] && grep -qE 'Red Hat Enterprise Linux.*Beta' /etc/redhat-release; }; then
```
copejon, line 27: I know we're not going to see a RHEL 9.5 for a minute, but could this be generalized so it's not something we have to continually update?  Maybe we could detect if it's a beta version?

ggiguash: Yes, we should probably change it to be more generic, something like the following
```
if grep -qE 'Red Hat Enterprise Linux.*Beta' /etc/redhat-release; then
   VID=$(source /etc/os-release && echo "${VERSION_ID}")
   sudo sed -i "s,dist/rhel9/${VID},beta/rhel9/9,g" /usr/share/osbuild-composer/repositories/rhel-${VID}.json
fi
```

PR #3225, opened 2024-04-08
USHIFT-2807: Add selected rpm-ostree scenario tests to bootc on CentOS

ggiguash:


test/bin/pyutils/build_bootc_images.py
```diff
@@ -199,8 +200,8 @@ def main():
         if args.group_dir:
             process_containerfiles(args.group_dir, args.dry_run)
             return
-        # Process a layer with group directories
-        for item in os.listdir(args.layer_dir):
+        # Process layer directory contents sorted by length and then alphabetically
+        for item in sorted(os.listdir(args.layer_dir), key=lambda i: (len(i), i)):
```
pmtk, line 204: It doesn't help with parallelization of the process, right? Similar to what we have now, seems like `cos9-bootc-source-optionals.containerfile` is material for `group2`

ggiguash: Very true. We have a separate issue on the parallelization [USHIFT-2929](https://issues.redhat.com/browse/USHIFT-2929). I think this is where we should do the restructuring and test the time gain, etc.

PR #3228, opened 2024-04-12
USHIFT-2806: Initial implementation of RHEL 9.4 bootc test

ggiguash:


test/bin/pyutils/common.py
```diff
@@ -24,12 +24,15 @@ def print_msg(msg: str, file=sys.stderr):
...
 
 
-def get_env_var(var_name: str):
+def get_env_var(var_name: str, def_val: str = None):
     """Get an environment variable or exit if not set."""
     value = os.environ.get(var_name)
     if value is None:
-        print_msg(f"Error: {var_name} environment variable not set.")
-        sys.exit(1)
+        if def_val is None:
```
pmtk, line 31: What do you think about reducing nesting?
```
    if value:
        return value

    if def_val:
        return def_value

    print_msg(f"Error: {var_name} environment variable not set.")
    sys.exit(1)
```

test/bin/pyutils/build_bootc_images.py
```diff
@@ -157,9 +159,16 @@ def process_containerfiles(groupdir, dry_run=False):
                 continue
 
         common.print_msg(f"Processing {cf_path}")
-        common.run_command(
-            ["podman", "build", "-t", cf_outname, "-f", cf_path,
-                os.path.join(IMAGEDIR, "rpm-repos")], dry_run)
+        # Run the container build command
+        build_args = [
+            "podman", "build",
+            "--authfile", PULL_SECRET,
```
pmtk, line 165: FYI: Alternative would be to copy the pull secret(s) to ~/.docker/config.json, but we already expect ~/.pull-secret.json in other places (e.g. rebase.sh) so I guess it's fine

pmtk (commented):
More questions than requests, looks pretty good

test/bin/pyutils/build_bootc_images.py
ggiguash: I'd prefer using the pull secret directly, not to mess up the user configuration files.

PR #3229, opened 2024-04-12
OCPBUGS-31739: Fix multus manifest: remove defaultMode

pmtk:


ggiguash:
/retitle OCPBUGS-31739: Fix multus manifest: remove defaultMode

PR #3236, opened 2024-04-15
USHFT-2969: Release notes job: use token for `git` commands

pmtk:
When ci-operator clones the repository it does not include any credentials to talk back to the repository.
To be able to execute `git push` we need to add a special remote that will use token that is either supplied (personal) or generated (Installation Access Token, IAT).

### DIFF
--- a/scripts/release-notes/gen_ec_release_notes.py
+++ b/scripts/release-notes/gen_ec_release_notes.py
@@ -66,6 +66,7 @@ URL_BASE = "https://mirror.openshift.com/pub/openshift-v4/aarch64/microshift"
 URL_BASE_X86 = "https://mirror.openshift.com/pub/openshift-v4/x86_64/microshift"
 GITHUB_ORG = 'openshift'
 GITHUB_REPO = 'microshift'
+REMOTE = "token-remote"
 
 # An EC RPM filename looks like
 # microshift-4.13.0~ec.4-202303070857.p0.gcf0bce2.assembly.ec.4.el9.aarch64.rpm
@@ -207,6 +208,7 @@ def main():
         return
 
     print()
+    add_token_remote()
 
     unique_releases = {
         r.commit_sha: r
@@ -404,6 +406,32 @@ def tag_exists(release_name):
         return False
 
 
+def add_token_remote():
+    """
+    Adds the Git remote to the given repository using
+    the provided installation (or personal) access token.
+    """
+    env = {}
+    env.update(os.environ)
+
+    print(f'git remote remove {REMOTE}')
+    subprocess.run(
+       ["git", "remote", "remove", REMOTE],
+       env=env,
+    )
+    print(f'git remote add {REMOTE} ~~REDACTED~~')
+    remote_url = f"https://x-access-token:{GITHUB_TOKEN}@github.com/{GITHUB_ORG}/{GITHUB_REPO}"
+    result = subprocess.run(
+       ["git", "remote", "add", REMOTE, remote_url],
+       env=env,
+       capture_output=True,
+       text=True,
+    )
+    if result.returncode != 0:
+        err = result.stderr.replace(GITHUB_TOKEN, "") if result.stderr else "stderr is empty"
+        raise Exception(f"Command `git remote add` failed: {err}")
+
+
 def tag_release(tag, sha, buildtime):
     env = {}
     # Include our existing environment settings to ensure values like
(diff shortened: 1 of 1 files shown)


scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -404,6 +407,28 @@ def tag_exists(release_name):
...
 
+def add_token_remote():
+    """
+    Returns the Git remote for the given repository using
+    the provided installation (or personal) access token.
+    """
+    env = {}
+    env.update(os.environ)
+
+    print(f'git remote remote {REMOTE}')
```
ggiguash, line 418:
```suggestion
    print(f'git remote {REMOTE}')
```

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -404,6 +407,28 @@ def tag_exists(release_name):
         return False
 
 
+def add_token_remote():
+    """
+    Returns the Git remote for the given repository using
```
ggiguash, line 412:
```suggestion
    Adds the Git remote to the given repository using
```

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -486,14 +511,18 @@ def publish_release(new_release, take_action):
     # Create draft release with message that includes download URLs and history
     try:
         github_release_create(release_name, notes)
+    except urllib.error.URLError as e:
+        print(f"Failed to create the release {release_name}: {e}")
+        print(f"Response: {str(e.fp.readlines())}")
```
ggiguash, line 516: Should we add response printout the to catch-all handler?

pmtk: I think it's a different type so it doesn't have these fields

ggiguash: Can you explain when this exception is thrown?

pmtk: For me it when I tested and the release was already present. Original log would only say: `Failed to create the release 4.16.0-ec.1-202401101824.p0: HTTP Error 422: Unprocessable Entity` and the hint that there already was a release was in the response body of the HTTP request.

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -404,6 +407,28 @@ def tag_exists(release_name):
...
+    subprocess.run(
+       ["git", "remote", "remove", REMOTE],
+       env=env,
+    )
+    print(f'git remote add {REMOTE} ~~REDACTED~~')
+    remote_url = f"https://x-access-token:{GITHUB_TOKEN}@github.com/{GITHUB_ORG}/{GITHUB_REPO}"
+    subprocess.run(
+       ["git", "remote", "add", REMOTE, remote_url],
+       env=env,
+       check=True,
```
dhellmann, line 428: If this throws an exception, does that exception include the command that ran? That would leak the token.

pmtk: Great catch. I believe it does:
```
>>> subprocess.run(["false", "dddd"], check=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['false', 'dddd']' returned non-zero exit status 1.
```

I changed it to raise the Exception by us. PTAL

pmtk: stderr however does not:
```
$ git remote add asd https://github.com/pmtk/microshift.git
$ git remote add asd https://github.com/pmtk/microshift.git
error: remote asd already exists.
```

dhellmann: That looks good, now. Thanks!

scripts/release-notes/gen_ec_release_notes.py
```diff
@@ -404,6 +407,32 @@ def tag_exists(release_name):
...
+    remote_url = f"https://x-access-token:{GITHUB_TOKEN}@github.com/{GITHUB_ORG}/{GITHUB_REPO}"
+    result = subprocess.run(
+       ["git", "remote", "add", REMOTE, remote_url],
+       env=env,
+       capture_output=True,
+       text=True,
+    )
+    if result.returncode != 0:
+        err = result.stderr.replace(GITHUB_TOKEN, "") if result.stderr else "stderr is empty"
+        raise Exception(f"Command `git remote add` failed: {err}")
```
dhellmann, line 433: You could use CalledProcessError, like on line 459, but I wouldn't hold this up over it.

dhellmann (commented):
/lgtm

scripts/release-notes/gen_ec_release_notes.py
pmtk: I noticed it, but didn't quite fit here - I didn't want to put `cmd` in there and redact two things

PR #3238, opened 2024-04-15
USHIFT-2968: Simplify router tests

pacevedom:
Sometimes the operations are not deterministic or take too long. Remove whatever is not mandatory for the test, as listening in the new ports also means not listening in the old ones, so the test may be simplified and the flakes should disappear.

Closes #<Issue Number>


pacevedom:
Problem unrelated to this fix and also being addressed in a different PR.

PR #3251, opened 2024-04-18
USHIFT-2992: Changes to scripts for RHEL9.4

pmtk:


scripts/image-builder/cleanup.sh
```diff
@@ -82,6 +82,7 @@ restart_osbuilder_services() {
     if ! systemctl is-active -q osbuild-composer.socket &>/dev/null ; then
         title "Starting osbuild services"
         sudo systemctl start osbuild-composer.socket
+        sleep 1 # Give composer some time to be ready for workers
```
ggiguash, line 85: I'd prefer not to have sleep. Can we try starting workers 3 times in a loop with sleep in between?

PR #3268, opened 2024-04-23
USHIFT-2929: Implement bootc container file parallel processing

ggiguash:


ggiguash:
The changes are only related to bootc - saving CI cycles
/override ci/prow/metal-periodic-test ci/prow/metal-periodic-test-arm

PR #3271, opened 2024-04-23
USHIFT-2992: Image builder cleanup: don't stop osbuild-composer.socket

pmtk:
Disabling socket can cause composer to fail the shutdown and result in multiple restarts that fail and block later start (at the end of the script).
Instead, stop workers first to keep composer from reactivating via socket, then stop the composer.

### DIFF
--- a/scripts/image-builder/cleanup.sh
+++ b/scripts/image-builder/cleanup.sh
@@ -59,15 +59,17 @@ clean_osbuilder_services() {
     fi
 
     title "Stopping osbuild services"
-    sudo systemctl stop --now osbuild-composer.socket osbuild-composer.service
     for n in $(seq 100) ; do
         worker=osbuild-worker@${n}.service
         if sudo systemctl status "${worker}" &>/dev/null ; then
-            sudo systemctl stop --now "osbuild-worker@${n}.service"
+            sudo systemctl stop "osbuild-worker@${n}.service"
         else
             break
         fi
     done
+    # Don't stop the .socket as it can cause composer.service to fail
+    # stopping resulting in failed restarts which cause problem starting later.
+    sudo systemctl stop osbuild-composer.service
 
     title "Cleaning osbuild worker cache"
     sleep 5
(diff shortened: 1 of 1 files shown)


scripts/image-builder/cleanup.sh
```diff
@@ -59,15 +59,15 @@ clean_osbuilder_services() {
     fi
 
     title "Stopping osbuild services"
-    sudo systemctl stop --now osbuild-composer.socket osbuild-composer.service
     for n in $(seq 100) ; do
         worker=osbuild-worker@${n}.service
         if sudo systemctl status "${worker}" &>/dev/null ; then
-            sudo systemctl stop --now "osbuild-worker@${n}.service"
+            sudo systemctl stop "osbuild-worker@${n}.service"
```
ggiguash, line 65: Why did we remove `--now`? Does it have no effect during stop?

scripts/image-builder/cleanup.sh
```diff
@@ -59,15 +59,15 @@ clean_osbuilder_services() {
...
         else
             break
         fi
     done
+    sudo systemctl stop osbuild-composer.service
```
ggiguash, line 70: Can we add a comment about stopping osbuild-composer.socket here?

scripts/image-builder/cleanup.sh
```diff
@@ -79,19 +79,10 @@ restart_osbuilder_services() {
...
-            if sudo systemctl start osbuild-worker@1.service; then
-                break
-            else
-                if [ "${try}" -eq 2 ]; then
-                    return 1
-                fi
-            fi
-        done
+        sudo systemctl start osbuild-worker@1.service
+        sudo systemctl start osbuild-composer.service
```
ggiguash, line 85: Should the start order be inverted here as we're first stopping workers now?

pmtk: Actually I found that we don't even need to start composer. Thanks to the socket, starting workers will cause composer to start but left it just in case

PR #3290, opened 2024-04-26
USHIFT-3099: wait_images.py: fix restart_job so it's aware of warnings printed to stdout

pmtk:


test/bin/wait_images.py
```diff
@@ -86,7 +86,14 @@ def restart_job(cmd):
...
     if result.returncode != 0:
         return ""
-    return result.stdout.split(" ")[1]
+    # Osbuild emits a warning on stdout about functionality we don't use in our blueprints,
+    # so we need to skip the lines we don't find interesting. Example output:
+    # > Warning: Please note that user customizations on "edge-commit" image type are deprecated and will be removed in the near future
+    # >
+    # > Compose 2c6a1ba2-4a18-49b8-a0f1-d103de1bd93a added to the queue
+    # Split output with \n and select line starting with "Compose", then split that line with space and select 2nd word.
+    line_with_id = [line for line in result.stdout.split("\n") if line.startswith("Compose")][0]
```
pacevedom, line 95: Not that its going to make a difference, but using generators is usually a good practice.
```suggestion
    line_with_id = next(line for line in result.stdout.split("\n") if line.startswith("Compose"))
```

pacevedom (approved):
/lgtm

PR #3292, opened 2024-04-26
NO-ISSUE: Upstream router crd path changed, breaking 4.16 rebase pipeline

copejon:
The upstream version of the file is now stored at https://github.com/openshift/api/blob/master/route/v1/zz_generated.crd-manifests/routes-Default.crd.yaml


pacevedom:
/close
https://github.com/openshift/microshift/pull/3302 addresses same issues and a few more, as that one is already merged.

PR #3317, opened 2024-05-03
USHIFT-3125: Switch to using non-beta RHEL 9.4 container images for bootc

ggiguash:
Also fixes [USHIFT-3131](https://issues.redhat.com//browse/USHIFT-3131)


pacevedom (approved):
/lgtm

PR #3320, opened 2024-05-04
USHIFT-2924: Add support for templates in bootc image and container files

ggiguash:
In addition, optimized the container files by:
* Merging multiple `RUN` commands into one to avoid linter errors
* Switching to using `fast-datapath` and `rhocp` repos in rhel images


test/bin/pyutils/build_bootc_images.py
```diff
@@ -180,10 +200,20 @@ def extract_container_images(version, repo_spec, outfile, dry_run=False):
...
+        "--out", ofile
+    ]
+    common.run_command_in_shell(gomplate_args, dry_run, logfile, logfile)
+
+
 def process_containerfile(groupdir, containerfile, dry_run):
     cf_path = os.path.join(groupdir, containerfile)
     cf_outname = os.path.splitext(containerfile)[0]
     cf_outdir = os.path.join(BOOTC_IMAGE_DIR, cf_outname)
+    cf_gomplate = os.path.join(BOOTC_IMAGE_DIR, f"{containerfile}.gomplate")
```
pmtk, line 216: Interpretation/readability nit: i would expect it to be other way round: `Containerfile.tpl` -> `Containerfile`

test/bin/pyutils/build_bootc_images.py
```diff
@@ -255,9 +289,22 @@ def process_image_bootc(groupdir, bootcfile, dry_run):
...
         # Redirect the output to the log file
         with open(bf_logfile, 'w') as logfile:
-            bf_imgref = common.read_file(bf_path).strip()
+            # Download the bootc image builder itself in case
+            # it requires authorization for accessing the image
+            pull_args = [
+                "sudo", "podman", "pull",
+                "--authfile", PULL_SECRET, BIB_IMAGE
+            ]
+            common.run_command_in_shell(pull_args, dry_run, logfile, logfile)
```
pmtk, line 298: Do we have to do it each time a .bootc image is processed?

test/image-blueprints/layer5-bootc/group1/rhel94-bootc-source.containerfile
```diff
@@ -1,31 +1,49 @@
...
+gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release\n\
+sslverify=1\n\
+sslcacert = /etc/rhsm/ca/redhat-uep.pem\n\
+sslclientkey = %s\n\
+sslclientcert = %s\n\
+skip_if_unavailable=0\n" "${sslkey}" "${sslcrt}" > "/etc/yum.repos.d/openshift-fast-datapath.repo" && \
+    printf "\
+[openshift-rhocp]\n\
+name=OpenShift Dependencies RHEL 9\n\
+baseurl=https://cdn.redhat.com/content/dist/layered/rhel9/{{ .Env.UNAME_M }}/rhocp/4.{{ .Env.PREVIOUS_MINOR_VERSION }}/os\n\
```
pmtk, line 38: What advantages have you noticed by changing using docker's ARG into `{{ .Env.VAR }}`? Beside substitution I don't see anything that would leverage gomplate or am I mistaken?

test/bin/pyutils/build_bootc_images.py
```diff
@@ -134,6 +139,21 @@ def set_rpm_version_info_vars():
...
+        'YMINUS2_MINOR_VERSION', 'FAKE_NEXT_MINOR_VERSION', 'SOURCE_VERSION_BASE',
+        'CURRENT_RELEASE_VERSION', 'CURRENT_RELEASE_REPO', 'PREVIOUS_RELEASE_VERSION',
+        'PREVIOUS_RELEASE_REPO', 'RHOCP_MINOR_Y', 'RHOCP_MINOR_Y1',
+        'RHOCP_MINOR_Y2', 'YMINUS2_RELEASE_VERSION', 'YMINUS2_RELEASE_REPO'
+    ]
+    for var in rpmver_globals_vars:
+        value = globals().get(var)
+        if value is None:
+            raise Exception(f"The '{var}' global variable does not exist")
+        os.environ[var] = str(value)
```
pmtk, line 155: TBH I'm not a fan of this solution. Wouldn't some python native templating be better fit than duplicating what we have in bash (what in my opinion is a bad approach) with some nice data structures?

test/image-blueprints/layer5-bootc/group2/rhel94-bootc-source-optionals.containerfile
```diff
@@ -1,31 +1,49 @@
...
 # Copy the MicroShift repository contents
 COPY $USHIFT_RPM_REPO_NAME $USHIFT_RPM_REPO_PATH
 
-# MicroShift local RPM repository
-RUN printf "\
+# Add the following repositories, using host entitlements for SSL keys:
+# - MicroShift local RPM repository
+# - Fast Datapath repository for MicroShift dependencies
+# - OpenShift previous minor version repository for MicroShift dependencies
+RUN sslkey=$(find /etc/pki/entitlement-host/ -type f -name "*-key.pem" -print -quit) && \
```
copejon, line 14: This run command is one the longer end of what I usually see.  Could we extract this to it's own script file to be `ADD`'d during build time?

test/bin/pyutils/build_bootc_images.py
ggiguash: Changed this to keep the same file name as we're not using `.tpl` in the source currently.

test/bin/pyutils/build_bootc_images.py
ggiguash: I could not find a better place to perform this operation because the `process_image_bootc` and the rest of processing are running in parallel. Putting image pull before the processing will slow down the overall procedure by creating a prerequisite for running `process_image_bootc`.

Do you have a suggestion on how to improve this?

test/image-blueprints/layer5-bootc/group1/rhel94-bootc-source.containerfile
ggiguash: At the moment, we're only using substitution, but I expect this to develop further in the future as more functionality is ported to bootc. 
We cannot use ARG instead of the generic templating because we need templating in other non-container-file templates as well.

copejon:
These changes LGTM. Will wait for @pmtk to chime in before labeling

ggiguash:
Failures not related to the PR changes - saving CI cycles.

PR #3326, opened 2024-05-06
USHIFT-3138: Refactor scenarios not to exceed 20m runtime in pre-submits

ggiguash:
* Splitting the osconfig tests to 3 scenarios in bootc only
* Moving certificate rotation test from standard1 to standard2 to balance them out

Note: when we create periodic tests for bootc, we will likely merge the osconfig tests back into one scenario, similar to what we have in composer-cli periodics.


test/suites/osconfig/lifecycle.robot
```diff
@@ -20,11 +20,10 @@ Restarting MicroShift In The Middle Of Startup Should Succeed
     ...    (for example due to not stopped microshift-etcd.scope).
     [Template]    Restart MicroShift ${TIME_TO_WAIT} Seconds After Starting
 
-    40
-    30
-    20
-    15
-    10
+    29
```
pmtk, line 23: I was under impression we'd have 23, 29, 37 - not drop the 40s one.
Honestly, there are no good or bad values, only not enough of them :D 
How much time do we gain by doing this?

ggiguash: Here's how much time each iteration takes. If we add one more timeout, it'll exceed the 20m "limit".
Another thing we may want to do it to reduce the number of iterations from 3 to 2- wdyt?
image

pmtk: I don't know, I don't like this imposed 20m limit we're imposing on ourselves. We should strive towards testing as comprehensively as possible without taking too many shortcuts. If the robustness/stability test needs more time, we should run it as such hammering it as much as possible.

ggiguash: Maybe, we should move this test to periodics then?

copejon:
Can we rename the scenario scripts to something more descriptive? Using osconfig1..5 doesn't indicate much outside of a general context for the scenario under test.

test/suites/osconfig/lifecycle.robot
ggiguash: I removed this change from the PR until we reach an agreement. Anyway, it does not exceed 20m runtime by far.

pmtk: > Maybe, we should move this test to periodics then?

It does run as periodic in composer tests. It's definitely not a presubmit test material

PR #3330, opened 2024-05-07
USHIFT-3139: Release notes job: create tag before referencing it

pmtk:


dhellmann:
This looks good. Another way to do it would be to use the SHA of the new release instead of the tag as the argument to `git describe`. I like splitting the steps of creating the local tag and pushing the tag, though.

pacevedom (approved):
/lgtm

PR #3340, opened 2024-05-09
OCPBUGS-30039: Switch to registry.redhat.io images for OpenSSL dependency in TopoLVM

ggiguash:
Also switched to the latest ubi9 image. 
> Note that neither ubi9-minimal nor ubi9-micro contain the `openssl` executable we need.


packaging/images/openshift-ci/Dockerfile.test-runtime
pacevedom, line 1: Cant delete this yet, there is a reference in release. Let me work it out because many steps depend on it.

pacevedom (commented):
/hold

ggiguash:
Removed the problematic change. Unholding
/unhold

pacevedom (commented):
/lgtm

PR #3350, opened 2024-05-13
OCPBUGS-33588: Upgrade openvswitch package for 4.16

igsilya:
4.16 ovn-kubernetes and RHCOS are both moved to OVS 3.3:
  - https://github.com/openshift/ovn-kubernetes/pull/2142
  - https://github.com/openshift/os/pull/1495

MicroShift should be kept in sync.

### DIFF
--- a/test/bootc-sources/microshift_repo_config.sh.template
+++ b/test/bootc-sources/microshift_repo_config.sh.template
@@ -5,6 +5,7 @@ USHIFT_LOCAL_REPO_FILE=/etc/yum.repos.d/microshift-local.repo
 OCP_MIRROR_REPO_FILE=/etc/yum.repos.d/openshift-mirror-beta.repo
 OCP_DTPATH_REPO_FILE=/etc/yum.repos.d/openshift-fast-datapath.repo
 OCP_RHOCP_REPO_FILE=/etc/yum.repos.d/openshift-rhocp.repo
+CENTOS_NFV_SIG_REPO_FILE=/etc/yum.repos.d/microshift-sig-nfv.repo
 
 usage() {
     echo "Usage: $(basename $0) <-create microshift_local_repo_path | -delete>"
@@ -36,6 +37,15 @@ config_centos9_repos() {
 [openshift-mirror-beta]
 name=OpenShift Mirror Beta Repository
 baseurl=https://mirror.openshift.com/pub/openshift-v4/{{ .Env.UNAME_M }}/dependencies/rpms/4.{{ .Env.PREVIOUS_MINOR_VERSION }}-el9-beta/
+enabled=1
+gpgcheck=0
+skip_if_unavailable=0
+EOF
+
+    cat > "${CENTOS_NFV_SIG_REPO_FILE}" <<EOF
+[nfv-sig]
+name=CentOS Stream 9 - SIG NFV
+baseurl=http://mirror.stream.centos.org/SIGs/9-stream/nfv/{{ .Env.UNAME_M }}/openvswitch-2/
 enabled=1
 gpgcheck=0
 skip_if_unavailable=0
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -108,7 +108,7 @@ The microshift-selinux package provides the SELinux policy modules required by M
 %package networking
 Summary: Networking components for MicroShift
 Requires: microshift = %{version}
-Requires: (openvswitch3.1 or openvswitch >= 3.1)
+Requires: (openvswitch3.3 or openvswitch >= 3.3)
 Requires: NetworkManager
 Requires: NetworkManager-ovs
 Requires: jq
@@ -470,6 +470,9 @@ fi
 # Use Git command to generate the log and replace the VERSION string
 # LANG=C git log --date="format:%a %b %d %Y" --pretty="tformat:* %cd %an <%ae> VERSION%n- %s%n" packaging/rpm/microshift.spec
 %changelog
+* Mon May 13 2024 Ilya Maximets <i.maximets@redhat.com> 4.16.0
+- Upgrade openvswitch package version to 3.3
+
 * Mon Apr 29 2024 Gregory Giguashvili <ggiguash@redhat.com> 4.16.0
 - Remove references to redundant files in selinux packaging
 
--- a/test/kickstart-templates/includes/post-cos9rpm.cfg
+++ b/test/kickstart-templates/includes/post-cos9rpm.cfg
@@ -8,7 +8,7 @@ gpgcheck=0
 skip_if_unavailable=0
 EOF
 
-# The openvswitch3.1 dependency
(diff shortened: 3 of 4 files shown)


igsilya:
CC: @zshi-redhat

zshi-redhat:
@igsilya  thanks for filing the PR for ovs bump! 
I assume there has at least one OVS 3.3 patch version been released in FDP,  could you confirm?  Since microshift rpm needs to be installed with released ovs versions.
@pmtk  which other rhel versions do we plan to support with microshift 4.16? (just want to make sure OVS 3.3 is also available in older rhel versions that microshift will support).

cc @pliurh

igsilya:
> @igsilya thanks for filing the PR for ovs bump! I assume there has at least one OVS 3.3 patch version been released in FDP, could you confirm? Since microshift rpm needs to be installed with released ovs versions. 

Yes, `3.3.0-2.el9fdp` was officially released a couple weeks back.

igsilya:
Upgrade tests are failing because it is not possible to simply update between major versions of `openvswitch` package.  The reason is that these are not the same package, but two different packages that obsolete each other, hence simple `dnf install -y` can't install them without `--allowerasing`.  Should I add  `--allowerasing` to the test?

@dhellmann

dhellmann:
> Upgrade tests are failing because it is not possible to simply update between major versions of `openvswitch` package. The reason is that these are not the same package, but two different packages that obsolete each other, hence simple `dnf install -y` can't install them without `--allowerasing`. Should I add `--allowerasing` to the test?
> 
> @dhellmann

Please work with @jerpeter1 and @ggiguash on this, I'm not on the MicroShift team any longer. Thank you!

ggiguash:
> Upgrade tests are failing because it is not possible to simply update between major versions of `openvswitch` package. The reason is that these are not the same package, but two different packages that obsolete each other, hence simple `dnf install -y` can't install them without `--allowerasing`. Should I add `--allowerasing` to the test?
> 
> @dhellmann



> Upgrade tests are failing because it is not possible to simply update between major versions of `openvswitch` package. The reason is that these are not the same package, but two different packages that obsolete each other, hence simple `dnf install -y` can't install them without `--allowerasing`. Should I add `--allowerasing` to the test?

Let's try adding `--allowerasing` in the test to see if this is working. We would have to also update the docs with this note.

ggiguash:
@ShaunaDiaz , we need documentation update for MicroShift upgrade from previous versions to 4.16. 
It's about adding `--allowerasing` options when doing `dnf install` on new packages.
Should we open a new ticket or use the attached bug for making the doc changes?

ShaunaDiaz:
> @ShaunaDiaz , we need documentation update for MicroShift upgrade from previous versions to 4.16. It's about adding `--allowerasing` options when doing `dnf install` on new packages. Should we open a new ticket or use the attached bug for making the doc changes?

Issue created here: https://issues.redhat.com/browse/OSDOCS-10554.

ggiguash:
@igsilya , it looks like CentOS 9 installation does not work
```
cos9-bootc-source: Package vim-minimal-2:8.2.2637-20.el9.aarch64 is already installed.
cos9-bootc-source: Error: 
cos9-bootc-source:  Problem: package microshift-4.16.0_0.nightly_arm64_2024_05_13_005028_20240513121611_5bc78baf3d3956865112e7b8d4598f8a50683832-1.el9.aarch64 from microshift-local requires microshift-networking = 4.16.0_0.nightly_arm64_2024_05_13_005028_20240513121611_5bc78baf3d3956865112e7b8d4598f8a50683832, but none of the providers can be installed
cos9-bootc-source:   - conflicting requests
cos9-bootc-source:   - nothing provides (openvswitch3.3 or openvswitch >= 3.3) needed by microshift-networking-4.16.0_0.nightly_arm64_2024_05_13_005028_20240513121611_5bc78baf3d3956865112e7b8d4598f8a50683832-1.el9.aarch64 from microshift-local
cos9-bootc-source: (try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
cos9-bootc-source: Error: building at STEP "RUN ${REPO_CONFIG_SCRIPT} -create ${USHIFT_RPM_REPO_PATH} &&     dnf install -y vi firewalld microshift &&     systemctl enable microshift &&     ${REPO_CONFIG_SCRIPT} -delete &&     rm -f ${REPO_CONFIG_SCRIPT} &&     rm -rf $USHIFT_RPM_REPO_PATH &&     dnf clean all": while running runtime: exit status 1
```

ggiguash:
Do we need to add [another](https://github.com/openshift/microshift/blob/main/test/kickstart-templates/includes/post-cos9rpm.cfg#L12) repo also [here](https://github.com/openshift/microshift/blob/main/test/bootc-sources/microshift_repo_config.sh.template#L35)?

igsilya:
@ggiguash enabling `openshift-mirror-beta` repo for centos seems generally wrong to me.  This repo only contains packages that are directly tagged into openshift and not available otherwise.  Since `openvswitch3.3` is available directly from `fast-datapath` repo, it is not explicitly tagged into that special repo.  Hence, CI can't find it.

The `openvswitch2` repo seems more reasonable.  Though a better solution might be to install `centos-release-nfv-openvswitch` package that will enable necessary repos.

ggiguash:
MicroShift needs the `openshift-client` package, which is only available from mirror for CentOS.
So, it looks like we need to enable both repos, to implement a workaround like [here](https://github.com/openshift/microshift/blob/main/test/kickstart-templates/includes/post-cos9rpm.cfg#L24-L34).

I think we just need to add the openvswitch repo...

igsilya:
OK, let me try to add the repo...

test/bootc-sources/microshift_repo_config.sh.template
```diff
@@ -5,6 +5,7 @@ USHIFT_LOCAL_REPO_FILE=/etc/yum.repos.d/microshift-local.repo
 OCP_MIRROR_REPO_FILE=/etc/yum.repos.d/openshift-mirror-beta.repo
 OCP_DTPATH_REPO_FILE=/etc/yum.repos.d/openshift-fast-datapath.repo
 OCP_RHOCP_REPO_FILE=/etc/yum.repos.d/openshift-rhocp.repo
+CENTOS_NFV_SIG_MIRROR_REPO_FILE=/etc/yum.repos.d/centos-sig-nfv.repo
```
ggiguash, line 8: We're deleting microshift* and openshift* repos in the cleanup. Please, name with repo with `microshift-` prefix.

ggiguash, line 8:
```suggestion
CENTOS_NFV_SIG_REPO_FILE=/etc/yum.repos.d/microshift-sig-nfv.repo
```

igsilya: Ack. Done.

igsilya: Done.

PR #3362, opened 2024-05-16
NO-ISSUE: `scenario.sh recreate` command

pmtk:
Add `recreate` command to scenario for single command to destroy and create scenario's VMs


pacevedom (approved):
/lgtm

PR #3365, opened 2024-05-16
USHIFT-3164: audit log defaults fix

copejon:


pkg/config/config.go
```diff
@@ -98,7 +98,12 @@ func (c *Config) fillDefaults() error {
 		URL:             "https://localhost:6443",
 		Port:            6443,
 	}
-	c.ApiServer.AuditLog.Profile = "Default"
+	c.ApiServer.AuditLog = AuditLog{
+		MaxFileAge:  10,
```
jerpeter1, line 102: In https://github.com/openshift/openshift-docs/pull/75233, we say that "The `maxFileAge` default is 0, meaning there is no default age limit." - we should make sure that the docs agree with the value being set here.

gangwgr:
@copejon Thanks for fix, with these changes, default size one issue is resolved, this one is left https://redhat-internal.slack.com/archives/C03CJTNLKAT/p1715963768827189
[redhat@dhcp-1-235-242 ~]$ sudo systemctl restart microshift
[redhat@dhcp-1-235-242 ~]$ sudo ls -ltrh /var/log/kube-apiserver
total 48M
-rw-------. 1 root root 200M May 17 12:42 audit-2024-05-17T16-43-10.869.log
-rw-------. 1 root root 200M May 17 12:46 audit-2024-05-17T16-46-51.339.log
-rw-------. 1 root root 2.2M May 17 12:50 audit.log

pkg/config/config.go
copejon: Fixed

copejon:
More infra failures

PR #3408, opened 2024-05-29
USHIFT-3159: Retry podman commands when building bootc images

ggiguash:


ggiguash:
The failure is not related to the change. It's a known problem when VMs get stuck on reboot on ARM platform.
/override metal-bootc-test-arm

PR #3424, opened 2024-06-02
USHIFT-3243: build_images.sh rewrite: App skeleton and templating data

pmtk:
Following PR focuses on templating data and preliminary program entrypoint.
It exposes invocation `compose templating-data` that can be used to obtain JSON form of data used for templating, which will be useful in other commands added in future to speed up local iterations thank to skipping most of the operations that query remote locations and are quite time consuming.


copejon:
USHIFT-3262

/jira-refresh

PR #3439, opened 2024-06-05
USHIFT-3315: Increase delete openshift-ingress pod timeout

agullon:


test/suites/standard1/router.robot
```diff
@@ -84,7 +84,7 @@ Router Disabled
     [Setup]    Run Keywords
     ...    Disable Router
 
-    Run With Kubeconfig    oc wait --for=delete namespace/openshift-ingress --timeout=60s
+    Run With Kubeconfig    oc wait --for=delete namespace/openshift-ingress --timeout=300s
```
ggiguash, line 87: Can you explain where we need this change? I do not remember this command expiring in MicroShift CI.

PR #3449, opened 2024-06-06
NO-ISSUE: Update asset.yaml source for the route crd

copejon:
The route API CRD was previously source from the route-controller-manager. This has proven to be problematic since the source file path has change in the last release.  MicroShift should source the CRD from the router repository instead, which should be safer in the future.

The source path used here is `amd64`.  It doesn't matter which arch is manifest is pulled from since the CRD is not arch-dependent.

### DIFF
--- a/scripts/auto-rebase/assets.yaml
+++ b/scripts/auto-rebase/assets.yaml
@@ -200,7 +200,8 @@ assets:
       - file: storage_version_migration.crd.yaml
         src:  0000_50_cluster-kube-storage-version-migrator-operator_01_storage_migration_crd.yaml
       - file: route.crd.yaml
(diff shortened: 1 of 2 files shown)


jerpeter1:
Are we sure the CRD in the router repo is the one we want? The PR where route/v1/route.crd.yaml was introduced claims that it superseded https://github.com/openshift/router/blob/master/deploy/route_crd.yaml . It also looks substantially different than e.g. https://github.com/openshift/api/blob/master/route/v1/zz_generated.crd-manifests/routes-Default.crd.yaml .

copejon:
@jerpeter1 I've updated the PR per your recommendation.  The rebase script will clone the openshift/api repository and copy the routes-Default.crd.yaml file.

PR #3463, opened 2024-06-10
NO-ISSUE: microshift-test-agent improvements

pmtk:
- Remove `/bin/bash` from the `ExecStart` so `microshift-test-agent.sh` is used in journal leading to easier debbuging.
- Add 5s stop timeout, to (hopefully) not receive two SIGTERMs in a row which can lead to _cleanup function to running completely and breaking the test.
- Don't fail if no file matches `/etc/ssh/ssh_host*key` wildcard

### DIFF
--- a/test/agent/microshift-test-agent.service
+++ b/test/agent/microshift-test-agent.service
@@ -3,7 +3,7 @@ Description=MicroShift Test Agent
 Before=microshift.service sshd.service
 
 [Service]
-ExecStart=/bin/bash /usr/bin/microshift-test-agent.sh
+ExecStart=/usr/bin/microshift-test-agent.sh
 
 [Install]
 WantedBy=multi-user.target
--- a/test/agent/microshift-test-agent.service
+++ b/test/agent/microshift-test-agent.service
@@ -4,6 +4,8 @@ Before=microshift.service sshd.service
 
 [Service]
 ExecStart=/usr/bin/microshift-test-agent.sh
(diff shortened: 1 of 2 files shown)


copejon:
this lgtm. can you squash these changes before we merge?

test/agent/microshift-test-agent.service
```diff
@@ -3,7 +3,8 @@ Description=MicroShift Test Agent
 Before=microshift.service sshd.service
 
 [Service]
-ExecStart=/bin/bash /usr/bin/microshift-test-agent.sh
+ExecStart=/usr/bin/microshift-test-agent.sh
+TimeoutStopSec=5s
```
ggiguash, line 7: @pmtk , can you add a comment on why we're adding a timeout?

test/agent/microshift-test-agent.sh
```diff
@@ -108,8 +108,9 @@ EOF
 # When going from "just RHEL" to RHEL+MicroShift+dependencies there might be a
 # problem with groups. Following line ensures that sshd's keys are owned by
 # ssh_keys group and not some other.
+# It should not fail if no file matching wildcard exist.
 # When removing, change Before= in microshift-test-agent.service
-chown -v root:ssh_keys /etc/ssh/ssh_host*key
+chown -v root:ssh_keys /etc/ssh/ssh_host*key || true
```
ggiguash, line 113: Let's try to delete this w/a. It may not be needed anymore because we're correctly using the same parent for subsequent ostree updates.

PR #3468, opened 2024-06-11
USHIFT-3177: Avoid updating default router ips if status is the same

pacevedom:
Updates to the network in the host trigger a possible update. This also triggers a log about an update that may be unnecessary (both the log and the update).
By comparing the current status and the new one we can avoid an API call and also log properly.

Closes #<Issue Number>

### DIFF
--- a/pkg/loadbalancerservice/controller.go
+++ b/pkg/loadbalancerservice/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"slices"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/loadbalancerservice/controller.go
```diff
@@ -235,13 +236,20 @@ func (c *LoadbalancerServiceController) updateDefaultRouterServiceStatus(ips []s
...
 		}
-		klog.Infof("Updating default router service status: %v", ips)
 		newStatus := &corev1.LoadBalancerStatus{}
 		for _, ip := range ips {
 			newStatus.Ingress = append(newStatus.Ingress, corev1.LoadBalancerIngress{
 				IP: ip,
 			})
 		}
+
+		equal := slices.EqualFunc(svc.Status.LoadBalancer.Ingress, newStatus.Ingress, func(oldIP, newIP corev1.LoadBalancerIngress) bool {
```
pmtk, line 246: Should we sort the slices before comparing? Are slices with the same elements considered equal or order makes a difference?

pacevedom: They are already sorted here: https://github.com/openshift/microshift/blob/main/pkg/loadbalancerservice/router.go#L119
This was done in a previous PR.

PR #3483, opened 2024-06-13
USHIFT-2475: The build_images.sh script exits with error if depsolve stage fails

ggiguash:


ggiguash:
Periodic test failures are due to FIPS tests and they are not related to the current change.
/test metal-bootc-test-arm

ggiguash:
FIPS problems not related to the current change.

PR #3492, opened 2024-06-15
USHIFT-3136: Fix remove_vm function to keep info files when not fully deleted

ggiguash:


ggiguash:
Failure because of the FIPS problem.
/test metal-bootc-test-arm

ggiguash:
/retitle USHIFT-3136: Fix remove_vm function to keep info files when not fully deleted

PR #3493, opened 2024-06-16
USHIFT-3494: Add log file printout to the console in case of anaconda errors

ggiguash:


ggiguash:
Saving CI cycles
/override ci/prow/microshift-metal-tests ci/prow/microshift-metal-tests-arm

PR #3506, opened 2024-06-18
USHIFT-3501: separate periodic upgrade tests to reduce resource pressure on hypervisor

copejon:
This PR handles the changes required in the microshift repo.  A PR in openshift/release will create a new parallel job to execute these scenarios on


ggiguash:
@copejon , can we use `scenarios-upgrades` directory name? The fact that it's run in periodics should not be reflected in the dir name.

PR #3521, opened 2024-06-20
USHIFT-3473: Add validations for network ipv4/ipv6 stacks

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -19,6 +19,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
+	netutils "k8s.io/utils/net"
 	"k8s.io/utils/ptr"
 
 	"github.com/apparentlymart/go-cidr/cidr"
@@ -295,6 +296,24 @@ func (c *Config) updateComputedValues() error {
 }
 
 func (c *Config) validate() error {
+	if !isValidIPAddress(c.ApiServer.AdvertiseAddress) {
+		return fmt.Errorf("error validating apiServer.advertiseAddress")
+	}
+	if c.ApiServer.SkipInterface {
+		err := checkAdvertiseAddressConfigured(c.ApiServer.AdvertiseAddress)
+		if err != nil {
+			return err
+		}
+	}
+
+	if !isValidIPAddress(c.Node.NodeIP) {
+		return fmt.Errorf("error validating node.nodeIP")
+	}
+
+	if err := validateNetworkStack(c); err != nil {
+		return fmt.Errorf("error validating networks: %w", err)
+	}
+
 	//nolint:nestif // extracting the nested ifs will just increase the complexity of the if expressions as validation expands
 	if len(c.ApiServer.SubjectAltNames) > 0 {
 		// Any entry in SubjectAltNames will be included in the external access certificates.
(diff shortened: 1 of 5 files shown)


pkg/config/network.go
```diff
@@ -58,3 +58,11 @@ func getClusterDNS(serviceCIDR string) (string, error) {
 
 	return dnsClusterIP.String(), nil
 }
+
+func isValidIPAddress(ipAddress string) error {
```
copejon, line 62: Being a validator function, it makes more since I think to simply return the bool value and generate the error at the caller's level.  Currently, the caller's already wrap this error with a more specific error anyway.

copejon (commented):
One nit, otherwise lgtm!

pkg/config/network.go
pacevedom: Ah wait, I had another function here that returned an error, hence the `isValidIPAddress` return type, then switched to `net.ParseIP` which does not return an error but somehow kept the old signature. Now its updated.

pkg/config/config.go
```diff
@@ -293,6 +294,24 @@ func (c *Config) updateComputedValues() error {
 }
 
 func (c *Config) validate() error {
+	if !isValidIPAddress(c.ApiServer.AdvertiseAddress) {
+		return fmt.Errorf("error validating apiServer.advertiseAddress")
```
pmtk, line 298: How about we add info about the value that failed validation?
```suggestion
		return fmt.Errorf("error validating apiServer.advertiseAddress (%q)", c.ApiServer.AdvertiseAddress)
```

pkg/config/config.go
```diff
@@ -293,6 +294,24 @@ func (c *Config) updateComputedValues() error {
...
+	}
+	if c.ApiServer.SkipInterface {
+		err := checkAdvertiseAddressConfigured(c.ApiServer.AdvertiseAddress)
+		if err != nil {
+			return err
+		}
+	}
+
+	if !isValidIPAddress(c.Node.NodeIP) {
+		return fmt.Errorf("error validating node.nodeIP")
```
pmtk, line 308:
```suggestion
		return fmt.Errorf("error validating node.nodeIP (%q)", c.Node.NodeIP)
```

pkg/config/config.go
```diff
@@ -560,3 +572,42 @@ func validateAuditLogConfig(cfg AuditLog) error {
...
+
+func validateNetworkStack(cfg *Config) error {
+	if len(cfg.Network.ClusterNetwork) != len(cfg.Network.ServiceNetwork) {
+		return fmt.Errorf("network.serviceNetwork and network.clusterNetwork have different cardinality")
+	}
+	if len(cfg.Network.ServiceNetwork) > 2 {
+		return fmt.Errorf("network.serviceNetwork can not have more than 2 entries")
+	}
+	if len(cfg.Network.ClusterNetwork) > 2 {
+		return fmt.Errorf("network.clusterNetwork can not have more than 2 entries")
```
pmtk, line 584: This will never happen. If both slices needs to be same length, only `if` above will effectively work

pacevedom: you are right. done.

pmtk (commented):
/lgtm

PR #3527, opened 2024-06-21
NO-ISSUE: Add runtime crun to crio for y-2 blueprint

pmtk:
microshift 4.15 RPM still specifies `crio >= 1.25` and doesn't ship its own `crun` runtime definition for crio. When it gets newest crio (>= 1.29), the test fails because crio exits shortly after startup.
This PR introduces a workaround: explicitly adds missing file to the blueprint.
When 4.15 RPM has its crio version fixed, we can remove the workaround.


pmtk:
/override ci/prow/microshift-metal-tests ci/prow/microshift-metal-tests-arm

Because test we're trying to fix here is using images from cache, it's not taking into consideration change in this PR. We need to merge this, so next night it's rebuilt.

pacevedom (approved):
/lgtm
Will open a ticket to remember this.

PR #3528, opened 2024-06-21
USHIFT-3482: IPv6 handling adjustments in configuration

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -23,7 +23,6 @@ import (
 	"net/http"
 	"os"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


pkg/util/net.go
```diff
@@ -117,15 +116,6 @@ func RetryTCPConnection(ctx context.Context, host string, port string) bool {
 	return status
 }
 
-func CreateLocalhostListenerOnPort(port int) (tcpnet.Listener, error) {
```
pmtk, line 120: I think etcd needs to re-vendor microshift

pacevedom: Done

PR #3534, opened 2024-06-24
USHIFT-3482: Add ipv6 advertiseAddress handling

pacevedom:

Closes #<Issue Number>


pkg/config/config.go
```diff
@@ -280,7 +280,11 @@ func (c *Config) updateComputedValues() error {
 		// host network pods trying to reach apiserver, as the VIP 10.43.0.1:443 is
 		// not translated to 10.43.0.1:6443. It remains unchanged and therefore
 		// connects to the ingress router instead, triggering all sorts of errors.
-		nextSubnet, exceed := cidr.NextSubnet(svcNet, 32)
+		prefix := 32
```
pmtk, line 283: I think we need to re-vendor microshift in etcd/

pacevedom: Done.

PR #3535, opened 2024-06-24
OCPEDGE-1126: Rebase LVMS in replacement of topolvm

jakobmoellerdev:
TODO:

- [X] ~Generate Lister / Typed Clients due to otherwise mandatory import on lvm-operator introducing tons of dependencies~ replaced with a generic unstructured client
- [X] ~Resolve vendor conflict (see above) with normal resolution instead of vendor patch~ resolved by using dynamic client with discovery API for LVMCluster
- [x] Create Cleanup Scenario that deletes old set of resources from installations that were present in 4.15/4.16
- [x] Add upgrade test scenario to verify that LVMS works as expected
- [x] ~Add webhook yaml generation as part of the rebase script so that the validatingwebhook configuration is applied to LVMCluster normally~ added opinionated webhook generation with notes for further maintenance
- [x] ~Use 4.17 Candidate instead of late 4.16 candidate once we have a CPaaS build out in the wild (need to trigger external replication that can be used during 4.17 testing cycle)~ rebased on 4.17 candidate replicated to quay as aligned. Binaries / shas are equivalent and once release is happening we just need to retrigger the rebase script with the registry.redhat.io url before pushing out the candidate for Microshift. Since the images are the same this is however not endangering breaking anything.
- [x] Add dynamic lvmd.yaml resolution via symlink into `/var/lib/microshift/lvms/lvmd.yaml`
- [x] Rebase on main (seems I have to do this a lot)


pkg/config/lvmd/lvmd.go
```diff
@@ -133,8 +133,20 @@ func NewLvmdConfigFromFile(p string) (*Lvmd, error) {
 	return l, nil
 }
 
-func LvmSupported() error {
-	if _, err := exec.LookPath("vgs"); err != nil {
+func SaveLvmdConfigToFile(l *Lvmd, p string) error {
```
pmtk, line 136: Just to persist what we discussed.
I think this file gets `p = /etc/microshift/lvmd.yaml` and that can be a problem in scenario such as:
1. User creates ostree commit without specifying `/etc/microshift/lvmd.yaml`
2. MicroShift starts, generates default file in `/etc/microshift/lvmd.yaml`
3. System works, everything is fine
4. User creates NEW ostree commit and specifies `/etc/microshift/lvmd.yaml`
5. User deploys the commit
6. OSTree performs its 3-way merge on /etc ([ref](https://containers.github.io/bootc/filesystem.html#etc)) and the result is that `the default file that microshift created` is not overwritten by `user supplied lvmd.yaml` thus breaking the possibility to update lvmd configuration using ostree.

In such case we need to avoid writing in /etc to files that user might edit.
What we came up with is:
1. setup an inotify on `/etc/microshift/lvmd.yaml`
2. if `/etc/microshift/lvmd.yaml` changes, copy it to `/var/lib/microshift/*/lvmd.yaml`
3. mount `/var/lib/microshift/*` for lvms

That way we still get hot reload of the file, but we don't mess up the /etc updates.

jakobmoellerdev:
Putting up as ready for review now even though cleanup discussion is still pending

pkg/components/storage.go
```diff
@@ -121,25 +124,164 @@ func startCSIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
...
+	return lvmd.NewLvmdConfigFromFile(runtimeCfg)
+}
+
+func copyUserCfg(userCfg string, runtimeCfg string, onFail context.CancelCauseFunc) {
+	userCfgFile, err := os.OpenFile(userCfg, os.O_RDONLY, 0)
+	defer func() {
+		if err := userCfgFile.Close(); err != nil {
+			onFail(fmt.Errorf("unable to close lvmd config file %q: %v", userCfg, err))
+		}
+	}()
```
pmtk, line 244: Should this defer be after `if err != nil`? I think that, if error is returned, file is not opened, so closing will err. Though it'll be probably too late for this error, to overwrite error from the `if err != nil` below.

pkg/components/storage.go
```diff
@@ -121,25 +124,164 @@ func startCSIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
...
+		defer func() {
+			if err := watcher.Close(); err != nil {
+				klog.Errorf("unable to close file watcher: %v", err)
+			}
+		}()
+		if err != nil {
+			klog.Errorf("unable to set up file watcher: %v", err)
+		}
+		if err = watcher.Add(usrCfgDir); err != nil {
+			klog.Errorf("unable to add file path to watcher: %v", err)
```
pmtk, line 195: Should we terminate the goroutine in case of errors? and maybe cancel the context?
```suggestion
			klog.Errorf("unable to set up file watcher: %v", err)
            return
		}
		if err = watcher.Add(usrCfgDir); err != nil {
			klog.Errorf("unable to add file path to watcher: %v", err)
            return
```

pkg/components/storage_test.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+		SocketName: "the socket should not matter!",
+	}
+
+	getRuntimeCfgAfterDelay := func() *lvmd.Lvmd {
+		// wait for the runtime config to be updated via the watch
+		time.Sleep(runtimeCfgUpdateDelay)
+		runtimeLvmd, err := lvmd.NewLvmdConfigFromFile(testRuntimeCfg)
+		if err != nil {
+			t.Fatalf("Failed to load user config: %v", err)
```
pmtk, line 42:
```suggestion
			t.Fatalf("Failed to load runtime config: %v", err)
```

scripts/auto-rebase/last_lvms_rebase.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash -x
-./scripts/auto-rebase/rebase-lvms.sh to "registry.redhat.io/lvms4/lvms-operator-bundle:v4.15.0"
+./scripts/auto-rebase/rebase-lvms.sh to "quay.io/lvms_dev/lvms4-lvms-operator-bundle:v4.17.0-18"
```
pmtk, line 2: Can we update https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase_job_entrypoint.sh#L49 to this image or it will fail due to access?

pkg/assets/storage.go
```diff
@@ -192,12 +142,12 @@ func applyVolumeSnapshotClass(ctx context.Context, handler resourceHandler, vcs
...
-// volumeSnapshot client to avoid vendoring
+// storageDynamicClient returns a generic kubernetes Client which can handle arbitrary API schemas.  We use this instead of th
+// volumeSnapshot Client to avoid vendoring
 func storageDynamicClient(kubeconfigPath string) *dynamic.DynamicClient {
 	restCfg, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 	if err != nil {
 		panic(err)
 	}
-	return dynamic.NewForConfigOrDie(rest.AddUserAgent(restCfg, "storage-dynamic-client"))
+	return dynamic.NewForConfigOrDie(rest.AddUserAgent(restCfg, "storage-dynamic-Client"))
```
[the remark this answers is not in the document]
jakobmoellerdev: not sure how that happened... anyhow fixed up :)

pkg/assets/unstructured.go
```diff
@@ -0,0 +1,208 @@
...
+	configClientCacheLock.RLock()
+	entry, ok := configClientCache[kubeconfigPath]
+	configClientCacheLock.RUnlock()
+	if ok {
+		return entry
+	}
+
+	restConfig, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
jakobmoellerdev: Agreed, reworked this setup

pkg/components/storage.go
jakobmoellerdev: good catch, did this in a few places, resolved.

pkg/components/storage.go
jakobmoellerdev: good catch, will do

scripts/auto-rebase/last_lvms_rebase.sh
jakobmoellerdev: changed to dynamically resolve based on the version (candidate or not) and also made it auto update itself on rebase so that this may never be forgotten again by me :)

pkg/components/storage.go
jakobmoellerdev: Btw just confirmed, Closing will not error if there is nothing to close, meaning a noop, so shouldnt change behavior.

jakobmoellerdev:
/cc @copejon
/cc @pmtk 

I feel like this is ready for review to be merged.

Regression test runs as well as migration test was included and I integrated all previous points from @pmtk

pkg/components/storage.go
```diff
@@ -2,101 +2,137 @@ package components
...
+	// TODO: Implement a mechanism to reload the configuration file even when enabling or disabling, this requires a mechanism in k8s to delete resources that are no longer wanted
+	// 3. If the plugin is already enabled, the configuration file is changed and the plugin is restarted with hot-reloading.
+	return setupPluginResources(ctx, cfg, kubeconfigPath)
 }
 
-func startCSIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath string) error {
+// deleteLegacyResources deletes the legacy resources of TopoLVM.
+// TODO: Remove this function in the next release.
+func deleteLegacyResources(ctx context.Context, kubeconfigPath string) error {
+	klog.Infof("Deleting legacy resources of TopoLVM (this is a migration operation and will disappear in the next release)")
```
pmtk, line 65:
```suggestion
	klog.Infof("Deleting legacy resources of TopoLVM (this is a migration operation and will disappear in the future release)")
```

pkg/components/storage.go
```diff
@@ -121,25 +157,168 @@ func startCSIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
...
+//
+// The function is designed to ensure that lvms always uses the most recent configuration through its hot-reload mechanism,
+// whether it's user-defined or the default, and reacts to changes in the configuration file in real-time.
+func loadCSIPluginConfig(ctx context.Context,
+	// usrCfg is the user's configuration file.
+	usrCfg string,
+	runtimeCfg string,
+) (*lvmd.Lvmd, error) {
+	usrCfgDir := filepath.Dir(usrCfg)
+	ctx, cancel := context.WithCancelCause(ctx)
```
pmtk, line 200: Just to be sure. AFAIK the context cancelation doesn't propagate back to parents, right?
So, if we get some error while watching, it won't be propagated back to main threat

jakobmoellerdev: thats right, cancel will only cancel the context return from context.WithCancelCause, not the parent ctx. Ive tried this and tested around and microshift ctx keeps on running

scripts/auto-rebase/rebase_job_entrypoint.sh
```diff
@@ -41,14 +61,24 @@ ARM64_RELEASE=${pullspec_release_arm64} \
...
 
 # Since LVMS is not part of the release payload, it is not kept in
 # CI. Use the latest z-stream that coincides with the release
 # payload's X.Y version
 pullspec_release_lvms="registry.redhat.io/lvms4/lvms-operator-bundle:${release_lvms}"
+# A unreleased candidate doesnt exist in the official registry, so fallback to the lvms_dev namespace, which contains
+# the latest lvms release candidate replicated from CPaaS into quay
+pullspec_release_lvms_fallback="quay.io/lvms_dev/lvms4-lvms-operator-bundle:${release_lvms}"
+
+# You can also pass the version as a command-line argument
```
copejon, line 74: Is this comment from a previous iteration? The script does not parse command line args.

scripts/auto-rebase/rebase-lvms.sh
```diff
@@ -153,15 +154,42 @@ download_lvms_operator_bundle_manifest(){
...
+  local kind=$1
+  local file=$2
+  local namespace=$3
+
+  if [[ $(yq e ".kind == \"${kind}\"" "${file}") == "true" ]]; then
+    # Check if the .metadata.namespace is not set or empty
+    if [[ $(yq e ".metadata.namespace == \"${namespace}\"" "${file}") == "false" ]]; then
+      echo "patching .metadata.namespace to \"${namespace}\" in ${file}"
+      # Set the .metadata.namespace to the specified value
+      yq e '.metadata.namespace = "'"${namespace}"'"' -i "${file}"
```
copejon, line 187: Nit - The bashy quote wrapping could be cleaned up as:
`ns=$namespace yq e '.metadata.namespace = strenv(ns)' -n`

pkg/components/storage.go
```diff
@@ -121,25 +157,168 @@ func startCSIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
...
+	} else {
+		copyDefaultCfg(runtimeCfg, cancel)
+	}
+
+	// Create new watcher.
+	go func(ctx context.Context) {
+		watcher, err := fsnotify.NewWatcher()
+		if err != nil {
+			klog.Errorf("unable to set up file watcher: %v", err)
+			cancel(err)
```
copejon, line 221: No point in passing err to cancel() if it's already logged, and no child contexts access the cause.

copejon:
A couple nits, but nothing to hold up the PR over

copejon:
Looks great!

/lgtm

PR #3536, opened 2024-06-24
NO-ISSUE: Add --showduplicates to `dnf repository-packages`

pmtk:
If newer CRI-O is installed than CRI-O present in the inspected repository then DNF will not consider it, meaning it will pretend like there is no CRI-O in the inspected repository.

By adding --showduplicates DNF will show information about the CRI-O in the repository regardless if newer is installed on the system or not.


pacevedom (approved):
/lgtm

PR #3547, opened 2024-06-26
specify the ocp version when running check-payload for 4.15

eslutsky:


eslutsky:
/close
no longer needed, the fix upstream handled the issues we were having.

PR #3557, opened 2024-06-28
NO-ISSUE: Add 4.15 rhocp repo to configure-vm.sh

pacevedom:
Manually add this repo to the configure-vm.sh script. The get_latest_rhocp_repo has been returning 4.15 and inadvertently enabling that repo until 4.16 went GA and the repo was publicly available. Now that this is the case, the scenario preparation for upgrade tests fails, as there is no repo configured to pull those packages.
This PR fixes that by adding not just 4.16 repo, but also 4.15. This is a manual cherry-pick of https://github.com/openshift/microshift/pull/3556

Closes #<Issue Number>


pacevedom:
/hold
rpm-install needs to update repos, as it does not get the right one now.

PR #3585, opened 2024-07-04
USHIFT-3730: Introduce common versions file for image builds

ggiguash:


test/scenarios/el94-src@rpm-install.sh
```diff
@@ -40,16 +37,12 @@ scenario_run_tests() {
...
-    local -r previous_version_repo=$(get_rel_version_repo "${previous_minor_version}")
-    local -r previous_version_repo_url=$(echo "${previous_version_repo}" | cut -d, -f2)
 
     # Enable the repositories with the dependencies.
 
     # RHOCP for current version might not exist yet. Then, just use whatever we can get for previous minor.
-    local -r rhocp_current=$("${ROOTDIR}/scripts/get-latest-rhocp-repo.sh" "${current_minor_version}" || true)
-    local -r rhocp_previous=$("${ROOTDIR}/scripts/get-latest-rhocp-repo.sh" "${previous_minor_version}")
+    local -r rhocp_current=$("${ROOTDIR}/scripts/get-latest-rhocp-repo.sh" "${MINOR_VERSION}" || true)
+    local -r rhocp_previous=$("${ROOTDIR}/scripts/get-latest-rhocp-repo.sh" "${PREVIOUS_MINOR_VERSION}")
```
pmtk, line 45: Would this work?

```suggestion
    local -r rhocp_current="${CURRENT_RELEASE_REPO}"
    local -r rhocp_previous="${PREVIOUS_RELEASE_REPO}"
```

and below:

```
-        if [[ "${rhocp}" =~ ^[0-9]{2} ]]; then
+        if [[ "${rhocp}" =~ ^rhocp ]]; then
```

test/bin/pyutils/build_bootc_images.py
```diff
@@ -111,78 +111,54 @@ def get_rhocp_beta_url_if_available(ver):
...
+    CURRENT_RELEASE_VERSION = common.get_env_var('CURRENT_RELEASE_REPO')
+    PREVIOUS_RELEASE_REPO = common.get_env_var('PREVIOUS_RELEASE_REPO')
+    PREVIOUS_RELEASE_VERSION = common.get_env_var('PREVIOUS_RELEASE_VERSION')
+    YMINUS2_RELEASE_REPO = common.get_env_var('YMINUS2_RELEASE_REPO')
+    YMINUS2_RELEASE_VERSION = common.get_env_var('YMINUS2_RELEASE_VERSION')
+    RHOCP_MINOR_Y = common.get_env_var('RHOCP_MINOR_Y')
+    RHOCP_MINOR_Y_BETA = common.get_env_var('RHOCP_MINOR_Y_BETA')
+    RHOCP_MINOR_Y1 = common.get_env_var('RHOCP_MINOR_Y1')
+    RHOCP_MINOR_Y1_BETA = common.get_env_var('RHOCP_MINOR_Y1_BETA')
+    RHOCP_MINOR_Y2 = common.get_env_var('RHOCP_MINOR_Y2')
```
pmtk, line 146: Do we need those? I believe we use gomplate with .Env and I don't see these being used anywhere else.

test/bin/build_images.sh
```diff
@@ -723,44 +679,11 @@ if [ -z "${release_info_rpm_base}" ]; then
...
-
-# For Y-2, there will always be a real repository, so we can always
-# set the template variable for enabling that package source and use
-# the well-known name of that repo instead of figuring out the URL.
-yminus2_version_repo=$(get_rel_version_repo "${YMINUS2_MINOR_VERSION}")
-YMINUS2_RELEASE_VERSION=$(echo "${yminus2_version_repo}" | cut -d, -f1)
-YMINUS2_RELEASE_REPO="$(get_ocp_repo_name_for_version ${YMINUS2_MINOR_VERSION})"
-RHOCP_MINOR_Y2="${YMINUS2_MINOR_VERSION}"
+export SOURCE_VERSION
+export SOURCE_VERSION_BASE
```
pmtk, line 686: Why are those variables not in common_version.sh? They're duplicated in build_bootc_images.py

test/bin/scenario.sh
```diff
@@ -692,31 +694,13 @@ current_version() {
...
-# version is "4.16.0", reports "14".
-yminus2_minor_version() {
-    echo $(( $(current_minor_version) - 2 ))
-}
-
-# Function to report the *next* minor version. If the current
-# version is "4.14.5", reports "15".
-next_minor_version() {
-    echo $(( $(current_minor_version) + 1 ))
+    local -r release_info_rpm=$(find "${LOCAL_REPO}" -name 'microshift-release-info-*.rpm' | sort | tail -n 1)
```
pmtk, line 698: It's not strictly part of the PR, but I find this function name confusing, especially with the comment. It sounds like it gets version of current release's released RPMs, not version of locally built RPMs from source.

test/bin/build_images.sh
ggiguash: To calculate these vars, we need access to LOCAL_REPO, so there will be dependency on common.sh file, which I wanted to avoid.

ggiguash: I can add it if you think it's a good idea.

test/bin/pyutils/build_bootc_images.py
ggiguash: We do not need those atm, but I added them just in case to cover the full list. 
We're about to add upgrade tests to bootc, and there vars may be required.

test/bin/scenario.sh
ggiguash: It's only used in one place. Let me rename it.

test/bin/pyutils/build_bootc_images.py
pmtk: In which place do you foresee this variables being used? If gomplate, then it already has access to them, right?

test/bin/build_images.sh
pmtk: I don't know. We removed duplication in other places so it's a bit asymmetrical, but I'll leave decision to you, you have better knowledge of the refactor

test/scenarios/el94-src@rpm-install.sh
```diff
@@ -36,37 +36,35 @@ scenario_remove_vms() {
...
+name=Beta rhocp RPMs for RHEL 9
+baseurl=${rhocp}
 enabled=1
 gpgcheck=0
 skip_if_unavailable=0
 EOF
             copy_file_to_vm host1 "${tmp_file}" "${tmp_file}"
-            run_command_on_vm host1 "sudo cp ${tmp_file} /etc/yum.repos.d/${OCP_REPO_NAME}.repo"
-            rm "${tmp_file}"
+            run_command_on_vm host1 "sudo cp ${tmp_file} /etc/yum.repos.d/${ocp_repo_name}.repo"
```
pmtk, line 66: Oh, this is not right.
We want to enable both repositories, but using the same filename we overwrite one with the other, rather than having both.

test/bin/build_images.sh
ggiguash: I think we should leave it the way it is. It's a minor code duplication, so I prefer not to introduce the dependency.
Maybe, we will do this later on.

test/bin/pyutils/build_bootc_images.py
ggiguash: I agree. Let me remove unused variables.

PR #3595, opened 2024-07-07
USHIFT-3734: Avoid using build image dnf wrapper script

ggiguash:
ART changed build images to use `dnf` wrapper script. 
Fix the MicroShift build image configuration to use the `dnf.real` executable directly.

Other minor enhancements include adding comments and build command optimization.


pacevedom (commented):
/lgtm

PR #3596, opened 2024-07-07
USHIFT-2925: Start using hypervisor container registry for ostreecontainer in kickstart

ggiguash:


test/bin/scenario.sh
```diff
@@ -17,6 +17,7 @@ source "${SCRIPTDIR}/common.sh"
 DEFAULT_BOOT_BLUEPRINT="rhel-9.2"
 LVM_SYSROOT_SIZE="10240"
 WEB_SERVER_URL="http://${VM_BRIDGE_IP}:${WEB_SERVER_PORT}"
+BOOTC_REGISTRY_URL="${VM_BRIDGE_IP}:5000"
```
pacevedom, line 20: I wonder if we could use the same registry for both mirrors, but this shouldnt stop the PR.

pacevedom (approved):
/lgtm

PR #3602, opened 2024-07-08
USHIFT-3736: Force NM configuration to allow ipv6

pacevedom:
/cc @pliurh 

Closes #<Issue Number>


pliurh:
LGTM. However, it may be worth mentioning why we want NM to ignore those interfaces in the commit info.

PR #3608, opened 2024-07-09
USHIFT-3742: test if all the running pods are annotated for workload partitioning

eslutsky:

Closes #<Issue Number>

### DIFF
--- a/test/suites/standard2/workload-partitioning.robot
+++ b/test/suites/standard2/workload-partitioning.robot
@@ -1,0 +1,47 @@
+*** Settings ***
+Documentation       Tests for Workload partitioning
+
+Resource            ../../resources/common.resource
+Resource            ../../resources/microshift-process.resource
+Resource            ../../resources/microshift-network.resource
+
+Suite Setup         Setup Suite With Namespace
+Suite Teardown      Teardown Suite With Namespace
+
+
+*** Test Cases ***
+Control Plane Pods Must Be Annotated
+    [Documentation]    Verify that all the Control Plane pods are properly annotated.
+    All Pods Should Be Annotated As Management
+
+
+*** Keywords ***
+All Pods Should Be Annotated As Management
+    [Documentation]    Obtains list of Deployments created by CSV.
+    ${pods_raw}=    Oc Get All Pods
+    @{pods}=    Split String    ${pods_raw}
+    FOR    ${pod}    IN    @{pods}
+        ${ns}    ${pod}=    Split String    ${pod}    \@
+        Pod Must Be Annotated    ${ns}    ${pod}
+    END
+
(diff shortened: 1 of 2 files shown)


test/suites/standard2/workload-partitioning.robot
```diff
@@ -0,0 +1,44 @@
...
+
+
+*** Keywords ***
+All Pods Should Be Annotated As Management
+    [Documentation]    Obtains list of Deployments created by CSV.
+    ${pods_raw}=    Oc Get All Pods
+    @{pods}=    Split String    ${pods_raw}
+    FOR    ${pod}    IN    @{pods}
+        ${ns}    ${pod}=    Split String    ${pod}    \@
+        ${csv_dss}=    Check Pod Management Annotation    ${ns}    ${pod}
```
ggiguash, line 25: Does not look like the check function returns any values, right?

test/suites/standard2/workload-partitioning.robot
```diff
@@ -0,0 +1,44 @@
...
+    END
+
+Check Pod Management Annotation
+    [Documentation]    Check management annotation for specified pod and namespace.
+    [Arguments]    ${ns}    ${pod}
+    ${networks_str}=    Oc Get JsonPath
```
ggiguash, line 31: Nit: Why the var name is `networks_str`?

test/suites/standard2/workload-partitioning.robot
```diff
@@ -0,0 +1,44 @@
...
+    ...    .metadata.annotations.target\\.workload\\.openshift\\.io/management
+    Should Not Be Empty    ${networks_str}
+
+Oc Get All Pods
+    [Documentation]    Runs 'oc get pods -A`
+    ...    Returns the command output as formatted string <name-space>@<pod-name>
+
+    ${data}=    Run With Kubeconfig
+    ...    oc get pods -A -o jsonpath='{range .items[*]}{.metadata.namespace}{"@"}{.metadata.name}{"\\n"}{end}'
+    RETURN    ${data}
```
ggiguash, line 44: Can we reuse the existing `Oc Get JsonPath` function from `oc.resource`?

eslutsky: we could, but the original function `Oc Get JsonPath` only works with a specified namespace, the idea here is that we want to get all the pods across all namespaces.

eslutsky: so we would need to refactor it to support `-A`

ggiguash: I think let's refactor rather than creating a copy of the code. 

The easiest way should be to check the namespace argument value and if it's empty, pass -A to the command.

PR #3625, opened 2024-07-12
USHIFT-3749: Prerequisites for ostree-to-bootc upgrade tests

ggiguash:
Before the actual tests are implemented, it is necessary to update ostree blueprints to contain `bootc` package. This needs to be done in a separate PR because those images are cached and otherwise, local changes are overwritten by the cache update.

This PR also contains the following minor fixes:
* ~~Additional image of rhel94 with y-2 released package including bootc~~
* Fixes in  `microshift-test-agent.sh` to prevent it from reporting errors on first boot
* Added `microshift-test-agent` service to the rhel94 bootc base image
* Removed workaround for MicroShift 4.13 backup as we no longer test this version


test/image-blueprints/layer1-base/group2/rhel94-microshift-yminus2.toml
```diff
@@ -0,0 +1,55 @@
+name = "rhel-9.4-microshift-4.{{ .Env.YMINUS2_MINOR_VERSION }}"
```
pmtk, line 1: Shouldn't this be 9.3?
4.15 is not supported on 9.4, but since bootc is only participating in the upgrade image (`9.3 + 4.15 -> bootc 9.4 + 4.17`), then it should be fine?

ggiguash: There's a problem with 9.3 OS because it does not container `bootc` RPM, so we have to start from 9.4
I guess, that answers the question, so I'd better remove this image at all?

pmtk: I think there's no point in testing unsupported scenario

PR #3634, opened 2024-07-15
USHIFT-3749: Implement ostree-to-bootc image basic upgrade tests

ggiguash:
Prerequisites required for image cache-related changes. Otherwise, the cache overwrites local images and packages will be missing in the ostree base commit.
* https://github.com/openshift/microshift/pull/3625
* https://github.com/openshift/microshift/pull/3633


test/image-blueprints/layer5-bootc/group0/rhel94-ostree-microshift-previous-minor.container-encapsulate
```diff
@@ -0,0 +1 @@
+rhel-9.4-microshift-4.16
```
ggiguash, line 1: Note that the build_bootc_images.py script does not support templating in container files yet.
We can implement it in another PR and change this and few other locations to make them more dynamic.
See https://issues.redhat.com/browse/USHIFT-3764

test/bin/pyutils/build_bootc_images.py
```diff
@@ -346,6 +346,100 @@ def process_image_bootc(groupdir, bootcfile, dry_run):
...
 
 
+def process_container_encapsulate(groupdir, containerfile, dry_run):
+    ce_path = os.path.join(groupdir, containerfile)
+    ce_outname = os.path.splitext(containerfile)[0]
+    ce_outdir = os.path.join(BOOTC_IMAGE_DIR, ce_outname)
+    ce_logfile = os.path.join(BOOTC_IMAGE_DIR, f"{ce_outname}.log")
+    ce_targetimg = os.path.join(ce_outdir, "manifest.json")
+    ce_tagname = "build_image_tag"
+    ce_tagval = f"localhost/{ce_outname}:latest"
```
pmtk, line 356: Do we need to prefix all of these variables?

test/bin/pyutils/build_bootc_images.py
```diff
@@ -346,6 +346,100 @@ def process_image_bootc(groupdir, bootcfile, dry_run):
...
+
+    def get_images_by_label():
+        images_args = [
+            "sudo", "podman", "images",
+            "--filter", f"label={ce_tagname}={ce_tagval}",
```
pmtk, line 361: What is advantage of using labels in this process instead of using tags directly? Is it producing more than 1 image?

test/suites/upgrade/upgrade-successful-bootc.robot
```diff
@@ -0,0 +1,53 @@
...
+    ${is_bootc}=    Is System Bootc
+    ${is_ostree}=    Is System OSTree
+    ${cond}=    Evaluate    ${is_bootc} == ${FALSE} and ${is_ostree} == ${FALSE}
+    Should Not Be True    ${cond}    The system should either be bootc or ostree
+
+    Wait Until Greenboot Health Check Exited
+
+    ${future_backup}=    Get Future Backup Name For Current Boot
+    Deploy Bootc Commit Not Expecting A Rollback    ${BOOTC_REGISTRY}    ${TARGET_REF}
+    Backup Should Exist    ${future_backup}
```
pmtk, line 36: Should we check greenboot for system readiness?

test/bin/pyutils/build_bootc_images.py
ggiguash: It's a copy-paste from the style of other similar functions. Strictly speaking, there's no need to prefix local vars.
Just a "safe-guard" from a copy-paste.

test/bin/pyutils/build_bootc_images.py
ggiguash: When running `rpm-ostree compose container-encapsulate` command, we need to somehow `mark` the produced image and it seems to only provides this functionality using `--label` option.
Once the image is produced, subsequent commands find it by the label.

pmtk: Gotcha, we cannot `podman load` with a `--tag` - without the labels we wouldn't find what we just loaded.

test/suites/upgrade/upgrade-successful-bootc.robot
pmtk: You are correct!

PR #3635, opened 2024-07-15
USHIFT-3476: Test harness ipv6 adaptations

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -83,9 +83,10 @@ copy_file_to_vm() {
     local -r local_filename="$2"
     local -r remote_filename="$3"
 
-    local ip=$(get_vm_property "${vmname}" ip)
-    if [ "$ip" != "${ip#*:[0-9a-fA-F]}" ]; then
-      ip="[$ip]"
+    local ip
+    ip=$(get_vm_property "${vmname}" ip)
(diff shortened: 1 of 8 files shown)


pacevedom:
/hold
just found an error coming from https://github.com/openshift/microshift/pull/3636

test/bin/scenario.sh
```diff
@@ -83,12 +83,31 @@ copy_file_to_vm() {
...
+    local -r vmname="$1"
+    local -r remote_filename="$2"
+    local -r local_filename="$3"
+
+    local ip
+    ip=$(get_vm_property "${vmname}" ip)
+    if [ "${ip}" != "${ip#*:[0-9a-fA-F]}" ]; then
+      ip="[${ip}]"
+    fi
+    local -r ssh_port=$(get_vm_property "${vmname}" ssh_port)
```
copejon, line 106: It's probably not likely to fail, but it'd be good to follow the same pattern used for `ip` to make sure the exit status of the subproc is caught.

copejon:
It looks good. A minor nit, but not a blocker. I'll label and hold incase you want to fix it, else feel free to drop the hold

/lgtm
/hold

test/bin/scenario.sh
pacevedom: Created https://issues.redhat.com/browse/USHIFT-3769 for this purpose. Thanks!

PR #3647, opened 2024-07-17
USHIFT-3761: Set default networks separately

pacevedom:
Allows configuring only one of them, keeping old behavior for both ipv4 and ipv6.

Closes #<Issue Number>


pacevedom:
need to vendor etcd, will fix.

PR #3653, opened 2024-07-18
USHIFT-3748: Check for oc utility on hypervisor host, install with fetch_tools.sh

vanhalenar:


test/bin/scenario.sh
```diff
@@ -772,6 +772,13 @@ run_tests() {
     fi
     record_junit "${vmname}" "robot_framework_installed" "OK"
 
+    # Make sure oc command is available
+    if ! command -v oc &> /dev/null
+    then
+        error "OpenShift Client package not installed, install it with ${ROOTDIR}/scripts/fetch_tools.sh oc"
+        exit 1
```
copejon, line 779: Please include the junit entry (e.g. L770):
```
record_junit "${vmname}" "robot_framework_installed" "FAILED"
```

scripts/fetch_tools.sh
```diff
@@ -216,6 +216,17 @@ gettool_awscli() {
     popd &>/dev/null
 }
 
+gettool_oc() {
+    local ver="4.16.3"
+    pushd "${WORK_DIR}" &>/dev/null
+
+    curl -s "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${ver}/openshift-client-linux-${ver}.tar.gz" -o "oc-${ver}-linux.tar.gz"
```
copejon, line 223: You'll need to dynamically set the arch here since MicroShift supports both x86_64 and arm64

copejon: Also, probably want to use the `-f` or `--fail-with-body` curl flags, otherwise http codes >400 will still be considered successes.

scripts/fetch_tools.sh
```diff
@@ -216,6 +216,25 @@ gettool_awscli() {
...
+    local ver="4.16.3"
+    
+    declare -A arch_map=(
+    ["x86_64"]="x86_64"
+    ["aarch64"]="arm64")
+
+    local arch="${arch_map[${ARCH}]}"
+
+    pushd "${WORK_DIR}" &>/dev/null
+    echo "https://mirror.openshift.com/pub/openshift-v4/${arch}/clients/ocp/${ver}/openshift-client-linux-${ver}.tar.gz"
```
pmtk, line 229: I think we can remove this line

scripts/fetch_tools.sh
```diff
@@ -216,6 +216,25 @@ gettool_awscli() {
...
+
+    curl -s -f "https://mirror.openshift.com/pub/openshift-v4/${arch}/clients/ocp/${ver}/openshift-client-linux-${ver}.tar.gz" -L -o "oc-${ver}-linux.tar.gz"
+    tar xvzf oc-${ver}-linux.tar.gz
+    sudo cp oc /usr/bin/oc
```
pmtk, line 233: I think the archive also contains `kubectl`, let's copy it as well

test/bin/scenario.sh
vanhalenar: Included, don't know why this was not marked as outdated

PR #3654, opened 2024-07-18
NO-ISSUE: Missing persistent journal on 1st boot when installing using rhel94 iso

pmtk:
For some reason, when using only `Storage=persistent`, the journal on first boot is not persistent. This causes some upgrade tests fail when they depend on counting amount of boots.


pacevedom (approved):
/lgtm

PR #3656, opened 2024-07-18
USHIFT-3463: Add microshift-tuned daemon for unattended TuneD profile activation

pmtk:


packaging/tuned/microshift-tuned.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=MicroShift TuneD helper
+Before=microshift.service
+After=tuned.service
+Requires=tuned.service
+
+[Service]
+ExecStart=/usr/bin/microshift-tuned
```
ggiguash, line 8: This script may reboot host. Should we provide a mandatory command line for the script to actually do something? Otherwise, if someone runs it, they can end up rebooting accidentally.

test/scenarios/el94-src@low-latency.sh
```diff
@@ -15,5 +15,14 @@ scenario_remove_vms() {
 }
 
 scenario_run_tests() {
-    run_tests host1 suites/tuned/
+    # Should not be ran immediately after creating VM because of
+    # microshift-tuned rebooting the node to activate the profile.
```
ggiguash, line 19: How do we time this after creating the VM, then?

ggiguash: Should we add some kind of a loop checking that the profile is active?

test/suites/tuned/microshift-tuned.robot
```diff
@@ -0,0 +1,97 @@
...
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+
+*** Test Cases ***
+Current Profile Should Be Hashed
+    [Documentation]    We expect that microshift-tuned keeps a hash of the profile and variables.
+
+    SSHLibrary.File Should Exist    /var/lib/microshift-tuned.yaml
```
ggiguash, line 17: I'm not sure how the name / docs reflect the contents of the test. Could you explain?

test/suites/tuned/microshift-tuned.robot
```diff
@@ -0,0 +1,97 @@
...
+
+Profile Is Already Active But Cache File Is Missing
+    [Documentation]    If profile is already active, but cache file is missing,
+    ...    we expect microshift-tuned to reactivate it, reboot, and store the hashes.
+
+    Command Should Work    rm /var/lib/microshift-tuned.yaml
```
ggiguash, line 23:
```suggestion
    Command Should Work    rm -f /var/lib/microshift-tuned.yaml
```

test/suites/tuned/microshift-tuned.robot
```diff
@@ -0,0 +1,97 @@
...
+    [Documentation]    Verify that microshift-tuned will activate any given tuned profile.
+
+    Command Should Work    sed -i 's/profile:.*/profile: virtual-guest/' /etc/microshift/tuned.yaml
+    Restart MicroShift-Tuned Not Expecting Reboot
+    ${active}=    Command Should Work    tuned-adm active
+    Should Contain    ${active}    Current active profile: virtual-guest
+
+MicroShift-Tuned Requires Config To Function
+    [Documentation]    Verify that missing configuration will be fatal.
+    Command Should Work    rm /etc/microshift/tuned.yaml
```
ggiguash, line 56:
```suggestion
    Command Should Work    rm -f /etc/microshift/tuned.yaml
```

packaging/tuned/microshift-tuned.service
pmtk: You mean that dry run is default and not dry run is protected with `--something`?

test/suites/tuned/microshift-tuned.robot
```diff
@@ -0,0 +1,97 @@
...
+    Sleep    30s
+
+    Wait Until Keyword Succeeds    5m    15s
+    ...    Is System Rebooted    ${bootid}
+
+Restart MicroShift-Tuned Not Expecting Reboot
+    [Documentation]    TODO
+    ${bootid}=    Get Current Boot Id
+    Command Should Work    systemctl restart microshift-tuned.service
+    Sleep    60s
```
ggiguash, line 95: Could we think of making this check a bit faster? Is there any other way we can get an indication that the microshift-tuned script "decided" not to reboot?

packaging/tuned/microshift-tuned.service
pmtk: There's a variable in the config which controls if reboot is allowed. Wouldn't it be slightly confusing if it was defined `False` but our service always had `--reboot-if-allowed`?

ggiguash: We could call it `--live-run` or `--execute`.

test/suites/tuned/microshift-tuned.robot
pmtk: Can you add a rationale for this change? I'm not opposing, just want to know full story

test/scenarios/el94-src@low-latency.sh
pmtk: Added a check to see if we have `>= 2` boots. `>` so we can run the script repeatedly in local env.

test/suites/tuned/microshift-tuned.robot
pmtk: Is it better now?

test/suites/tuned/microshift-tuned.robot
pmtk: I changed it but I'm not sure about robustness, we'll see

test/suites/tuned/microshift-tuned.robot
ggiguash: Sometime rm is aliased in the root account as `rm -i`

```
# alias rm
alias rm='rm -i'
```

packaging/tuned/microshift-tuned.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        prog='MicroShift TuneD',
+        formatter_class=argparse.RawTextHelpFormatter,
+        description=f"""Daemon for unattended TuneD profile activation.
+
+When program starts, it compares configuration and system state.
+If the requested profile is not activate, it will be activated,
```
ggiguash, line 183:
```suggestion
If the requested profile is not active, it will be activated,
```

PR #3658, opened 2024-07-18
USHIFT-3001: Add initial RF ipv6 tests

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/test/suites/ipv6/singlestack.robot
+++ b/test/suites/ipv6/singlestack.robot
@@ -7,15 +7,16 @@ Resource            ../../resources/ostree-health.resource
 Resource            ../../resources/microshift-network.resource
 Library             ../../resources/libipv6.py
 
-
 Suite Setup         Setup
 Suite Teardown      Teardown
 
-Test Tags           network    ipv6
+Test Tags           ipv6    network
+
 
 *** Variables ***
 ${USHIFT_HOST}      ${EMPTY}
 ${HOSTNAME}         hello-microshift.cluster.local
+
 
 *** Test Cases ***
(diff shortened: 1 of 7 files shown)


pacevedom:
/hold
Need to fix a routing issue.

test/suites/ipv6/singlestack.robot
```diff
@@ -0,0 +1,102 @@
...
+    ${match_string}=    Remove String    ${match_string}    "
+    ${response}=    Run With Kubeconfig    oc get svc,ep -l app\=hello-microshift -n ${NAMESPACE}
+    Should Be Equal As Strings    ${match_string}    ${response}    strip_spaces=True
+
+DNS Entry For Route Should Resolve
+    [Documentation]    Resolve hello-microshift route via mDNS from the hypervisor/RF runner.
+    ...    Expects RF runner host has opened port 5353 for libvirt zone.
+
+    ${result}=    Run Process
+    ...    avahi-resolve-host-name ${HOSTNAME}
```
pmtk, line 87: I see it returned `hello-microshift.cluster.local	2001:db8:ca7:ff::1522`. Do think it's possible to have a race condition with another test that would also expose `hello-microshift.cluster.local`?

Should we check if the IP we got is indeed IPv6?

pacevedom: Good points, did both.

pacevedom:
Periodic tests failing because of the crel upgrade issue described here: https://redhat-internal.slack.com/archives/C0310LGMQQY/p1721797490189859

PR #3672, opened 2024-07-23
USHIFT-1545: Full VM name validation

vanhalenar:


test/bin/scenario.sh
```diff
@@ -38,6 +38,48 @@ full_vm_name() {
...
 }
 
+# hostname validation
+# based on https://github.com/rhinstaller/anaconda/blob/c95142f76735a2e9ae6d845f8569d46632ddd619/pyanaconda/network.py#L96-L120
+validate_full_vm_name() {
+    local vm_name="$1"
+
+    if [ -z "${vm_name}" ]; then
+        error "VM hostname cannot be empty string" 
+        record_junit "${vmname}" "vm_name_validation" "FAILED"
```
pmtk, line 48: Similar thing in other places
```suggestion
        record_junit "${vm_name}" "vm_name_validation" "FAILED"
```

test/bin/scenario.sh
```diff
@@ -227,6 +269,8 @@ prepare_kickstart() {
     local -r vm_hostname="${full_vmname/./-}"
     local -r hostname=$(hostname)
 
+    validate_full_vm_name $full_vmname
```
pmtk, line 272: I think we want to validate `vm_hostname` since it'll be used for hostname.
```suggestion
    validate_full_vm_name "${vm_hostname}"
```

test/bin/scenario.sh
```diff
@@ -38,6 +38,48 @@ full_vm_name() {
...
+    fi
+
+    if [[ $vm_name =~ ^\. ]]; then
+        error "VM hostname cannot start with '.'" 
+        record_junit "${vmname}" "vm_name_validation" "FAILED"
+        exit 1
+    fi
+
+    # split by '.' and check every section separately
+    IFS='.' read -ra ARR <<< "$vm_name"
```
pmtk, line 65: If we validate `$vm_hostname` instead of `$full_vmname`, then we can remove this loop and flatten and merge the checks.

test/bin/scenario.sh
```diff
@@ -38,6 +38,38 @@ full_vm_name() {
...
+        error "VM hostname is too long" 
+        record_junit "${vm_name}" "vm_hostname_validation" "FAILED"
+        exit 1
+    fi
+
+    if [[ "${vm_name}" =~ ^\. ]]; then
+        error "VM hostname cannot start with '.'" 
+        record_junit "${vm_name}" "vm_hostname_validation" "FAILED"
+        exit 1
+    fi
```
ggiguash, line 62: Do we need this check? I mean if the name starts from `.`, wouldn't the next check fail?

pmtk: Yes, however this check has explicit error message about what is wrong

pmtk: On the other hand: before passing the hostname to this function, we replace all `.` with `-`, so this won't be executed unless we do some refactoring and mess up

vanhalenar: Okay, I removed it

PR #3690, opened 2024-07-29
USHIFT-3008: Getting started bootc documentation

ggiguash:


docs/contributor/image_mode.md
```diff
@@ -0,0 +1,211 @@
...
+deploy, and manage Red Hat Enterprise Linux as a bootc container image. MicroShift
+build and deployment procedures can utilize bootc containers to benefit from this
+technology.
+
+See [Image mode for Red Hat Enterprise Linux](https://developers.redhat.com/products/rhel-image-mode/overview)
+for more information.
+
+The procedures described in this document require the following setup:
+* A `physical hypervisor host` with the [libvirt](https://libvirt.org/) virtualization
+platform for starting virtual machines that run RHEL OS containing MicroShift binaries
```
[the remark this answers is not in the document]
ggiguash: The current style mimics other already-existing documents in the upstream repo. I will work with the docs team to make production docs more generic whenever necessary.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,211 @@
...
+
+```
+sudo podman run --rm -it --privileged \
+    -v /var/lib/containers:/var/lib/containers \
+    --name "${IMAGE_NAME}" \
+    "${IMAGE_NAME}"
+```
+
+Log into the container host when prompted using the `redhat:redhat` credentials and
+save your pull secret in the `/etc/crio/openshift-pull-secret` file.
```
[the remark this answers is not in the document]
ggiguash: There are a few ways to implement this requirement. The other getting started document adds a pull secret as a post-install stage, so I tried to mimic the same approach. 

It helps to simplify the flow and concentrate on the main bootc-related parts.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,211 @@
...
+
+Run the following commands to start the MicroShift service and verify that it is
+running properly.
+
+```
+sudo systemctl start microshift
+sudo oc get pods -A -w \
+    --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig
+```
```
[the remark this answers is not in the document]
ggiguash: I would like to avoid adding pods output, but I'll rephrase the wording a bit to make it clearer.

docs/contributor/image_mode.md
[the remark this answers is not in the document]
DanielFroehlich: for me, it makes the flow more complex, and distract from the bootc related parts. When I read from top to bottom, I am surprised on why the container is started. 
And: how and why is the working anyway? I thought container images are immutable? How is this change retained after shutdown?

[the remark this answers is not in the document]
ggiguash: The change are not retained after the shutdown. I'll think on how to add this file to the image in a more user-friendly way.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,211 @@
...
+RUN dnf config-manager \
+        --set-enabled rhocp-4.16-for-rhel-9-x86_64-rpms \
+        --set-enabled fast-datapath-for-rhel-9-x86_64-rpms
+RUN dnf install -y firewalld microshift && \
+    systemctl enable microshift && \
+    dnf clean all
+
+# Add 'redhat:redhat' user
+RUN useradd -G wheel redhat && \
+    echo "redhat:redhat" | chpasswd
```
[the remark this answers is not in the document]
ggiguash: The rest of the docs use user:pass for login. I'll use an ancrypted password, like in other places.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
+# Image Mode for MicroShift
+
+Image mode is a new approach to operating system deployment that lets users build,
+deploy, and manage Red Hat Enterprise Linux as a bootable container (`bootc`) image.
```
pmtk, line 4: Should we mentioned that "bootable container images" are based on OCI images so it's explicit that there's nothing new about build process (docker/podman build) or transport (regular docker registries) and it's more about how bootc handles all of it?

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
...
+    dnf clean all
+
+# Create a default 'redhat' user, allowing it to run sudo commands
+ARG USER_PASSWD=\$5\$XDVQ6DxT8S5YWLV7\$8f2om5JfjK56v9ofUkUAwZXTxJl3Sqnc9yPnza4xoJ0
+RUN useradd -m -d /var/home/redhat -G wheel -p ${USER_PASSWD} redhat
+
+# Configure OpenShift pull secret
+ARG PULL_SECRET
+RUN echo $PULL_SECRET > /etc/crio/openshift-pull-secret && \
+    chmod 600 /etc/crio/openshift-pull-secret
```
pmtk, line 51: I would expect an `ADD` in such context but I guess for such guide it's easier to pass with --build-arg rather than instruct to copy the file into the build context.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
...
+# Configure firewall
+RUN firewall-offline-cmd --zone=public --add-port=22/tcp && \
+    firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16 && \
+    firewall-offline-cmd --zone=trusted --add-source=169.254.169.1 && \
+    firewall-offline-cmd --zone=public --add-port=80/tcp && \
+    firewall-offline-cmd --zone=public --add-port=443/tcp && \
+    firewall-offline-cmd --zone=public --add-port=30000-32767/tcp && \
+    firewall-offline-cmd --zone=public --add-port=30000-32767/udp
+
+# Recursively make the root filesystem subtree shared as required by OVN images
```
pmtk, line 62: Is it requirement of OVN images or containers? Do you know what's so special about them? I think lvms also mounts considerable amount of `/` subpaths

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+Log into the container host when prompted using the `redhat:redhat` credentials.
+Run the following commands to start the MicroShift service and verify that all
+the pods are up and running without errors.
+
+```
+sudo systemctl start microshift
+sudo oc get pods -A -w \
+    --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig
```
pmtk, line 116: Does this mean that we can run microshift inside a container? I thought ovn wanted to mess with networking to much so it wasn't possible?

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
...
+Log into the `development virtual machine` with the `microshift` user credentials.
+
+Run the following commands to log into the remote registry and publish the image.
+
+```
+REGISTRY_URL=quay.io
+REGISTRY_IMG=myorg/mypath/"${IMAGE_NAME}"
+
+sudo podman login "${REGISTRY_URL}"
+sudo podman push localhost/"${IMAGE_NAME}" "${REGISTRY_URL}/${REGISTRY_IMG}"
```
pmtk, line 134: FYI (but you probably know this already): `podman build --tag "${REGISTRY_URL}/${REGISTRY_IMG}"` to avoid intermediate `localhost/` tag.

pmtk (commented):
only minor comments

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
...
+
+Log into the container host when prompted using the `redhat:redhat` credentials.
+Run the following commands to start the MicroShift service and verify that all
+the pods are up and running without errors.
+
+```
+sudo systemctl start microshift
+sudo oc get pods -A -w \
+    --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig
+```
```
praveenkumar, line 117: I can see all the pods are running except the CSI one and this is because the running container doesn't have the required the lvm layout.

praveenkumar: Another thing which I observe is the size of the image which we end up having because of bootc is around 2.18GB and this is justified if this is used to boot it up as VM or run it on hardware but if we are trying to run it as container only then it is not justified because image which I created for our POC was ~ 1GB.

praveenkumar: `podman run` command doesn't start the VM if the image is bootc I think that is different project https://github.com/containers/podman-bootc and this is mostly targeted for podman desktop as extension which need to have podman machine.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Build Image
+
+Create the `Containerfile` file with the following contents.
+
+```
+FROM registry.redhat.io/rhel9/rhel-bootc:9.4
+
+RUN dnf config-manager \
+        --set-enabled rhocp-4.16-for-rhel-9-x86_64-rpms \
```
praveenkumar, line 38: for arch we can have use it like `rhocp-4.16-for-rhel-9-$(uname -m)-rpms` so we can build it for arm64 also without much change (like on mac M1). Also version we can take as argument with default value set to 4.16.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
+# Image Mode for MicroShift
+
+Image mode is a new approach to operating system deployment that lets users build,
+deploy, and manage Red Hat Enterprise Linux as a bootable container (`bootc`) image.
+MicroShift build and deployment procedures can utilize bootable containers to
+benefit from this technology.
+
+This document demonstrates how to:
+* Create a bootable container image and boot a RHEL operating system from this
+image using `podman`
```
praveenkumar, line 10: It is not booting the RHEL operating system using `podman run` command but starting the container.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,217 @@
...
+    --name "${IMAGE_NAME}" \
+    "${IMAGE_NAME}"
+```
+
+Log into the container host when prompted using the `redhat:redhat` credentials.
+Run the following commands to start the MicroShift service and verify that all
+the pods are up and running without errors.
+
+```
+sudo systemctl start microshift
```
praveenkumar, line 114: We don't need this step because in our container file we are already enabling `microshift` service so it is going to start eventually. 
```
● microshift.service - MicroShift
     Loaded: loaded (/usr/lib/systemd/system/microshift.service; enabled; preset: disabled)
     Active: activating (start) since Wed 2024-07-31 11:49:12 UTC; 23s ago
   Main PID: 585 (microshift)
      Tasks: 19 (limit: 1638)
     Memory: 215.6M
        CPU: 10.931s
     CGroup: /system.slice/microshift.service
             └─585 microshift run
```

docs/contributor/image_mode.md
ggiguash: Yes, that was the original implementation by copy, but I got comments to make it easier.

docs/contributor/image_mode.md
ggiguash: One of the ovn-master containers fails if the mount is not `rshared`. 
I still need to investigate the required LVMS configuration.

docs/contributor/image_mode.md
ggiguash: We can only run MicroShift inside a `bootc` container, which includes kernel modules, so OVN can function properly.

docs/contributor/image_mode.md
ggiguash: Yes, I do this in 2 steps because I only introduce remote registry in the second section. 
Let me tweak it a bit.

docs/contributor/image_mode.md
ggiguash: 1. I'm about to add a section on CSI
2. The size of the resulting image is not something we can control much, but this is a good catch!
3. I tweaked the wording to make this more precise.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,310 @@
...
+logvol / --vgname=rhel --fstype=xfs --size=10240 --name=root
+
+# Lock root user account
+rootpw --lock
+
+# Configure network to use DHCP and activate on boot
+network --bootproto=dhcp --device=link --activate --onboot=on
+
+# Pull a `bootc` images from a remote registry
+ostreecontainer --url quay.io/myorg/mypath/microshift-4.16-bootc --no-signature-verification
```
mrguitar, line 265: You can drop --no-signature-verification

ggiguash: May I ask why it is no longer necessary?

mrguitar: I think it's this issue: https://github.com/containers/bootc/issues/218 Colin will know for sure. I can tell you if you drop that option, it will just work.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+RUN dnf install -y firewalld microshift && \
+    systemctl enable microshift && \
+    dnf clean all
+
+# Create a default 'redhat' user, allowing it to run sudo commands
+ARG USER_PASSWD=\$5\$XDVQ6DxT8S5YWLV7\$8f2om5JfjK56v9ofUkUAwZXTxJl3Sqnc9yPnza4xoJ0
+RUN useradd -m -d /var/home/redhat -G wheel -p ${USER_PASSWD} redhat
+
+# Configure OpenShift pull secret
+ARG PULL_SECRET
```
praveenkumar, line 52: After putting some thoughts, I think we shouldn't make the pull secret as part of image build process but it should be part of container run because image can be shared across and we don't want that pull secret part of that shared image. 

We can use `podman secret` to have pull secret which is mounted to `/run/secrets/pull-secret` by default and in this container file we can just as part of CMD copy that to `/etc/crio/openshift-pull-secret` location.

following I tried during POC
```
cat > entrypoint.sh << EOF
#!/bin/bash
set -e
mount --make-rshared /
cp /run/secrets/pull-secret /etc/crio/openshift-pull-secret
exec /usr/sbin/init
EOF
```
that script was part of entry point but I think we can use something for `CMD` also and then with following we can use with podman commands
```
sudo podman secret create pull-secret <pull-secret-file>
sudo podman run  --secret pull-secret ....
```
```

File: docs/contributor/image_mode.md [hunk 16]
```diff
@@ -0,0 +1,308 @@
...
+VGLOOP=$(losetup -j ${VGFILE} | cut -d: -f1)
+sudo vgcreate -f -y rhel "${VGLOOP}"
+```
+
+The device will now be shared with the newly created containers as described in
+the next section.
+
+> The following commands can be run to detach the loop device and delete the LVM
+> volume group file.
+> ```
```
miabbott, line 192: nit: add a line break before starting the "```" section

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+    -v /lib/modules:/lib/modules:ro \
+    --name "${IMAGE_NAME}" \
+    "${IMAGE_NAME}"
+```
+
+> If additional LVM volume group device was allocated as described in the
+> [Configure CSI](#configure-csi) section, the loop device should automatically
+> be shared with the container and used by the MicroShift CSI driver.
+
+Log into the container host when prompted using the `redhat:redhat` credentials.
```
miabbott, line 217: "container host" isn't defined previously and is somewhat confusing. Maybe this can be reworded such as:

> After the MicroShift `bootc` image has been successfully started, there will be a login prompt presented. Login to this running container using the `redhat:redhat` credentials.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+Run the following command to verify that all the MicroShift pods are up and running
+without errors.
+
+```
+sudo oc get pods -A -w \
+    --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig
+```
+
+> Run the `sudo shutdown now` command to stop the container host.
```
miabbott, line 226: Similar to my comment above, suggest rewording this to:

> Run the `sudo shutdown now` command to stop the MicroShift `bootc` container.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+# Configure network to use DHCP and activate on boot
+network --bootproto=dhcp --device=link --activate --onboot=on
+
+# Pull a `bootc` images from a remote registry
+ostreecontainer --url quay.io/myorg/mypath/microshift-4.16-bootc --no-signature-verification
+```
+
+The kickstart file uses a special [ostreecontainer](https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#ostreecontainer)
+directive to pull a `bootc` image from the remote registry and install it on top
+of the base RHEL operating system.
```
miabbott, line 268: > on top of the base RHEL operating system.

I don't think this is correct. When Anaconda runs and installs the operating system, it is doing a fresh install. The `ostreecontainer` directive doesn't add additional content on top of an existing OS.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Create Virtual Machine
+
+Download a RHEL boot ISO image from https://developers.redhat.com/products/rhel/download.
+Copy the downloaded file to the `/var/lib/libvirt/images` directory.
+
+Run the following commands to create a RHEL virtual machine with 2 cores, 2GB of
+RAM and 20GB of storage. The command uses the kickstart file prepared in the
+previous step to pull a `bootc` image from the remote registry and install it on
+top of the base RHEL operating system.
```
miabbott, line 280: Same comment as above re: "on top of the base RHEL operating system"

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+```
+VGFILE=/var/lib/microshift-lvm-storage.img
+VGSIZE=1G
+
+sudo truncate --size="${VGSIZE}" "${VGFILE}"
+sudo losetup -f "${VGFILE}"
+```
+
+Query the loop device name and create a free volume group on the device according
+to the MicroShift CSI driver requirements.
```
ShaunaDiaz, line 180: Where does the user find the requirements?

ShaunaDiaz (commented):
Some questions from the "eventually-docs" side.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+ARG USHIFT_VER=4.16
+RUN dnf config-manager \
+        --set-enabled rhocp-${USHIFT_VER}-for-rhel-9-$(uname -m)-rpms \
+        --set-enabled fast-datapath-for-rhel-9-$(uname -m)-rpms
+RUN dnf install -y firewalld microshift && \
+    systemctl enable microshift && \
+    dnf clean all
+
+# Create a default 'redhat' user, allowing it to run sudo commands
+ARG USER_PASSWD=\$5\$XDVQ6DxT8S5YWLV7\$8f2om5JfjK56v9ofUkUAwZXTxJl3Sqnc9yPnza4xoJ0
```
cgwalters, line 48: The chance that someone copy-pastes this is quite high; I don't think it's a good idea to have default passwords. You can make it a required parameter, or just refer to the docs for users.

docs/contributor/image_mode.md
cgwalters: But this image won't actually be *run* as a container...

There's docs for this: https://docs.fedoraproject.org/en-US/bootc/container-pull-secrets/

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+RUN useradd -m -d /var/home/redhat -G wheel -p ${USER_PASSWD} redhat
+
+# Configure OpenShift pull secret
+ARG PULL_SECRET
+RUN echo $PULL_SECRET > /etc/crio/openshift-pull-secret && \
+    chmod 600 /etc/crio/openshift-pull-secret
+
+# Configure firewall
+RUN firewall-offline-cmd --zone=public --add-port=22/tcp && \
+    firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16 && \
```
cgwalters, line 58: At least some of these firewall rules seem like something that could be part of a setup script shipped by microshift itself? Or is this intended to just be a generic starting point?

Maybe `RUN /usr/sbin/microshift-setup --default-firewall-rules` or something?

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+    firewall-offline-cmd --zone=trusted --add-source=169.254.169.1 && \
+    firewall-offline-cmd --zone=public --add-port=80/tcp && \
+    firewall-offline-cmd --zone=public --add-port=443/tcp && \
+    firewall-offline-cmd --zone=public --add-port=30000-32767/tcp && \
+    firewall-offline-cmd --zone=public --add-port=30000-32767/udp
+
+# Recursively make the root filesystem subtree shared as required by OVN images
+CMD ["/bin/sh", "-c", "mount --make-rshared / ; exec /sbin/init"]
```
cgwalters, line 66: Hmmm...interesting. I guess there's actually an argument that we should do that by default...I thought systemd did that, but perhaps not.

Note that, instead of wrapping init, this logic could actually run as a systemd service early in boot and I think that's a much better way to do it.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+
+When a `bootc` image is started, it uses the host kernel, which is not necessarily
+the same one used for building the image. This means that the `openvswitch` module
+cannot be loaded in the container due to the kernel version mismatch with the
+modules present in the `/lib/modules` directory.
+
+To work around this problem, it is necessary to share the host `/lib/modules`
+directory with the container when it is started. This is accomplished by providing
+the `-v /lib/modules:/lib/modules:ro` option to the `podman run` command as
+described below.
```
cgwalters, line 150: Mmmm. If we're running privileged I think it's saner honestly to just escape the mount namespace early on and run `modprobe` in the host mountns...

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+### Run Container
+
+Run the following command to start the MicroShift `bootc` image in an interactive
+terminal session. Note that the host container storage and kernel modules paths
+are shared with the container.
+
+```
+IMAGE_NAME=microshift-4.16-bootc
+
+sudo podman run --rm -it --privileged \
```
cgwalters, line 206: After everything described here I'd be a bit wary of running all this stuff as a privileged container...how sure are you it's not doing unexpected things in the host?

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+text
+reboot
+
+# Partition disk with a 1GB boot XFS partition and an LVM volume containing a 10GB+
+# system root. The remainder of the volume is for the CSI driver for storing data.
+zerombr
+clearpart --all --initlabel
+part biosboot --fstype=biosboot --size=1 --asprimary
+part /boot/efi --fstype=efi --size=200
+part /boot --fstype=xfs --asprimary --size=800
```
cgwalters, line 251: I would recommend just: `reqpart --add-boot`. xref https://github.com/osbuild/images/pull/363/commits/79662c40ea25b4e456e05b4684ec2c0bab6f322a for example

cgwalters (commented):
Thanks for this!

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+* A `remote registry` (i.e. `quay.io`) for storing and accessing `bootc` images
+
+## Build MicroShift Bootc Image
+
+Log into the `RHEL 9.4 host` using the user credentials that have SUDO
+permissions configured.
+
+### Build Image
+
+Create the `Containerfile` file with the following contents.
```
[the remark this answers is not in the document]
ggiguash: No extension. Containerfile is a standard file name recognized by podman command.
The directory does not matter as long as all the commands are run in the current directory. This is standard podman behavior.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+
+ARG USHIFT_VER=4.16
+RUN dnf config-manager \
+        --set-enabled rhocp-${USHIFT_VER}-for-rhel-9-$(uname -m)-rpms \
+        --set-enabled fast-datapath-for-rhel-9-$(uname -m)-rpms
+RUN dnf install -y firewalld microshift && \
+    systemctl enable microshift && \
+    dnf clean all
+
+# Create a default 'redhat' user, allowing it to run sudo commands
```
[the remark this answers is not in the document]
ggiguash: These commands are run to build a container image. They are not related to the host.

docs/contributor/image_mode.md
ggiguash: MicroShift does not configure OS as a basic principle. These instructions for setting firewall ports are a standard step everywhere else during MicroShift bootstrap. If we change the "standard", we can change it everywhere,

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+The MicroShift CNI driver (OVN) requires the Open vSwitch service to function
+properly. The service depends on the `openvswitch` kernel module to be available
+in the `bootc` image.
+
+Run the following commands on the host to check the `openvswitch` module presence
+and the version of `kernel-core` package used in the `bootc` image. Note that the
+kernel versions are different.
+
+```
+$ find /lib/modules/$(uname -r) -name "openvswitch*"
```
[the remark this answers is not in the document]
ggiguash: The module is part of the `kernel-core` package, so if it does not exist, something is terribly wrong. I do not think we should extrapolate on this in the current doc.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+When a `bootc` image is started, it uses the host kernel, which is not necessarily
+the same one used for building the image. This means that the `openvswitch` module
+cannot be loaded in the container due to the kernel version mismatch with the
+modules present in the `/lib/modules` directory.
+
+To work around this problem, it is necessary to share the host `/lib/modules`
+directory with the container when it is started. This is accomplished by providing
+the `-v /lib/modules:/lib/modules:ro` option to the `podman run` command as
```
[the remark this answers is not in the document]
ggiguash: It's part of running the container, not CSI. I see the wording is confusing. Let me change it.

docs/contributor/image_mode.md
ggiguash: Can you explain a bit more what you mean?

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+
+```
+$ sudo vgs
+  VG   #PV #LV #SN Attr   VSize   VFree
+  rhel   1   1   0 wz--n- <91.02g <2.02g
+```
+
+Additional configuration is required for using storage in `bootc` MicroShift
+images when the host does not have the volume group setup as required by the
+MicroShift CSI driver.
```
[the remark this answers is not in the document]
ggiguash: It is required if one wants to allocate storage. I'll improve the wording.

docs/contributor/image_mode.md
ggiguash: That's an interesting question and I do not have a good answer for it, except that this is a development flow, not a production one. I will discuss with MicroShift PM on whether we'd like to add this to downstream docs. 

The production flow is running a VM and applying an image on top of it.

docs/contributor/image_mode.md
ggiguash: Great suggestion! If this works for MicroShift, I would replace it "everywhere". 
Created [USHIFT-3789](https://issues.redhat.com/browse/USHIFT-3789) issue to follow-up on this.

docs/contributor/image_mode.md
ggiguash: This needs to be fixed in a few other places. 
Created [USHIFT-3790](https://issues.redhat.com/browse/USHIFT-3790) to follow-up

docs/contributor/image_mode.md
ggiguash: Reworded. Is this better?

docs/contributor/image_mode.md
ggiguash: Reworded. Is this better?

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,308 @@
...
+# Create a default 'redhat' user, allowing it to run sudo commands
+ARG USER_PASSWD=\$5\$XDVQ6DxT8S5YWLV7\$8f2om5JfjK56v9ofUkUAwZXTxJl3Sqnc9yPnza4xoJ0
+RUN useradd -m -d /var/home/redhat -G wheel -p ${USER_PASSWD} redhat
+
+# Configure OpenShift pull secret
+ARG PULL_SECRET
+RUN echo $PULL_SECRET > /etc/crio/openshift-pull-secret && \
+    chmod 600 /etc/crio/openshift-pull-secret
+
+# Configure firewall
```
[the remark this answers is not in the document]
ggiguash: The first 3 are mandatory. The rest are application specific. 
I'll add a comment in the file, but it's a standard MicroShift configuration everywhere

docs/contributor/image_mode.md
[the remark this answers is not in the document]
ggiguash: Also the subsequent podman build command references the Containerfile explicitly, even though it's a default.

docs/contributor/image_mode.md
ggiguash: I updated the pull secret handling not to include it in the image.
Please, review and post new comments if necessary. I'm resolving this thread.

docs/contributor/image_mode.md
ggiguash: Done

docs/contributor/image_mode.md
```diff
@@ -63,8 +63,19 @@ RUN firewall-offline-cmd --zone=public --add-port=80/tcp && \
...
     firewall-offline-cmd --zone=public --add-port=30000-32767/udp
 
-# Recursively make the root filesystem subtree shared as required by OVN images
-CMD ["/bin/sh", "-c", "mount --make-rshared / ; exec /sbin/init"]
+# Create a systemd unit to recursively make the root filesystem subtree
+# shared as required by OVN images
+RUN cat > /etc/systemd/system/microshift-make-rshared.service <<'EOF'
+[Unit]
+Description=Make root filesystem shared
+Before=microshift.service
```
cgwalters, line 71: I forgot to mention this in my previous comment but we should have `ConditionVirtualization=container` here; don't want to affect the main booted path!

docs/contributor/image_mode.md
ggiguash: After confirming with MicroShift PM, we are not going to publish how to run MicroShift containers in product docs, only in the repo contributor docs.

docs/contributor/image_mode.md
ggiguash: image

cgwalters: Hmm this should work but I admit to not testing it locally, did you get it to work or not?

ggiguash: Yes, it's working OK according to my tests. 
I provided the service status above that comes from a VM. For a container, the service runs successfully.
All good, IMO.

docs/contributor/image_mode.md
cgwalters: > The main problem is that the right openvswitch module does not exist as part of the container image.

It does exist - but *only for the kernel embedded in the container* which is almost always going to differ from the kernel for the host. This is just trying to fix the case of running as a container, not the production deploy path, right?

Is the openvswitch userspace doing an explicit `modprobe`? If so, then sure, this will work...but it's...well, slightly hacky and I think may fail in some corner cases because note that we're using the `modprobe` binary from the container to insert a host kernel module, which is...not tested very often.

In the general case, I'd suggest changing the setup instructions to say to `modprobe openvswitch` in the host shell, and just requiring it to be present when starting the container.

What I was talking about at first is simple: This is a privileged container, we can escape it and just run `modprobe` directly in the host context (one mechanism is via `systemd-run` using the host's systemd socket).

ggiguash: Understood. I fixed the instructions by explicit module preload before running the container.
The `/lib/modules` directory is not shared anymore.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,352 @@
...
+
+> See [Image mode for Red Hat Enterprise Linux](https://developers.redhat.com/products/rhel-image-mode/overview)
+for more information.
+
+The procedures described below require the following setup:
+* A `RHEL 9.4 host` with an active Red Hat subscription for building MicroShift `bootc`
+images and running containers
+* A `physical hypervisor host` with the [libvirt](https://libvirt.org/) virtualization
+platform for starting virtual machines that run RHEL OS containing MicroShift binaries
+* A `remote registry` (i.e. `quay.io`) for storing and accessing `bootc` images
```
pmtk, line 25: nit, i.e. means "that is" - unless using quay.io is absolutely necessary?
```suggestion
* A `remote registry` (e.g. `quay.io`) for storing and accessing `bootc` images
```

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,352 @@
...
+$ find /lib/modules/$(uname -r) -name "openvswitch*"
+/lib/modules/6.9.9-200.fc40.x86_64/kernel/net/openvswitch
+/lib/modules/6.9.9-200.fc40.x86_64/kernel/net/openvswitch/openvswitch.ko.xz
+
+$ IMAGE_NAME=microshift-4.16-bootc
+$ sudo podman inspect "${IMAGE_NAME}" | grep kernel-core
+        "created_by": "kernel-core-5.14.0-427.26.1.el9_4.x86_64"
+```
+
+When a `bootc` image is started, it uses the host kernel, which is not necessarily
```
pmtk, line 157:
```suggestion
When a `bootc` image is started as a container, it uses the host kernel, which is not necessarily
```

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,352 @@
...
+Run the following command to determine if the volume group exists and it has the
+necessary free space.
+```
+$ sudo vgs
+  VG   #PV #LV #SN Attr   VSize   VFree
+  rhel   1   1   0 wz--n- <91.02g <2.02g
+```
+
+Otherwise, a new volume group should be set up for MicroShift CSI driver to allocate
+storage in `bootc` MicroShift images.
```
pmtk, line 180: I guess we already built the image and these things are at runtime?
```suggestion
storage in `bootc` MicroShift containers.
```

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,352 @@
...
+logvol / --vgname=rhel --fstype=xfs --size=10240 --name=root
+
+# Lock root user account
+rootpw --lock
+
+# Configure network to use DHCP and activate on boot
+network --bootproto=dhcp --device=link --activate --onboot=on
+
+# Pull a 'bootc' image from a remote registry
+ostreecontainer --url quay.io/myorg/mypath/microshift-4.16-bootc
```
pmtk, line 298: How does this authenticate? Does image needs to be public? Isn't this leaking subscription protected assets?

ggiguash: Good catch. I'll make some changes.

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,369 @@
...
+starting the container as described in the [Run Container](#run-container) section.
+
+### Configure CSI
+
+If the host is already configured to have a `rhel` volume group with free space,
+this configuration is inherited by the container so that it can be used by the
+MicroShift CSI driver to allocate storage.
+
+Run the following command to determine if the volume group exists and it has the
+necessary free space.
```
dhensel-rh, line 172: Is 2G enough free space ? Not to get into the weeds, but is there a general rules like you need 90% free space ?

docs/contributor/image_mode.md
```diff
@@ -0,0 +1,369 @@
...
+WantedBy=multi-user.target
+EOF
+RUN systemctl enable microshift-make-rshared.service
+```
+
+> **Important:**<br>
+> When building a container image, podman uses host subscription information and
+> repositories inside the container. With only `BaseOS` and `AppStream` system
+> repositories enabled by default, we need to enable the `rhocp` and `fast-datapath`
+> repositories for installing MicroShift and its dependencies. These repositories
```
dhensel-rh, line 86: Does it need to be mentioned here that an additional subscription that includes "Red Hat Enterprise Linux for Real Time" might be needed, or is it out of scope  ?

ggiguash: I think this is out of scope for getting started because we're not installing low-latency RPM and profile

PR #3716, opened 2024-08-04
USHIFT-3150: Restructure scenarios to presubmits and periodics folders

ggiguash:
Blocked by https://github.com/openshift/release/pull/55162, which implements scenario directory existence check with fallback to old paths.


pacevedom (commented):
/lgtm

PR #3700, opened 2024-07-31
USHIFT-3788: Workload partitioning RF tests

eslutsky:
**Which issue(s) this PR addresses**:

Closes #<Issue Number>

PR #3726, opened 2024-08-06
Run  Workload partitioning after the real-time suite

eslutsky:


eslutsky:
/close
will be integrated in https://github.com/openshift/microshift/pull/3700

PR #3733, opened 2024-08-07
USHIFT-3781: Merge build_rpms.sh and create_local_repo.sh

vanhalenar:


test/bin/build_rpms.sh
```diff
@@ -9,70 +9,112 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-fi
-
-if ! ${BUILD_OK} ; then
-    echo "RPM build failed"
-    exit 1
-fi
-
-MAX_RUNTIME=$(jq -s 'max_by(.JobRuntime) | .JobRuntime | floor' "${BUILD_RPMS_LOG}")
-echo "RPM build complete in ${MAX_RUNTIME}s"
+build_rpms() {
```
ggiguash, line 12: When moving code from MAIN to a function, we need to take care of variables:
- Declare local variables explicitly using `local` or `local -r` for read-only
- UPPER case variables are assumed to be global

test/bin/build_rpms.sh
```diff
@@ -9,70 +9,112 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
+        exit 1
+    fi
+
+    MAX_RUNTIME=$(jq -s 'max_by(.JobRuntime) | .JobRuntime | floor' "${BUILD_RPMS_LOG}")
+    echo "RPM build complete in ${MAX_RUNTIME}s"
+}
+
+make_repo() {
+    local repodir="$1"
+    local builddir="$2"
```
ggiguash, line 84:
```suggestion
    local -r repodir="$1"
    local -r builddir="$2"
```

test/bin/build_rpms.sh
vanhalenar: My bad. Fixed

test/bin/build_rpms.sh
vanhalenar: Done

PR #3749, opened 2024-08-11
USHIFT-4025: Implement MicroShift image mode container build procedure

ggiguash:


packaging/imagemode/Makefile
```diff
@@ -0,0 +1,139 @@
...
+# using NAME=value make arguments
+#
+PULL_SECRET ?= $(HOME)/.pull-secret.json
+BASE_IMAGE_URL ?= registry.redhat.io/rhel9/rhel-bootc
+BASE_IMAGE_TAG ?= 9.4
+
+#
+# Define the main targets
+#
+all:
```
pmtk, line 12: Should `all`, `rhocp` and other target have `.PHONY`? Outcomes of these targets are not files named like these targets, so I guess they should?

packaging/imagemode/Makefile
```diff
@@ -0,0 +1,139 @@
...
+		--name "$${IMAGE_NAME}" \
+	    "$${IMAGE_NAME}") ; \
+	echo "sudo podman exec -it $${CONT_ID} bash"
+
+stop:
+	CONT_IDS=$$(sudo podman ps --format "{{.Names}}" | grep '^microshift-' || true) ; \
+	if [ -n "$${CONT_IDS}" ] ; then sudo podman stop $${CONT_IDS} ; fi
+
+clean:
+	sudo podman rmi -af
```
pmtk, line 91: Can we do something more sophisticated? We cannot assume that this only runs on a microshift dev vm where someone can not care about the images and we can wipe everything

packaging/imagemode/Makefile
```diff
@@ -0,0 +1,139 @@
...
+# 		--rm -i --privileged \
+# 		--pull=never \
+# 		--security-opt "label=type:unconfined_t" \
+# 		-v "$${BIB_OUTDIR}:/output" \
+# 		-v "/var/lib/containers/storage:/var/lib/containers/storage" \
+# 		"${BIB_IMAGE_URL}:${BIB_IMAGE_TAG}" \
+# 		--type anaconda-iso \
+# 		--local \
+# 		"localhost/$${IMAGE_NAME_URL}" && \
+# 	sudo mv -f "$${BIB_OUTDIR}/bootiso/install.iso" "${BIB_OUTBASE}/$${ISO_FNAME}"
```
pmtk, line 139: I presume this will be doable in near future and we'll uncomment it?

packaging/imagemode/Makefile
ggiguash: Do you have a suggestion what to do?
Probably, we should just remove the `clean` target, which is not very helpful in its current form.

packaging/imagemode/Makefile
ggiguash: Yes, our follow-up project would be to create ISO images from the container images.
So, these are the commands we can use. Do you think it's a problem to merge this code in its current form?
I can remove those.

ggiguash: BTW, it is already doable now. I just did not want to create unnecessary "noise" in the code and docs.

joepvd:
[packaging/imagemode/Containerfile.repobase](https://github.com/openshift/microshift/pull/3749/files#diff-17faf852f6c6f7c33e115d2641f3bd5bd58953779058557b918135b9c919cfa8) looks like a usable Containerfile to productize this.

packaging/imagemode/Makefile
pmtk: We're producing images with names starting with `microshift-` so maybe we could clean only those?

PR #3768, opened 2024-08-14
USHIFT-4061: change imagePullPolicy from Always to IfNotPresent

eslutsky:

Closes #<Issue Number>

### DIFF
--- a/assets/optional/multus/07-daemonset-dhcp.yaml
+++ b/assets/optional/multus/07-daemonset-dhcp.yaml
@@ -40,7 +40,7 @@ spec:
       containers:
       - name: dhcp-daemon
         image: containernetworking-plugins-microshift
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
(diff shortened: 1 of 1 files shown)


assets/optional/multus/07-daemonset-dhcp.yaml
```diff
@@ -40,7 +40,7 @@ spec:
       containers:
       - name: dhcp-daemon
         image: containernetworking-plugins-microshift
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
ggiguash, line 43: Do we need to update the rebase script as well?

eslutsky: AFAIK we dont  rebase multus from upstream,
this is the reference:
https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/assets.yaml#L257

@pmtk , correct me if am wrong..

PR #3775, opened 2024-08-15
USHIFT-3784: Removed manage_vm_connections.sh and PUBLIC_IP

vanhalenar:


ggiguash:
The `public_ip` file seems to be also used in `test/scenarios/periodics/el94-src@cncf-conformance.sh`. 
The error did not come up in the tests because we explicitly exclude CNCF test from presubmits because it takes over 2h.

We need to also fix the test.

ggiguash:
The `public_ip` file seems to be also used in `test/scenarios/periodics/el94-src@cncf-conformance.sh`. 
The error did not come up in the tests because we explicitly exclude CNCF test from presubmits because it takes over 2h.

We need to also fix the test.
/hold

PR #3777, opened 2024-08-15
OCPBUGS-34705: Enable crio-wipe service to handle container corruption on reboot

copejon:


packaging/rpm/microshift.spec
```diff
@@ -394,6 +394,7 @@ if [ $1 -eq 1 ]; then
 	# if crio was already started, restart it so it will catch /etc/crio/crio.conf.d/10-microshift.conf
 	systemctl is-active --quiet crio && systemctl restart --quiet crio || true
 fi
+systemctl enable --now crio-wipe
```
ggiguash, line 397: For crio, we do not enable it explicitly, just note the dependency.
I think we might want to do the same for crio-wipe.
See https://github.com/openshift/microshift/blob/main/packaging/systemd/microshift.service#L3

packaging/rpm/microshift.spec
```diff
@@ -394,6 +394,7 @@ if [ $1 -eq 1 ]; then
 	# if crio was already started, restart it so it will catch /etc/crio/crio.conf.d/10-microshift.conf
 	systemctl is-active --quiet crio && systemctl restart --quiet crio || true
```
ggiguash, line 395: We need to add a similar check for crio-wipe here, so that it's restarted if already running.

packaging/systemd/microshift-containerized.service
ggiguash, line 1: This file is obsolete. Can we delete it instead of editing it?

ggiguash:
@copejon , I checked how the `crio-wipe.service` works and it seems to delete everything (not just pods) under `/var/lib/containers` on dirty shutdown.
I think we need to discuss this further with the team because deleting all downloaded images on reboot maybe too much for MicroShift

/hold

copejon:
CI is undergoing a rollback that affects the way secret data is populated. This is reflected in tests as greenboot failure, due to a malformed pull-secret. Need to wait a bit before retrying

packaging/crio.conf.d/10-microshift_amd64.conf
```diff
@@ -1,3 +1,10 @@
+[crio]
+# the presence of a version_file path will be result in
+# only the wipe of containers, only. Image deletion is
+# controlled by version_file_persist, which is unset by
+# default and thus will not cause images to be deleted.
+version_file="/var/run/crio/version"
```
ggiguash, line 6: I tried to test this, but it does not seem to work as we want it. Here's the sequence I used.
* Install and enable crio + crio-wipe
* Enable version_file option
* Install and start microshift (not enable)
* Wait until all pods are up and running
* Run `sudo reboot` - normal reboot
* When the system is up, run `crictl images` and `crictl pods` - images are there, but all pods get deleted
* Start (not enable) microshift again 
* Wait until all pods are recreated
* Run `sudo reboot -f` - failure simulation
* When the system is up, run `crictl images` and `crictl pods` - **both pods and images are gone**

We never want to lose images...

copejon:
After some more experimentation, it's clear that the crio-wipe service may be checking the `version_file` value, but will wipe the entire storage directly anyway. I've confirmed that the existing crio defaults, where `internal_wipe = true`, will do what this PR is trying to achieve via the existing (though deprecated) feature.

PR #3784, opened 2024-08-18
USHIFT-2804: Instructions for layered product teams for testing the products with MicroShift

ggiguash:


docs/contributor/layered_product_ci.md
```diff
@@ -0,0 +1,191 @@
...
+> MicroShift includes comprehensive [Automated Functional and Integration Tests](../../test/README.md)
+> for features developed within its source repository.
+
+## Build and Publish MicroShift Container Image
+
+Follow the instructions in the [Build Image](./image_mode.md#build-image) section
+to implement the MicroShift container image layer build procedure.
+
+> Prebuilt MicroShift bootc container images are not currently available for
+> download.
```
DanielFroehlich, line 21: Can we fix this, at least for internal use? 
It cant be that hard to find a private registry/repo where we could push our image to. E.g. quay.io can be used for that using employee accounts. I would really make it as simple as possible for layered product teams. Each step we can elimnate lowers the bar to adoption.

docs/contributor/layered_product_ci.md
```diff
@@ -0,0 +1,191 @@
...
+
+Customize the `Containerfile` file according to the requirements of the layered
+product to be tested. A typical customization would be to select a custom version
+of MicroShift, which may also include pre-released ones that are published at
+[OpenShift Mirror](mirror.openshift.com).
```
DanielFroehlich, line 26: Good that we have EC/RC build here, too. I guess GA is in the referebce "build-image doc".
Maybe we should provide a small snipped on when to use which version, e.g. use the GA if layered product has an independant release cycle of OpenShift/MicroShift, use EC/RC build when release in lockstep.

DanielFroehlich (approved):

docs/contributor/layered_product_ci.md
ggiguash: We can consider this, but it's not part of this task / PR.
As soon as we have something, we can change the wording appropriately.

DanielFroehlich (approved):
/LGTM

PR #3806, opened 2024-08-22
4.18 rebase with kubelet.yaml fix

pmtk:
Original rebase PR: https://github.com/openshift/microshift/pull/3804


copejon:
CI team has moved builds to another farm while they sort out the bug.  Tests shouldn't encounter the secret injection error any more

/retest

PR #3830, opened 2024-08-29
USHIFT-4335 USHIFT-4354 USHIFT-4356: build runnable microshift-okd bootc container

eslutsky:

Closes #<Issue Number>


test/bin/ci_phase_iso_build.sh
```diff
@@ -124,6 +124,28 @@ cd "${ROOTDIR}/test/"
...
+okd_releaseTag=4.17.0-0.okd-scos-2024-08-21-100712
+
+oc adm release info ${okd_url}:${okd_releaseTag} >all_images
+
+for op in $(cat ${ROOTDIR}/assets/release/release-x86_64.json  | jq -e -r  '.images | keys []') 
+do
+    image=$(oc adm release info --image-for=${op} ${okd_url}:${okd_releaseTag}) 
+    if [ $? -eq 0 ] ; then
+        echo "${op} ${image}"
+        jq --arg a "${op}" --arg b "${image}"  '.images[$a] = $b' assets/release/release-x86_64.json >/tmp/release-x86_64.json.tmp
```
ggiguash, line 138: Should you concat lines here using `>>`?

eslutsky: its not concatenating, its replacing each time one image and overriding the file.

eslutsky: see for example:
```
+ 10:57:33.558234231 ./test/bin/ci_phase_iso_build.sh:138 	jq --arg a cli --arg b quay.io/okd/scos-content@sha256:580848455b5c4f5a28b2dc4fa32fa11ead577f02e5649b660491ba7bab31542b '.images[$a] = $b' ../assets/release/release-x86_64.json
+ 10:57:33.562627453 ./test/bin/ci_phase_iso_build.sh:139 	mv /tmp/release-x86_64.json.tmp ../assets/release/release-x86_64.json
```

https://prow.ci.openshift.org/log?container=test&id=1829467008731713536&job=pull-ci-openshift-microshift-release-4.17-microshift-metal-tests

eslutsky:
so most of the assets  can be  replaced and passes our tests.

except two which are missing in the okd  release:
```
error: no image tag "lvms_operator" exists in the release image quay.io/okd/scos-release:4.17.0-0.okd-scos-2024-08-21-
error: no image tag "openssl" exists in the release image quay.io/okd/scos-release:4.17.0-0.okd-scos-2024-08-21-100712
```

copejon:
> so most of the assets can be replaced and passes our tests.
> 
> except two which are missing in the okd release:
> 
> ```
> error: no image tag "lvms_operator" exists in the release image quay.io/okd/scos-release:4.17.0-0.okd-scos-2024-08-21-
> error: no image tag "openssl" exists in the release image quay.io/okd/scos-release:4.17.0-0.okd-scos-2024-08-21-100712
> ```

There isn't a community repo for LVMS at the moment. Previously, MicroShift deployed the internal build of topolvm. The last commit where this was the case is 68437d73b0ccbe1606be92c60a6c020ef72c41cf.  We'll need to grab the microshift code, manifests, and rebase data files from there.

eslutsky:
@ggiguash , looks like ovn-kubernetes-microshift in  also part of the okd upstream release, so we dont need to replace that.

packaging/systemd/microshift.service
ggiguash, line 1: This one we cannot merge. It has to be a systemd drop-in, ont a rewrite of the existing file

eslutsky:
@praveenkumar take a look

microshift-okd/src/README.md
copejon, line 1: I think we should drop the `microshift-` from the parent directory name - it's already in the microshift repo

microshift-okd/src/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+
+- running the container 
+  > `sudo podman run --privileged --rm --name microshift-okd -d microshift-okd`
+
+- connect to the container
+  > `sudo podman exec -ti microshift-okd /bin/bash`
+
+## configuration customization
+1. storage driver disabled (there is no lvms images upstream)
```
copejon, line 19: topolvm is the upstream variant of lvms, and so should be added to the accepted values under .storage

Leveraging the config to turn off storage has some downsides:
1. community users won't get storage provisioning out of the box, unlike downstream
1. inverts the "opt-out" component design. A stable upstream cluster MUST be configured to opt out of storage. So it's really opting-in for a stable cluster - the opposite of downstream.
1. Future customers may be confused by the inverted meaning of .storage between upstream and downstream. Documentation could help, but the config is designed around an "opt-out" philosophy.
1. (Ignoring topolvm) the community build would be missing a key part of the microshift experience, which only complicates the out of box experience.

microshift-okd/src/README.md
ggiguash: +1 on this

microshift-okd/src/README.md
ggiguash: I do not think we're ignoring CSI. Let's change the comment that it will be implemented in the next stages of this project. 
We should take it step-wise - the current change is already a great progress...

microshift-okd/src/build.sh
```diff
@@ -0,0 +1,41 @@
...
+for op in $(cat assets/release/release-x86_64.json  | jq -e -r  '.images | keys []') 
+do
+    image=$(oc adm release info --image-for=${op} ${okd_url}:${okd_releaseTag} || true) 
+    if [ -n "${image}" ] ; then
+        echo "${op} ${image}"
+        jq --arg a "${op}" --arg b "${image}"  '.images[$a] = $b' assets/release/release-x86_64.json >/tmp/release-x86_64.json.tmp
+
+
+        # delete openssl image from assets - just to verify if we still need it,since it doesnt referenced anywhere
+        jq '. | del (.images["openssl"])'  assets/release/release-x86_64.json >/tmp/release-x86_64.json.tmp2
```
ggiguash, line 19: We are removing the `openssl` dependency in https://github.com/openshift/microshift/pull/3874.
Let's drop this code once the PR is merged.

microshift-okd/src/create_repos.sh
```diff
@@ -0,0 +1,37 @@
...
+EOF
+
+    cat > "${CENTOS_NFV_SIG_REPO_FILE}" <<EOF
+[nfv-sig]
+name=CentOS Stream 9 - SIG NFV
+baseurl=http://mirror.stream.centos.org/SIGs/9-stream/nfv/x86_64/openvswitch-2/
+enabled=1
+gpgcheck=0
+skip_if_unavailable=0
+EOF
```
ggiguash, line 35: This one should not be necessary for OKD.

microshift-okd/src/build.sh
```diff
@@ -0,0 +1,41 @@
...
+done
+
+
+sudo podman stop microshift-okd
+rm -rf microshift-okd/src/rpmbuild
+rm -rf _output/rpmbuild
+make build 
+make rpm
+createrepo _output/rpmbuild/
+cp -rf _output/rpmbuild microshift-okd/src/
```
ggiguash, line 32: It should be done the other way around. We should not copy binaries into the source directory.
Instead, copy `okd/src` to `_output/okd/...` and run podman builds from there.

microshift-okd/src/create_repos.sh
ggiguash: The openvswitch package is pulled in the RPM spec, but it's part of the mirror repo, so we could use it from there.

microshift-okd/src/build.sh
eslutsky: yes, i had to do it because of podman build context.
 see the other approach using multi-build.

microshift-okd/src/build.sh
eslutsky: since we are based on 4.17
this is the relevant change  https://github.com/openshift/microshift/pull/3884

okd/src/README.md
```diff
@@ -0,0 +1,31 @@
...
+  > `cd microshift-okd/src && ./build.sh` 
+  - this script will:
+    1. replace microshift assets images to OKD  upstream images
+    1. will build microshift RPMs and repo based on current sources.
+    1. will build micrsoshift_okd bootc container based on `centos-bootc:stream9`
+    1. apply upstream customization  (see below)
+
+- building the container with podman multistage build :
+  > `git clone https://github.com/openshift/microshift.git ~/microshift`
+  > `cd ~/microshift && podman build --no-cache --volume $(pwd)/:/src:Z -f okd/src/microshift-okd-multi-build.Containerfile`
```
eslutsky, line 13: i had to add the  `--no-cache`  because microshift rpms are the artifacts of the `builder` stage and if we re-run the build it will not produce new RPMs (it will assume there is no change),.

packaging/systemd/microshift.service
eslutsky: okay. lets see how the Flannel PR will handle it, since we already relying on it.

scripts/devenv-builder/configure-vm.sh
```diff
@@ -136,8 +136,8 @@ if ${INSTALL_BUILD_DEPS} || ${BUILD_AND_RUN}; then
     if ${DNF_UPDATE}; then
         "${DNF_RETRY}" "update"
     fi
-    "${DNF_RETRY}" "install" "gcc git golang cockpit make jq selinux-policy-devel rpm-build jq bash-completion avahi-tools createrepo"
-    sudo systemctl enable --now cockpit.socket
+    # "${DNF_RETRY}" "install" "gcc git golang cockpit make jq selinux-policy-devel rpm-build jq bash-completion avahi-tools createrepo"
+    # sudo systemctl enable --now cockpit.socket
```
eslutsky, line 140: @ggiguash , we need this part handled on different PR,
we dont want to enable cockpit when it is running in build-only mode.
its creating issue for containers build (systemd is not running during build time)

okd/src/README.md
pmtk: How does the `microshift-okd-multi-build.Containerfile` gets the `/src`?

eslutsky: the current source  mounted  throughout  `--volume`

okd/src/configure.sh
```diff
@@ -0,0 +1,16 @@
...
+
+
+    cat > /etc/microshift/config.yaml <<EOF
+storage:
+    driver: "none"
+EOF
+
+# disable ovsdb
+rm -rf /usr/lib/systemd/system/ovs-vswitchd.service
+rm -rf /usr/lib/systemd/system/openvswitch.service
```
ggiguash, line 11:
```suggestion
rm -f /usr/lib/systemd/system/ovs-vswitchd.service
rm -f /usr/lib/systemd/system/openvswitch.service
```

okd/src/configure.sh
```diff
@@ -0,0 +1,16 @@
...
+
+#removes ExecStart=/usr/bin/grub2-editenv - set boot_success=1
+rm -rf /usr/lib/systemd/system/greenboot-grub2-set-success.service
```
ggiguash, line 14:
```suggestion
rm -f /usr/lib/systemd/system/greenboot-grub2-set-success.service
```

okd/src/create_repos.sh
```diff
@@ -0,0 +1,28 @@
+#/bin/bash
+
+
+repo_path=$1
+
+USHIFT_LOCAL_REPO_FILE=/etc/yum.repos.d/microshift-local.repo
+OCP_MIRROR_REPO_FILE=/etc/yum.repos.d/openshift-mirror-beta.repo
+CENTOS_NFV_SIG_REPO_FILE=/etc/yum.repos.d/microshift-sig-nfv.repo
```
ggiguash, line 8:
```suggestion
```

okd/src/README.md
pmtk: That's why you need `--no-cache` - if there was `ADD/COPY . .` for source, then changing the files would invalidate the cache. Have you tried copying the sources to the container? I think it would be more aligned with how DockerContainer files are expected to be

eslutsky: originally i've started with COPY, but the problem if your DockerFile inside sub-dir  then :+1: - you must  must run the build in the context of MICROSHIFT root.
 IE:
`cd ~/microshift && podman build okd/src/microshift-okd-multi-build.Containerfile .`

the problem is that microshift root ,  can become realy big , i can add `.dockerignore` to ignore few dirs ,
ie:
```
_output
.git
.github
```

eslutsky: @pmtk , see the updated version

okd/src/build.sh
```diff
@@ -0,0 +1,41 @@
+#/bin/bash
+
+
+# replace microshift assets to Upsteam from OKD
+okd_url=quay.io/okd/scos-release
+okd_releaseTag=4.17.0-0.okd-scos-2024-08-21-100712
+
+oc adm release info ${okd_url}:${okd_releaseTag} >all_images
```
praveenkumar, line 8: Is this step only to check if we are able to get the release info for the okd release or `all_images` file is used somewhere?

okd/src/configure.sh
praveenkumar: Instead of removing the service, we could just mask it right?

okd/src/configure.sh
praveenkumar: Same here, we can just mask those.

okd/src/README.md
pmtk: It looks better.
We could also specify want we want to copy, although it's bit lengthy... On the other hand, we would be sure that there's no need to remember specific podman build invocation
```
COPY assets cmd etcd packaging pkg scripts vendor go.mod go.sum Makefile Makefile.kube_git.var Makefile.version.aarch64.var Makefile.version.x86_64.var /src
```

okd/src/build.sh
```diff
@@ -0,0 +1,41 @@
...
+sudo podman stop microshift-okd
+rm -rf okd/src/rpmbuild
+rm -rf _output/rpmbuild
+make build 
+make rpm
+createrepo _output/rpmbuild/
+cp -rf _output/rpmbuild microshift-okd/src/
+
+# build the container
+sudo podman build -f okd/src/microshift-okd-source.containerfile -t microshift-okd
```
pmtk, line 35: I'm still not sure what's the difference between two containerfiles and when should they be used? They seems totally unrelated.

okd/src/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# run container on the background
+sudo podman run --privileged --rm --name microshift-okd -d microshift-okd
+
+# connect the container
+sudo podman exec -ti microshift-okd /bin/bash
```
pmtk, line 41: Script is called `build.sh` but we're starting and attaching to the container. Should we split the operations?

okd/src/build.sh
eslutsky: no its not used, it was for debugging, so i will remove it.

okd/src/build.sh
eslutsky: well its two different approaches , so we can decide:
1. microshift-okd-source.containerfile -   when building  your microshift RPM on the dev Machine , so it requires installing bunch of pre-req (gcc,linker,make,go ) on the dev.so  during container build-time the RPMs copied ..
2.  microshift-okd-multi-build.Containerfile - RPM build  done in container using multi-stage dockerfile.

the second approach is more community friendly , since it only requires a container engine , so i vote for it.

okd/src/build.sh
eslutsky: `build.sh` was my first attempt, i'll remove it once  we decide to use `microshift-okd-multi-build` only approach.

okd/src/build.sh
pmtk: I think keeping only multi-build Containerfile is better - it's a complete standalone file that can be build on any host

packaging/systemd/microshift.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=MicroShift
-Wants=network-online.target crio.service openvswitch.service microshift-ovs-init.service
-After=network-online.target crio.service openvswitch.service microshift-ovs-init.service
+Wants=network-online.target crio.service
+After=network-online.target crio.service
```
praveenkumar, line 4: I think we should allow user to use OVN-k also (upstream component , build for OKD) and different CNI plugin (flannel in this case) and this different CNI plugin should have updated unit file instead we doing it here.

eslutsky: yea, its not going to be merged like that,  yes that part will be handled by the CNI (flannel).
I'll remove it .

okd/src/README.md
```diff
@@ -0,0 +1,30 @@
...
+- running the container 
+  `sudo podman run --privileged --rm --name microshift-okd -d microshift-okd`
+
+- connect to the container
+  `sudo podman exec -ti microshift-okd /bin/bash`
+
+## configuration customization
+1. storage driver disabled (there is no lvms images upstream) - will be added in the stage of the project.
+1. network CNI disabled (requires kernel modules)
+    - microshift service is not dependent on ovs
```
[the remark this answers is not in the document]
eslutsky: I'll try it it again it also with OVN-k, and update  the result here.
I still think that the upstream defaults  should provide runnable container with out any hard requirement from the host.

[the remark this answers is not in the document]
praveenkumar: > I still think that the upstream defaults should provide runnable container with out any hard requirement from the host.

I am curious about the way we want to have it delivered to users, if it is runnable container then I agree to have defaults not depend of the host but if we want upstream also follow what we have in downstream supported way of doing it then may be have the OVN-K makes more sense as default.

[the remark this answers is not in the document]
eslutsky: I've made the changes so it will work with the OVN-k,
so currently its running all the pods except storage.

okd/src/use_okd_assets.sh
```diff
@@ -0,0 +1,26 @@
...
+# replace microshift assets to Upsteam from OKD
+okd_url=$1
+okd_releaseTag=$2
+
+
+[[ -n $okd_url  && -n $okd_releaseTag ]] || { echo missing vars ;  exit 2 ; }
+
+
+# replace Microshift images with upstream (from OKD release)
+for op in $(cat assets/release/release-x86_64.json  | jq -e -r  '.images | keys []')
```
ggiguash, line 13: Please, use `$(uname -m)` everywhere instead of the hardcoded platform name.

okd/src/use_okd_assets.sh
```diff
@@ -0,0 +1,26 @@
+#/bin/bash
+
+
+# replace microshift assets to Upsteam from OKD
+okd_url=$1
+okd_releaseTag=$2
+
+
+[[ -n $okd_url  && -n $okd_releaseTag ]] || { echo missing vars ;  exit 2 ; }
```
ggiguash, line 9: Can we make this more "official" by adding a usage message?

okd/src/use_okd_assets.sh
ggiguash, line 13: The way this script is written, it assumes to be executed from a certain location in the tree. Let's make it more robust please.

okd/src/microshift-okd-multi-build.Containerfile
```diff
@@ -0,0 +1,56 @@
...
+ARG OKD_REPO=quay.io/okd/scos-release
+ARG OKD_VERSION_TAG=4.17.0-0.okd-scos-2024-08-21-100712
+ARG REPO_DIR=/src/_output/rpmbuild/RPMS/
+ENV USER=microshift
+ENV HOME=/microshift
+ENV GOPATH=/microshift
+ENV GOMODCACHE=/microshift/.cache
+
+# Adding non-root user for building microshift
+run useradd -m -s /bin/bash -p "$(openssl passwd -1 redhat)" microshift
```
ggiguash, line 13: This might generate a prodsec warning after we merge this code. We're not allowed to have hardcoded passwords in the repo.

okd/src/microshift-okd-multi-build.Containerfile
```diff
@@ -0,0 +1,56 @@
...
+run echo -e 'microshift\tALL=(ALL)\tNOPASSWD: ALL' | sudo tee /etc/sudoers.d/microshift
+RUN mkdir /microshift
+copy . /src
```
ggiguash, line 16:
```suggestion
COPY . /src 
```

okd/src/microshift-okd-multi-build.Containerfile
```diff
@@ -0,0 +1,56 @@
...
+ENV HOME=/microshift
+ENV GOPATH=/microshift
+ENV GOMODCACHE=/microshift/.cache
+
+# Adding non-root user for building microshift
+run useradd -m -s /bin/bash -p "$(openssl passwd -1 redhat)" microshift
+run echo -e 'microshift\tALL=(ALL)\tNOPASSWD: ALL' | sudo tee /etc/sudoers.d/microshift
+RUN mkdir /microshift
+copy . /src 
+RUN chown -R 1000:1000 /microshift /src
```
ggiguash, line 17: Why are we using hardcoded uid and gid? Can we use `microshift:microshift` instead?

okd/src/use_okd_assets.sh
```diff
@@ -0,0 +1,68 @@
...
+
+set -eo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+MICROSHIFT_ROOT="${SCRIPTDIR}/../.."
+
+
+verify(){
+    okd_url=$1
+    okd_releaseTag=$2
```
ggiguash, line 11:
```suggestion
    local -r okd_url=$1
    local -r okd_releaseTag=$2
```

okd/src/use_okd_assets.sh
```diff
@@ -0,0 +1,68 @@
...
+        if [ -n "${image}" ] ; then
+            echo "${op} ${image}"
+            jq --arg a "${op}" --arg b "${image}"  '.images[$a] = $b' "${MICROSHIFT_ROOT}/assets/release/release-${arch}.json" >"/tmp/release-${arch}.json.tmp"
+            mv "/tmp/release-${arch}.json.tmp" "${MICROSHIFT_ROOT}/assets/release/release-${arch}.json"
+        fi
+    done
+
+    pod_image=$(oc adm release info --image-for=pod "${okd_url}:${okd_releaseTag}" || true) 
+    # update the infra pods for crio
+    sed -i 's,pause_image .*,pause_image = '"\"${pod_image}\""',' packaging/crio.conf.d/10-microshift_amd64.conf
```
ggiguash, line 40: This code needs to be updated to use $arch

okd/src/use_okd_assets.sh
```diff
@@ -0,0 +1,68 @@
...
+    if ! stdout=$(oc adm release info "${okd_url}:${okd_releaseTag}" 2>&1)  ; then
+        echo -e "error verifying okd release (URL: ${okd_url} , TAG: ${okd_releaseTag}) \nERROR: ${stdout}"
+        exit 1
+    fi
+}
+
+replace_assets(){
+    okd_url=$1
+    okd_releaseTag=$2
+    arch=$(uname -m)
```
ggiguash, line 23:
```suggestion
    local -r okd_url=$1
    local -r okd_releaseTag=$2
    local -r arch=$(uname -m)
```

okd/src/use_okd_assets.sh
```diff
@@ -0,0 +1,68 @@
...
+
+    oc adm release info --image-for="${op}" "${okd_url}:${okd_releaseTag}"
+
+    # replace Microshift images with upstream (from OKD release)
+    for op in $(jq -e -r  '.images | keys []' "${MICROSHIFT_ROOT}/assets/release/release-${arch}.json") 
+    do
+        image=$(oc adm release info --image-for="${op}" "${okd_url}:${okd_releaseTag}" || true)
```
ggiguash, line 30:
```suggestion
        local image
        image=$(oc adm release info --image-for="${op}" "${okd_url}:${okd_releaseTag}" || true) 
```

okd/src/use_okd_assets.sh
```diff
@@ -0,0 +1,68 @@
...
+
+    oc adm release info --image-for="${op}" "${okd_url}:${okd_releaseTag}"
+
+    # replace Microshift images with upstream (from OKD release)
+    for op in $(jq -e -r  '.images | keys []' "${MICROSHIFT_ROOT}/assets/release/release-${arch}.json") 
+    do
+        image=$(oc adm release info --image-for="${op}" "${okd_url}:${okd_releaseTag}" || true) 
+        if [ -n "${image}" ] ; then
+            echo "${op} ${image}"
+            jq --arg a "${op}" --arg b "${image}"  '.images[$a] = $b' "${MICROSHIFT_ROOT}/assets/release/release-${arch}.json" >"/tmp/release-${arch}.json.tmp"
```
ggiguash, line 33: Please, use mktemp to create temporary files.

okd/src/microshift-okd-multi-build.Containerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN useradd -m -s /bin/bash microshift -d /microshift && \
+    echo 'microshift  ALL=(ALL)  NOPASSWD: ALL' >/etc/sudoers.d/microshift 
+COPY . /src 
+RUN chown -R microshift:microshift /microshift /src
+
+USER 1000:1000
+WORKDIR /src
+# Preparing for the build
+RUN echo '{"auths":{"fake":{"auth":"aWQ6cGFzcwo="}}}' > /tmp/.pull-secret && \
+   /src/scripts/devenv-builder/configure-vm.sh --no-build /tmp/.pull-secret && \
```
ggiguash, line 22: Should we also add the --no-set-release-version and --skip-dnf-update options to configure?

okd/src/use_okd_assets.sh
eslutsky: its trickier, i'll need to convert to it with `UNAME_TO_GOARCH_MAP=( ["x86_64"]="amd64" ["aarch64"]="arm64" )`

ggiguash: Yes, please. We should not hardcode archs.

PR #3831, opened 2024-08-29
NO-ISSUE: Skip prow job name conformance tests

pacevedom:

Closes #<Issue Number>


pacevedom:
The `api-unreachable-from-client-metrics` are failing because the promoted image for origin does not yet contain the fix for it. The target of this PR is fulfilled, as the prow job names tests are not failing.

PR #3874, opened 2024-09-06
USHIFT-4279: Remove the unnecessary openssl image dependency

ggiguash:


pacevedom (approved):
/lgtm
The RHEL boot issue is impacting some of these tests. Bootc and conformance worked, as did rebase, so we can merge this one.

PR #3885, opened 2024-09-09
USHIFT-2909: Implement downgrade block test to verify greenboot bootc system rollback functionality

ggiguash:


ggiguash:
Waiting for the new greenboot package before review / merge.
See [THEEDGE-4010](https://issues.redhat.com/browse/THEEDGE-4010) and [THEEDGE-4011](https://issues.redhat.com/browse/THEEDGE-4011)
/hold

packaging/greenboot/functions.sh
```diff
@@ -39,18 +40,34 @@ function print_boot_status() {
...
-    ostr_stat=$(ostree admin status 2>/dev/null || true)
 
     [ -z "${grub_vars}" ] && grub_vars="None"
     [ -z "${boot_vars}" ] && boot_vars="None"
-    [ -z "${ostr_stat}" ] && ostr_stat="Not an ostree system"
+
+    # Assume RPM system installation type
+    local system_type
+    local system_stat
+    system_type="RPM"
```
pmtk, line 52: I don't see it being overwritten anywhere

test/suites/upgrade/downgrade-block.robot
```diff
@@ -26,7 +26,17 @@ Downgrade Is Blocked
...
+    IF    ${is_bootc}
+        Deploy Bootc Commit Expecting A Rollback
+        ...    ${BOOTC_REGISTRY}
+        ...    ${OLDER_MICROSHIFT_REF}
+        ...    write_agent_cfg=False
+    ELSE
+        Deploy Commit Expecting A Rollback
+        ...    ${OLDER_MICROSHIFT_REF}
+        ...    write_agent_cfg=False
+    END
```
pmtk, line 39: I'm wondering if we could move bootc-or-ostree` logic to different place? Like "Deploy Commit Expecting A Rollback" could check if system is bootc and we could pass the bootc url+ref as single string in the scenario script instead of joining them later.
If it's about allowing for insecure registry, we could use optional arguments like:
"if registry_url is given, it's a bootc, allow insecure, then concat url+ref".

Because doing this in every upgrade test individually scales poorly imho

packaging/greenboot/functions.sh
ggiguash: Oops

test/suites/upgrade/downgrade-block.robot
ggiguash: Let me try this.

PR #3911, opened 2024-09-12
USHIFT-4385: Add devenv instructions on mounting shared directory

ggiguash:
This change also contains an optimization for how hadolint is used. This is necessary to avoid SELinux context setting on an NFS-mounted partition. 

Note that `container_file_t` SELinux context is still required for running container images off the mounted directory. But, unlike the context for hadolint that was always set on verify, the `container_file_t` is only set in case of NFS mount.

### DIFF
--- a/docs/contributor/devenv_setup.md
+++ b/docs/contributor/devenv_setup.md
@@ -414,30 +414,15 @@ Reboot the development virtual machine to finalize the configuration.
 
 #### Using Shared Directory
 
-One way of using the shared directory would be to keep the whole source tree in
-it for easy access from the host.
+Run the following commands to move the original source tree to a shared location
+and link it from the `microshift` user home directory. Note that this allows to
+keep all the user configuration files locally while sharing the source tree.
 
 ```
 MOUNT_DIR=/home/microshift-shared
 
-sudo mkdir -p -m 0755 "${MOUNT_DIR}/microshift"
-sudo chown $(whoami). "${MOUNT_DIR}/microshift"
+sudo mv ~/microshift "${MOUNT_DIR}/"
 
 # Do not attempt to create the link if the target already exists
 [ ! -e ~/microshift ] && ln -s "${MOUNT_DIR}/microshift" ~/microshift
-```
-
-Another way of using the shared directory would be to keep the whole `microshift`
-user home directory on the host and mount it from the virtual machine.
-
-```
-MOUNT_DIR=/home/microshift-shared
-
-sudo mkdir -p -m 0755 "${MOUNT_DIR}/microshift"
-sudo chown $(whoami). "${MOUNT_DIR}/microshift"
-
-# Move the existing home directory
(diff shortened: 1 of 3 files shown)


docs/contributor/devenv_setup.md
```diff
@@ -307,3 +307,108 @@ sudo lvextend -l +95%FREE /dev/mapper/rhel-root
...
+One way of using the shared directory would be to keep the whole source tree in
+it for easy access from the host.
+
+```
+MOUNT_DIR=/home/microshift-shared
+
+sudo mkdir -p -m 0755 "${MOUNT_DIR}/microshift"
+sudo chown $(whoami). "${MOUNT_DIR}/microshift"
+
+ln -s "${MOUNT_DIR}/microshift" ~/microshift
```
pmtk, line 413: I think this misses git clone? Or at least move of the repo?
I already have source in ~/microshift, so the last command creates a symlink at `/home/microshift/microshift/microshift` pointing to `/home/microshift-shared/microshift` so it's invalid.

Also is it a bit weird that dev vm populates host's share? Why not share the microshift repo from the host directly mounted as ~/microshift in the VM?

docs/contributor/devenv_setup.md
```diff
@@ -307,3 +307,108 @@ sudo lvextend -l +95%FREE /dev/mapper/rhel-root
...
+```
+
+#### Synchronize User and Group Identifiers
+
+For read-write to work properly both on the hypervisor host and on the virtual
+machine, it is recommended to edit the `microshift` user and group
+identifiers on the virtual machine to have the value as on the hypervisor.
+
+Edit the `/etc/passwd` and `/etc/group` files on the virtual machine to replace
+the `microshift` user and group identifiers.
```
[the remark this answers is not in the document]
ggiguash: Added more explanation.

docs/contributor/devenv_setup.md
ggiguash: Having the whole home directory remotely mounted maybe a bit too intrusive. I have not tested it, but I can add a note that this is another sharing option.

docs/contributor/devenv_setup.md
```diff
@@ -307,3 +307,137 @@ sudo lvextend -l +95%FREE /dev/mapper/rhel-root
...
+```
+MOUNT_DIR=/home/microshift-shared
+
+sudo mkdir -p -m 0755 "${MOUNT_DIR}/microshift"
+sudo chown $(whoami). "${MOUNT_DIR}/microshift"
+
+# Move the existing home directory
+[ -d /home/microshift ] && sudo mv /home/microshift /home/microshift.local
+# Do not attempt to create the link if the target already exists
+[ ! -L /home/microshift ] && sudo ln -s "${MOUNT_DIR}" /home/microshift
```
pmtk, line 442: I think this goes too far. I don't think it's a good advice to move existing home directory and mount whole share as home dir. What about all the dotfiles there, like ~/.config, ~/.ssh, ~/.local/bin, and others?

PR #3914, opened 2024-09-12
USHIFT-4278: Specify default value for .storage.optionalCsiComponents that is an accepted value

copejon:


packaging/microshift/config.yaml
```diff
@@ -131,6 +131,9 @@ storage:
     # storage driver. Like CSIStorageDriver, uninstallation is not supported as this can lead to orphaned storage
     # objects.
     # Allowed values are: unset, [], or one or more of ["snapshot-controller", "snapshot-webhook"]
+    # example:
+    #   - snapshot-controller
+    # - snapshot-webhook
```
pmtk, line 136: Is this auto-generated?
```suggestion
    # Example:
    # - snapshot-controller
    # - snapshot-webhook
```

copejon: I know, controller-gen _insists_ on generating it this way. I tried a handful of permutations on the tag's value but it always comes out this way.

pmtk: `ci/prow/verify` failed, I don't think we should include this "suggestions"

copejon:
@pmtk I reverted the comment fix. The format was ping as an error by the verify script 🤷🏻

pacevedom (approved):
/lgtm

PR #3952, opened 2024-09-22
USHIFT-4518: Add retries on rebasing bootc or ostree systems

ggiguash:
An attempt to work around https://github.com/ostreedev/ostree-rs-ext/issues/657 when running tests.
Note: The problem still persists when running rebase from kickstart.


pacevedom (approved):
/lgtm

PR #3968, opened 2024-09-24
NO-ISSUE: Remove unused rhel92 and rhel93 blueprints

ggiguash:


ggiguash:
The cache has been rebuilt in the main branch on both platforms. Let's make sure the tests pass now without the rhel92.iso and rhel93.iso in the cache.
image
image

PR #3975, opened 2024-09-25
USHIFT-4116: Auto recovery: backup

pmtk:


test/suites/backup/auto-recovery.robot
```diff
@@ -0,0 +1,68 @@
...
+    Verify Backup Version Against The System    ${backup[1]}
+
+Verify Backup Datetime Against Host Time
+    [Documentation]    Compare given datetime against current time on the VM
+    [Arguments]    ${backup_datetime}
+    ${backup_dt}=    Convert Date    ${backup_datetime}    datetime
+    ${vm_datetime}=    Command Should Work    date +'%Y-%m-%d %H:%M:%S'
+    ${vm_dt}=    Convert Date    ${vm_datetime}    datetime
+    ${time_diff}=    Subtract Date From Date    ${vm_dt}    ${backup_dt}
+    Should Be True    ${time_diff}<5
```
ggiguash, line 54: This might be a bit unstable. Can we think of another verification like this?

test/suites/backup/auto-recovery.robot
```diff
@@ -0,0 +1,68 @@
...
+Suite Teardown      Teardown
+
+
+*** Test Cases ***
+Backups Are Created Successfully With Expected Schema
+    [Documentation]    Verify that backups intended for auto-recovery are created successfully
+    ...    and they are named with expected schema of DATETIME_DEPLOY-ID for ostree/bootc and DATETIME_MICROSHIFT-VERSION.
+    Stop MicroShift
+    Command Should Work    microshift backup --auto-recovery /var/lib/microshift-auto-recovery
+    ${backup_name}=    Command Should Work    ls /var/lib/microshift-auto-recovery
```
ggiguash, line 20: I'm not sure what the main purpose of this check is, but seeing file attributes might be helpful in general.
```suggestion
    ${backup_name}=    Command Should Work    ls -l /var/lib/microshift-auto-recovery
```

test/suites/backup/auto-recovery.robot
```diff
@@ -0,0 +1,68 @@
...
+    Verify Backup Name    ${backup_name}
+
+    [Teardown]    Run Keywords
```
ggiguash, line 23: What's the main goal of putting this in Teardown? 
If we want to leave MicroShift running, we should probably remove `rm` command into the main function.

test/scenarios/periodics/el94-src@auto-recovery-rpm.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+
+scenario_create_vms() {
+    prepare_kickstart host1 kickstart-liveimg.ks.template "" true
+    launch_vm --boot_blueprint rhel-9.4-microshift-source-isolated
```
ggiguash, line 7: What's the reason for using non-default blueprint here?

pkg/cmd/admin.go
```diff
@@ -105,6 +106,14 @@ func backupRestorePreRun(backingUp bool) func(*cobra.Command, []string) error {
 			return err
 		}
 
+		if autorec, err := cmd.Flags().GetBool("auto-recovery"); err != nil {
+			return fmt.Errorf("failed to get `auto-recovery` flag: %w", err)
+		} else if autorec {
+			// For auto-recovery don't care about existence of the storage,
+			// if it doesn't exists before backup, microshift will create it.
```
ggiguash, line 113:
```suggestion
			// For auto-recovery, the existence of the storage is not important.
			// If it doesn't exist before backup, MicroShift will create it.
```

pkg/cmd/admin.go
```diff
@@ -151,15 +162,40 @@ func NewBackupCommand() *cobra.Command {
...
+			}
+			fmt.Printf("%s\n", backupPath)
+			return nil
 		},
 	}
 
+	cmd.Flags().BoolVar(&autorec, "auto-recovery", false,
+		`Changes behavior of the command.
+Argument PATH will be treated as directory where backups are
+created with naming scheme compatible with "restore --auto-recovery"`)
```
ggiguash, line 197:
```suggestion
		`Changes the behavior of the backup command.
The PATH argument will be treated as a directory where backups are
created using the naming scheme compatible with "restore --auto-recovery"`)
```

pkg/admin/data/data_manager.go
```diff
@@ -235,10 +244,25 @@ func copyPath(src, dest string) error {
 		klog.InfoS("Failed to copy", "cmd", cmd,
 			"stdout", strings.ReplaceAll(outb.String(), "\n", `, `),
 			"stderr", errb.String())
-		return fmt.Errorf("failed to copy %q to %q: %w", src, dest, err)
+		copyErr := fmt.Errorf("failed to copy %q to %q: %w", src, dest, err)
+		if err := os.RemoveAll(tmpDest); err != nil {
```
ggiguash, line 248: We should probably add "removing intermediate path" message similar to below.

pkg/admin/data/data_manager.go
```diff
@@ -235,10 +244,25 @@ func copyPath(src, dest string) error {
...
+			return errors.Join(copyErr, fmt.Errorf("failed to remove %q: %w", tmpDest, err))
+		}
+		return copyErr
+	}
+
+	klog.InfoS("Renaming path", "src", tmpDest, "dest", dest)
```
ggiguash, line 254: Please, add a code comment here explaining why we're doing this.

pkg/admin/data/data_manager.go
```diff
@@ -223,7 +224,15 @@ func checkDirectoryContents(existing sets.Set[string]) error {
 }
 
 func copyPath(src, dest string) error {
-	cmd := exec.Command("cp", append(cpArgs, src, dest)...) //nolint:gosec
+	tmpDest := fmt.Sprintf("%s.tmp", dest)
+	if exists, err := pathExists(tmpDest); err != nil {
+		return err
+	} else if exists {
+		if err := os.RemoveAll(tmpDest); err != nil {
```
ggiguash, line 231: There are quite a few attempts in this function to remove tmpDest data. Is it better handled once at the function exit, as a deferred operation?

pkg/admin/autorecovery/backup.go
```diff
@@ -0,0 +1,56 @@
...
+	} else {
+		execVer, err := prerun.GetVersionOfExecutable()
+		if err != nil {
+			return "", fmt.Errorf("failed to get version of MicroShift executable: %w", err)
+		}
+		versionPart = execVer.String()
+	}
+
+	// Format current time as yyyymmddHHMMSS
+	t := time.Now().Format("20060102150405")
```
ggiguash, line 35: Out of my ignorance, how does the format "know" which is a month and which is a day?
Should it be `20060130...` to be move explicit?

test/suites/backup/auto-recovery.robot
pmtk: Perhaps we can could rethink it when it starts making problems? So far it was behaving fine.

test/suites/backup/auto-recovery.robot
pmtk: Actually the command is there to get the name of the backup - there's one subdir in the dir and we want its name to decompose into DATE and deploy ID

pkg/admin/autorecovery/backup.go
pmtk: These values are defined in the Go source code and it knows how to search for them: https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/time/format.go;l=57-71

pkg/admin/data/data_manager.go
pmtk: But they have different purposes at different times.
This one is to remove it if already exists (like it was interrupted etc.) so it needs to happen in this place.
Others are cleaning up in case of a problem and immediately exiting. I don't want to add a defer that "runs wild".

test/scenarios-bootc/presubmits/el94-src@auto-recovery.sh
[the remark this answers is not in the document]
pmtk: It's not rpm-ostree, but regular RPM. I thought we're not adding more scenarios for ostree?

test/scenarios/periodics/el94-src@auto-recovery-rpm.sh
pmtk: Because it's not ostree - it's regular RPM

test/suites/backup/auto-recovery.robot
pmtk: I don't understand. It starts microshift because it was stopped. It removes the dir to clean up.

pkg/admin/data/data_manager.go
pmtk: I added the information to the error message, I don't think we need extra log for this

pkg/admin/autorecovery/backup.go
pmtk: Longer answer would be that it has some mapping between meaning and values, like:
01/1 means month, some form of `2` is day, hour is 3/15/03, minute is 4/04, and so on

test/suites/backup/auto-recovery.robot
ggiguash: We need to make sure MicroShift is running after this. Let's test it by putting `false` command in the first statement to verify that `Start MicroShift` is "always" executed.

test/suites/backup/auto-recovery.robot
ggiguash: Let's not spend more time on this and just increase the timeout to a higher value, i.e. 60?

pkg/admin/data/data_manager.go
pmtk: I thought about this, but:
- in next PR I plan to refactor this into different way to accommodate "copy all", then "rename all" we'll need for restore
- if we just `defer RemoveAll(tmpDest)`, it'll run even if we don't want it (like the Rename was successful) - as errors should not be ignored, if we log it, it'll be misleading (wanting to remove non-existent file when we don't need to)

test/suites/backup/auto-recovery.robot
pmtk: I verified following and even if the 1st command fails, the `Start MicroShift` will still run. I changed it though for explicitness 
```
    [Teardown]    Run Keywords
    ...    Command Should Work    false
    ...    AND
    ...    Start MicroShift
```

test/suites/backup/auto-recovery.robot
```diff
@@ -0,0 +1,68 @@
...
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+
+*** Test Cases ***
+Backups Are Created Successfully With Expected Schema
+    [Documentation]    Verify that backups intended for auto-recovery are created successfully
+    ...    and they are named with expected schema of DATETIME_DEPLOY-ID for ostree/bootc and DATETIME_MICROSHIFT-VERSION.
+    Stop MicroShift
```
pacevedom, line 18: Wondering if this should be part of the test setup?

test/suites/backup/auto-recovery.robot
```diff
@@ -0,0 +1,68 @@
...
+    Verify Backup Version Against The System    ${backup[1]}
+
+Verify Backup Datetime Against Host Time
+    [Documentation]    Compare given datetime against current time on the VM
+    [Arguments]    ${backup_datetime}
+    ${backup_dt}=    Convert Date    ${backup_datetime}    datetime
+    ${vm_datetime}=    Command Should Work    date +'%Y-%m-%d %H:%M:%S'
+    ${vm_dt}=    Convert Date    ${vm_datetime}    datetime
+    ${time_diff}=    Subtract Date From Date    ${vm_dt}    ${backup_dt}
+    Should Be True    ${time_diff}<60
```
pacevedom, line 54: I guess this is huge for the time it actually takes to calculate the name once you issue the command, right?

pacevedom (approved):
/lgtm

pacevedom (commented):
/lgtm

PR #4000, opened 2024-09-30
USHIFT-4378: introduce ingress customization fields

eslutsky:

Closes #<Issue Number>


pkg/config/ingress.go
```diff
@@ -13,9 +17,10 @@ type IngressStatusEnum string
 type IngressConfig struct {
 	// Default router status, can be Managed or Removed.
 	// +kubebuilder:default=Managed
-	Status          IngressStatusEnum    `json:"status"`
-	AdmissionPolicy RouteAdmissionPolicy `json:"routeAdmissionPolicy"`
-	Ports           IngressPortsConfig   `json:"ports"`
+	Status          IngressStatusEnum                         `json:"status"`
+	AdmissionPolicy RouteAdmissionPolicy                      `json:"routeAdmissionPolicy"`
+	Ports           IngressPortsConfig                        `json:"ports"`
+	TuningOptions   operatorv1.IngressControllerTuningOptions `json:"tuningOptions"`
```
pacevedom, line 23: Good idea! Remember to regenerate the config files.

pkg/config/config.go
```diff
@@ -147,6 +149,20 @@ func (c *Config) fillDefaults() error {
 			Http:  ptr.To[int](80),
 			Https: ptr.To[int](443),
 		},
+		TuningOptions: operatorv1.IngressControllerTuningOptions{
+			HeaderBufferBytes:           32768,
+			HeaderBufferMaxRewriteBytes: 8192,
+			//ClientTimeout:               time.Duration(5 * time.Second),
```
pacevedom, line 155: Typo? It is configured 2 lines below.

pkg/config/config.go
```diff
@@ -271,6 +287,40 @@ func (c *Config) incorporateUserSettings(u *Config) {
...
+	}
+	if u.Ingress.TuningOptions.ServerFinTimeout != nil {
+		c.Ingress.TuningOptions.ServerFinTimeout = u.Ingress.TuningOptions.ServerFinTimeout
+	}
+	if u.Ingress.TuningOptions.TunnelTimeout != nil {
+		c.Ingress.TuningOptions.TunnelTimeout = u.Ingress.TuningOptions.TunnelTimeout
+	}
+	if u.Ingress.TuningOptions.TLSInspectDelay != nil {
+		c.Ingress.TuningOptions.TLSInspectDelay = u.Ingress.TuningOptions.TLSInspectDelay
+	}
```
pacevedom, line 317: Do we want to follow the same validation pattern as the ingress operator does? https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go#L611-L634

assets/components/openshift-router/deployment.yaml
```diff
@@ -54,7 +54,7 @@ spec:
             - name: ROUTER_TCP_BALANCE_SCHEME
               value: source
             - name: ROUTER_THREADS
-              value: "4"
+              value: '{{.ThreadCount}}'
```
pacevedom, line 57: You need to edit the rebase for this to take effect or it will be overwritten.

test/suites/standard1/router-config.robot
```diff
@@ -0,0 +1,94 @@
+*** Settings ***
```
pacevedom, line 1: What do you think about including this in the already existing router test file? https://github.com/openshift/microshift/blob/main/test/suites/standard1/router.robot

packaging/microshift/config.yaml
```diff
@@ -47,11 +47,82 @@ etcd:
...
 ingress:
+    # Determines default http version should be used for the ingress backends
+    # By default,  using version 1.
+    defaultHTTPVersion: 1
+    # forwardedHeaderPolicy specifies when and how ingress router
+    # sets the Forwarded, X-Forwarded-For, X-Forwarded-Host,
+    # X-Forwarded-Port, X-Forwarded-Proto, and X-Forwarded-Proto-Version
+    # HTTP headers.  The value may be one of the following:
+
+
```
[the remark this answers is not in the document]
eslutsky: this comment is auto generated from the sources, we can check it outside of this PR.

pkg/config/config.go
```diff
@@ -519,6 +591,28 @@ func (c Config) IsIPv6() bool {
...
+		}
+		// A single quote doesn't get escaped by strconv.Quote, so do it explicitly
+		if strings.Contains(mimeType, "'") {
+			mimeType = strings.ReplaceAll(mimeType, "'", "\\'")
+		}
+		mimes = append(mimes, mimeType)
+	}
+
+	return mimes
+}
```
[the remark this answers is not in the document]
eslutsky: since its a helper function can we place it in 
 https://github.com/openshift/microshift/blob/main/pkg/util/util.go 
 @pmtk , WDYT?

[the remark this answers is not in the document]
eslutsky: probably not.. so  i'll move it to controllers.go

ShudiLi:
the default maxConnections should be 50000, not 5000, not sure it is expected. Others look good to me, thanks.

```
1.
% oc -n openshift-ingress rsh router-default-64785749b4-zz84t
sh-5.1$ env | grep -i conn
ROUTER_MAX_CONNECTIONS=5000
sh-5.1$
sh-5.1$  cat haproxy.config | grep -i maxconn
  maxconn 5000
  maxconn 5000
sh-5.1$

2. in config.yaml.default, the default of maxConnections is 50000
        # maxConnections defines the maximum number of simultaneous
        # connections that can be established per HAProxy process.
        # Increasing this value allows each ingress controller pod to
        # handle more connections but at the cost of additional
        # system resources being consumed.


        # Permitted values are: empty, 0, -1, and the range 
        # 2000-2000000.


        # If this field is empty or 0, the IngressController will use
        # the default value of 50000, but the default is subject to
        # change in future releases.
```

ShudiLi:
Configure the custom parameters with 2 times of the default values, restart the  microshift service(sudo systemctl restart microshift), then check the router-default deployment, haproxy.conf and env under a router pod. Are are expected except connectTimeout and reloadInterval

1```
.
sh-5.1# cat /etc/microshift/config.yaml | grep connectTimeout:
        connectTimeout: 10s
sh-5.1#
sh-5.1$ grep "timeout connect" haproxy.config
  timeout connect 5s
sh-5.1$
sh-5.1$ env | grep CONN
ROUTER_MAX_CONNECTIONS=100000
sh-5.1$

2.
sh-5.1# cat /etc/microshift/config.yaml | grep reloadInterval:
        reloadInterval: 10s
sh-5.1#
% oc -n openshift-ingress get deployment router-default -oyaml | grep -A1 RELOAD_INTERVAL
        - name: RELOAD_INTERVAL
          value: 5s

sh-5.1$ env | grep RELOAD_INTERVAL
RELOAD_INTERVAL=5s
sh-5.1$
```

**Comment by eslutsky** [comment]
@ShudiLi , connectTimeout and reloadInterval wasnt meant to be exposed as ingress configuration option, we inheritted them because we used IngressControllerTuningOptions from the ingress operator [API](https://github.com/openshift/microshift/blob/main/vendor/github.com/openshift/api/operator/v1/types_ingress.go#L1715)
@pacevedom , it looks like we have to duplicate that struct into our code base.

**Comment by ShudiLi** [comment]
@eslutsky some timeout parameters weren't updated in the haproxy.config file, thanks.

```
1.
% oc -n openshift-ingress get deployment -oyaml | grep -A1 TIMEOUT
          - name: ROUTER_DEFAULT_CLIENT_TIMEOUT
            value: 2m0s
          - name: ROUTER_CLIENT_FIN_TIMEOUT
            value: 1m0s
          - name: ROUTER_DEFAULT_SERVER_TIMEOUT
            value: 2m0s
          - name: ROUTER_DEFAULT_SERVER_FIN_TIMEOUT
            value: 2m0s
          - name: ROUTER_DEFAULT_TUNNEL_TIMEOUT
            value: 2h0m0s

2. in the router pod
sh-5.1$ env | grep -i timeout
ROUTER_DEFAULT_TUNNEL_TIMEOUT=2h0m0s
ROUTER_DEFAULT_SERVER_TIMEOUT=2m0s
ROUTER_DEFAULT_CLIENT_TIMEOUT=2m0s
ROUTER_DEFAULT_SERVER_FIN_TIMEOUT=2m0s
ROUTER_CLIENT_FIN_TIMEOUT=1m0s
sh-5.1$

3. in the router pod, check the timeout parameters, "timeout tunnel", "timeout server", "timeout client", "timeout server-fin", and "timeout client-fin" were the default values, weren't updated.
sh-5.1$ cat haproxy.config | grep -i "timeout "
  stats timeout 2m
  timeout connect 5s
  timeout client 30s
  timeout client-fin 1s
  timeout server 30s
  timeout server-fin 1s
  timeout http-request 10s
  timeout http-keep-alive 300s
  # Long timeout for WebSocket connections.
  timeout tunnel 1h
sh-5.1$

```

**Comment by eslutsky** [comment]
good catch ! 
looks like [timeSpecPattern](https://github.com/openshift/router/blob/master/images/router/haproxy/conf/haproxy-config.template#L26) is not accepting  all the [metav1.Duration](https://pkg.go.dev/k8s.io/apimachinery@v0.31.1/pkg/apis/meta/v1#Duration) variables like  2m0s,so  it has to be a single value ie:
2h0m0s  should be just `2h` , we will have to use some other validation type. not sure how openshift accept those..

**Comment by eslutsky** [comment]
> good catch ! looks like [timeSpecPattern](https://github.com/openshift/router/blob/master/images/router/haproxy/conf/haproxy-config.template#L26) is not accepting all the [metav1.Duration](https://pkg.go.dev/k8s.io/apimachinery@v0.31.1/pkg/apis/meta/v1#Duration) variables like 2m0s,so it has to be a single value ie: 2h0m0s should be just `2h` , we will have to use some other validation type. not sure how openshift accept those..

i've checked again against the operator code, and we were missing the [durationToHAProxyTimespec](https://github.com/openshift/cluster-ingress-operator/blob/aa8860a85b9c74a821d1658487704ca0320308e5/pkg/operator/controller/ingress/deployment.go#L1761C6-L1761C31) function

**Comment by eslutsky** [comment]
@ShudiLi , the issue that you found has been addressed.

**Comment by eslutsky** [comment]
> ROUTER_HTTP_IGNORE_PROBES
thanks @ShudiLi 
it appears that haproxy.config expects "true" in `ROUTER_HTTP_IGNORE_PROBES` and `ROUTER_DONT_LOG_NULL` ENV, fixed.

File: pkg/components/controllers.go [hunk 8]
```diff
@@ -185,12 +195,52 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		}
 	}
 
+	routerEnableCompression := "false"
```
pacevedom, line 198: I wonder if we should have a separate function to generate all these parameters, now that there are 20 of them.

eslutsky: @pacevedom , so the idea is to create seperate function to genereate extraParams map[string] and call it generateParamsForIngress for example?

pacevedom: Yes, thats what I meant. It is getting lengthy here, wdyt?

eslutsky: yes it makes sense. i'll do it.

eslutsky: @pacevedom , resolved, please take a look.

pmtk:
this looks good to me

assets/components/openshift-router/deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
             - name: ROUTER_CIPHERSUITES
               value: TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256
             - name: ROUTER_DISABLE_HTTP2
-              value: "true"
+              value: '{{.RouterDisableHttp2}}'
```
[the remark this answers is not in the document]
pacevedom: Is it still missing?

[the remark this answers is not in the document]
eslutsky: yeap, it was hardcoded in rebase.sh - **fixed**

eslutsky:
squashed the extra commits

pacevedom (approved):
/lgtm

PR #4008, opened 2024-10-02
USHIFT-4534: Add rebase capabilities for gateway api

pacevedom:

Closes #<Issue Number>


pacevedom:
/hold
Will add the new manifests in new commits instead of a new PR.

scripts/auto-rebase/rebase_gateway_api.sh
```diff
@@ -0,0 +1,470 @@
...
+    fi
+  fi
+}
+
+patch_image_values() {
+  local kind=$1
+  local file=$2
+
+  if [[ $(yq e ".kind == \"${kind}\"" "${file}") == "true" ]]; then
+    yq 'with(.spec.template.spec.containers[]; .image=.name)' -i "${file}"
```
pmtk, line 193: I think we generally place `-i` right after `yq` so it's more visible it's editing the file

scripts/auto-rebase/rebase_gateway_api.sh
```diff
@@ -0,0 +1,470 @@
...
+rebase_ossm_to() {
+    local operator_bundle_manifest="$1"
+    local gateway_api_version="$2"
+
+    title "# Rebasing OSSM to ${operator_bundle_manifest}"
+
+    download_ossm_operator_bundle_manifest "${operator_bundle_manifest}" "${gateway_api_version}"
+
+    # OSSM image names may include `/` and `:`, which make messy branch names.
+    rebase_branch="rebase-ossm-${operator_bundle_manifest//[:\/]/-}"
```
pmtk, line 88: Should this include the "base branch"? Like main, release-4.18 and so on? Or is it unnecessary? Will this be executed like normal rebase (periodically) or like lvms (on-demand)?

pacevedom: This will be executed on demand because the bundles are artifacts from brew (they get published as catalogs, this is an optional operator).

scripts/auto-rebase/rebase_gateway_api.sh
pacevedom: Done

PR #4048, opened 2024-10-10
USHIFT-4349: Fixed --osinfo=generic warning

vanhalenar:
`virt-install` can't detect the OS automatically, most likely because we're building the images ourselves, therefore the OS needs to be specified manually.


test/bin/scenario.sh
```diff
@@ -533,13 +533,19 @@ launch_vm() {
     vm_extra_args="fips=${fips_mode}"
     vm_initrd_inject=""
 
+    declare -A os_map=(
+        ["rhel94-bootc"]="rhel9.4"
+        ["rhel-9.4"]="rhel9.4"
+        ["centos9-bootc"]="centos-stream9"
+        ["centos9"]="centos-stream9")
+
```
ggiguash, line 541: Let's use switch statement to:
- Make name check more generic
- Report an error in case of no match

vanhalenar: done

PR #4061, opened 2024-10-14
USHIFT-4117: Auto Recovery: Restore

pmtk:
Adds two new options to `microshift restore` command:
- `--auto-recovery` to run the procedure of restoring the most recent backup from provided directory
- `--save-failed` to opt-out from making a copy of current MicroShift data to "failed" subdirectory

### DIFF
--- a/pkg/admin/autorecovery/restore.go
+++ b/pkg/admin/autorecovery/restore.go
@@ -1,0 +1,180 @@
+package autorecovery
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/openshift/microshift/pkg/admin/data"
+	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/util"
+	"k8s.io/klog/v2"
+)
+
+const (
+	failedSubstorageName   = "failed"
+	restoredSubstorageName = "restored"
+)
+
+type Manager struct {
+	storage    data.StoragePath
+	saveFailed bool
+}
+
+func NewManager(storage data.StoragePath, saveFailed bool) (*Manager, error) {
+	if storage == "" {
+		return nil, fmt.Errorf("`storage` argument is empty")
+	}
+
+	return &Manager{storage: storage, saveFailed: saveFailed}, nil
+}
+
+func (m *Manager) PerformRestore() error {
+	if err := storageShouldExist(m.storage); err != nil {
+		return err
+	}
+
+	existingState, err := GetState(m.storage)
+	if err != nil {
+		return err
+	}
+
+	restoreCandidate, err := getCandidateForRestore(m.storage, existingState)
+	if err != nil {
+		return err
+	}
+
+	/*
+		Preparations - initial file system operations:
+		  COPY    $STORAGE/$CANDIDATE -> /var/lib/microshift.tmp
+		  COPY    /var/lib/microshift -> $STORAGE/failed/$DATE_DEPLOY-ID.tmp
+		  CREATE  $STORAGE/state.json.new
+		  -       $STORAGE/$PREVIOUSLY_RESTORED									Should already exist, nothing to do.
+
+		"Closing the transaction"
+		  RENAME    /var/lib/microshift.tmp -> /var/lib/microshift
+						This operation is first, because it's the most important one - whole point of the restore is to get to running state.
+		  RENAME    $STORAGE/failed/DATE_DEPLOY-ID.tmp -> $STORAGE/failed/DATE_DEPLOY-ID
+		  RENAME    $STORAGE/state.json.new -> $STORAGE/state.json
+		  RENAME    $STORAGE/$PREVIOUSLY_RESTORED -> $STORAGE/restored/$PREVIOUSLY_RESTORED
+	*/
+
+	// Copies/creations into intermediate destinations
+	var oldData *data.AtomicDirCopy
+	if m.saveFailed {
+		oldDataBackupName, err := GetBackupName()
+		if err != nil {
+			return err
+		}
+		failedStorage := m.storage.SubStorage(failedSubstorageName)
+		if err := os.MkdirAll(string(failedStorage), 0600); err != nil {
+			return fmt.Errorf("failed to create %q subdirectory: %w", failedSubstorageName, err)
+		}
+		oldData = &data.AtomicDirCopy{Source: config.DataDir, Destination: failedStorage.GetBackupPath(oldDataBackupName)}
+		if err := oldData.CopyToIntermediate(); err != nil {
+			return fmt.Errorf("old microshift data: %w", err)
+		}
+	}
+
+	newData := data.AtomicDirCopy{Source: m.storage.GetBackupPath(restoreCandidate.Name()), Destination: config.DataDir}
+	if err := newData.CopyToIntermediate(); err != nil {
+		if rollbackErr := oldData.RollbackIntermediate(); rollbackErr != nil {
+			klog.ErrorS(rollbackErr, "Failed to rollback intermediate state for old data")
+		}
+		return fmt.Errorf("new microshift data: %w", err)
+	}
+
+	newState := NewState(m.storage, restoreCandidate.Name())
+	if err := newState.SaveToIntermediate(); err != nil {
+		if rollbackErr := oldData.RollbackIntermediate(); rollbackErr != nil {
+			klog.ErrorS(rollbackErr, "Failed to rollback intermediate state for old data")
+		}
+		if rollbackErr := newData.RollbackIntermediate(); rollbackErr != nil {
+			klog.ErrorS(rollbackErr, "Failed to rollback intermediate state for new data")
+		}
+		return fmt.Errorf("new state file: %w", err)
+	}
+
+	// Renames into final destinations
+	if err := newData.RenameToFinal(); err != nil {
+		return fmt.Errorf("new microshift data: %w", err)
+	}
+	// Update the state file right after /var/lib/microshift is restored from a backup
+	if err := newState.MoveToFinal(); err != nil {
+		return fmt.Errorf("new state file: %w", err)
+	}
+	if err := oldData.RenameToFinal(); err != nil {
+		return fmt.Errorf("old microshift data: %w", err)
+	}
+
+	//nolint:nestif
+	if existingState != nil {
+		previous := m.storage.GetBackupPath(existingState.LastBackup)
+		exists, err := util.PathExists(previous)
+		if err != nil {
+			klog.ErrorS(err, "Failed to check if previously restored backup exists")
+			return err
+		}
+		if exists {
+			restoredStorage := m.storage.SubStorage(restoredSubstorageName)
+			if err := os.MkdirAll(string(restoredStorage), 0600); err != nil {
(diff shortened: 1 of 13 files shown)


pkg/util/atomic_copier.go
```diff
@@ -0,0 +1,132 @@
...
+var (
+	cpArgs = []string{
+		"--verbose",
+		"--recursive",
+		"--preserve",
+		"--reflink=auto",
+	}
+)
+
+// AtomicCopier performs a two operation: copies source path to an intermediate location,
```
ggiguash, line 23: To be more precise, it's not any intermediate location. It's the same destination path, but with temporary name. Can we make it clearer?

pkg/util/atomic_copier.go
```diff
@@ -0,0 +1,132 @@
...
+// is either fully renamed or not at all, which helps prevent issues like partial
+// file copies in cases of power failure or unexpected program termination.
+type AtomicCopier struct {
+	Source      string
+	Destination string
+
+	intermediatePath string
+}
+
+func (c *AtomicCopier) CopyToIntermediate() error {
```
ggiguash, line 35: Should we allow optional cpArgs override here?

ggiguash: The default copy technique and args need to be documented.

pkg/util/atomic_copier.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	c.intermediatePath = fmt.Sprintf("%s.tmp", c.Destination)
+	if exists, err := PathExists(c.intermediatePath); err != nil {
+		return err
+	} else if exists {
+		if err := os.RemoveAll(c.intermediatePath); err != nil {
+			klog.Errorf("Failed to remove intermediate path %q which already existed: %v", c.intermediatePath, err)
+			return fmt.Errorf("failed to remove %q: %w", c.intermediatePath, err)
+		}
+	}
```
ggiguash, line 47: I'm not living in peace with this approach of adding `.tmp` suffix. It seems to be too common for the generic `util` code. 
Can we do something like `mktemp` and return a fatal error if the destination exists? I'm a bit wary of deleting files while copying them.

pkg/util/atomic_copier.go
```diff
@@ -0,0 +1,132 @@
...
+		klog.ErrorS(nil, "Failed to make an intermediate copy", "cmd", cmd,
+			"stdout", strings.ReplaceAll(outb.String(), "\n", `, `),
+			"stderr", errb.String())
+		copyErr := fmt.Errorf("failed to copy %q to %q: %w", c.Source, c.intermediatePath, err)
+		if err := os.RemoveAll(c.intermediatePath); err != nil {
+			return errors.Join(copyErr, fmt.Errorf("failed to remove intermediate path %q: %w", c.intermediatePath, err))
+		}
+		return copyErr
+	}
+	klog.InfoS("Made an intermediate copy", "cmd", cmd)
```
ggiguash, line 65: Do you mean this?
```suggestion
	klog.InfoS("Completed an atomic file copy operation", "cmd", cmd)
```

ggiguash, line 65: Do you mean this?
```suggestion
	klog.InfoS("Completed an atomic file copy operation", "cmd", cmd)
```

pkg/util/atomic_copier.go
```diff
@@ -0,0 +1,132 @@
...
+		// Now it needs to be renamed into final destination.
+		// This two-step operation should provide a high guarantee that
+		// copying is complete and not partial thanks to rename being OS/filesystem atomic.
+		src = c.intermediatePath
+		dest = c.Destination
+	}
+
+	// Delete the destination if it's a non-empty directory.
+	// This is a limitation of the POSIX's rename.
+	if err := removeDirIfExists(dest); err != nil {
```
ggiguash, line 103: I think this should be a fatal error. We should not delete leftover paths as a byproduct of copy operation.

pkg/util/atomic_copier.go
ggiguash, line 1: Nit: what are we copying here? Is it `atomic_dir_copy.go`, or more generic `atomic_file_copy.go`?

pkg/admin/data/data_manager.go
```diff
@@ -224,48 +221,13 @@ func checkDirectoryContents(existing sets.Set[string]) error {
 }
 
 func copyPath(src, dest string) error {
-	tmpDest := fmt.Sprintf("%s.tmp", dest)
-	if exists, err := pathExists(tmpDest); err != nil {
+	copier := util.AtomicCopier{Source: src, Destination: dest}
+	if err := copier.CopyToIntermediate(); err != nil {
```
ggiguash, line 225: What's the reason for exposing the intermediate / final implementation of the object? 
Should we just have a public `Copy` function and hide the implementation specifics?

pkg/util/atomic_copier.go
pmtk: This is not leftover path - this is final path. And without this `mv /var/lib/microshift.tmp /var/lib/microshift` doesn't work because of `Rename()` limitations in POSIX and UNIX.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/rename.html
> If new names an existing directory, it shall be required to be an empty directory.

```
$ mkdir 1 2; touch 1/a 2/b
$ mv 1 2
$ ls
2/
$ ls 2
1/  b
```
From `mv` man (same thing in the `cp` as they use the same library underneath):
```
       -T, --no-target-directory
              treat DEST as a normal file
```

```
$ mkdir 1 2; touch 1/a 2/b
$ mv -T 1 2
mv: cannot move '1' to '2': File exists
```

pkg/util/atomic_copier.go
pmtk: Not really, not yet at least. The "atomic file copy operation" will be complete after rename, it's just a first part.

pkg/admin/data/data_manager.go
pmtk: Do you mean the "ToIntermediate" part of the name?

test/suites/backup/auto-recovery.robot
```diff
@@ -11,29 +11,114 @@ Suite Setup         Setup
...
+State File Is Created
+    [Documentation]    Restoring causes a state file creation.
+
+    ${last_backup}=    Create Backups    1
+    ${backup_name}=    Remove String    ${last_backup}    ${WORKDIR}
+
+    Command Should Work    microshift restore --auto-recovery ${WORKDIR}
+    ${state_json}=    Command Should Work    cat ${WORKDIR}/state.json
+    ${state}=    Json Parse    ${state_json}
+    Should Be Equal As Strings    ${backup_name}    ${state}[LastBackup]
```
raz126, line 107: gmail

raz126 (approved):

pkg/cmd/admin.go
```diff
@@ -218,5 +229,15 @@ func NewRestoreCommand() *cobra.Command {
...
 	}
 
+	cmd.Flags().BoolVar(&autorec, "auto-recovery", false,
+		`Changes the behavior of the restore command.
+The PATH argument will be treated as a directory where backups are
+created with "backup --auto-recovery" command.`)
+
+	cmd.Flags().BoolVar(&saveFailed, "save-failed", true,
+		`Only applicable if --auto-recovery is also specified.
+It true (default), a copy of MicroShift data directory will be made inside
```
agullon, line 239:
```suggestion
If true (default), a copy of MicroShift data directory will be made inside
```
is this a typo, right?

pmtk: Yes, thx

pkg/admin/data/atomic_dir_copy.go
```diff
@@ -0,0 +1,149 @@
...
+	cmd.Stdout = &outb
+	cmd.Stderr = &errb
+	err = cmd.Run()
+
+	if err != nil {
+		klog.ErrorS(nil, "Failed to make an intermediate copy", "cmd", cmd,
+			"stdout", strings.ReplaceAll(outb.String(), "\n", `, `),
+			"stderr", errb.String())
+		copyErr := fmt.Errorf("failed to copy %q to %q: %w", c.Source, c.intermediatePath, err)
+		if err := os.RemoveAll(c.intermediatePath); err != nil {
```
pacevedom, line 61: Could you call RollbackIntermediate?

pkg/admin/autorecovery/restore.go
```diff
@@ -0,0 +1,180 @@
...
+		if rollbackErr := oldData.RollbackIntermediate(); rollbackErr != nil {
+			klog.ErrorS(rollbackErr, "Failed to rollback intermediate state for old data")
+		}
+		if rollbackErr := newData.RollbackIntermediate(); rollbackErr != nil {
+			klog.ErrorS(rollbackErr, "Failed to rollback intermediate state for new data")
+		}
+		return fmt.Errorf("new state file: %w", err)
+	}
+
+	// Renames into final destinations
```
pacevedom, line 97: If any of the following fail the rollbacks are performed within their own functions, right? What about the other variables (newData, newState, oldData)? Would a defer help with the data about what to rollback help?

pkg/admin/data/atomic_dir_copy.go
```diff
@@ -0,0 +1,149 @@
...
+// file copies in cases of power failure or unexpected program termination.
+type AtomicDirCopy struct {
+	Source      string
+	Destination string
+
+	intermediatePath string
+}
+
+func (c *AtomicDirCopy) CopyToIntermediate() error {
+	if c == nil {
```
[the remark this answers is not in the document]
pmtk: Yes, exactly

pkg/admin/autorecovery/storage.go
```diff
@@ -0,0 +1,104 @@
...
+			continue
+		}
+		dt := splitName[0]
+		ver := splitName[1]
+		creationTime, err := time.Parse(backupCreationTimeFormat, dt)
+		if err != nil {
+			klog.ErrorS(err, "Failed to parse datetime part of the backup name", "name", file)
+			continue
+		}
+		// TODO: check if creationDate is older than Now?
```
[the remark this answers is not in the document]
pmtk: Then the user has bigger problems than backups :D
Forgot to remove the comment, I don't think there's need to be so strict. And I think I meant "what if creationDate is newer than Now" :D

pkg/admin/data/atomic_dir_copy.go
pmtk: Yes, good catch

pkg/admin/autorecovery/restore.go
pmtk: When this fails in the middle of renaming, I don't think there's much to do...
For example, if `oldData.RenameToFinal()` fails, then to roll everything back we'd have to move `oldData.Intermediate` -> `/var/lib/microshift` (because it already contains the data we wanted to restore). I feel like this would be very volatile twister game.
If the FS would offer transactions, we'd be golden, but I think renaming microshift data into place first is good enough (although we still return error and rc=1 so it might not be interpreted that way, maybe we need more exit codes?)

pacevedom (approved):
/lgtm

PR #4066, opened 2024-10-15
USHIFT-4565: Gateway API rpm

pacevedom:
Includes:
- Rebase to latest released tech preview OSSM3.
- New gateway api rpm.
- Greenboot script.
- Some fixes to the automatic rebase for OSSM3.

Closes #<Issue Number>


packaging/rpm/microshift.spec
```diff
@@ -445,6 +453,19 @@ install -p -m644 packaging/tuned/runtime-class/kustomization.yaml %{buildroot}/%
 install -p -m644 packaging/tuned/microshift-tuned.service %{buildroot}%{_unitdir}/microshift-tuned.service
 install -p -m755 packaging/tuned/microshift-tuned.py %{buildroot}%{_bindir}/microshift-tuned
 
+# gateway-api
+install -d -m755 %{buildroot}/%{_prefix}/lib/microshift/manifests.d/000-microshift-gateway-api
+install -p -m644 assets/optional/gateway-api/0* %{buildroot}/%{_prefix}/lib/microshift/manifests.d/000-microshift-gateway-api
+install -p -m644 assets/optional/gateway-api/kustomization.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/000-microshift-gateway-api
+install -p -m755 packaging/greenboot/microshift-running-check-gateway-api.sh %{buildroot}%{_sysconfdir}/greenboot/check/required.d/41_microshift_running_check_gateway_api.sh
```
pmtk, line 460: multus uses 41, 42 will work?
```suggestion
install -p -m755 packaging/greenboot/microshift-running-check-gateway-api.sh %{buildroot}%{_sysconfdir}/greenboot/check/required.d/42_microshift_running_check_gateway_api.sh
```

packaging/rpm/microshift.spec
```diff
@@ -445,6 +453,19 @@ install -p -m644 packaging/tuned/runtime-class/kustomization.yaml %{buildroot}/%
 install -p -m644 packaging/tuned/microshift-tuned.service %{buildroot}%{_unitdir}/microshift-tuned.service
 install -p -m755 packaging/tuned/microshift-tuned.py %{buildroot}%{_bindir}/microshift-tuned
 
+# gateway-api
+install -d -m755 %{buildroot}/%{_prefix}/lib/microshift/manifests.d/000-microshift-gateway-api
```
pmtk, line 457: Does it have to 000? Ideally each manifest would have different number

scripts/auto-rebase/rebase_gateway_api.sh
```diff
@@ -91,20 +91,12 @@ rebase_ossm_to() {
...
-    if [[ -n "$(git status -s pkg/release)" ]]; then
-        title "## Committing changes to pkg/release"
-        git add pkg/release
-        git commit -m "update OSSM images"
-    else
-        echo "No changes in OSSM images."
-    fi
-
     update_ossm_manifests "${operator_bundle_manifest}"
+    update_ossm_images
```
pmtk, line 95: What about `release-gateway-api-$arch.json` file for OSSM?

packaging/rpm/microshift.spec
pacevedom: I understand these prefixes as "dependency ordering". Since gateway api does not depend on anything else I thought this was good enough?

packaging/rpm/microshift.spec
pacevedom: This is the same case as above. With greenboot you need the first one to be `40_` because thats the baseline. From there you may use prefixes as dependencies. Since gateway api does not depend on any other components it can share the prefix, it could run at the same time as multus, for example.

scripts/auto-rebase/rebase_gateway_api.sh
pacevedom: This one is going to be part of a different PR, I wanted to establish this one, then tests, then the release-info. Those files need to be created manually and updated automatically (they are not included in the bundle, but there is data there to feed them)

PR #4075, opened 2024-10-17
[release-4.17] Fix microshift-test-agent race conditions

pmtk:


ggiguash:
Should this be a cherry pick?
/retitle [release-4.17] Fix microshift-test-agent race conditions

PR #4080, opened 2024-10-19
USHIFT-4704: Port bootc build procedure to directly use mirror registry instead of local storage

ggiguash:
The functional changes include:
* Consolidation of Web and Registry URL definition using VM bridge interface, or host name if libvirt is not configured
* Mirror registry script uses an external storage and configuration file
* Bootc containerfile and container encapsulate builds save artifacts to the mirror registry directly bypassing file system export.
* Mirror registry startup is implicit when running bootc builds
* Mirror registry is no longer optional

These changes should result in reduced resource consumption and reduced runtime of the builds.

**Test prerequisites**
* Delete `_output`
* `sudo podman rmi -af && podman rmi -af && podman volume rm -af`
* Run `./test/bin/build_rpms.sh` and download cache.
* Compare bootc build times and disk space

**Test results on AWS `c5n.metal` host**
* Registry creation of 1m is included in the build time and 34G of disk saved.
* Registry recreation is fast due to external persistent storage

|What                       | Before|After|
|--------------------------|-----------|-------|
|Bootc build time   | 10m    | 10m|
|Create mirror reg | 1.5m   | 20s  |
|sudo podman df  | 11G     |  11G |
| podman df           | 16G     | 0G  |
|_output du            | 53G     | 35G |

### DIFF
--- a/test/bin/common.sh
+++ b/test/bin/common.sh
@@ -119,9 +119,6 @@ RF_VENV=${RF_VENV:-${OUTPUTDIR}/robotenv}
 # shellcheck disable=SC2034  # used elsewhere
 export GOMPLATE=${OUTPUTDIR}/bin/gomplate
 
-# Which port the web server should run on.
-WEB_SERVER_PORT=${WEB_SERVER_PORT:-8080}
-
 title() {
     # Only use color when reporting to a terminal
     if [ -t 1 ]; then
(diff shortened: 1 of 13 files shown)


test/bin/common.sh
```diff
@@ -190,6 +191,22 @@ get_vm_bridge_ip() {
...
 
+# Web server port number
+WEB_SERVER_PORT=8080
+
+# Web server URL using VM bridge IP with fallback to host name
+# shellcheck disable=SC2034  # used elsewhere
+WEB_SERVER_URL="http://${VM_BRIDGE_IP:-$(hostname)}:${WEB_SERVER_PORT}"
+
+# Mirror registry port number
+# shellcheck disable=SC2034  # used elsewhere
```
pmtk, line 202: Probably unnecessary as the variable is used couple lines below?

pmtk (commented):
One nit, can be fix some other time
/lgtm

test/bin/common.sh
ggiguash: I will fix it when rebasing https://github.com/openshift/microshift/pull/3991

PR #4101, opened 2024-10-24
USHIFT-4705: Check disk space before backing up / restoring

pmtk:


jerpeter1:
/hold /lgtm
@pmtk - you might consider padding the "space required" that's reported by a bit to account for any ongoing changes that might be in flight while it's being calculated - but otherwise looks fine - unhold if you feel like it

pacevedom (approved):
Padding some extra space was added, so lgtm again.

/lgtm

PR #4120, opened 2024-10-30
USHIFT-4330: Running RF tests against a MicroShift in container

pmtk:


test/bin/manage_hypervisor_config.sh
```diff
@@ -8,6 +8,9 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=test/bin/common.sh
 source "${SCRIPTDIR}/common.sh"
 
+VGFILE=/var/lib/microshift-lvm-storage.img
```
ggiguash, line 11: I'm not sure this is the "right" place for the storage file. 
Could it be that it fits better in per-scenario directories? Otherwise, different tests can affect each other by adjusting the same file, right?

test/bin/manage_hypervisor_config.sh
```diff
@@ -123,6 +126,13 @@ action_create() {
 
     # Start nginx web server
     "${TESTDIR}/bin/manage_webserver.sh" "start"
+
+    if ! sudo vgdisplay rhel; then
```
ggiguash, line 130: As above, this might need to be part of `scenario.sh` script.

test/bin/manage_hypervisor_config.sh
```diff
@@ -159,6 +169,13 @@ action_cleanup() {
 
     # Stop nginx web server
     "${TESTDIR}/bin/manage_webserver.sh" "stop"
+
+    if [ -e "${VGFILE}" ]; then
```
ggiguash, line 173: As above, this might need to be part of `scenario.sh` script.

test/bin/scenario.sh
```diff
@@ -16,6 +16,8 @@ source "${SCRIPTDIR}/common.sh"
 # shellcheck source=test/bin/common_versions.sh
 source "${SCRIPTDIR}/common_versions.sh"
 
+source "${SCRIPTDIR}/scenario_container.sh"
```
ggiguash, line 19: Do we not need shellcheck directive here?

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
+
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
```
ggiguash, line 3: That's a great idea. Can you add it to the other two files we source in `tests/bin`?

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+    echo "This script must be sourced, not executed."
+    exit 1
+fi
+
+process_kickstart_into_script() {
```
ggiguash, line 8: Please, add a few lines of docs explaining what the function does.

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+    while IFS= read -r line; do
+        if [[ $line =~ ^%include[[:space:]]+(.+) ]]; then
+            process_kickstart_into_script "${container_name}" "${BASH_REMATCH[1]}"
+        else
+            echo "$line" >> "${output_dir}/init.sh"
+        fi
+    done < "${output_dir}/${file}"
+}
+
+create_container_entrypoint() {
```
ggiguash, line 23: Please, add a few lines of docs explaining what the function does.

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+    local -r name="$1"
+    local -r output_dir="${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${name}/"
+
+    cat << EOF > "${output_dir}/init.sh"
+#!/bin/bash
+set -xeuo pipefail
+
+cat /proc/sys/kernel/random/uuid > /proc/sys/kernel/random/boot_id
+
+if [ -e /var/container-init-ok ]; then
```
ggiguash, line 33: Should it be in /var/run so that file is not persisted after reboot?

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+
+    cat << EOF >> "${output_dir}/init.sh"
+touch /var/container-init-ok
+
+EOF
+
+    cat << 'EOF' > "${output_dir}/entrypoint.sh"
+#!/bin/bash
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+bash -x "${SCRIPTDIR}/init.sh" &> /var/log/container_init.log
```
ggiguash, line 50: Let's make the log file name less generic by adding `microshift_` prefix?

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+exec /sbin/init
+EOF
+    chmod +x "${output_dir}/entrypoint.sh"
+}
+
+launch_container() {
```
ggiguash, line 56: Please, add a few lines of docs explaining what the function does.

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+    fi
+
+    local -r full_image="${MIRROR_REGISTRY_URL}/${image}"
+
+    record_junit "${name}" "ctr-launch-args" "OK"
+
+    local -r full_ctr_name=$(full_vm_name "${name}")
+
+    if sudo podman inspect -f "{{.Id}}" "${full_ctr_name}" &> /dev/null; then
+        echo "Container '${full_ctr_name}' already exists."
```
ggiguash, line 95: Should this be an error to be consistent with startup of VMs?

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+        record_junit "${name}" "launch_ctr" "SKIP"
+        exit 0
+    fi
+
+    sudo modprobe openvswitch
+
+    prepare_kickstart "${name}" kickstart-bootc-container.ks.template ""
```
ggiguash, line 102: We run prepare_kickstart in scenarios. Can we move it for consistency?

test/bin/scenario_container.sh
```diff
@@ -0,0 +1,162 @@
...
+        --cpus "${vcpus}" \
+        --name "${full_ctr_name}" \
+        --tls-verify=false \
+        "${image}"; then
+        record_junit "${name}" "launch_ctr" "FAILED"
+        exit 1
+    fi
+
+    local -r ip=$(sudo podman inspect "${full_ctr_name}" | jq -r '.[0].NetworkSettings.IPAddress')
+    echo "Container ${full_ctr_name} has IP ${ip}"
```
ggiguash, line 127: Is there not a race here with the container startup in the detached mode?

test/kickstart-templates/includes/post-root-fs-shared.cfg
ggiguash, line 1: We have this as part of the bootc images. Why do we need it again in kickstart?

pmtk: Can't find it in the images. How does it look like?

test/bin/scenario_container.sh
pmtk: Haven't had a problem honestly. I think `-d` is different from e.g. systemd's --no-block - the container is created and started, it's just isn't attached to.
But we can poll for IP first for safety

test/bin/scenario_container.sh
pmtk: What do you mean? Remove "Container" so it's exactly the same?
```
    if sudo virsh dominfo "${full_vmname}" 2>/dev/null; then
        echo "${full_vmname} already exists"
        record_junit "${vmname}" "install_vm" "SKIP"
        return 0
    fi
```

test/bin/scenario_container.sh
pmtk: We don't want these on each reboot. Only the first one. These kickstart commands are written to be executed once during installation, they don't handle rerunning well.

test/bin/manage_hypervisor_config.sh
pmtk: Not really. VG is something shared across the containers, so I didn't want to `scenario.sh cleanup` to remove a VG that might be used by different container.
I'm curious though if we could run these commands inside a container to get isolated VGs.

test/bin/manage_hypervisor_config.sh
pmtk: It might be, but see other comment for nuance around cleanup - it would be asymmetrical.

test/bin/manage_hypervisor_config.sh
pmtk: Yeah, but the VGs are shared between host and containers.

test/bin/scenario_container.sh
ggiguash: I see. Can you store the file under a directory, e.g. `/var/adm` or like?

test/bin/scenario_container.sh
ggiguash: Ah, I see [here](https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L503-L507) that we're skipping VM creation the same way. All good.

test/bin/scenario_container.sh
ggiguash: Yeah, let's poll. It may be necessary when running tests in parallel on an overloaded system.

test/bin/manage_hypervisor_config.sh
ggiguash: I'm concerned how this is going to work if we have multiple containers accessing the same vg.
Apparently, there is --device option in podman that allows mapping only specific devices.
Can you try to see if this addresses our concerns?

test/suites/optional/multus.robot
```diff
@@ -9,6 +9,8 @@ Resource            ../../resources/microshift-rpm.resource
 Suite Setup         Setup
 Suite Teardown      Teardown Suite With Namespace
 
+Test Tags           vm
```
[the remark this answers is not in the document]
ggiguash: If the meaning is that we can only run this test in vm, should it be `vm-only`?

test/kickstart-templates/includes/post-system.cfg
```diff
@@ -1,9 +1,16 @@
 # Create a default redhat user, allowing it to run sudo commands without password
-useradd -m -d /home/redhat -p \$5\$XDVQ6DxT8S5YWLV7\$8f2om5JfjK56v9ofUkUAwZXTxJl3Sqnc9yPnza4xoJ0 redhat
+useradd -m -p \$5\$XDVQ6DxT8S5YWLV7\$8f2om5JfjK56v9ofUkUAwZXTxJl3Sqnc9yPnza4xoJ0 redhat
 echo -e 'redhat\tALL=(ALL)\tNOPASSWD: ALL' > /etc/sudoers.d/microshift
 
 # Make the KUBECONFIG from MicroShift directly available for the root user
-echo -e 'export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig' >> /root/.profile
+if [ ! -d /root ]; then
+    # Workaround for bootc container init. /var/roothome (which /root symlinks to) is not yet present.
```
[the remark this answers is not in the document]
pmtk: I think systemd or some service sets it all up. If we create it here, it'll stay.
That file doesn't exist on normal install anyway, so it's not like we lose any content that might be there.

Also, was the intention that a manual `source ~/.profile` is needed each time? Because it's not automatically sourced when logging using `sudo su`

test/suites/optional/multus.robot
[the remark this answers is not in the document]
pmtk: It's used with `--exclude` in the scenario script

test/kickstart-templates/includes/post-system.cfg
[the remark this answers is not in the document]
ggiguash: Does it not work with `sudo -i`? 
P.S. `sudo su -` should be the `su` equivalent.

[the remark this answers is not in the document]
ggiguash: In any case, let's expand on the comment, saying that we expect the file to be there after the /root is mounted.

test/suites/optional/multus.robot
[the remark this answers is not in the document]
ggiguash: Great idea!

PR #4156, opened 2024-11-07
OCPBUGS-42864: Add sample kickstart files to microshift-release-info RPM and document usage

ggiguash:


packaging/kickstart/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+Log into the virtual machine using the `redhat:<password>` credentials.
+Run the following command to verify that all the MicroShift pods are up and running
+without errors.
+
+```bash
+watch sudo oc get pods -A \
+    --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig
+```
```
ShaunaDiaz, line 101: Would it be useful to link to the kickstart reference in the RHEL docs? e.g., https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/automatically_installing_rhel/kickstart-script-file-format-reference_rhel-installer

ShaunaDiaz, line 101: Wonder if we should also state which sections are required for MicroShift, since the RHEL docs say "All sections except the command section are optional and can be present multiple times." WDYT?

ShaunaDiaz, line 101: Is this true for MicroShift installations? "Omitting any required item results in the installation program changing to the interactive mode so that the user can provide an answer to the related item..." or is that only for a RHEL install? Wondering if we need to let users know that if the kickstart is incomplete, the VM and MicroShift won't start.

ggiguash: There may be lots of potential errors in kickstart, but I was trying to build a complete example. So, knowingly, we should not end up with incomplete kickstart.

packaging/kickstart/README.md
```diff
@@ -0,0 +1,101 @@
...
+## Procedure Overview
+
+* Follow the instructions in [Prerequisites](#prerequisites).
+* Depending on the desired installation type, follow the instructions one of the
+  [RPM](#RPM), [Image Mode](#image-mode) or [OSTree](#OSTree) sections to create
+  a working kickstart file from a template.
+* Follow the instructions in [Create Virtual Machine](#create-virtual-machine)
+  to bootstrap a host using the kickstart file from the previous step.
+
+## Prepare Kickstart File
```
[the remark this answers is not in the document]
ggiguash: We can do this in the production docs. I'd like to keep the upstream docs simple.

packaging/kickstart/README.md
```diff
@@ -0,0 +1,101 @@
...
+
```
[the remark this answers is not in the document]
ggiguash: We can do this in the production docs. I'd like to keep the upstream docs simple.

packaging/kickstart/README.md
```diff
@@ -0,0 +1,101 @@
...
+### Prerequisites
+
+Install the utilities used for kickstart file creation.
```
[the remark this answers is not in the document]
ggiguash: Those instructions go as far as validating the kickstart contents. I think this is out of scope for the current doc.

packaging/kickstart/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+```bash
+export PULL_SECRET="$(cat ~/.pull-secret.json)"
+# Make sure to encrypt the password
+PASSWD_TEXT=<my_redhat_user_password>
+export USER_PASSWD=$(echo -n ${PASSWD_TEXT} | mkpasswd -m yescrypt -s)
+```
+
+### RPM
+
```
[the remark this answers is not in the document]
ggiguash: Yes, that's the idea. I've already added RPM part. The ostree is coming soon.

packaging/kickstart/kickstart-bootc.ks.template
```diff
@@ -0,0 +1,68 @@
+lang en_US.UTF-8
+keyboard us
+timezone UTC
+text
+reboot
+
+# Partition the disk with hardware-specific boot and swap partitions, adding an
```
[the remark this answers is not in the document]
ggiguash: All the comments in the file are in this style, e.g. partition disk, lock root user, register host, etc.

packaging/kickstart/kickstart-bootc.ks.template
```diff
@@ -0,0 +1,68 @@
...
+# LVM volume that contains a 10GB+ system root. The remainder of the volume will
+# be used by the CSI driver for storing data.
+zerombr
+clearpart --all --initlabel
+# Create boot and swap partitions as required by the current hardware platform
+reqpart --add-boot
+# Add an LVM volume group and allocate a system root logical volume
```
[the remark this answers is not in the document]
ggiguash: No minimum size. I'd like to avoid complicating instructions further. What we have here is a sensible start, I think. Users can then change the values if necessary.

packaging/kickstart/kickstart-ostree.ks.template
```diff
@@ -0,0 +1,58 @@
...
+# Configure network to use DHCP and activate on boot
+network --bootproto=dhcp --device=link --activate --onboot=on
+
+%pre-install --log=/dev/console --erroronfail
+
+# Create an rpm-ostree authentication file
+mkdir -p /etc/ostree/
+cat > /etc/ostree/auth.json <<'EOF'
+${AUTH_CONFIG}
+EOF
```
DanielFroehlich, line 33: I like we are doing private registry with auth here! How about also adding a custom CA trust root, in case the private registry has a custom CA cert?
```
# Add custom ca cert to trust root:
mkdir -p /etc/pki/ca-trust/source/anchors/
cat > /etc/pki/ca-trust/source/anchors/2023CertificateAuthorityRHCSv2_Chain.pem << EOF
${AUTH_ADDITIONAL_CA_TRUST}
EOF
update-ca-trust

```

DanielFroehlich (approved):
LGTM

packaging/kickstart/kickstart-ostree.ks.template
ggiguash: If we add this, it should go into a generic section common to all kickstart files, right? 
If we consider this mandatory, I can add this. Otherwise, users can extrapolate on the template and implement this and many other customizations in a similar manner.
WDYT?

DanielFroehlich: Its mandatory only if the the registry (bootc) or repo-server (ostree-commit) is using TLS with a custom CA certificate. Which happens quite frequently in enterprise grade use cases / deployments. And it took me a while to figure out where / how to do this.
Its a tricky balance between keeping the docs/example simple and easy, but also realistic and helpfull to use.

ggiguash: I'd suggest keeping the current "simplified" content and add appendix in the production docs about any additions we'd like to users to pay attention to.

packaging/kickstart/README.md
```diff
@@ -0,0 +1,172 @@
...
+> The subscription must include access to the `rhocp-4.x-for-rhel-9-$(uname -m)-rpms`
+> and `fast-datapath-for-rhel-9-$(uname -m)-rpms` RPM repositories.
+
+* `RHSM_ORG` contains an RHSM organization ID for the subscription registration
+  command in kickstart.
+* `RHSM_KEY` contains an RHSM activation key for the subscription registration
+  command in kickstart.
+* `MICROSHIFT_VER` references the MicroShift version to install using the `4.x`
+  format. Note that the latest available `.y` version will be installed.
```
pmtk, line 61:
```suggestion
* `MICROSHIFT_VER` references the MicroShift version to install using the `4.y`
  format. Note that the latest available `.z` version will be installed.
```

packaging/kickstart/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+```bash
+envsubst < \
+    /usr/share/microshift/kickstart/kickstart-ostree.ks.template > \
+    "${HOME}/kickstart.ks"
+```
+
+## Create Virtual Machine
+
+Download a RHEL boot ISO image from https://developers.redhat.com/products/rhel/download.
```
pmtk, line 141: Would https://access.redhat.com/downloads/content/rhel be more generic link? Is developers portal accessible with customer subscription?

packaging/kickstart/kickstart-ostree.ks.template
DanielFroehlich: fair! that works for me

packaging/kickstart/README.md
DanielFroehlich: Uh, not 100%, worst case customer with sub gets an addition developer sub when using the developer. link.
How about: we use the developer... link for the upstream/community doc, and the access.... link for the enterprise doc? Because that would match quite nice the intention of the two different sites.

ggiguash: Yeap, that's a good point. The current doc is more upstream-oriented.

ggiguash:
The failures are not related to the current PR changes.
/override ci/prow/e2e-aws-tests ci/prow/e2e-aws-tests-arm

PR #4167, opened 2024-11-11
OCPBUGS-43471: Service restart when upgrading MicroShift RPM

ggiguash:
Added a test verifying the `crio` and `microshift` services actually get restarted on upgrade.

### DIFF
--- a/test/suites/rpm/install-and-upgrade-successful.robot
+++ b/test/suites/rpm/install-and-upgrade-successful.robot
@@ -68,6 +68,31 @@ Upgrade From Previous Version
     Wait Until Greenboot Health Check Exited
     [Teardown]    Clean Up Test
 
+Verify Service Restart On Upgrade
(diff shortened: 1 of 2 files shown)


test/suites/rpm/install-and-upgrade-successful.robot
```diff
@@ -68,6 +68,31 @@ Upgrade From Previous Version
     Wait Until Greenboot Health Check Exited
     [Teardown]    Clean Up Test
 
+Verify Service Restart On Upgrade
```
pmtk, line 71: This test takes almost 7 minutes and it's very similar to the test above.
What do you think about merging them?

ggiguash: Good idea!

pacevedom (approved):
/lgtm

PR #4168, opened 2024-11-11
NO-ISSUE: Sprint metrics tooling

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/scripts/jira/scrum.py
+++ b/scripts/jira/scrum.py
@@ -1,0 +1,405 @@
+#!/usr/bin/python3
+
+import csv
+from jira import JIRA
+import json
+import logging
+
+#TODO investigate whats wrong with maxResults from sprints in the api.
+SKIP_FIRST_N_SPRINTS = 50
+
+# Sprint name
+CSV_FIELD_SPRINT_NAME = 'Sprint'
+# Planned story points for a sprint
+CSV_FIELD_PLANNED_TOTAL = 'Planned story points'
+# Resolved story points at the end of a sprint
+CSV_FIELD_RESOLVED_TOTAL = 'Velocity'
+# Not started story points at the end of a sprint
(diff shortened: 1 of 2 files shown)


scripts/jira/scrum.py
```diff
@@ -0,0 +1,420 @@
...
+    CSV_FIELD_PREDICTABILITY_RATIO,
+    CSV_FIELD_FLOW_EFFICIENCY,
+    CSV_FIELD_WIP_RATIO,
+    CSV_FIELD_SCOPE_VOLATILITY,
+    CSV_FIELD_DEFECT_DENSITY_RATIO,
+    CSV_FIELD_DEFECT_REMOVAL_EFFICIENCY,
+    CSV_FIELD_FLOW_EFFICIENCY_BUGS,
+]
+
+AI_PROMPTS = [
```
ggiguash, line 96: Do we want to add relevant text to the readme file?

scripts/jira/scrum.py
ggiguash, line 1: Let's add a .sh file that configures the requirements?

scripts/jira/sprint_metrics_config.json
ggiguash, line 1: Let's change the file name to reflect to which script and project it's related? 
Would something like `scrum_ushift_config.json` be OK?

scripts/jira/scrum.py
pacevedom: Done

scripts/jira/scrum.py
pacevedom: You mean the config file? Good idea.

scripts/jira/sprint_metrics_config.json
pacevedom: Ended up changing the scrum.py name instead, as it was too "generic".

scripts/jira/scrum.py
ggiguash: There are other .sh scripts in the same directory.

PR #4173, opened 2024-11-13
USHIFT-4742: Bootc and ostree blueprint structure reorganization

ggiguash:
Attempt to optimize and document ostree and bootc blueprint hierarchy. 
> CentOS image fetch will be removed as a follow-up in [USHIFT-4868](https://issues.redhat.com/browse/USHIFT-4868).
> Bootc build times may be further optimized as a follow-up in [USHIFT-4881](https://issues.redhat.com/browse/USHIFT-4881) and [USHIFT-4720](https://issues.redhat.com/browse/USHIFT-4720)

**OSTree Summary**
I could not make the overall build times faster. Layer 1 is the heaviest (~30m), but it's cached. Pre-submits and periodics take up to 10 and 15 minutes respectively.
I hope the "optimized" hierarchy is clearer and it is extensible if we add another OS version for instance.

**Bootc Summary**
Created a separate `image-blueprints-bootc` directory to be consistent with other bootc-specific sources.
The new hierarchy only results in a couple of minutes of time saving for builds, which are not significant. The main benefit atm is like for ostree - clear and extensible hierarchy.


pmtk (commented):
I noticed that following is duplicated:
- test/image-blueprints/layer2-presubmit/group1/rhel94-source-with-optionals.toml
- test/image-blueprints/layer3-periodic/group1/rhel94-source-with-optionals.toml

Why do we need
- test/image-blueprints-bootc/layer1-base/group1/rhel94-bootc-base.containerfile
- test/image-blueprints-bootc/layer1-base/group2/rhel94-bootc.image-bootc

PR #4181, opened 2024-11-14
Set fixed RHEL version for DNF

pmtk:


eslutsky:
can you add a comment with a reference to here  https://access.redhat.com/solutions/4251981

PR #4189, opened 2024-11-15
USHIFT-4889: Cache builds: use rhocp-y only for crel image

pmtk:


ggiguash:
@pmtk , we seem to have a problem in periodics because `fips-mode-setup` utility is not pulled as an implicit dependency when `rhocp` is not used. We might need to add an explicit reference to `crypto-policies-scripts` in blueprints.

```
$ sudo dnf whatprovides fips-mode-setup
Updating Subscription Management repositories.
Last metadata expiration check: 2:29:43 ago on Mon 18 Nov 2024 06:22:24 AM IST.
crypto-policies-scripts-20220223-1.git5203b41.el9_0.1.noarch : Tool to switch between crypto policies
Repo        : rhel-9-for-x86_64-baseos-rpms
Matched from:
Provide    : fips-mode-setup = 20220223-1.git5203b41.el9_0.1
```

PR #4196, opened 2024-11-18
USHIFT-4519: Recover and re-panic certain subcomponents

pmtk:
Go's recover() only works within a go routine.
If the service spawns a Go routine to run the component, the recover() in service manager won't catch potential panics.


pacevedom (approved):
/lgtm

PR #4201, opened 2024-11-18
USHIFT-4895: Add ipv6 bootc scenario

pacevedom:

Closes #<Issue Number>


test/scenarios-bootc/presubmits/el94-src@ipv6.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Sourced from scenario.sh and uses functions defined there.
+
+# Redefine network-related settings to use the dedicated IPv6 network bridge
+# shellcheck disable=SC2034  # used elsewhere
+VM_BRIDGE_IP="$(get_vm_bridge_ip "${VM_IPV6_NETWORK}")"
+# shellcheck disable=SC2034  # used elsewhere
+WEB_SERVER_URL="http://[${VM_BRIDGE_IP}]:${WEB_SERVER_PORT}"
+# shellcheck disable=SC2034  # used elsewhere
+MIRROR_REGISTRY_URL="${VM_BRIDGE_IP}:${MIRROR_REGISTRY_PORT}"
```
ggiguash, line 11: We probably also need to fix the registry URL by adding `[]`, or we're getting the following error.
```
18:26:16,940 INF program: Running... ostree container image deploy --sysroot=/mnt/sysimage --image=2001:db8:ca7:ff::1:5000/rhel94-bootc-source --transport=registry
18:26:17,203 INF program: error: Performing deployment: Creating importer: Failed to invoke skopeo proxy method OpenImage: remote error: invalid reference format
```

pacevedom: Not enough, skopeo [does not work with ipv6](https://github.com/containers/image/issues/1004) addresses so I will try to see if /etc/hosts resolution works. Also asked the bootc team about their recommendation so this is just a test to verify my understanding.

test/kickstart-templates/includes/post-containers.cfg
```diff
@@ -51,5 +51,5 @@ cat > /etc/containers/policy.json <<EOF
 EOF
 
 cat >> /etc/hosts <<EOF
-REPLACE_VM_BRIDGE_IP REPLACE_MIRROR_HOSTNAME
+REPLACE_VM_BRIDGE_IP REPLACE_MIRROR_HOSTNAME bootc-registry
```
ggiguash, line 54: - Can we add a comment explaining what we're trying to do and why?
- Do we want to limit this setting to IPv6 case, or we'd better change [this template](https://github.com/openshift/microshift/blob/main/test/kickstart-templates/includes/main-ostreecontainer.cfg#L15)? 

What you're doing here might obsolete BOOTC_REGISTRY_URL setting completely.

pacevedom: This was just a quick test (was using the same branch for my own local tests), it is changed to something (hopefully) simpler and cleaner now.

test/scenarios-bootc/presubmits/el94-src@ipv6.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Sourced from scenario.sh and uses functions defined there.
+
+# Redefine network-related settings to use the dedicated IPv6 network bridge
+# shellcheck disable=SC2034  # used elsewhere
+VM_BRIDGE_IP="$(get_vm_bridge_ip "${VM_IPV6_NETWORK}")"
+# shellcheck disable=SC2034  # used elsewhere
+WEB_SERVER_URL="http://[${VM_BRIDGE_IP}]:${WEB_SERVER_PORT}"
+# shellcheck disable=SC2034  # used elsewhere
+MIRROR_REGISTRY_URL="$(hostname):${MIRROR_REGISTRY_PORT}"
```
ggiguash, line 11: Let's add a comment here on why we must use the host name here, and not VM_BRIDGE_IP?

test/kickstart-templates/includes/main-host-resolution.cfg
ggiguash, line 1: Strictly speaking, this is `pre-host-resolution.cfg`. We have a few other places when we use `%pre`, but it's always together with the main pieces. 
Do you think it's worth renaming?

pacevedom: Yeah, let me do it

test/scenarios-bootc/presubmits/el94-src@ipv6.sh
pacevedom: Done

PR #4211, opened 2024-11-21
USHIFT-4924: Measure bootc build time and include build log in junit for failed builds

pmtk:


test/bin/pyutils/common.py
```diff
@@ -43,15 +43,15 @@ def close_junit():
...
+def record_junit(object, step, status, duration=0.0):
     """Add a message for the specified object and step with OK, SKIP or FAIL status.
     Recording messages is synchronized and it can be called from different threads.
     """
     try:
         # BEGIN CRITICAL SECTION
         JUNIT_LOCK.acquire()
 
-        append_file(JUNIT_LOGFILE, f'<testcase classname="{object}" name="{step}">')
+        append_file(JUNIT_LOGFILE, f'<testcase classname="{object}" name="{step}" time="{duration}">')
```
ggiguash, line 54: Let's make it conditional not to append duration if it's 0.0?

test/bin/pyutils/common.py
```diff
@@ -43,22 +43,31 @@ def close_junit():
...
             pass
         elif status.startswith("SKIP"):
             append_file(JUNIT_LOGFILE, f'<skipped message="{status}" type="{step}-skipped" />')
         elif status.startswith("FAIL"):
-            append_file(JUNIT_LOGFILE, f'<failure message="{status}" type="${step}-failure" />')
+            desc = ''
+            if log_filepath:
+                with open(log_filepath) as f:
+                    desc = f.read()
+            append_file(JUNIT_LOGFILE, f'<failure message="{status}" type="${step}-failure">\n{desc}\n</failure>')
```
ggiguash, line 70: Log files can be rather large. Do you see a problem with that?

pmtk: I changed to get only last 20 lines of the log. PTAL at the result with the debug commit: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_microshift/4211/pull-ci-openshift-microshift-main-e2e-aws-tests-bootc/1859888074561949696

test/bin/pyutils/build_bootc_images.py
```diff
@@ -239,18 +240,20 @@ def process_containerfile(groupdir, containerfile, dry_run):
...
                 cf_outname,
                 f"{MIRROR_REGISTRY}/{cf_outname}"
             ]
+            start = time.time()
             common.run_command_in_shell(push_args, dry_run, logfile, logfile)
-            common.record_junit(cf_path, "push-container", "OK")
+            common.record_junit(cf_path, "push-container", "OK", start)
     except Exception:
-        common.record_junit(cf_path, "process-container", "FAILED")
+        common.record_junit(cf_path, "process-container", "FAILED", log_filepath=cf_logfile)
```
ggiguash, line 256: Let's pass the start here as well? It might sometimes be interesting how long an operation takes until it fails.

test/bin/pyutils/common.py
```diff
@@ -163,7 +174,7 @@ def read_file(file_path: str):
 def append_file(file_path: str, content: str):
     """Append the specified content to a file"""
     with open(file_path, 'a') as file:
-        file.write(content)
+        file.write(content + '\n')
```
ggiguash, line 177: Nit: This is a generic function and we might want to use it for other purpose. Let's not add `\n` unconditionally, but when it's called in record_junit?

PR #4212, opened 2024-11-21
NO-ISSUE: Install nfv-openvswitch package to get openvswitch-selinux-extra-policy package

praveenkumar:
openvswitch-selinux-extra-policy package is part of nfv-openvswitch repo for centos stream and without it building image for microshift-okd fails with following error:

```
Error:
 Problem: package microshift-4.18.0_0.nightly_2024_11_15_113437_20241120110835_10cf3b14d_dirty-1.el9.x86_64 from microshift-local requires microshift-networking = 4.18.0_0.nightly_2024_11_15_113437_20241120110835_10cf3b14d_dirty, but none of the providers can be installed
  - package microshift-networking-4.18.0_0.nightly_2024_11_15_113437_20241120110835_10cf3b14d_dirty-1.el9.x86_64 from microshift-local requires (openvswitch3.4 or openvswitch >= 3.4), but none of the providers can be installed
  - conflicting requests
  - nothing provides openvswitch-selinux-extra-policy needed by openvswitch3.4-3.4.0-1.el9fdp.x86_64 from openshift-mirror-beta
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
Error: building at STEP "RUN ${REPO_CONFIG_SCRIPT} ${USHIFT_RPM_REPO_PATH} &&     dnf install -y microshift &&     if [ "$WITH_FLANNEL" -eq 1 ]; then       dnf install -y microshift-flannel;       systemctl disable openvswitch;     fi &&     ${REPO_CONFIG_SCRIPT} -delete &&     rm -f ${REPO_CONFIG_SCRIPT} &&     rm -rf $USHIFT_RPM_REPO_PATH &&     dnf clean all": while running runtime: exit status 1
```

### DIFF
--- a/okd/src/microshift-okd-multi-build.Containerfile
+++ b/okd/src/microshift-okd-multi-build.Containerfile
@@ -39,6 +39,9 @@ COPY --chmod=755 ./okd/src/create_repos.sh ${REPO_CONFIG_SCRIPT}
 COPY --chmod=755 ./okd/src/configure.sh ${OKD_CONFIG_SCRIPT}
 COPY --from=builder /src/_output/rpmbuild/RPMS ${USHIFT_RPM_REPO_PATH}
 
+# Install nfv-openvswitch repo which provides openvswitch extra policy package
+RUN dnf install -y centos-release-nfv-openvswitch
+
 # Installing MicroShift and cleanup
 # In case of flannel we don't need openvswitch service which is by default enabled as part
 # once microshift is installed so better to disable it because it cause issue when required


okd/src/microshift-okd-multi-build.Containerfile
```diff
@@ -39,6 +39,9 @@ COPY --chmod=755 ./okd/src/create_repos.sh ${REPO_CONFIG_SCRIPT}
 COPY --chmod=755 ./okd/src/configure.sh ${OKD_CONFIG_SCRIPT}
 COPY --from=builder /src/_output/rpmbuild/RPMS ${USHIFT_RPM_REPO_PATH}
 
+# Install nfv-openvswitch repo which provides openvswitch extra policy package
+RUN dnf install -y centos-release-nfv-openvswitch
```
ggiguash, line 43: Can we add this to the RUN command below, not to create yet another layer?
Also, please explain in comment why this is necessary. 

P.S. In general, it's strange that the package deoendency is not pulled automatically. Is the dependency missing?

praveenkumar: > Can we add this to the RUN command below, not to create yet another layer? Also, please explain in comment why this is necessary.

I tried to add this in RUN command below, but that doesn't work because this package add the nfv-openshfit repo and once repo is added then only it able to pull packages from it.

> 
> P.S. In general, it's strange that the package deoendency is not pulled automatically. Is the dependency missing?

openvswitch package which comes from openshift-mirror-beta repo (https://mirror.openshift.com/pub/openshift-v4/dependencies/rpms/4.17-el9-beta/ ) which doesn't have it dependency `openvswitch-selinux-extra-policy` so either we need to update this mirror repo (not sure how) to add this dep package or use nfv repo which have required dependency.

PR #4214, opened 2024-11-22
USHIFT-4737: Add tag and digest capabilities to mirror script

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/scripts/mirror-images.sh
+++ b/scripts/mirror-images.sh
@@ -18,20 +18,6 @@ function usage() {
     exit 1
 }
 
-function skopeo_retry() {
-    for attempt in $(seq 3) ; do
-        if ! skopeo "$@" ; then
-            echo "WARNING: Failed to run skopeo, retry #${attempt}"
-        else
-            return 0
-        fi
-        sleep $(( "${attempt}" * 10 ))
-    done
-
-    echo "ERROR: Failed to run skopeo, quitting after 3 tries"
-    return 1
-}
-
 function mirror_registry() {
     local -r img_pull_file=$1
     local -r img_file_list=$2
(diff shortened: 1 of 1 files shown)


scripts/mirror-images.sh
```diff
@@ -119,46 +104,40 @@ function dir_to_registry() {
...
 
         # Remove the manifest.json file name
         local src_img
         src_img=$(dirname "${src_manifest}")
         # Add the target registry prefix and remove SHA
-        local dst_img
-        dst_img="${dest_registry}/${src_img}"
-        dst_img=$(echo "${dst_img}" | awk -F'@' '{print $1}')
+        local -r dst_img="${dest_registry}/${src_img}"
+        local -r dst_img_no_tag="${dest_registry}/$(echo "${src_img}" | awk -F'@|:' '{print $1}')"
```
ggiguash, line 114: Can it be done using bash string manipulation, like:
```
result=${input_string%%[@:]*}
```

ggiguash:
@pacevedom , can we verify that, besides CI, https://github.com/openshift/microshift/blob/main/docs/user/howto_mirror_images.md instructions work after this change?

ggiguash:
@pacevedom , we also need to verify that this production doc is correct and open a bug otherwise:
https://docs.redhat.com/en/documentation/red_hat_build_of_microshift/4.17/html/embedding_in_a_rhel_for_edge_image/microshift-deploy-with-mirror-registry#microshift-configuring-hosts-for-mirror_microshift-deployment-mirror

scripts/mirror-images.sh
pacevedom: Done

pacevedom:
> @pacevedom , can we verify that, besides CI, https://github.com/openshift/microshift/blob/main/docs/user/howto_mirror_images.md instructions work after this change?

Yes, this is also why the dir-to-reg was changed, to preserve the tag capabilities that the one-shot mirror has. I tested with the output from a build phase in CI (container-images-list), both reg-to-dir and dir-to-reg to a local registry.


> @pacevedom , we also need to verify that this production doc is correct and open a bug otherwise: https://docs.redhat.com/en/documentation/red_hat_build_of_microshift/4.17/html/embedding_in_a_rhel_for_edge_image/microshift-deploy-with-mirror-registry#microshift-configuring-hosts-for-mirror_microshift-deployment-mirror

This needs an update now, as there are no counters for temporary tags anymore and its also capable of mirroring tags, not just sha's.

scripts/mirror-images.sh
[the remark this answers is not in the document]
pacevedom: Changed it to extend tag mirroring to the 2-phase approach as well. See https://github.com/openshift/microshift/pull/4214#issuecomment-2493688385

ggiguash:
I opened the [OCPBUGS-44921](https://issues.redhat.com/browse/OCPBUGS-44921) issue for production doc follow-up.

ggiguash:
I opened the OCPBUGS-44921 issue for production doc follow-up.

PR #4225, opened 2024-11-26
USHIFT-4881: Optimize bootc image builds by extending local mirror usage

ggiguash:
The following enhancements were implemented to speed up mirroring and builds in general:
* Add `cache-to` and `cache-from` options to bootc image builds for storing layers in the local registry rather than local storage
* Remove elevation from dnf query and download commands to avoid **potentially slow** subscription updates (for common environment initialization and also in ostree builds)
* The mirror registry script now explicitly adds non-localhost `FROM` container images to the list
* The mirror registry script now checks the local mirror before attempting to download the images

### DIFF
--- a/test/bin/mirror_registry.sh
+++ b/test/bin/mirror_registry.sh
@@ -11,28 +11,21 @@ REGISTRY_HOST=${REGISTRY_HOST:-${MIRROR_REGISTRY_URL}}
 PULL_SECRET=${PULL_SECRET:-${HOME}/.pull-secret.json}
 LOCAL_REGISTRY_NAME="microshift-local-registry"
 
-retry_pull_image() {
-    for attempt in $(seq 3) ; do
-        if ! podman pull "$@" ; then
-            echo "WARNING: Failed to pull image, retry #${attempt}"
-        else
-            return 0
-        fi
-        sleep $(( "${attempt}" * 10 ))
-    done
-
-    echo "ERROR: Failed to pull image, quitting after 3 tries"
-    return 1
-}
-
 prereqs() {
     # Install packages if not yet available locally
     if ! rpm -q podman skopeo jq &>/dev/null ; then
         "${SCRIPTDIR}/../../scripts/dnf_retry.sh" "install" "podman skopeo jq"
     fi
-    podman stop "${LOCAL_REGISTRY_NAME}" || true
-    podman rm "${LOCAL_REGISTRY_NAME}" || true
-    retry_pull_image "${REGISTRY_IMAGE}"
+    # Pull the registry image locally
+    podman rm -f "${LOCAL_REGISTRY_NAME}" || true
+    skopeo copy \
+        --quiet \
+        --retry-times 3 \
+        --preserve-digests \
+        --authfile "${PULL_SECRET}" \
+        "docker://${REGISTRY_IMAGE}" \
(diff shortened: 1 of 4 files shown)


test/bin/mirror_registry.sh
```diff
@@ -79,10 +85,22 @@ EOF
 mirror_images() {
     local -r ifile=$1
     local -r ofile=$(mktemp /tmp/container-list.XXXXXXXX)
+    local -r ffile=$(mktemp /tmp/from-list.XXXXXXXX)
+
+    # Add the distribution registry and non-localhost-FROM images
+    # to the mirrored list explicitly
+    echo "${REGISTRY_IMAGE}" > "${ffile}"
```
pacevedom, line 92: This image is already local to the hypervisor, right? https://github.com/openshift/microshift/pull/4225/files#diff-25478684142fbd29b423928d694845c40d6092a0fda08b781e0a70a770808c30R26
Do we need it in the registry itself? Its the running image for that.

ggiguash: Yes, it makes sense only if we're caching images in S3.
Since we decided not to do the S3 caching, I will remove it.

test/bin/mirror_registry.sh
```diff
@@ -11,28 +11,21 @@ REGISTRY_HOST=${REGISTRY_HOST:-${MIRROR_REGISTRY_URL}}
...
 prereqs() {
     # Install packages if not yet available locally
     if ! rpm -q podman skopeo jq &>/dev/null ; then
         "${SCRIPTDIR}/../../scripts/dnf_retry.sh" "install" "podman skopeo jq"
     fi
-    podman stop "${LOCAL_REGISTRY_NAME}" || true
-    podman rm "${LOCAL_REGISTRY_NAME}" || true
-    retry_pull_image "${REGISTRY_IMAGE}"
+    # Pull the registry image locally
+    podman rm -f "${LOCAL_REGISTRY_NAME}" || true
```
pmtk, line 20: Do we really need to recreate the container every time this script runs? Can we make it a bit smarter: if the container exists, don't recreate?

test/bin/mirror_registry.sh
```diff
@@ -41,8 +34,21 @@ setup_registry() {
...
+[[registry.mirror]]
+    location = "${REGISTRY_HOST}"
+    insecure = true
+[[registry]]
+    prefix = ""
+    location = "registry.redhat.io"
+[[registry.mirror]]
+    location = "${REGISTRY_HOST}"
+    insecure = true
 EOF
```
pmtk, line 52: Can we add some newlines so the sections (`registry` + its respective `mirror`) are grouped together?

```suggestion
[[registry]]
    prefix = ""
    location = "${REGISTRY_HOST}"
    insecure = true

[[registry]]
    prefix = ""
    location = "quay.io"
[[registry.mirror]]
    location = "${REGISTRY_HOST}"
    insecure = true

[[registry]]
    prefix = ""
    location = "registry.redhat.io"
[[registry.mirror]]
    location = "${REGISTRY_HOST}"
    insecure = true
EOF
```

test/bin/mirror_registry.sh
pmtk: In such case we can leave it as it is.

ggiguash: Let me revert my last commit then.

PR #4228, opened 2024-11-26
USHIFT-4921: Remove ostree-to-bootc build dependency chain

ggiguash:
After the [discussion](https://redhat-internal.slack.com/archives/C02CU30L7GF/p1732286634666629?thread_ts=1732262956.003349&cid=C02CU30L7GF) with the bootc team, the guidance was to stop relying on container image build chain because it's hardly supportable.

Instead, the deployed layer should use systemd services to "fix" UID / GUI skew after the deployment.


test/image-blueprints-bootc/README.md
```diff
@@ -13,11 +13,11 @@ Artifacts built in this layer are cached.
 |group1| Short    | Basic prerequisites
 |group2| Long     | Artifacts independent of current sources
 
-The `.container-encapsulate` files depend on the `rpm-ostree` commits built in
-the `image-blueprints/layer1-base` layer. These commits must be downloaded from
-cache or pre-built locally for `bootc` base layer builds to be successful.
+The `y-2` and `y-1` upgrade tests depend on `ostree` commits when running
+scenarios, not during container image builds. These commits must be downloaded
+from cache locally for all `bootc` test scenarios to execute successfully.
```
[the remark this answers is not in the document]
ggiguash: It's not in the diffs, but upgrade tests use ostree commits, like [this one](https://github.com/openshift/microshift/pull/4228/files#diff-95b3396c0a59f5a4eb77aff2336b3a8d42f1fc5372230fcb02fcceeaa1afe464R8) for instance.

test/image-blueprints-bootc/layer2-source/group1/microshift-sshd-ownership.conf.template
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStartPre=/usr/bin/find /etc/ssh/ -name "ssh_*_key" -exec /bin/chgrp ssh_keys {} +
```
cgwalters, line 2: I think it's a bit more subtle than this; we only want to change the group if it's not already owned by `root:root` because that's the new expected default in current Fedora and C10S.

test/image-blueprints-bootc/layer2-source/group1/rhel94-bootc-source.containerfile
```diff
@@ -35,3 +35,11 @@ RUN firewall-offline-cmd --zone=public --add-port=22/tcp && \
     firewall-offline-cmd --zone=public --add-port=6443/tcp && \
     firewall-offline-cmd --zone=public --add-port=30000-32767/tcp && \
     firewall-offline-cmd --zone=public --add-port=30000-32767/udp
+
+# Install systemd configuration drop-ins to fix potential permission problems
+# when upgrading from older ostree commits to bootc container layers
+RUN mkdir -p /etc/systemd/system/sshd.service.d && \
```
cgwalters, line 41: Two things:

- It's preferred to use `usr/` over `etc/` for these things
- I think using the `COPY usr/ /usr/` pattern is cleaner (but it applies equally well to `COPY etc/ /etc/` as in other examples where config files need to be in `/etc`)

test/image-blueprints-bootc/layer2-source/group1/microshift-ovsdb-ownership.conf.template
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStartPre=/bin/chown -R openvswitch. /etc/openvswitch
```
cgwalters, line 2: Any `chown -R` without `-h` is almost always a bug

cgwalters: What's up with the `.`? Do you mean `:`?

test/image-blueprints-bootc/layer2-source/group1/microshift-sshd-ownership.conf.template
cgwalters: Possibly `/usr/bin/find /etc/ssh/ -name "ssh_*_key" -a '!' -gid 0 -exec /bin/chgrp ssh_keys {} +` or so

cgwalters: While it'd have a large blast radius, maybe we should add this ssh one to the base image actually.

test/image-blueprints-bootc/README.md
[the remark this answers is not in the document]
pmtk: Oh, so lack of parentage isn't a problem for bootc?

test/image-blueprints-bootc/layer2-source/group1/microshift-ovsdb-ownership.conf.template
ggiguash: The `username<dot>` syntax means the ownership is changed for the default group of the user.

test/image-blueprints-bootc/layer2-source/group1/rhel94-bootc-source.containerfile
ggiguash: We have templating mechanism in the build system, which does not support directory structure.
The build is run in an _output directory, so the .template file explicit copy is more straightforward because of that limitation.

test/image-blueprints-bootc/layer2-source/group1/microshift-sshd-ownership.conf.template
ggiguash: I'd prefer not to do this in the base image because only specific source images are used in ostree-to-bootc upgrade procedures. The rest of the build still maintains the proper build hierarchy.

test/image-blueprints-bootc/README.md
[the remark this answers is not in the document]
ggiguash: It is a problem, which we're trying to work around until it's fixed in RHEL in a generic manner. 
For now, we decided that maintaining ostree-to-bootc hierarchy is too much of an overhead when upgrading cross OS, i.e. rhel94-ostree to rhel94-bootc to rhel95-bootc for example.

test/image-blueprints-bootc/layer2-source/group1/microshift-sshd-ownership.conf.template
```diff
@@ -0,0 +1,4 @@
+# Change group of ssh key files if they are not owned by root already because
+# that is the new expected default upstream
+[Service]
+ExecStartPre=/usr/bin/find /etc/ssh/ -name "ssh_*_key" -a '!' -gid 0 -exec /bin/chgrp ssh_keys {} +
```
pmtk, line 4: Test agent already does it, right?

ggiguash: Ah, that's a good point!
I will remove the SSH settings and fix the test agent not to do update according to the new guidelines.

ggiguash: Done

test/image-blueprints-bootc/layer2-source/group1/microshift-ovsdb-ownership.conf.template
ggiguash: Added -h option.

test/image-blueprints-bootc/layer2-source/group1/microshift-sshd-ownership.conf.template
ggiguash: Addressed part of the comments.

PR #4229, opened 2024-11-26
USHIFT-4916: `microshift healthcheck` - wait for `microshift.service`

pmtk:


pkg/healthcheck/greenboot.go
```diff
@@ -0,0 +1,130 @@
+package healthcheck
```
pacevedom, line 1: What do you think about dropping this file and replace it with flags/options to the command instead? You would decouple the healthcheck from greenboot/grub conf reading and make things easier to reason about. Also you would be able to run the command standalone.

pmtk: That's an excellent idea! No need to forcefully integrate with greenboot, we can keep existing timeout calculation we have in `functions.sh`

pacevedom (approved):
/lgtm

PR #4244, opened 2024-11-28
USHIFT-4932: TLS config options for api server

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -32,6 +32,8 @@ type ApiServer struct {
 	// advertising because of dual stack limitations there. This is only to
 	// make ovnk work properly.
 	AdvertiseAddresses []string `json:"-"`
+
+	TLS TLSConfig `json:"tls"`
 }
 
 // NamedCertificateEntry provides certificate details
@@ -54,4 +56,16 @@ type AuditLog struct {
 	// profile is the OpenShift profile specifying a specific logging policy
 	// +kubebuilder:default=Default
 	Profile string `json:"profile"`
+}
+
+type TLSConfig struct {
+	// CipherSuites lists the allowed cipher suites that the API server will
+	// accept and serve.
+	CipherSuites []string `json:"cipherSuites"`
+
+	// MinVersion specifies which TLS version is the minimum version of TLS
+	// to serve from the API server.
+	// +kubebuilder:validation:Enum:=v1.2;v1.3
+	// +kubebuilder:default=v1.2
+	MinVersion string `json:"minVersion"`
 }
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -25,6 +25,8 @@ type ApiServer struct {
 
 	AuditLog AuditLog `json:"auditLog"`
 
+	TLS TLSConfig `json:"tls"`
+
 	// The URL and Port of the API server cannot be changed by the user.
 	URL  string `json:"-"`
 	Port int    `json:"-"`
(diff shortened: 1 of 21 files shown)


pmtk (commented):
What about `RouteControllerManager` and `ClusterPolicyController` (there might be other)?
`openshiftcontrolplanev1.OpenShiftControllerManagerConfig` -> `configv1.HTTPServingInfo` -> `configv1.ServingInfo`:
```
type ServingInfo struct {
	...

	// MinTLSVersion is the minimum TLS version supported.
	// Values must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants
	MinTLSVersion string `json:"minTLSVersion,omitempty"`
	// CipherSuites contains an overridden list of ciphers for the server to support.
	// Values must match cipher suite IDs from https://golang.org/pkg/crypto/tls/#pkg-constants
	CipherSuites []string `json:"cipherSuites,omitempty"`
}
```

pkg/config/apiserver.go
```diff
@@ -55,3 +65,58 @@ type AuditLog struct {
...
 }
+
+type TLSConfig struct {
+	// CipherSuites lists the allowed cipher suites that the API server will
+	// accept and serve.
+	CipherSuites []string `json:"cipherSuites"`
+
+	// MinVersion specifies which TLS version is the minimum version of TLS
+	// to serve from the API server.
+	// +kubebuilder:validation:Enum:=VersionTLS12;VersionTLS13
```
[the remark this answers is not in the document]
pacevedom: Done

pacevedom:
> What about `RouteControllerManager` and `ClusterPolicyController` (there might be other)? `openshiftcontrolplanev1.OpenShiftControllerManagerConfig` -> `configv1.HTTPServingInfo` -> `configv1.ServingInfo`:
> 
> ```
> type ServingInfo struct {
> 	...
> 
> 	// MinTLSVersion is the minimum TLS version supported.
> 	// Values must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants
> 	MinTLSVersion string `json:"minTLSVersion,omitempty"`
> 	// CipherSuites contains an overridden list of ciphers for the server to support.
> 	// Values must match cipher suite IDs from https://golang.org/pkg/crypto/tls/#pkg-constants
> 	CipherSuites []string `json:"cipherSuites,omitempty"`
> }
> ```

Added `RouteControllerManager`, as `ClusterPolicyController` does not expose and endpoint where it listens. Good catch!

pkg/config/apiserver.go
```diff
@@ -55,3 +65,60 @@ type AuditLog struct {
...
+		// Golang will ignore whatever is configured.
+		t.CipherSuites = getIANACipherSuites(configv1.TLSProfiles[configv1.TLSProfileModernType].Ciphers)
+	}
+}
+
+func (t *TLSConfig) Validate() error {
+	if len(t.CipherSuites) == 0 {
+		return fmt.Errorf("unsupported empty cipher suites")
+	}
+	var cipherSuites []string
```
pmtk, line 105: This might make it more obvious
```suggestion
	var allowedCipherSuites []string
```

pacevedom: Was keeping the syntax from other control plane components, such as kubelet or apiserver. All of them use `--tls-cipher-suites` and `-tls-min-version`. The ingress operator [also uses](https://github.com/openshift/cluster-ingress-operator/blob/8be174945db79b0660bd0fd21ed279ab7b6aafbf/manifests/02-deployment.yaml#L98-L100) this syntax. Should we change it to make it different?

PR #4259, opened 2024-12-02
NO-ISSUE: Rebase main 4.19.0 with updates

pmtk:


test/scenarios-bootc/periodics/el94-crel@el94-src@optional-upgrade.sh
```diff
@@ -2,16 +2,30 @@
...
 # Sourced from scenario.sh and uses functions defined there.
 
+start_image=rhel94-bootc-crel-optionals
+
 scenario_create_vms() {
-    prepare_kickstart host1 kickstart-bootc.ks.template rhel94-bootc-crel-optionals
+    if ! does_image_exist "${start_image}"; then
+        echo "Image '${start_image}' not found - skipping test"
+        return 0
+    fi
```
copejon, line 11: If the image does not exist, could skipping this test mask a buildtime error?

pmtk: I think only if we don't catch the build time error. Shouldn't happen in CI, might be unnoticed in local dev env.
To do this "properly" we'd need a much more complicated mechanism to first decide if we should expect `crel` and then propagate it down to build and test phases.

PR #4293, opened 2024-12-09
OCPBUGS-38468: Revert "use default route mtu"

pacevedom:
This reverts commit 6d13426260cd293baf6600307779b1dcb011178d.

**Which issue(s) this PR addresses**:

Closes #<Issue Number>

PR #4295, opened 2024-12-09
USHIFT-5101: Move offline tests from periodics to presubmits

pacevedom:

Closes #<Issue Number>


pacevedom:
/hold
Until https://github.com/openshift/microshift/pull/4293 is merged

test/scenarios/presubmits/el94-src@offline.sh
ggiguash, line 1: This might not work for ostree builds because we only create the necessary offline image in periodics job. 
In bootc, it should be OK, so should we just move the bootc version?

ggiguash: See https://github.com/openshift/microshift/tree/main/test/image-blueprints

ggiguash: If we want to do this for ostree, it'll require changes to the layer structure, similar to what we have for bootc: base and source layers.

pacevedom: You are right, moving only the bootc version should be enough. Thanks!

PR #4302, opened 2024-12-10
USHIFT-4918 USHIFT-4920: microshift healthcheck - wait for workloads

pmtk:


pkg/healthcheck/workloads.go
```diff
@@ -0,0 +1,245 @@
...
+			DaemonSets: []string{
+				"dns-default",
+				"node-resolver",
+			},
+		},
+	}
+	fillOptionalWorkloadsIfApplicable(cfg, workloads)
+
+	return workloads, nil
+}
```
ggiguash, line 67: I'm a bit concerned that this information is "hidden" deep in the code. 
Should we cancel the default list and provide it all explicitly from greenboot?

packaging/greenboot/microshift-running-check.sh
```diff
@@ -93,102 +16,7 @@ echo "STARTED"
...
-
-        name=${pid2name["${pid}"]}
-        echo "Error: Pods are restarting too frequently in the '${name}' namespace"
-    fi
-done
-
-# Exit with an error code if the pod restart check failed
-if ${check_failed} ; then
-    exit 1
-fi
```
ggiguash, line 194: Should we not have a call to check microshift core namespaces here?
If it still does not work, let's add a TODO comment?

pkg/healthcheck/workloads.go
pmtk: The idea was that `microshift healthcheck` is something that works standalone, without greenboot.
It can be executed via greenboot scripts, but can be without.
This would break that assumption and make `microshift healthcheck` useless for running the command manually.

packaging/greenboot/microshift-running-check.sh
pmtk: `microshift healthcheck` without options checks the core namespaces so it works by default

pkg/cmd/healthcheck.go
```diff
@@ -2,25 +2,70 @@ package cmd
...
+1. Use --namespace option together with --deployments, --daemonsets, and --statefulsets
+   For example:
+      $ microshift healthcheck --namespace openshift-storage --deployments lvms-operator --daemonsets vg-manager
+      $ microshift healthcheck --namespace openshift-ovn-kubernetes --daemonsets ovnkube-master,ovnkube-node
+
+2. Provide --custom option with a JSON argument. It can wait for several namespaces in single execution.
+   The schema is: '{ "NAMESPACE": {"deployments: ["NAMES"...], "daemonsets: ["NAMES"...], "statefulsets: ["NAMES"...]}}'
+   For example:
+      $ microshift healthcheck --custom '{"openshift-storage":{"deployments": ["lvms-operator"], "daemonsets": ["vg-manager"]}, "openshift-ovn-kubernetes":{"daemonsets": ["ovnkube-master", "ovnkube-node"]}}'
+`,
```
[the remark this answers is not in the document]
pmtk: Json option provides a way to run a healthcheck for multiple namespaces in parallel but the other option is more user friendly if you just need to check a single namespace

pkg/healthcheck/workloads.go
ggiguash: Currently, the information on the checked namespaces is hidden. We could still leave the option to run healthcheck on core namespaces by default, but keep this information external to be loaded and used by the command.

pkg/cmd/healthcheck.go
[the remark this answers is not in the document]
ggiguash: Should we not use this option for our checks to speed up the overall process?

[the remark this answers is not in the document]
pacevedom: Under what circumstances would someone issue a custom healthcheck command?
MicroShift is already aware of what is core and what is optional (it applies the kustomize manifests so it knows everything that should be up and running when starting), so who is better suited to do these checks than MicroShift itself?

Why would a user try to perform a custom check? It must be for something that is not core and/or optional packages for MicroShift, which falls in the application realm. Should the user not have their own healthcheck, instead of intertwining with the one built for core MicroShift?

This command could even become part of the MicroShift readiness towards systemd.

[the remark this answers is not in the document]
pmtk: So this is just a convenience thing - currently we have functions.sh which customers can use. This --custom and --namespace is just something to give a migration path.
I don't know if anyone is using function.sh, but if they would need to wait for their DS,deploy,STS, then we can easily provide helpers.

pacevedom (approved):
/lgtm

PR #4315, opened 2024-12-12
OCPBUGS-38468: use default route mtu

eslutsky:
determine if the host has a default gateway  based on IPv4 or IPv6 routes, then use its MTU,
the original solution was based on net.GetHostIP which didn't cover cases when we  dont have any network connection (hardcoded nodeIP with loopback) .

### DIFF
--- a/pkg/components/networking.go
+++ b/pkg/components/networking.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/microshift/pkg/assets"
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/config/ovn"
+	"github.com/vishvananda/netlink"
 	"k8s.io/klog/v2"
 )
 
@@ -52,7 +53,16 @@ func startCNIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
 		}
 	}
 
-	ovnConfig, err := ovn.NewOVNKubernetesConfigFromFileOrDefault(filepath.Dir(config.ConfigFile), cfg.MultiNode.Enabled)
+	ipFamily := netlink.FAMILY_ALL
+	if cfg.IsIPv4() && !cfg.IsIPv6() {
+		ipFamily = netlink.FAMILY_V4
+	}
+
+	if cfg.IsIPv6() && !cfg.IsIPv4() {
+		ipFamily = netlink.FAMILY_V6
+	}
+
+	ovnConfig, err := ovn.NewOVNKubernetesConfigFromFileOrDefault(filepath.Dir(config.ConfigFile), cfg.MultiNode.Enabled, ipFamily)
 	if err != nil {
 		return fmt.Errorf("failed to create OVN-K configuration from %q: %w", config.ConfigFile, err)
 	}
--- a/pkg/config/ovn/ovn.go
+++ b/pkg/config/ovn/ovn.go
@@ -8,6 +8,8 @@ import (
 	"path/filepath"
 	"regexp"
 
+	"github.com/openshift/microshift/pkg/util"
+	"github.com/vishvananda/netlink"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 )
@@ -40,10 +42,6 @@ func (o *OVNKubernetesConfig) Validate() error {
 	if err != nil {
 		return fmt.Errorf("failed to validate OVS bridge: %w", err)
 	}
-	err = o.validateConfig()
-	if err != nil {
-		return fmt.Errorf("failed to validate OVN-K configuration: %w", err)
-	}
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


eslutsky:
logs related to those failures ,
from the journal:
```Jan 13 17:33:43 i-0f899d7309062d8a9 microshift[57347]: infrastructure-services-manager I0113 17:33:43.951911   57347 ovn.go:119] OVNKubernetes config file not found, assuming default values
Jan 13 17:33:43 i-0f899d7309062d8a9 microshift[57347]: infrastructure-services-manager W0113 17:33:43.952196   57347 components.go:37] Failed to start CNI plugin: failed to validate ovn-kubernetes configuration: failed to validate OVN-K configuration: interface MTU (1500) is too small for specified overlay (9001)
Jan 13 17:33:43 i-0f899d7309062d8a9 microshift[57347]: infrastructure-services-manager I0113 17:33:43.952225   57347 startuprecorder.go:58] "SERVICE READY" service="infrastructure-services-manager" since-start="21.508052857s"
Jan 13 17:33:43 i-0f899d7309062d8a9 microshift[57347]: infrastructure-services-manager E0113 17:33:43.952219   57347 manager.go:138] "SERVICE FAILED - stopping MicroShift" err="failed to validate ovn-kubernetes configuration: failed to validate OVN-K configuration: interface MTU (1500) is too small for specified overlay (9001)" service="infrastructure-services-manager" since-start="21.508041643s"
```

eslutsky:
@pacevedom , looks like it works now, but  default route MTU disccovery is not working with IPv6. 
i guess we need to adapt it to ipv6.
from [log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_microshift/4315/pull-ci-openshift-microshift-main-e2e-aws-tests/1879272407659712512/artifacts/e2e-aws-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el94-src@ipv6/vms/host1/sos/journal_2025-01-14_21:43:43.log)
```
 Jan 14 16:40:02 el94-src-ipv6-host1 microshift[1372]: infrastructure-services-manager I0114 16:40:02.931900    1372 ovn.go:59] getClusterMTU: finding default route interface
Jan 14 16:40:02 el94-src-ipv6-host1 microshift[1372]: infrastructure-services-manager I0114 16:40:02.932051    1372 ovn.go:66] getClusterMTU: Couldnt extract MTU from the default route interface , Using Default MTU.```

File: pkg/util/net.go [hunk 1]
```diff
@@ -266,3 +274,55 @@ func GetHostIPv6(ipHint string) (string, error) {
 
 	return "", fmt.Errorf("unable to find host IPv6 address")
 }
+
+// Find the Default route Interface based on ipv4 or ipv6 routes.
+func FindDefaultRouteIface() (iface *tcpnet.Interface, err error) {
```
pacevedom, line 279: I think we need to pass the configuration here to determine if MicroShift is running on ipv4, ipv6, or both.
Imagine you have different NICs in the host for each IP family. Default routes would be configured on different devices, therefore they could have different MTUs.
If MicroShift is running on ipv6 check only ipv6 routes.
If MicroShift is running on ipv4 check only ipv4 routes.
If MicroShift is running on both, check both and pick the smallest.
Or is there something I am missing?

eslutsky: I'm not sure its a common scenario for edge devices when you have different Default routes set across different IP families .

eslutsky: we are calling FindDefaultRouteIface from the ovn package which doesn't have its own configuration. the IPv4/6 detection logic is in the config [struct](https://github.com/praveenkumar/microshift/blob/offline_images/etcd/vendor/github.com/openshift/microshift/pkg/config/config.go#L582) .

pkg/components/networking.go
```diff
@@ -52,7 +53,16 @@ func startCNIPlugin(ctx context.Context, cfg *config.Config, kubeconfigPath stri
 		}
 	}
 
-	ovnConfig, err := ovn.NewOVNKubernetesConfigFromFileOrDefault(filepath.Dir(config.ConfigFile), cfg.MultiNode.Enabled)
+	ip_family := netlink.FAMILY_ALL
```
pacevedom, line 56: Nit: Can we use camelCase instead?

pkg/config/ovn/ovn.go
```diff
@@ -56,28 +54,31 @@ func (o *OVNKubernetesConfig) validateOVSBridge() error {
...
+	// if configure both IPV4 and IPV6 check the smallest
+	//nolint:nestif
+	if ip_family == netlink.FAMILY_ALL {
+		mtu, err := util.FindDefaultRouteMinMTU()
+		if err == nil {
+			o.MTU = mtu
+		} else {
+			klog.Infof("getClusterMTU: error %s.", err)
+		}
+		klog.Infof("getClusterMTU: using Dual stack MTU discovery found - %d", mtu)
```
pacevedom, line 72: What is the value of `mtu` if there is an error? Should this be `o.MTU`?

pkg/config/ovn/ovn.go
```diff
@@ -56,28 +54,31 @@ func (o *OVNKubernetesConfig) validateOVSBridge() error {
...
-		o.MTU = link.MTU
 	} else {
-		o.MTU = defaultMTU
+		mtu, err := util.FindDefaultRouteMTU(ip_family)
+		if err == nil {
+			o.MTU = mtu
+		} else {
+			klog.Infof("getClusterMTU: error %s.", err)
+		}
+		klog.Infof("getClusterMTU: using ip version: %d stack MTU discovery found - with MTU %d", ip_family, mtu)
```
pacevedom, line 81: Same as above

pkg/util/net.go
```diff
@@ -266,3 +275,74 @@ func GetHostIPv6(ipHint string) (string, error) {
...
 	return "", fmt.Errorf("unable to find host IPv6 address")
 }
+
+func FindDefaultRouteMinMTU() (mtu int, err error) {
+	ip_families := []int{netlink.FAMILY_V4, netlink.FAMILY_V6}
+
+	mtu_slice := []int{}
+
+	for _, ip_family := range ip_families {
+		new_mtu, err := FindDefaultRouteMTU(ip_family)
```
pacevedom, line 285: Nit: Can we use camel case?

pkg/util/net.go
```diff
@@ -266,3 +275,74 @@ func GetHostIPv6(ipHint string) (string, error) {
...
+	return 0, fmt.Errorf("could not find minimal MTU")
+}
+
+func FindDefaultRouteMTU(ip_family int) (mtu int, err error) {
+	link, err := FindDefaultRouteIface(ip_family)
+	if err != nil && link.MTU == 0 {
+		return 0, err
+	}
+
+	klog.Infof("using IP %d on Interface %s with MTU %d ", ip_family, link.Name, link.MTU)
```
pacevedom, line 303: This log may be shown twice if using dual stack and its already printing in `getClusterMTU` function, isnt it?

pacevedom (approved):
/lgtm

PR #4318, opened 2024-12-13
USHIFT-5202: Optimize CI runtime for pre-submit jobs to run faster and more balanced

ggiguash:
The goal of this change is to complete the VM creation and test phase under 45m. 

> Periodic jobs may also be triggered in the PR, thus it's important to balance them out by having each test run under 30m.
> CNCF is the longest periodic job taking ~2h, but it only runs in nightlies.

The following changes were implemented:
* The `osconfig` tests that take 1.5h were split into 3 parts for ostree, similar to what already exists for bootc.
* Reduced cos9 jobs number to improve the overall VM startup time and stability
  * Do not run osconfig tests on cos9, it's enough to test this on RHEL
  * Remove cos9 non-essential tests (e.g. backup, storage, log-scan, etc.) from presubmits
  * Move "special" configuration tests (e.g. multi-nic, optional, etc.) to periodics

**The metal-tests stage now takes around 45m, which is ~15m for creating VMs and ~30m for running tests.**


test/scenarios/periodics/el94-src@osconfig-cleanup-data.sh
ggiguash, line 1: Note to reviewers: the diff is misleading  by looking like files were moved from bootc to ostree hierarchies. The logic of the change is not to delete bootc cos9 osconfig tests and split ostree rhel osconfig tests into 3 parts.

PR #4326, opened 2024-12-16
USHIFT-5204: Detect optional MicroShift workloads based on namespace existence

pmtk:


packaging/rpm/microshift.spec
ggiguash, line 1: Some of the removed scripts have already been made part of the previous releases. Are they removed when we upgrade say from 4.18 to 4.19 packages?

pkg/healthcheck/microshift_core_workloads.go
```diff
@@ -68,11 +84,58 @@ func getExpectedCSIComponents(cfg *config.Config) []string {
...
+		"openshift-gateway-api": {
+			Deployments: []string{"servicemesh-operator3", "istiod-openshift-gateway-api"},
+		},
+		"kube-flannel": {
+			DaemonSets: []string{"kube-flannel-ds"},
+		},
+		"kube-proxy": {
+			DaemonSets: []string{"kube-proxy"},
+		},
+	}
```
ggiguash, line 118: I think it would be a mishap for us to hardcode the optional package names and resources in the MicroShift code. The existing implementation provides a clear separation between the package and the main code by naming the resources in the scripts that are delivered together with the optional component.

pmtk: Maybe, but how would you express "ovnk should not be checked, we use flannel"?
These optional packages reside in microshift's repository anyway, so by hardcoding them we get better integration without performing hacks.

PR #4330, opened 2024-12-18
USHIFT-5219: microshift healthcheck - extend the decision for expecting LVMS

pmtk:


pkg/healthcheck/microshift_core_workloads.go
```diff
@@ -30,15 +32,76 @@ func getCoreMicroShiftWorkloads() (map[string]NamespaceWorkloads, error) {
...
+		klog.Infof("LVMS is disabled via config. Configured value: %q", string(cfg.Storage.Driver))
+		return false, nil
+	}
+
+	cmd := exec.Command("vgs", "--readonly", "--options=name", "--reportformat=json")
+	output, err := cmd.Output()
+	if err != nil {
+		return false, err
+	}
+	klog.V(2).Infof("vgs reported: %s", strings.ReplaceAll(strings.ReplaceAll(string(output), "\n", ""), " ", ""))
```
ggiguash, line 66: Should we use https://pkg.go.dev/encoding/json#Compact instead?

pkg/healthcheck/microshift_core_workloads.go
```diff
@@ -30,15 +32,76 @@ func getCoreMicroShiftWorkloads() (map[string]NamespaceWorkloads, error) {
...
+		return false, err
+	}
+
+	if len(report.Report) == 0 || len(report.Report[0].VGs) == 0 {
+		klog.Infof("Detected 0 volume groups - LVMS is not expected")
+		return false, nil
+	}
+
+	if len(report.Report[0].VGs) == 1 {
+		klog.Infof("Detected 1 volume groups - LVMS is expected")
```
ggiguash, line 87: Could we print the name of the detected group?

pkg/healthcheck/microshift_core_workloads.go
```diff
@@ -30,15 +32,76 @@ func getCoreMicroShiftWorkloads() (map[string]NamespaceWorkloads, error) {
...
+		klog.Infof("%s exists - expecting LVMS to be deployed", cfgFile)
+		return true, nil
+	}
+
+	if !cfg.Storage.IsEnabled() {
+		klog.Infof("LVMS is disabled via config. Configured value: %q", string(cfg.Storage.Driver))
+		return false, nil
+	}
+
+	cmd := exec.Command("vgs", "--readonly", "--options=name", "--reportformat=json")
```
ggiguash, line 61: Could we simplify the code by running the following command? 
`vgs --readonly --noheadings --options=name`

It's just a list of names then, so no need for json parsing.

pkg/healthcheck/microshift_core_workloads.go
pmtk: It's in the `vgs reported` log

pkg/healthcheck/microshift_core_workloads.go
pmtk: But the output of vgs is lines with extra whitespace. So we'd have to split on new lines then remove the whitespace. So one way or another we need to do some processing

pkg/healthcheck/microshift_core_workloads.go
pmtk: Wasn't aware of it, thx

pkg/healthcheck/microshift_core_workloads.go
pacevedom: Is there a way to do this natively in go? Does lvms do something similar to check it out?

pacevedom: Patryk provided the [source](https://github.com/openshift/lvm-operator/blob/89234f2d32e1df7bf7dc6326f09fa68407d235e1/internal/controllers/vgmanager/lvm/lvm.go#L348-L395) from lvms but its in the internal libraries.
This is lgtm to me.

pacevedom (approved):
/lgtm
/hold
Feel free to unhold, this is just in case you want more reviews.

PR #4331, opened 2024-12-18
USHIFT-5221: Fix StartupRecorder race condition

vanhalenar:


pkg/servicemanager/startuprecorder/startuprecorder.go
```diff
@@ -32,11 +32,15 @@ type StartupData struct {
 type StartupRecorder struct {
 	Data StartupData
 
-	m sync.Mutex
+	ServiceCount int
+	AllLogged    chan struct{}
```
pmtk, line 36: Can be made private
```suggestion
	allLogged    chan struct{}
```

pkg/servicemanager/startuprecorder/startuprecorder.go
```diff
@@ -77,18 +87,23 @@ func (l *StartupRecorder) ServicesStart(start time.Time) {
...
 
 func (l *StartupRecorder) OutputData() {
-	jsonOutput, err := json.Marshal(l.Data)
-	if err != nil {
-		klog.Error("Failed to marshal startup data")
-	}
-
-	klog.Infof("Startup data: %s", string(jsonOutput))
+	go func() {
+		<-l.AllLogged
```
pmtk, line 91: FYI, you could use `select` to simultaneously wait on this channel and for timeout, so if this takes too long for some reason, we'll get some data.
I'm not saying we need it here, but for the future (or if you want to play with it now)

From https://gobyexample.com/timeouts:
```
    select {
    case res := <-c1:
        fmt.Println(res)
    case <-time.After(1 * time.Second):
        fmt.Println("timeout 1")
    }
```

vanhalenar: Added here as well, sounds like a good idea

PR #4339, opened 2024-12-19
USHIFT-5202: Add elapsed timestamps for python shell commands

ggiguash:


ggiguash:
The failure does not have anything to do with the change. 
Overriding the job to save on CI cycles.
/test e2e-aws-tests-bootc-periodic

ggiguash:
The failure does not have anything to do with the change.
Overriding the job to save on CI cycles.

PR #4345, opened 2024-12-20
USHIFT-5202: Use offline images for standard suites to speed up service startup

ggiguash:
The standard tests runtime is around 25 and 20 minutes. These are the longest tests for pre-submits. 

* Using preloaded container images for bootc pre-submit tests to speed up their startup
  * Note that this is not possible for ostree tests because we only build offline layer in periodic jobs.


ggiguash:
Does not look like there is much difference in total runtime.
/close

PR #4387, opened 2025-01-07
USHIFT-5286: Update image mode document with bootc image builder (BIB)

praveenkumar:

Closes #<Issue Number>

### DIFF
--- a/docs/contributor/image_mode.md
+++ b/docs/contributor/image_mode.md
@@ -337,6 +337,119 @@ watch sudo oc get pods -A \
     --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig
 ```
 
+## Using Bootc Image Builder (BIB)
+
+The [bootc-image-builder](https://github.com/osbuild/bootc-image-builder), is a containerized tool to
+create disk images from bootc images. You can use the images that you build to deploy disk images in
+different environments, such as the edge, server, and clouds.
+
+### Prepare Build Config File
+A build config is a Toml (or JSON) file with customizations for the resulting image.
+The config file is mapped into the container directory to /config.toml. The customizations are
+specified under a customizations object.
+
+Set variables pointing to secret files that are included in `config.toml` for
+gaining access to private container registries:
+* `PULL_SECRET` file contents are copied to `/etc/crio/openshift-pull-secret`
+  at the post-install stage to authenticate OpenShift registry access
+
+```bash
+PULL_SECRET=~/.pull-secret.json
+IMAGE_NAME=microshift-4.17-bootc
+```
+
+Run the following command to create the `kickstart.ks` file to be used during
+the virtual machine installation. If you want to embed the kickstart file directly
+to iso using BIB please refer to [upstream docs](https://osbuild.org/docs/bootc/#anaconda-iso-installer-options-installer-mapping)
+
+```bash
+cat > kickstart.ks <<EOFKS
+lang en_US.UTF-8
+keyboard us
+timezone UTC
+text
+reboot
+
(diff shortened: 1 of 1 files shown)


ggiguash:
> you mean those script should be part of docs/config instead as part of scripts directory?

Such scripts are intended for devenv setup and we have previous experience with build.sh I would not want to repeat. Users took it as a base for their procedures, but it was very contextual and not correct for all the cases.

For what I'd like to see is an augmentation of the [image_mode.md](https://github.com/openshift/microshift/blob/main/docs/contributor/image_mode.md) page with this w/f, using the same style - instructions rather than a full script.

praveenkumar:
> > you mean those script should be part of docs/config instead as part of scripts directory?
> 
> Such scripts are intended for devenv setup and we have previous experience with build.sh I would not want to repeat. Users took it as a base for their procedures, but it was very contextual and not correct for all the cases.
> 
> For what I'd like to see is an augmentation of the [image_mode.md](https://github.com/openshift/microshift/blob/main/docs/contributor/image_mode.md) page with this w/f, using the same style - instructions rather than a full script.

@ggiguash Understood, will move the script to our project and update the image mode doc with BIB info in this PR.

praveenkumar:
@ggiguash Please have a look, I updated the docs and remove script part.

docs/contributor/image_mode.md
```diff
@@ -337,6 +337,119 @@ watch sudo oc get pods -A \
...
+    --config /config.toml \
+    localhost/${IMAGE_NAME}:latest
+```
+
+### Create Virtual Machine
+
+Copy the `install.iso` file to the `/var/lib/libvirt/images` directory.
+
+```bash
+sudo cp -Z ./output/bootiso/install.iso /var/lib/libvirt/images
```
ggiguash, line 428: Can we rename the install.iso to `${VMNAME}.iso` during the copy?

docs/contributor/image_mode.md
```diff
@@ -337,6 +337,119 @@ watch sudo oc get pods -A \
...
+```bash
+PULL_SECRET=~/.pull-secret.json
+IMAGE_NAME=microshift-4.17-bootc
+```
+
+Run the following commands to create the `config.toml` file to be used during
+the bootc image builder process.
+
+```bash
+cat > config.toml <<EOFCONF
```
ggiguash, line 365: I'm not sure about using the config.tomp file for only embedding kickstart.
What's the main point here to deviate from the kickstart settings in previous sections? 

P.S. I'm not decided on which approach is better, so I'm looking for justification to introduce config.toml notion in this document.

praveenkumar: @ggiguash Use of kickstart file is required because of providing the required disk layout to work for microshift lvm part otherwise we can directly use BiB on the generated bootc microshift image.

ggiguash: I agree that kickstart is required. I'm asking if we need propagate kickstart via config.toml rather than from virt-install command line, like in previous section.

praveenkumar: Sure, we can use virt-install cli to pass the kickstart but we should mention or link to the doc that how it can also be embedded with iso when using BIB so that created ISO can be used on other hypervisor?

ggiguash: Yeap, that would work. Let's add a note pointing to upstread docs [here](https://osbuild.org/docs/bootc/#anaconda-iso-installer-options-installer-mapping)?

ggiguash:
/retitle USHIFT-5286: Update image mode document with bootc image builder (BIB)

ggiguash:
Thank you, @praveenkumar !

/lgtm

PR #4425, opened 2025-01-17
USHIFT-4926: Gantt chart generator script for MicroShift startup data

vanhalenar:


scripts/startup_graph.py
```diff
@@ -0,0 +1,113 @@
...
+)
+
+fig.add_vline(x=microshift_start, line_dash="dash")
+fig.add_vline(x=microshift_ready, line_dash="dash")
+
+fig.add_annotation(
+    x=microshift_start,
+    y=1.01,
+    yref="paper",
+    text="Microshift start",
```
ggiguash, line 95:
```suggestion
    text="MicroShift start",
```

scripts/startup_graph.py
```diff
@@ -0,0 +1,113 @@
...
+    xanchor="center",
+    yanchor="bottom",
+    font=dict(size=12, color="black")
+)
+
+fig.add_annotation(
+    x=microshift_ready,
+    y=1.01,
+    yref="paper",
+    text=f"Microshift ready ({microshift_timeToReady})",
```
ggiguash, line 106:
```suggestion
    text=f"MicroShift ready ({microshift_timeToReady})",
```

scripts/startup_graph.py
```diff
@@ -0,0 +1,113 @@
...
+    elif nanoseconds >= 1e6:
+        return f"{nanoseconds / 1e6:.2f} ms"
+    elif nanoseconds >= 1e3:
+        return f"{nanoseconds / 1e3:.2f} μs"
+    else:
+        return f"{nanoseconds} ns"
+
+
+parser = argparse.ArgumentParser(description="Generate Gantt chart of startup times of MicroShift services.")
+parser.add_argument("file_path", help="The path to the JSON file containing MicroShift startup data")
```
ggiguash, line 20: Let's add a short desciption how such a file can be generated?

scripts/startup_graph.py
```diff
@@ -0,0 +1,113 @@
...
+        return f"{nanoseconds / 1e9:.2f} s"
+    elif nanoseconds >= 1e6:
+        return f"{nanoseconds / 1e6:.2f} ms"
+    elif nanoseconds >= 1e3:
+        return f"{nanoseconds / 1e3:.2f} μs"
+    else:
+        return f"{nanoseconds} ns"
+
+
+parser = argparse.ArgumentParser(description="Generate Gantt chart of startup times of MicroShift services.")
```
ggiguash, line 19: Not to confuse with systemd services.
```suggestion
parser = argparse.ArgumentParser(description="Generate a Gantt chart of MicroShift internal services startup times .")
```

scripts/startup_graph.py
```diff
@@ -0,0 +1,113 @@
+import json
+import pandas as pd
+import plotly.express as px
+import numpy as np
```
pmtk, line 4: How do we get those dependencies? Is it system's package manager or requirements.txt?
I think more python-ish would be to create requirements.txt and run the script inside a virtual env (venv).
Take a look at `gettool_robotframework()` from `scripts/fetch_tools.sh` - it creates venv and gets the deps using the requirements.txt

vanhalenar: so create a bash script that creates venv, gets the deps and runs the python script then?

vanhalenar:
microshift_internal_services
this is what the graph looks like, it is interactive, so you can hover over each bar and see the startup time and dependencies of each service. Here is a sample JSON file:
[startup_times.json](https://github.com/user-attachments/files/18453089/startup_times.json)

scripts/startup_graph.py
vanhalenar: done

scripts/startup_graph.py
vanhalenar: I added a description, I will also be adding a bash script that creates the venv and gets the dependencies, so I will add an action that generates the file from microshift logs for convenience.

scripts/startup_graph.py
vanhalenar: done

scripts/startup_graph.py
vanhalenar: I'll make a bash script for this in the next commit.

scripts/startup_graph.py
vanhalenar: done

PR #4472, opened 2025-01-27
wip debug do not merge | re-enable CRIO's watchdog

pmtk:


pmtk:
@sohankunkerkar I can confirm that reintroducing Watchdog setting for crio.service works as intended. I don't see buggy behavior of CRIO being restarted because CNI is not ready.

```
Jan 27 03:55:02 el94-src-standard-suite1-host1 crio[1255]: time="2025-01-27T03:55:02.482902684-05:00" level=info msg="Starting systemd watchdog using interval: 30s"
```

PR #4475, opened 2025-01-27
USHIFT-5351: Enable sigstore verification for selected scenarios

ggiguash:
* Published bootc images
* ostree image with optional components
* bootc image with optional components

Testing optional components gives us level of confidence that all container images we depend on are signed in ec / rc / z-stream.


ggiguash:
Test flakes. 
/test e2e-aws-tests

PR #4481, opened 2025-01-28
USHIFT-4726: Attempt console login in offline tests

vanhalenar:
I had to add a trailing slash to the `wget` command, because it would get redirected and wouldn't download the `.cfg` files otherwise, I'm not sure how it has worked so far.

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -622,7 +622,7 @@ launch_vm() {
     vm_extra_args+=" inst.ks=file:/$(basename "${kickstart_file}")"
     vm_initrd_inject+=" --initrd-inject ${kickstart_file}"
     # Download and inject all the kickstart include files
-    wget -r -q -nd -A "*.cfg" -P "${kickstart_idir}" "$(dirname "${kickstart_url}")"
+    wget -r -q -nd -A "*.cfg" -P "${kickstart_idir}" "$(dirname "${kickstart_url}")/"
     for cfg_file in "${kickstart_idir}"/*.cfg ; do
         vm_initrd_inject+=" --initrd-inject ${cfg_file}"
     done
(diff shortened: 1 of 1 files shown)


test/bin/scenario.sh
```diff
@@ -622,7 +622,7 @@ launch_vm() {
     vm_extra_args+=" inst.ks=file:/$(basename "${kickstart_file}")"
     vm_initrd_inject+=" --initrd-inject ${kickstart_file}"
     # Download and inject all the kickstart include files
-    wget -r -q -nd -A "*.cfg" -P "${kickstart_idir}" "$(dirname "${kickstart_url}")"
+    wget -r -q -nd -A "*.cfg" -P "${kickstart_idir}" "$(dirname "${kickstart_url}")/"
```
ggiguash, line 625: Why do we need this change?

vanhalenar: The command would otherwise not work for me, it would redirect to the url with the slash but not download any files. This way it doesn't have to redirect.

ggiguash: This is a bit strange as it's been working for some time already and it's not related to the current changes.
What's version of wget you have installed? This is what I have, for example.
```
GNU Wget 1.21.1 built on linux-gnu.
``

vanhalenar: I agree it is strange. I have `GNU Wget2 2.2.0`. I'll try downgrading then.

ggiguash: Please, verify that it works for you in the older version with and without changes. 
You're probably using Fedora, which has a newer wget version. I'm fine with this change - just need to make sure we're not breaking anything for RHEL 9.4 environments.

vanhalenar: I tested on `Wget 1.21.4` on Fedora and it works fine with and without the slash.

ggiguash: OK, so it's the new package update than that introduced this issue.
Let's merge it.

PR #4517, opened 2025-02-06
USHIFT-5213: Fix host name setting to persist when applied in bootc kickstart

ggiguash:
The `--no-hostname` argument is required to avoid embedding `/etc/hostname` file into the image. 
This allows for a proper host name setting by the anaconda installer if the image is used for installing a new operating system.

### DIFF
--- a/docs/contributor/image_mode.md
+++ b/docs/contributor/image_mode.md
@@ -60,9 +60,14 @@ USER_PASSWD="<your_redhat_user_password>"
 IMAGE_NAME=microshift-4.17-bootc
 
 sudo podman build --authfile "${PULL_SECRET}" -t "${IMAGE_NAME}" \
+    --no-hostname \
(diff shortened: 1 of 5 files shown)


ggiguash:
The decision was to fix this problem on the import of bootc images. See the following issues / PRs:
* https://gitlab.com/fedora/bootc/tracker/-/issues/25 
* https://github.com/containers/bootc/issues/1064 
* https://github.com/containers/bootc/pull/1096

Dianne087:
Micromedia or called fusion Mx pls allowing me to pass maybe one of my friend inserting to Database if u will listining to me i will make also as part of fusion, were not far using java (printf)let me to arrange it

Dianne087:
Python be hold monitor always

Dianne087 (commented):
Python always specifies i need your help to guard,

PR #4531, opened 2025-02-10
USHIFT-5352: Use EUS bootc base images and prevent unintended OS version upgrade

ggiguash:


ggiguash:
Transient subscription management issues
```
+ 12:19:06.792586087 ./scripts/devenv-builder/configure-vm.sh:128 	sudo subscription-manager release --set 9.4
Remote server error. Please check the connection details, or see /var/log/rhsm/rhsm.log for more information.
```

/test e2e-aws-tests-bootc
/test e2e-aws-tests

packaging/imagemode/Containerfile.repobase
```diff
@@ -9,6 +9,8 @@ FROM ${BASE_IMAGE_URL}:${BASE_IMAGE_TAG}
 
 ARG DNF_OPTIONS
 
+# Do not run 'dnf upgrade' command to avoid overrides of the base
+# images specified by the container image build pipelines.
```
pmtk, line 13: Is this the best advice we can give? Are those images rebuilt daily so we don't need to upgrade the packages?

ggiguash: It goes back to the comment in the beginning of this Containerfile. It's crafted for ART pipeline requirements.

pmtk: Ack, I didn't see the comment at the top of the file, so I assumed this is an example for the user (just like the blueprint example in release-info rpm)

PR #4538, opened 2025-02-12
USHIFT-5381: Check microshift.service for all healthchecks

pmtk:
If microshift optional RPMs were installed, but microshift is not enabled, primary healthcheck will exit early without an error.

This doesn't hold for other healthchecks - there's no check if microshift.service is enabled, it just proceeds to querying the API server.

Result is that optional healthchecks fail greenboot even if MicroShift is disabled.

### DIFF
--- a/test/suites/optional/healthchecks-disabled-service.robot
+++ b/test/suites/optional/healthchecks-disabled-service.robot
@@ -1,0 +1,38 @@
+*** Settings ***
+Documentation       Test if healthcheck exits quickly when MicroShift service is disabled
+
+Resource            ../../resources/common.resource
+Resource            ../../resources/microshift-process.resource
+Library             DateTime
+
+Suite Setup         Setup Suite
+Suite Teardown      Teardown Suite
+
+
+*** Test Cases ***
+Healthchecks Should Exit Fast And Successful When MicroShift Is Disabled
+    [Documentation]    When microshift.service is disabled, the healtchecks should
+    ...    exit quickly and without an error. They should not attempt to query API Server.
+    [Tags]    ushift-5381
+    [Setup]    Disable MicroShift
+
+    ${stdout}    ${stderr}    ${rc}=    SSHLibrary.Execute Command
+    ...    find /etc/greenboot/check/required.d/ -iname '*_microshift_running_check_*'
+    ...    sudo=True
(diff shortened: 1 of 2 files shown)


ggiguash:
@pmtk , should we add a test for this in the `optionals` suite?

test/suites/optional/healthchecks-disabled-service.robot
```diff
@@ -0,0 +1,31 @@
...
+
+*** Test Cases ***
+Healthchecks Should Exit Fast And Successful When MicroShift Is Disabled
+    [Documentation]    When microshift.service is disabled, the healtchecks should
+    ...    exit quickly and without an error. They should not attempt to query API Server.
+    [Tags]    ushift-5381
+    [Setup]    Disable MicroShift
+    ${before}=    Get Current Date
+    ${stdout}    ${stderr}    ${rc}=    SSHLibrary.Execute Command
+    ...    /etc/greenboot/check/required.d/41_microshift_running_check_multus.sh
```
ggiguash, line 20: Should we do this for all the scripts in that directory except the main one?

test/suites/optional/healthchecks-disabled-service.robot
```diff
@@ -0,0 +1,31 @@
...
+    ...    sudo=True
+    ...    return_rc=True
+    ...    return_stderr=True
+    ...    return_stdout=True
+    Should Be Equal As Integers    0    ${rc}
+    ${after}=    Get Current Date
+    Should Contain    ${stderr}    microshift.service is not enabled
+    ${diff}=    Subtract Date From Date    ${after}    ${before}
+    # Verify that the command returned very quickly - it didn't waste time trying to access API Server.
+    Should Be True    ${diff} < 5    Multus healthcheck script took too long to finish
```
ggiguash, line 30: Why do we need the time check? Is getting the right error message not enough?

test/suites/optional/healthchecks-disabled-service.robot
pmtk: Done

test/suites/optional/healthchecks-disabled-service.robot
pmtk: You're right, removed

PR #4539, opened 2025-02-12
USHIFT-5201: Add new RPM with AI Model Serving for MicroShift

pmtk:


packaging/rpm/microshift.spec
```diff
@@ -221,6 +221,24 @@ The microshift-gateway-api-release-info package provides release information fil
 release. These files contain the list of container image references used by Gateway API
 and can be used to embed those images into osbuilder blueprints.
 
+%package ai-model-serving
+Summary: AI Model Serving for MicroShift
+ExclusiveArch: x86_64
+Requires: microshift = %{version}
+
+%description ai-model-serving
+The microshift-ai-model-serving package provides manifests for the RHOAI based kserve and serving runtimes.
```
ggiguash, line 230: Please, work with docs team on the description line. "kserve and serving runtimes" seems to be a bit cryptic

packaging/rpm/microshift.spec
```diff
@@ -221,6 +221,24 @@ The microshift-gateway-api-release-info package provides release information fil
...
+
+%package ai-model-serving-release-info
+Summary: Release information for AI Model Serving for MicroShift
+BuildArch: noarch
+Requires: microshift = %{version}
+
+%description ai-model-serving-release-info
+The microshift-ai-model-serving-release-info package provides release information files for this
+release. These files contain the list of container image references used by Model Serving
+and can be used to embed those images into osbuilder blueprints.
```
ggiguash, line 240: As we now also have bootc, we should probably changing the wording by removing "osbuilder"?

packaging/rpm/microshift.spec
```diff
@@ -484,6 +502,58 @@ cat assets/optional/gateway-api/kustomization.x86_64.yaml >> %{buildroot}/%{_pre
 mkdir -p -m755 %{buildroot}%{_datadir}/microshift/release
 install -p -m644 assets/optional/gateway-api/release-gateway-api-{x86_64,aarch64}.json %{buildroot}%{_datadir}/microshift/release/
 
+# ai-model-serving
+# Currently only x86_64 is supported. Following `ifarch` prevents building aarch64 RPM by not specifying the files for the aarch64 architecture.
+%ifarch x86_64
```
ggiguash, line 507: This is a long list. Do we need to do this for each directory?
I forgot if there's an option in RPM spec to do file copy recursively including the dir creation.

packaging/rpm/microshift.spec
ggiguash: Whatever we decide to do, please update all the relevant places.

packaging/rpm/microshift.spec
pmtk: Actually this has nothing to do with the rpm spec, it just runs shell commands, in this particular case it's `install` command and it does not support recursive copying

packaging/rpm/microshift.spec
pmtk: Changed it to be less cryptic. Given our vast documentation, I doubt we need to go into much detail in RPM descriptions.

PR #4569, opened 2025-02-18
NO-ISSUE: Enable flannel RPM build by default in build_rpms.sh

ggiguash:
This should not interfere with ART pipelines because those are not using MicroShift build_rpms.sh script.


ggiguash:
The FIPS errors are not related to this PR changes. Overriding to save CI cylces.
/override ci/prow/e2e-aws-tests-bootc-periodic ci/prow/e2e-aws-tests-periodic

PR #4570, opened 2025-02-18
USHIFT-4725: Allow running scenarios with similar name but different type

ggiguash:
This is particularly useful in dev and QE environments when it is necessary to test both bootc and ostree on the same hypervisor.


ggiguash:
Subscription flake, retesting
/test e2e-aws-tests-bootc-periodic-arm

PR #4575, opened 2025-02-19
USHIFT-5099: SOS report support for offline VMs

vanhalenar:


test/bin/scenario.sh
```diff
@@ -159,24 +159,25 @@ sos_report() {
...
                 record_junit "${vmname}" "sos-report" "FAILED"
-            fi
-        else
-            if "${junit}"; then
+            else
                 record_junit "${vmname}" "sos-report" "OK"
             fi
         fi
+
+
```
pmtk, line 180: Unnecessary double new line

test/bin/scenario.sh
```diff
@@ -159,24 +159,25 @@ sos_report() {
...
-            # FIXME: use virsh to copy sos report files
-            if "${junit}"; then
-                record_junit "${vmname}" "sos-report" "SKIP"
+            echo "Creating sos report offline"
+            if ! sos_report_for_vm_offline "${vmdir}" "${vmname}"; then
+                scenario_result=1
+            fi
+        else
+            if ! sos_report_for_vm "${vmdir}" "${vmname}"; then
+                scenario_result=1
```
pmtk, line 168: I think we could do something like:
```
sos_func="sos_report_for_vm"
if [ -z "${ip}" ]; then
    sos_func="sos_report_for_vm_offline"
fi

if ! "${sos_func}" "${vmdir}" "${vmname}"; then
    scenario_result=1
fi
```

test/bin/scenario.sh
```diff
@@ -232,6 +218,91 @@ EOF
     }
 }
 
+invoke_qemu_script() {
+    "${ROOTDIR}/_output/robotenv/bin/python" "${ROOTDIR}/test/resources/qemu-guest-agent.py" "$@"
+}
+
+wait_for_qemu_agent() {
```
pmtk, line 225: How does this differ from `guest_agent_is_ready` from py file? Is the problem that it doesn't return a value?

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +433,106 @@ def guest_agent_is_ready(vm_name: str):
...
+    cp_from_parser.add_argument("--src_dir", required=True, help="source path on VM")
+    cp_from_parser.add_argument("--dst_dir", required=True, help="local destination path")
+    cp_from_parser.add_argument("--pat", required=True, help="filename or pattern")
+
+    cp_to_parser = subparsers.add_parser("upload", help="Copy file to VM")
+    cp_to_parser.add_argument("--vm", required=True, help="domain name")
+    cp_to_parser.add_argument("--src", required=True, help="path to local file")
+    cp_to_parser.add_argument("--dst", required=True, help="destination path on VM")
+
+    bash_parser = subparsers.add_parser("bash", help="run bash command on VM")
```
pmtk, line 522:
```suggestion
    bash_parser = subparsers.add_parser("bash", help="Run bash command on VM")
```

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +433,106 @@ def guest_agent_is_ready(vm_name: str):
...
+    :type src_dir:      str
+    :param dst_dir:     Destination directory where the downloaded files should be saved
+    :type dst_dir:      str
+    :param pattern:     Pattern or filename, e.g. sosreport-*, journal.log
+    :type pattern:      str
+    """
+    if (not src_dir.endswith('/')):
+        src_dir = src_dir + '/'
+    if (not dst_dir.endswith('/')):
+        dst_dir = dst_dir + '/'
```
pmtk, line 479: Would this work?
```suggestion
    if not src_dir.endswith('/'):
        src_dir = src_dir + '/'
    if not dst_dir.endswith('/'):
        dst_dir = dst_dir + '/'
```

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +433,106 @@ def guest_agent_is_ready(vm_name: str):
...
+
+    files = _find_files(vm_name, src_dir, pattern).splitlines()
+    for file in files:
+        filename = basename(file)
+        src = src_dir + filename
```
pmtk, line 484: Isn't `file` already a `src_dir + filename`?

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +433,106 @@ def guest_agent_is_ready(vm_name: str):
...
+def main():
+    parser = argparse.ArgumentParser()
+
+    subparsers = parser.add_subparsers(dest="command", required=True)
+
+    cp_from_parser = subparsers.add_parser("download", help="Copy file(s) from VM")
+    cp_from_parser.add_argument("--vm", required=True, help="domain name")
+    cp_from_parser.add_argument("--src_dir", required=True, help="source path on VM")
+    cp_from_parser.add_argument("--dst_dir", required=True, help="local destination path")
+    cp_from_parser.add_argument("--pat", required=True, help="filename or pattern")
```
pmtk, line 515: Not sure about `--pat` but no ideas for replacement

pmtk (commented):
Nice! I like the ideas

test/bin/scenario.sh
vanhalenar: yes, and also it uses the `_execute` function, which does not let you add params to the `virsh` command, and I wanted to use the `--timeout 2` param there

vanhalenar: I felt it was better to just print a message if the agent is not yet ready because it's expected to not be ready right away instead of raising scary errors

test/assets/sos-wrapper.sh
```diff
@@ -0,0 +1,14 @@
...
+    /usr/bin/microshift-sos-report || {
+        echo "WARNING: The /usr/bin/microshift-sos-report script failed"
+    }
+else
+    chmod +x /tmp/microshift-sos-report.sh
+    PROFILES=network,security /tmp/microshift-sos-report.sh || {
+        echo "WARNING: The /tmp/microshift-sos-report.sh script failed"
+    }
+fi
+chmod +r /tmp/sosreport-* || echo "WARNING: The sos report files do not exist in /tmp"
```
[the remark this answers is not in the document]
vanhalenar: done

test/bin/scenario.sh
vanhalenar: fixed

test/bin/scenario.sh
vanhalenar: changed

test/resources/qemu-guest-agent.py
vanhalenar: changed

test/resources/qemu-guest-agent.py
vanhalenar: lol yes, fixed :D

test/resources/qemu-guest-agent.py
vanhalenar: yeah I don't like it much either, but I didn't have a better idea, should I just leave it `--pat` then?

vanhalenar: `--ptrn` ?

test/bin/scenario.sh
ggiguash: Should we then fix the `guest_agent_is_ready` function instead of writing this code in bash?

vanhalenar: I would rather not modify the `guest_agent_is_ready` function, it might be used elsewhere, I can add a similar function to the python file instead of bash, but I would have to forgo the `--timeout 2` parameter, or not use the `_execute` function to execute guest commands

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +434,129 @@ def guest_agent_is_ready(vm_name: str):
...
+
+    print(f"Waiting for QEMU agent to become available on VM: {vm_name}...")
+    for i in range(1, max_retries):
+        try:
+            guest_agent_is_ready(vm_name)
+            print("QEMU agent is now available")
+            return
+        except Exception:
+            print(f"Attempt {i}/{max_retries}: QEMU agent not ready, retrying in {sleep_time} seconds")
+            sleep(sleep_time)
```
ggiguash, line 522: We should add a `finally` check here to abort execution if max_retries were reached.

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +434,129 @@ def guest_agent_is_ready(vm_name: str):
...
+        upload_file(args.vm, args.src, args.dst)
+    elif args.command == "bash":
+        print(f"Running {args.args}")
+        run_guest_process(args.vm, "/bin/bash", "-c", args.args)
+    elif args.command == "wait":
+        wait_for_guest_agent(args.vm)
+
+
+if __name__ == "__main__":
+    main()
```
ggiguash, line 562: Do we need the `main` function at all? I mean, it does not do much except passing arguments to other functions. What about just calling those functions from bash directly?

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +434,129 @@ def guest_agent_is_ready(vm_name: str):
...
+                })
+                f.write(b64decode(content['buf-b64']))
+
+                if content['eof'] is True:
+                    break
+    finally:
+        _close_file(vm_name, handle)
+
+
+def download_files(vm_name: str, src_dir: str, dst_dir: str, pattern: str):
```
ggiguash, line 466: Please, update the description in the top of the file as you added new functionality.

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +434,129 @@ def guest_agent_is_ready(vm_name: str):
     _execute(vm_name, {
         'execute': 'guest-ping',
     })
+
+
+def _find_files(vm_name: str, dir: str, pattern: str):
+    content, _ = run_guest_process(vm_name, "/bin/find", dir, "-maxdepth", "1", "-name", pattern)
+    return content['stdout']
```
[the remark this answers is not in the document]
ggiguash: Yeah, if we're looking for files, we should probably use `-type f` find argument.

test/resources/qemu-guest-agent.py
vanhalenar: I felt like that would make it ugly in the `scenario.sh` script, doing it this way makes it more readable.

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +434,129 @@ def guest_agent_is_ready(vm_name: str):
...
+    :type src_dir:      str
+    :param dst_dir:     Destination directory where the downloaded files should be saved
+    :type dst_dir:      str
+    :param pattern:     Pattern or filename, e.g. sosreport-*, journal.log
+    :type pattern:      str
+    """
+    if not src_dir.endswith('/'):
+        src_dir = src_dir + '/'
+    if not dst_dir.endswith('/'):
+        dst_dir = dst_dir + '/'
```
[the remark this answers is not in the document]
vanhalenar: I guess they aren't, just guardrails in case someone forgets to add the slash there. Removed

test/resources/qemu-guest-agent.py
vanhalenar: Is a `finally` block that is executed each iteration necessary if the loop only repeats `max_retries` times?  I did add better error handling in case the QEMU agent is unavailable.

test/resources/qemu-guest-agent.py
vanhalenar: done

test/resources/qemu-guest-agent.py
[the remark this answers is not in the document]
vanhalenar: done

test/resources/qemu-guest-agent.py
vanhalenar: changed

test/resources/qemu-guest-agent.py
```diff
@@ -431,3 +434,129 @@ def guest_agent_is_ready(vm_name: str):
...
+                        'handle': handle,
+                        'count': 10240
+                    }
+                })
+                f.write(b64decode(content['buf-b64']))
+
+                if content['eof'] is True:
+                    break
+    finally:
+        _close_file(vm_name, handle)
```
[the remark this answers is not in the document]
vanhalenar: This should be fixed, I extracted the base functionality to a different function to avoid duplication

test/resources/qemu-guest-agent.py
```diff
@@ -516,11 +513,13 @@ def wait_for_guest_agent(vm_name: str):
         try:
             guest_agent_is_ready(vm_name)
             print("QEMU agent is now available")
-            return
+            return 0
```
copejon, line 516: Undocumented return value. That said, is it actually necessary? The value is ignored by the caller.

test/bin/scenario.sh
```diff
@@ -226,9 +226,13 @@ sos_report_for_vm_offline() {
...
     "${ROOTDIR}/scripts/fetch_tools.sh" "robotframework"
 
-    #wait_for_qemu_agent "${vmname}"
     invoke_qemu_script "wait" \
         "--vm" "${full_vmname}"
+    ready="$?"
+    if [ "${ready}" -ne 0 ]; then
+        echo "QEMU agent not available, exiting"
+        exit 1
+    fi
```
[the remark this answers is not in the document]
vanhalenar: oh right, in that case this block is unnecessary, removed

test/resources/qemu-guest-agent.py
vanhalenar: removed

copejon (approved):
Great work! This looks good to me

PR #4586, opened 2025-02-22
USHIFT-5402: Handle DUID in ipv6 test harness

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/test/kickstart-templates/includes/post-network.cfg
+++ b/test/kickstart-templates/includes/post-network.cfg
@@ -23,11 +23,17 @@ done
 # will issue a lease for that DUID and MAC address. When the system boots into anaconda to install
 # the OS with kickstart files, the DUID is automatically generated. After the system boots into the
 # OS a new DUID may be generated, causing DHCP to identify the VM as a new system, thus allocating
-# a new IP address. In order to avoid this, the DUID generated during the installarion is saved and
+# a new IP address. In order to avoid this, the DUID generated during the installation is saved and
 # configured in NetworkManager to use it when the system boots into the OS.
-# The DUID is extracted from the journal logs to avoid pinning to a specific NIC name (which would
-# use nmcli commands).
-DUID=$(journalctl -u NetworkManager | grep duid | grep -Eo "([0-9a-f]{2}:){17}[0-9a-f]{2}" | uniq)
+# The DUID is unique per host, and is extracted from the DHCP6 options of the active connections
+# from NetworkManager.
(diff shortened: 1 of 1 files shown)


test/kickstart-templates/includes/post-network.cfg
```diff
@@ -17,3 +17,18 @@ find /etc/NetworkManager -name '*.nmconnection' -print0 | while IFS= read -r -d
         sed -i 's/method=.*/method=auto/g' "${file}"
     fi
 done
+
+# IPv6 only feature. An IPv6 VM will use DHCPv6 to get an IP address. To identify itself to a DHCP
+# server it uses something called DHCP Unique Identifier (DUID), and based on that the DHCP server
+# will issue a lease for that DUID and MAC address. When the system boots into anaconda to install
+# the OS with kickstart files, the DUID is automatically generated. After the system boots into the
+# OS a new DUID may be generated, causing DHCP to identify the VM as a new system, thus allocating
+# a new IP address. In order to avoid this, the DUID generated during the installarion is saved and
```
agullon, line 26:
```suggestion
# a new IP address. In order to avoid this, the DUID generated during the installation is saved and
```

test/kickstart-templates/includes/post-network.cfg
```diff
@@ -17,3 +17,24 @@ find /etc/NetworkManager -name '*.nmconnection' -print0 | while IFS= read -r -d
...
+# configured in NetworkManager to use it when the system boots into the OS.
+# The DUID is unique per host, and is extracted from the DHCP6 options of the active connections
+# from NetworkManager.
+DUID=$(nmcli con show --active | \
+  awk '{print $1}' | \
+  grep -v NAME | \
+  xargs nmcli --fields DHCP6.OPTION con show | \
+  grep dhcp6_client_id | \
+  awk '{print $4}' | \
+  uniq)
```
ggiguash, line 36: Do you think converting it to a for loop would be more straightforward? 
Also, the `uniq` is a bit confusing - may the process return more than one DUID? If not, a break in for-loop could do the job.

pacevedom: Not sure if the for loop would be more confusing, after all its iterating over all connections to determine which one is using DHCPv6 and take the id from there. This one liner does it in one shot.
About the `uniq`: there is only one DUID per host, but if there is more than one NIC connected to ipv6 the command might return it multiple times. We know it must be the same because its system wide, hence the uniq to remove duplicates.

PR #4596, opened 2025-02-24
OCPBUGS-51184: Remove MicroShift Greenboot Fail Marker on first check start

pmtk:


packaging/greenboot/microshift-running-check.sh
```diff
@@ -11,6 +11,11 @@ if [ "$(id -u)" -ne 0 ] ; then
     exit 1
 fi
 
+if [ -f "${MICROSHIFT_GREENBOOT_FAIL_MARKER}" ]; then
```
ggiguash, line 14: Let's hide it in functions.sh and expose as a function like for the other operations?
Something like `clear_fail_marker_status`?

PR #4600, opened 2025-02-25
USHIFT-5401: Add telemetry configuration parameters

pacevedom:

Closes #<Issue Number>


test/kickstart-templates/includes/post-microshift.cfg
```diff
@@ -18,3 +18,9 @@ else
     MAIN_HOST_IP="$(hostname -I | awk '{print $1}')"
     echo "    - ${MAIN_HOST_IP}" >>/etc/microshift/config.yaml
 fi
+
+# Configure MicroShift to not send telemetry data as default
+cat - >>/etc/microshift/config.yaml <<EOF
```
ggiguash, line 23: Should we discuss this on the arch meeting? Would it not be interesting to see this info from CI?

PR #4623, opened 2025-03-03
USHIFT-5297: Add telemetry client

pacevedom:

Closes #<Issue Number>


pkg/controllers/telemetry.go
```diff
@@ -0,0 +1,126 @@
...
+		defer func() {
+			if r := recover(); r != nil {
+				panicChannel <- r
+			}
+		}()
+	}()
+
+	close(ready)
+
+	if t.config.Telemetry.Status == config.StatusDisabled {
```
agullon, line 74: Would it be better to move this check at the begining of the Run() func? To skip getClusterId() and readPullSecret() funcs when no need to run them.

pkg/controllers/telemetry.go
```diff
@@ -0,0 +1,126 @@
...
+
+	_ = telemetry.NewTelemetryClient(t.config.Telemetry.Endpoint, clusterId, pullSecret)
+	go func() {
+		klog.Infof("First metrics collection")
+		for {
+			select {
+			case <-ctx.Done():
+				klog.Infof("collect and send for the last time")
+				return
+			case <-time.After(time.Hour):
```
agullon, line 86: Is it possible that clusterId or pullSecret values change during 1 hour wait?

pkg/controllers/telemetry.go
pacevedom: Done.

pkg/controllers/telemetry.go
pacevedom: The cluster id does not change during the lifespan of MicroShift.
The pull secrets would need a crio restart to take effect afaik. I dont know if we have any documents on the procedure to update the pull secrets. @ShaunaDiaz do we have something for that?

pacevedom: Switched to this after some brief testing.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,141 @@
...
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", t.encodeAuth(pullSecret)))
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("unable to do the request: %v", err)
+	}
+	defer func() {
+		if _, err := io.Copy(io.Discard, resp.Body); err != nil {
+			klog.Error(err, "error discarding body")
```
pmtk, line 93: Why do we need to do this?

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("unable to read body: %v", err)
+	}
+	return fmt.Errorf("request unsuccessful. Status code: %v. Body: %v", resp.StatusCode, string(body))
+}
+
+func convertMetricsToWriteRequest(metrics []Metric) *prompb.WriteRequest {
```
pmtk, line 107: How about some unit tests for this file?

pkg/controllers/telemetry.go
```diff
@@ -0,0 +1,129 @@
...
+		collectAndSend := func() {
+			pullSecret, err := readPullSecret()
+			if err != nil {
+				klog.Errorf("Unable to get pull secret: %v", err)
+				return
+			}
+			//TODO swap with collected metrics from client.
+			metrics := []telemetry.Metric{}
+			if err := client.Send(ctx, pullSecret, metrics); err != nil {
+				klog.Errorf("Failed to send metrics: %v", err)
```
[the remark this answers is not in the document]
pacevedom: Also thought about this and ended up discarding the idea.
Initially the feature intended to provide one single write request per day. In that case I would vouch for retries because you only have one chance every 24h.
Later on we decided to send them every hour, so worst case scenario you have 24 retries in a single day. Our dashboards tolerate gaps and as long as there is at least one data point per day we should be ok.
Also having one hour gaps in between metrics (for those that change) does not make them critical to fill in the information blackouts.
We want this to be as lean as possible since it is outgoing traffic that is not valuable to the applications themselves.

pkg/telemetry/telemetry.go
pacevedom: This is more of a good practice in this particular case. If you close the body you allow the underlying TCP connection to get re-used on the next request, if you dont it creates a new one. They are 1h apart here, but this is an old habit I have :)

pkg/controllers/telemetry.go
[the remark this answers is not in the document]
pmtk: Ack, sounds good to me

pkg/telemetry/telemetry.go
pmtk: Let's add a comment? :)

pkg/controllers/telemetry.go
```diff
@@ -0,0 +1,129 @@
...
+				klog.Infof("collect and send for the last time")
+				collectAndSend()
+				return
+			case <-time.After(time.Hour):
+				klog.Infof("collect and send again")
+				collectAndSend()
+			}
+		}
+	}()
+	return nil
```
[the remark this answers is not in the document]
pacevedom: Done.

pkg/controllers/telemetry.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func getClusterId() (string, error) {
+	data, err := os.ReadFile(ClusterIDFilePath)
```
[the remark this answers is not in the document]
pacevedom: Actually it makes more sense, done.

pkg/telemetry/telemetry.go
pacevedom: Done

PR #4632, opened 2025-03-05
USHIFT-5373: Replace GenerateUniqueTempPath by Go library functions

vanhalenar:


pkg/util/util.go
```diff
@@ -144,19 +143,24 @@ func (l LogFilePath) Remove() error {
...
-		if exists, err := PathExists(newPath); err != nil {
-			return "", err
-		} else if !exists {
-			return newPath, nil
-		}
-	}
-	return "", fmt.Errorf("attempted to generate unique temporary path (%q) for 1000 tries - giving up", path)
+// GetTempPathArgs returns the directory in which to create a temp file
+// along with the pattern for the temp file name.
+func GetTempPathArgs(path string) (string, string) {
```
ggiguash, line 148: This function should probably be private and not exported from the module?

vanhalenar: yes, fixed

PR #4636, opened 2025-03-05
NO-ISSUE: Fix typo in scenario.sh script

ggiguash:


kasturinarra:
I understand  full_vmname is a variable and we needed to substitute its value, so based on that adding lgtm to the PR here.

PR #4656, opened 2025-03-10
AUTH-482: Add required-scc annotation to router

jacobsee:
Adds the `required-scc` annotation to the default router (set to `restricted`) so that it can't be preempted by any other custom SCCs that might exist with a higher priority.

Test (flaking) showing that `restricted` is the correct SCC for the router: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-microshift-release-4.19-periodics-e2e-aws-ovn-ocp-conformance-optional-components-arm/1899015011318108160

**Which issue(s) this PR addresses**: AUTH-482

Closes #none

### DIFF
--- a/assets/components/openshift-router/deployment.yaml
+++ b/assets/components/openshift-router/deployment.yaml
@@ -7,6 +7,7 @@ spec:
     metadata:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: restricted
       labels:
         ingresscontroller.operator.openshift.io/deployment-ingresscontroller: default
     spec:


kramaranya (commented):
/lgtm

assets/components/openshift-router/deployment.yaml
```diff
@@ -7,6 +7,7 @@ spec:
     metadata:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: restricted
```
pmtk, line 10: Can you tell why it isn't included in the upstream file (https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/manifests/assets/router/deployment.yaml#L10-L11)? It is set during runtime by the operator?

This file is copied when microshift updates its dependencies, so this change might be overwritten in few days if we merge this. If this is runtime addition, then we also need a change in rebase.sh script ([see this](https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/rebase.sh#L684-L751)).

I believe this should do the job:
```
yq '.spec.template.metadata.annotations += {"openshift.io/required-scc": "restricted"}' ./assets/components/openshift-router/deployment.yaml
```

pacevedom:
/hold
Holding because of https://github.com/openshift/microshift/pull/4656#pullrequestreview-2678327979. This change will get overwritten in the next rebase if not included in `rebase.sh`.

assets/components/openshift-router/deployment.yaml
jacobsee: Yes, the operator is setting it dynamically based on whether it needs host networking or not here:  https://github.com/openshift/cluster-ingress-operator/blob/7dc12cf28557f8c80460435649cfc393761e0285/pkg/operator/controller/ingress/deployment.go#L1212-L1215

I don't see where Microshift has the same host networking config, so just setting it to `restricted` here.

I was not aware of the rebase script - made the change there. Thanks!

PR #4659, opened 2025-03-11
USHIFT-5339: Test AI Model Serving with NVIDIA GPU

pmtk:


scripts/ci-ai-model-serving/tests/04-verify-device-plugin.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -xeuo pipefail
+
+oc rollout status -n nvidia-device-plugin daemonset/nvidia-device-plugin-daemonset
+
+capacity=$(oc get node -o json | jq -r '.items[0].status.capacity')
+gpus=$(echo "${capacity}" | jq -r '."nvidia.com/gpu"')
+
+if [[ "${gpus}" != "1" ]]; then
```
ggiguash, line 10: Should we check that it's not empty? I mean there may be more than one GPUs on the node.

scripts/ci-ai-model-serving/tests/05-test-cuda-vector-add.sh
```diff
@@ -0,0 +1,45 @@
...
+          seccompProfile:
+            type: "RuntimeDefault"
+EOF
+
+# Wait till complete
+oc wait --for=condition=complete --timeout=120s -n cuda-test job/test-cuda-vector-add
+
+# Get the logs
+pod=$(oc get pods -n cuda-test --selector=batch.kubernetes.io/job-name=test-cuda-vector-add --output=jsonpath='{.items[*].metadata.name}')
+logs=$(oc logs -n cuda-test "${pod}")
```
ggiguash, line 37: I'm a bit wary of storing entire log outputs in a variable. Should we use a tmp file instead?

scripts/ci-ai-model-serving/tests/05-test-cuda-vector-add.sh
```diff
@@ -0,0 +1,45 @@
...
+
+oc delete job -n cuda-test test-cuda-vector-add
+oc delete ns cuda-test
+
+if ! echo "${logs}" | grep -q PASSED; then
```
ggiguash, line 42: Could we make it more precise, i.e. ^PASSED or like?

scripts/ci-ai-model-serving/tests/06-test-vllm.sh
```diff
@@ -0,0 +1,119 @@
...
+# Because of temperature=0.5, the responses vary between calls.
+curl -X POST \
+    granite-predictor.apps.example.com/v1/completions \
+    --connect-to "granite-predictor.apps.example.com::$(hostname -i):" \
+    -H "Content-Type: application/json" \
+    --data '{
+        "model": "granite-3b-code-base-2k",
+        "prompt": "Once upon a time,",
+        "max_tokens": 256,
+        "temperature": 0.5}' | jq
```
ggiguash, line 111: Should we check the answer is not empty?

scripts/ci-ai-model-serving/tests/vllm-image/Containerfile
```diff
@@ -0,0 +1,23 @@
...
+# podman push quay.io/microshift/ai-testing-model:vllm-granite-3b-code-base-2k
+
+FROM registry.access.redhat.com/ubi9/python-311:latest as base
+
+USER root
+
+RUN pip install huggingface-hub
+
+# Download the model file from hugging face
+WORKDIR /tmp
```
ggiguash, line 13: Is the workdir change necessary? The model seems to be downloaded to /model

scripts/ci-ai-model-serving/tests/05-test-cuda-vector-add.sh
pmtk: It's only like 6 lines, but sure we can do it
```
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
```

scripts/ci-ai-model-serving/tests/05-test-cuda-vector-add.sh
pmtk: This is the output.
```
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
```

Would `^Test PASSED$` be better?

scripts/ci-ai-model-serving/tests/06-test-vllm.sh
pmtk: Yeah, that's a good idea for double checking it

scripts/ci-ai-model-serving/tests/vllm-image/Containerfile
pmtk: hadolint complaint: 
```
 ./scripts/ci-ai-model-serving/tests/vllm-image/Containerfile:13 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set. 
```

So it was either `WORKDIR /tmp` or
```
COPY download_model.py /tmp
RUN python /tmp/download_model.py
```
but I figured that `WORKDIR` would work without retesting (laziness).
Do you prefer adding `/tmp` to COPY and RUN?

PR #4673, opened 2025-03-13
USHIFT-5473: Doc for AI Model Serving

pmtk:


docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
+# AI Model Serving on MicroShift
+
+AI Model Serving on MicroShift is a platform for serving AI models.
+It includes limited subset of Red Hat OpenShift AI (RHOAI): [kserve](https://github.com/opendatahub-io/kserve) and ServingRuntimes.
+Now you can train your models in the cloud and serve them on the edge.
+
+MicroShift is intended for the edge (small footprint, low resources),
+therefore kserve is configured to use a "Raw Deployment" mode which means that:
+- Kubernetes Deployments and Services will be created, and
+- neither Service Mesh (Istio) nor Serverless (Knative) are required.
```
ggiguash, line 10: If this is a user doc, I'm not sure how to interpret this info, mostly the first bullet.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+Additionally, automatic creation of the Ingress objects is disabled. If you want to expose your model outside the cluster,
+create a Route object instead.
```
ggiguash, line 13: This should probably go with the above info in a more user-friendly way.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+[Learn more about Raw deployment mode from RHOAI documentation](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#raw_deployment_mode)
+
+## Limitations
+- Currently, AI Model Serving on MicroShift is only available on x86_64 platform.
+- AI Model Serving on MicroShift only includes very specific subset of RHOAI.
+- Securing (e.g. OAUTH2) exposed model server's endpoint is user's responsibility
```
ggiguash, line 20: Did we not discover that not all the models / servers work on IPv6?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+## Known issues
+- Because of [bug in kserve](https://github.com/kserve/kserve/pull/4274)
+  ([to be ported to RHOAI](https://issues.redhat.com/browse/RHOAIENG-21106)),
+  rebooting a MicroShift host can result model server not coming back up if it was using ModelCar (model in OCI image).
```
ggiguash, line 25: Let's rephrase it in more user-friendly way, please.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+- Because of MicroShift's architecture, installing microshift-ai-model-serving RPM before running `systemctl start microshift` for the first time,
+  can cause a failure to start up, however MicroShift will restart and then start successfully. See [OCPBUGS-51365](https://issues.redhat.com/browse/OCPBUGS-51365).
+  - **It might also cause the ServingRuntimes to not be deployed because kustomizer gives up before webhook is available**
+- Currently, ClusterServingRuntimes are not supported by RHOAI which means that
+  user needs to copy ServingRuntime shipped within the package to their workload's namespace.
+  In future releases we're hoping to transition to ClusterServingRuntime which is ServingRuntime usable from any namespace
+  (if users will need to override some options, they still will be able to create ServingRuntime which take precedence over ClusterServingRuntime).
```
ggiguash, line 32: I'm not sure making future looking statements is necessary here.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+It's possible to add extra arguments that will be passed to the model server using `.spec.predictor.model.args`.
+
+Below is example of InferenceService with model in `openvino_ir` format.
+It features additional argument `--layout=NHWC:NCHW` to make OVMS accept the
+request data in different format than usual for demonstration purposes.
+Extra args are passed through to the OVMS container.
+
+For more information about InferenceService CR refer to
+[RHOAI documentation](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#inferenceservice).
+
```
ShaunaDiaz, line 260:
```suggestion
.Example `InferenceService` object with an `openvino_ir` model format
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+oc create -n ai-demo -f ./FILE.yaml
+```
+
+Output:
+```
+inferenceservice.serving.kserve.io/ovms-resnet50 created
+```
+
+Soon, a Deployment and a Pod should appear in that namespace.
+Depending on the size of ServingRuntime's image and size of the ModelCar OCI image,
```
ShaunaDiaz, line 288:
```suggestion
Depending on the size of ServingRuntime's image and the size of the ModelCar OCI image,
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+it may take a while for the Pod to become ready.
+
```
ShaunaDiaz, line 290: You can check the status of the pod with the following commands:

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```sh
+$ oc get svc -n ai-demo
+NAME                      TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
+ovms-resnet50-predictor   ClusterIP   None         <none>        80/TCP    119s
+```
+
+#### Specifying hardware accelerators
+
+InferenceService can also include plethora of different options.
```
ShaunaDiaz, line 313: nice use of "plethora"

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+For example it can contain `resources` section that is passed to the Deployment and then to the Pod,
+so the model server gets access to the NVIDIA hardware (thanks to the device plugin):
```
ShaunaDiaz, line 315:
```suggestion
so that the model server gets access to the hardware (thanks to the device plugin). In this example, an NVIDIA device:
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+      resources:
+        limits:
+          nvidia.com/gpu: 1
+        requests:
+          nvidia.com/gpu: 1
+```
+
+For complete InferenceService specification refer to [kserve API Reference](https://kserve.github.io/website/latest/reference/api/).
+
+### Create Route
```
ShaunaDiaz, line 329:
```suggestion
### Creating a Route
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+> Note: You don't need to wait for model server Pod's readiness before creating Route.
```
ShaunaDiaz, line 331:
```suggestion
> Note: You don't need to wait for the model server Pod's readiness before creating a Route.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+Finally, we can create an OpenShift Route CR to expose the service.
+You can either use `oc expose svc` command or create definition in a YAML file and apply it.
```
ShaunaDiaz, line 334:
```suggestion
You can either use the `oc expose svc` command or create a definition in a YAML file and apply it.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+          nvidia.com/gpu: 1
+```
+
+For complete InferenceService specification refer to [kserve API Reference](https://kserve.github.io/website/latest/reference/api/).
+
+### Create Route
+
+> Note: You don't need to wait for model server Pod's readiness before creating Route.
+
+Finally, we can create an OpenShift Route CR to expose the service.
```
ShaunaDiaz, line 333:
```suggestion
Create an OpenShift Route CR to expose the service.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+route.route.openshift.io/ovms-resnet50-predictor exposed
+
+$ oc get route -n ai-demo
+NAME                      HOST                                               ADMITTED   SERVICE                   TLS
+ovms-resnet50-predictor   ovms-resnet50-predictor-ai-demo.apps.example.com   True       ovms-resnet50-predictor
+```
+
+### Querying the model server
+
+Finally, we'll check if our model is ready for inference and we'll reuse OVMS examples to test the inference.
```
ShaunaDiaz, line 347:
```suggestion
You are ready to check if your model is ready for inference. We'll reuse the OVMS examples to test the inference.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+Get the IP of the MicroShift cluster and assign it to the `IP` variable.
```
ShaunaDiaz, line 349:
```suggestion
Get the IP of the MicroShift cluster and assign it to the `IP` variable of the Route you just created.
```
I think?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Use the HOST value of the Route created in previous step and assign to the `DOMAIN` variable.
```
ShaunaDiaz, line 350:
```suggestion
Use the `HOST` value of the Route and assign to the `DOMAIN` variable.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Next, run the following `curl` command.
```
ShaunaDiaz, line 351:
```suggestion
Next, run the following `curl` command:
```
I know I didn't note all of these, but you can use a colon after every instruction.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Alternatively, instead of using `--connect-to "${DOMAIN}::${IP}:"`,
```
ShaunaDiaz, line 352:
```suggestion
Alternatively, instead of using the `--connect-to "${DOMAIN}::${IP}:"` flag,
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+you can use real DNS, or add the IP and the domain to the `/etc/hosts` file.
+```sh
```
ShaunaDiaz, line 354: so right here I think you're going to give me the real DNS way; but it's the curl command; perhaps just give the curl and then show the DNS as an optional step

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Response code 200 is expected. Example output:
+```
+HTTP/1.1 200 OK
+content-type: application/json
+date: Wed, 12 Mar 2025 16:01:32 GMT
+content-length: 0
+set-cookie: 56bb4b6df4f80f0b59f56aa0a5a91c1a=4af1408b4a1c40925456f73033d4a7d1; path=/; HttpOnly
+```
+
+We can also query model's metadata:
```
ShaunaDiaz, line 370:
```suggestion
We can also query the model's metadata:
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```sh
+curl "${DOMAIN}/v2/models/ovms-resnet50" \
+    --connect-to "${DOMAIN}::${IP}:"
+```
+Output:
```
ShaunaDiaz, line 375:
```suggestion
Example output:
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Let's try querying actual model - following example verifies if inference is in accordance to the training data.
+
+First, download an image of a bee from the OpenVino examples:
+```sh
+curl -O https://raw.githubusercontent.com/openvinotoolkit/model_server/main/demos/common/static/images/bee.jpeg
+```
+
+Next, we need to create the request data.
+1. Start with an inference header in JSON format.
+1. Get the size of the header. It needs to be passed to the OVMS later in form of HTTP header.
```
ShaunaDiaz, line 389:
```suggestion
1. Get the size of the header. It needs to be passed to the OVMS later in form of an HTTP header.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```
+
+Let's try querying actual model - following example verifies if inference is in accordance to the training data.
+
+First, download an image of a bee from the OpenVino examples:
+```sh
+curl -O https://raw.githubusercontent.com/openvinotoolkit/model_server/main/demos/common/static/images/bee.jpeg
+```
+
+Next, we need to create the request data.
```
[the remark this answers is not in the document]
ShaunaDiaz: numbering....

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+1. Start with an inference header in JSON format.
+1. Get the size of the header. It needs to be passed to the OVMS later in form of HTTP header.
+1. Append size of the image to the request file. OVMS expects 4 bytes (little endian).
```
ShaunaDiaz, line 390:
```suggestion
1. Append the size of the image to the request file. OVMS expects 4 bytes (little endian).
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+   Command below uses `xxd` utility which is part of `vim-common` package.
+1. Append the image to the request file.
```
ShaunaDiaz, line 392:
```suggestion
1. Append the image to the request file:
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+HEADER_LEN="$(stat -c %s "${REQ}")"
+
+# Add size of the data (image) in binary format (4 bytes, little endian)
+printf "%08X" $(stat --format=%s "${IMAGE}") | sed 's/\(..\)/\1\n/g' | tac | tr -d '\n' | xxd -r -p >> "${REQ}"
+
+# Add the data, i.e. the image
+cat "${IMAGE}" >> "${REQ}"
+```
+
+Now we can make an inference request against the model server using `ovms-resnet50` model.
```
ShaunaDiaz, line 411:
```suggestion
Now we can make an inference request against the model server that is using the `ovms-resnet50` model.
```
it's the server using the model; not us using the model to make the inference request, right?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```sh
+curl \
+    --data-binary "@./request.json" \
+    --header "Inference-Header-Content-Length: ${HEADER_LEN}" \
+    "${DOMAIN}/v2/models/ovms-resnet50/infer" \
+    --connect-to "${DOMAIN}::${IP}:" > response.json
+```
+
+Response saved to a `response.json` is a JSON object which a structure as shown below.
```
ShaunaDiaz, line 420:
```suggestion
A response saved to a `response.json` is a JSON object which has the following structure:
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Contents of `.outputs[0].data` were omitted from the example for brevity.
```
ShaunaDiaz, line 421:
```suggestion
The contents of `.outputs[0].data` were omitted from the example for brevity.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+    "outputs": [{
+            "name": "1463",
+            "shape": [1, 1000],
+            "datatype": "FP32",
+            "data": [ ....... ]
+        }]
+}
+```
+
+To verify the response, we'll use python.
```
ShaunaDiaz, line 435:
```suggestion
To verify the response, we'll use Python.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+We need to obtain index of the highest element in the `.outputs[0].data`.
```
ShaunaDiaz, line 436:
```suggestion
We need to obtain the index of the highest element in the `.outputs[0].data`.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+with open('response.json') as f:
+    response = json.load(f)
+
+data = response["outputs"][0]["data"]
+argmax = data.index(max(data))
+print(argmax)
+```
+
+The output of the script above should be `309`.
```
ShaunaDiaz, line 449:
```suggestion
The output of the Python script we just ran should be `309`.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+# HELP ovms_requests_success Number of successful requests to a model or a DAG.
+# TYPE ovms_requests_success counter
+ovms_requests_success{api="KServe",interface="REST",method="ModelReady",name="ovms-resnet50"} 4
+ovms_requests_success{api="KServe",interface="REST",method="ModelMetadata",name="ovms-resnet50",version="1"} 1
+```
+
+## Appendix
+### Overriding kserve configuration
+
+If you wish to override kserve settings, you need to make a copy of existing ConfigMap, tweak desired settings, and overwrite existing ConfigMap.
```
ShaunaDiaz, line 476:
```suggestion
If you wish to override kserve settings, you need to make a copy of existing ConfigMap, tweak the desired settings, and overwrite the existing ConfigMap.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+Settings are stored in a ConfigMap named `inferenceservice-config` in namespace `redhat-ods-applications`.
```
ShaunaDiaz, line 478:
```suggestion
Settings are stored in a ConfigMap named `inferenceservice-config` in the `redhat-ods-applications` namespace .
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Alternatively, you can copy it from `/usr/lib/microshift/manifests.d/001-microshift-ai-model-serving/kserve/inferenceservice-config-microshift-patch.yaml`.
```
ShaunaDiaz, line 479:
```suggestion
Alternatively, you can copy the ConfigMap from `/usr/lib/microshift/manifests.d/001-microshift-ai-model-serving/kserve/inferenceservice-config-microshift-patch.yaml`.
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+After tweaking it, you must apply it and restart kserve (e.g. deleting Pod or scaling the Deployment down to 0 and back to 1).
```
ShaunaDiaz, line 481:
```suggestion
After tweaking it, you must apply the ConfigMap and restart kserve (e.g. by deleting Pod or scaling the Deployment down to 0 and back to 1).
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+For OSTree/bootc systems, create a new manifest making sure it's applied after `/usr/lib/microshift/manifests.d/001-microshift-ai-model-serving`.
```
ShaunaDiaz, line 482:
```suggestion
For RHEL for Edge and bootc systems, create a new manifest making sure it's applied after `/usr/lib/microshift/manifests.d/001-microshift-ai-model-serving`.
```

ShaunaDiaz (commented):
Lot of comments! Everything your judgement, of course.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
+# AI Model Serving on MicroShift
+
+AI Model Serving on MicroShift is a platform for serving AI models.
+It includes limited subset of Red Hat OpenShift AI (RHOAI): [kserve](https://github.com/opendatahub-io/kserve) and ServingRuntimes.
```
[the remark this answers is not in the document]
pmtk: Kserve has several modes of operation: serverless, model mesh, and raw deployment.

For microshift we opted for Raw Deployment mode because we don't need to install extra dependencies (istio and knative).

There's nothing preventing users from changing the kserve's settings and installing istio+knative, but from MicroShift's perspective I believe we only want to support RawDeployment? @DanielFroehlich wdyt?

docs/user/ai_model_serving.md
pmtk: Yeah. I wrote this doc from a perspective of someone already knowing kserve so wanted to write out the differences, but maybe we need to show in simpler way how we intend it to be used on microshift

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+- Currently, AI Model Serving on MicroShift is only available on x86_64 platform.
+- AI Model Serving on MicroShift only includes very specific subset of RHOAI.
+- Securing (e.g. OAUTH2) exposed model server's endpoint is user's responsibility
+
+## Known issues
+- Because of [bug in kserve](https://github.com/kserve/kserve/pull/4274)
+  ([to be ported to RHOAI](https://issues.redhat.com/browse/RHOAIENG-21106)),
+  rebooting a MicroShift host can result model server not coming back up if it was using ModelCar (model in OCI image).
+- Because of MicroShift's architecture, installing microshift-ai-model-serving RPM before running `systemctl start microshift` for the first time,
+  can cause a failure to start up, however MicroShift will restart and then start successfully. See [OCPBUGS-51365](https://issues.redhat.com/browse/OCPBUGS-51365).
```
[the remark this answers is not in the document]
pmtk: Yes, it is

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+  - **It might also cause the ServingRuntimes to not be deployed because kustomizer gives up before webhook is available**
```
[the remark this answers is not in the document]
pmtk: This is a bit more nuanced and I think I'll remove it from the doc as I plan to work on this next week.
MicroShift will fail to apply ServingRuntimes only if the kserve Pod isn't running yet.
So e.g. first start after install, first start after reboot (if it didn't happen the previous boot). But if MicroShift runs for 10 minutes, kserve is up, and someone runs `systemctl restart microshift`, then it would work because the webhook is up.
Just FYI, I don't think we need to document it.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+## Definitions
+
+- [ServingRuntime](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#servingruntime)
+- [InferenceService](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#inferenceservice)
+
+## Supported ServingRuntimes
+
+For the list of supported Serving Runtimes refer to
+[RHOAI documentation](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#supported-model-serving-runtimes_serving-large-models).
```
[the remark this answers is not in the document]
pmtk: I believe so. I haven't tested all of them, but none of these are for 'multi model' deployments, so they should work with any kserve.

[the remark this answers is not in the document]
pmtk: On second thought: I guess we want to say we support only those runtimes we tested with microshift, right? So at the time it means OpenVino Model Server and vLLM (for nvidia, not guadi or amd)

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+Keep in mind that in MicroShift they are created directly as ServingRuntimes in `redhat-ods-applications`, not as templates,
```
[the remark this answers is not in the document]
pmtk: Not really. It's more of a implementation detail and diff to RHOAI.
Just copying the SR to custom namespace looks different compared to RHOAI.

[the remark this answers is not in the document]
pmtk: Depends whether you're applying it against cluster directly or creating a manifest

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+so they need to be copied to workload's namespace.
+
+## General usage overview
+
```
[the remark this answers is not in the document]
pmtk: Yes, but the markdown rendering fixes it for "Rich view/diff"

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+1. Prepare model for serving
+1. Configure OS and MicroShift for the hardware - driver & device plugin
```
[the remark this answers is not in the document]
pmtk: @hgeaydem, @DanielFroehlich, @ggiguash do you know anything about "image mode" and accelerator drivers? Is there any procedure? Especially matching driver against bootc image kernel (as during build time I think it'll use host's kernel).
I don't know anything about bootc and RT kernel either tbh...

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+1. Install AI Model Serving for MicroShift
+1. (Optional) Package model into an OCI image (ModelCar)
```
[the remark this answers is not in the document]
pmtk: TBH best info is the upstream kserve doc: https://kserve.github.io/website/latest/modelserving/storage/oci/
There are other ways of delivering the model (S3, URI, PVC) but so far I only tested ModelCar (OCI).
In general ModelCar is very convenient for offline deployments: you can embed the OCI image just like any other container image.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+1. Select suitable Serving Runtime (Model Server)
```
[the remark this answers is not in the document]
pmtk: Yeah, but RHOAI doc has more info about them, like formats, etc.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+1. Copy ServingRuntime Custom Resource from `redhat-ods-applications` to your own namespace
```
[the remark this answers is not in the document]
pmtk: What do you mean by "application namespace"?
You create ServingRuntime and InferenceService in namespace X, and in the same namespace the model server serving the model will run.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+1. Create InferenceService
+1. Create Route
```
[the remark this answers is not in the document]
pmtk: Yes. If the workload that will ask model server about stuff is inside the microshift cluster, it does not require the route

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+1. Make requests against model server
```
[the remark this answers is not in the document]
pmtk: This is just very general overview, more information below

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+## Setting up hardware - drivers and device plugins
+
+In order to enable GPU/hardware accelerator for Red Hat Device Edge,
+you must follow partner's guidance on installing Operator, or driver for RHEL and Device Plugin for Kubernetes.
+Operators may be more convenient, but installing only the driver and device plugin may be more resource efficient.
+
+MicroShift does not provide support for partner's procedure.
+In case of any problems consult partner's documentation or support.
+Following links serve only as examples / pointers and may not include everything needed.
```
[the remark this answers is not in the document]
pmtk: Yes

docs/user/ai_model_serving.md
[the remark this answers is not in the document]
agullon: Totally agree on this. It may be other relevant limitations, for example currently under investigation: OpenVino Model is supported on IPv6?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+- AMD
+  - [Operator](https://instinct.docs.amd.com/projects/gpu-operator/en/latest/)
+  - [Device Plugin](https://github.com/ROCm/k8s-device-plugin)
+  - [Driver](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/install-overview.html)
+
+## Step by step guide
+
+Below is an example usage of AI Model Serving for MicroShift.
+It uses OpenVino Model Server (OVMS) and resnet50 model.
+OVMS can run on CPU, no hardware accelerator is needed for this example.
```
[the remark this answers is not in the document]
pmtk: Configuring here means installing drivers and device plugins.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+For the list of supported Serving Runtimes refer to
+[RHOAI documentation](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#supported-model-serving-runtimes_serving-large-models).
+Keep in mind that in MicroShift they are created directly as ServingRuntimes in `redhat-ods-applications`, not as templates,
+so they need to be copied to workload's namespace.
+
+## General usage overview
+
+1. Prepare model for serving
+1. Configure OS and MicroShift for the hardware - driver & device plugin
+1. Install AI Model Serving for MicroShift
```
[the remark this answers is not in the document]
agullon: Should we mention here to install both RPMs?
 - `microshift-ai-model-serving`
 - `microshift-ai-model-serving-release-info`

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+### Installing AI Model Serving for MicroShift
+
+`microshift-ai-model-serving` RPM contains manifests that deploy `kserve` and `ServingRuntimes` in `redhat-ods-applications` namespace.
+
+To install AI Model Serving for MicroShift run following command:
+```bash
+sudo dnf install microshift-ai-model-serving
+```
+
```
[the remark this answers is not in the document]
pmtk: How would that look like?

BTW. Not sure if we need to apply the same level of hand-holding in upstream docs.
I'm kind of assuming (perhaps wrongly) that upstream docs are more for tinkerers and perhaps dev reference.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+After installing the package (and restarting MicroShift),
+there should be new Pod running in `redhat-ods-applications` namespace:
+```sh
+$ oc get pods -n redhat-ods-applications
+NAME                                        READY   STATUS    RESTARTS   AGE
+kserve-controller-manager-7fc9fc688-kttmm   1/1     Running   0          1h
+```
+
+You can also install the release info package. It contains JSON file with image references
```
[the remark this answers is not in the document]
pmtk: It all depends.
User could have some build pipeline that prepares RPM with manifests that is installed in the ostree/bootc image. So the release info package is installed in CI when they prepare the manifest and they don't need it in the bootc image.
They could also do what we do - assemble the ServingRuntime during container image build ("deploying serving runtime to workloads namespace" > "automatically..." section).

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+It uses OpenVino Model Server (OVMS) and resnet50 model.
+OVMS can run on CPU, no hardware accelerator is needed for this example.
+
+### Installing AI Model Serving for MicroShift
+
+`microshift-ai-model-serving` RPM contains manifests that deploy `kserve` and `ServingRuntimes` in `redhat-ods-applications` namespace.
+
+To install AI Model Serving for MicroShift run following command:
+```bash
+sudo dnf install microshift-ai-model-serving
```
agullon, line 93: Should we mention here to install both RPMs?
 - `microshift-ai-model-serving`
 - `microshift-ai-model-serving-release-info`

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+useful for offline procedures or deploying copy of a ServingRuntime to your namespace during
+bootc image build:
+```bash
+sudo dnf install microshift-ai-model-serving-release-info
+```
+
+### Package model into an OCI image (ModelCar)
+
+You can package your model into an OCI image and make use of what is known as [ModelCar](https://kserve.github.io/website/latest/modelserving/storage/oci/) approach.
+This can help you set up offline environments because the model can be embedded just like any other container image.
```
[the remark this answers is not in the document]
pmtk: I guess it depends, if user already has their model in S3 then it might be easier to continue using that.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```sh
+podman build -t IMAGE_REF .
+podman push IMAGE_REF
+```
+
+For purposes of this documentation, we'll build it locally and use without pushing to registry first.
+`sudo` is required to make it part of the root's container storage and usable by MicroShift
+as CRI-O and Podman share the storage.
+
+For offline use cases make sure to include a tag other than `latest`.
```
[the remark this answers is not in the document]
pmtk:
```
sudo podman build -t ovms-resnet50:test .
                                   ----
```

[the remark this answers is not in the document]
pmtk: If it's unspecified like `-t ovms-resnet50` then it defaults to `latest`

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```sh
+sudo podman images | grep ovms-resnet50
+```
+
+Example output:
+```
+localhost/ovms-resnet50                                     test          ac4606eb6cb3  3 minutes ago  27.5 MB
+```
+
+### Create namespace
```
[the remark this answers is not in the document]
pmtk: It's the first place where we actually interact with the cluster. Previous steps don't need it.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+For more information about ServingRuntimes refer to
+[RHOAI documentation](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#model-serving-runtimes_serving-large-models)
+
+#### Manually - from running cluster
+
+You can use following command to save ServingRuntime definition to a yaml file:
+```sh
+oc get -n redhat-ods-applications servingruntimes kserve-ovms -o yaml > ovms.yaml
+```
+Then, you need to remove following metadata:
```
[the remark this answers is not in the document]
pmtk: If you forget to remove `namespace` then command `oc create -n NAMESPACE -f ./file` will fail:
```
error: the namespace from the provided object "redhat-ods-applications" does not match the namespace "ai-demo". You must pass '--namespace=redhat-ods-applications' to perform this operation.
```

If you remove `namespace` and leave others... well, it creates the object. So I guess only removing namespace is necessary.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```
+
+#### Automatically based on installed manifests and release info
+
+Overview of the procedure:
+1. Install `microshift-ai-model-serving-release-info` RPM
+1. Extract image reference of particular ServingRuntime from the release info file
+1. Make a copy of chosen ServingRuntime yaml file
+1. Substitute placeholder used for `image:` field with actual image reference
+1. Create the object using file or make it part of manifest (kustomization)
```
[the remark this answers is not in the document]
pmtk: There's more info below.

I'm wondering if we should have this mini TOC? I think they generate questions before reader moves on to remainder of the section.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+the ServingRuntime ends up as part of new manifest, the namespace can set in `kustomization.yaml`:
+```yaml
+namespace: ai-demo
+```
+
+### Create InferenceService
+
+Next, we need to create InferenceService custom resource.
+InferenceService instructs kserve on how to create a Deployment for serving the model.
+Kserve uses ServingRuntime based on `modelFormat` specified in InferenceService.
```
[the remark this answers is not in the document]
pmtk: First, you need a model and it has some kind of format.
You need right ServingRuntime for the format.
You also need to specify the modelFormat in the InferenceService so kserve knows which ServingRuntime to pick.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+It's possible to add extra arguments that will be passed to the model server using `.spec.predictor.model.args`.
```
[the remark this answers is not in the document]
pmtk: Not sure really

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+Below is example of InferenceService with model in `openvino_ir` format.
+It features additional argument `--layout=NHWC:NCHW` to make OVMS accept the
+request data in different format than usual for demonstration purposes.
```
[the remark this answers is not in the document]
pmtk: I believe it depends on the actual model. 

See: https://github.com/openvinotoolkit/model_server/blob/e5bdea160cefd6c7fc122503eefaff1149816192/docs/shape_batch_size_and_layout.md#changing-model-inputoutput-layout

docs/user/ai_model_serving.md
pmtk: I don't think that's a valid markdown?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```
+
+Output:
+```
+inferenceservice.serving.kserve.io/ovms-resnet50 created
+```
+
+Soon, a Deployment and a Pod should appear in that namespace.
+Depending on the size of ServingRuntime's image and size of the ModelCar OCI image,
+it may take a while for the Pod to become ready.
```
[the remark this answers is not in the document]
pmtk: I think couple minutes is good estimate. In the CI, Pod with ~6GiB Model OCI takes ~3 minutes to start.

It's possible to use
```
oc rollout status -n ai-demo deployment ovms-resnet50-predictor
```
to wait for Deployment to be ready

docs/user/ai_model_serving.md
pmtk: I don't think so

docs/user/ai_model_serving.md
pmtk: If you have DNS configured so that `ovms-resnet50-predictor-ai-demo.apps.example.com` points to the MicroShift cluster, the this is enough:
```
curl -i "ovms-resnet50-predictor-ai-demo.apps.example.com/v2/models/ovms-resnet50/ready"
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+First, download an image of a bee from the OpenVino examples:
+```sh
+curl -O https://raw.githubusercontent.com/openvinotoolkit/model_server/main/demos/common/static/images/bee.jpeg
+```
+
+Next, we need to create the request data.
+1. Start with an inference header in JSON format.
+1. Get the size of the header. It needs to be passed to the OVMS later in form of HTTP header.
+1. Append size of the image to the request file. OVMS expects 4 bytes (little endian).
+   Command below uses `xxd` utility which is part of `vim-common` package.
```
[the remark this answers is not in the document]
pmtk: Depends in `vim` is installed. I cannot guarantee it - in bootc images it had to be installed explicitly.

docs/user/ai_model_serving.md
pmtk: Yes, we create an input data and send it to the model server, model server takes a request and uses model to process the requests.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+
+To obtain metrics of the model server simply make a request on `/metrics` endpoint:
+```sh
+curl "${DOMAIN}/metrics" \
+    --connect-to "${DOMAIN}::${IP}:"
+```
+
+Partial example output:
+```
+# HELP ovms_requests_success Number of successful requests to a model or a DAG.
```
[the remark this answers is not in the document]
pmtk: There's a document about that: https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md
I guess they call it text-based format?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,482 @@
...
+```sh
+$ oc expose svc -n ai-demo ovms-resnet50-predictor
+route.route.openshift.io/ovms-resnet50-predictor exposed
+
+$ oc get route -n ai-demo
+NAME                      HOST                                               ADMITTED   SERVICE                   TLS
+ovms-resnet50-predictor   ovms-resnet50-predictor-ai-demo.apps.example.com   True       ovms-resnet50-predictor
+```
+
+### Querying the model server
```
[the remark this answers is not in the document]
pmtk: Yeah, you might be right.
This is much simpler but also raw. It might be better to reuse real openvino procedure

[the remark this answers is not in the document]
agullon: From my point of view, as a tech person, I prefer to see the `curl` commands. To understand better how it's working under the hood.
Because this is just an example guide, I'd say we should only add the `curl` command to the metrics and infer endpoint (happy path). And remove the rest of curl commands (to ready, live, etc) in favor of the python scripts from ovms official repo.
Either way, I don't have a strong opinion. It's up to you.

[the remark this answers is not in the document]
pmtk: I'd rather not defer to OVMS example for the metadata, ready, or metrics, because the endpoints are actually "coming" from kserve.

See [v1](https://kserve.github.io/website/latest/modelserving/data_plane/v1_protocol/) and [v2](https://kserve.github.io/website/latest/modelserving/data_plane/v2_protocol/#open-inference-protocol-v2-inference-protocol) protocols.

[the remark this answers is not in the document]
pmtk: Added a new subsection "Other Inference Protocol endpoints"

docs/user/ai_model_serving.md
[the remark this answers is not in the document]
pmtk: Changed above
```
To obtain metrics of the model server simply make a request on `/metrics` endpoint:
```
to
```
To obtain Prometheus metrics of the model server simply make a request on `/metrics` endpoint:
```

AFAIK OTEL can work with prometheus metrics

docs/user/ai_model_serving.md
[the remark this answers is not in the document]
ShaunaDiaz: @agullon Will we test all of these in the future? Wondering if we consider additional model use Tech Preview. @DanielFroehlich WDYT?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,517 @@
...
+  - For more information about Serving Runtimes refer to [upstream Kserve documentation](https://kserve.github.io/website/latest/modelserving/servingruntimes/)
+- [InferenceService](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#inferenceservice)
+
+## Supported model-serving runtimes
+
+Currently MicroShift supports following model-serving runtimes:
+- OpenVINO Model Server
+- vLLM ServingRuntime for KServe
+
+Following runtimes are included but not supported at the time:
```
ggiguash, line 21: What does it mean? We cannot separate those out from the image? 
The question that requires explaining is: why do we include those if we do not support them?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,517 @@
...
+- vLLM ServingRuntime with Gaudi accelerators support for KServe
+- vLLM ROCm ServingRuntime for KServe
+
+Refer to
+[RHOAI documentation](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.17/html/serving_models/serving-large-models_serving-large-models#supported-model-serving-runtimes_serving-large-models).
+for details about the model-serving runtimes.
+
+## General usage overview
+
+1. Develop, train, test, and prepare model for serving
```
ggiguash, line 34: Should these be links to the appropriate chapters in the doc, or there's no one-to-one mapping?
Also, why do we need Optional sections?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,517 @@
...
+$ oc get pods -n redhat-ods-applications
+NAME                                        READY   STATUS    RESTARTS   AGE
+kserve-controller-manager-7fc9fc688-kttmm   1/1     Running   0          1h
+```
+
+You can also install the release info package. It contains JSON file with image references
+useful for offline procedures or deploying copy of a ServingRuntime to your namespace during
+a bootc image build:
+```bash
+sudo dnf install microshift-ai-model-serving-release-info
```
ggiguash, line 97:
```suggestion
sudo dnf install -y microshift-ai-model-serving-release-info
```

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,517 @@
...
+- `generation`
+- `resourceVersion`
+- `uid`
+
+Re-create the ServingRuntime in a custom namespace:
+```sh
+oc create -n ai-demo -f ./ovms.yaml
+```
+
+#### Create ServingRuntime based on installed manifests and release info
```
ggiguash, line 211: I'm not sure we want this in our docs. Can we stay with the live-cluster-only part?

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,517 @@
...
+Output:
+```
+inferenceservice.serving.kserve.io/ovms-resnet50 created
+```
+
+Soon, a Deployment and a Pod should appear in that namespace.
+Depending on the size of ServingRuntime's image and the size of the ModelCar OCI image,
+it may take a while for the Pod to become ready.
+
+You can check the status of the Deployment with following command:
```
ggiguash, line 293: I think we're too nice here :). These are pure kube commands, so let's try to shorten it so that we only include info relevant to the current topic.

docs/user/ai_model_serving.md
pmtk: For example, that first point doesn't have a section. It's more of a prerequisite.
Also, the list is an high level step by step to give idea about the order of actions, not really a ToC

ggiguash: Yes, optional may be confusing, especially in the overview. We should attempt simplifying as much as we can.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,517 @@
...
+The exact directory structure depends on the model server.
+Below is an example Containerfile with a ResNet-50 model compatible with OpenVino Model Server (OVMS)
+used in the OVMS' examples.
+
+See "How to build a ModelCar container" section of
+[Build and deploy a ModelCar container in OpenShift AI](https://developers.redhat.com/articles/2025/01/30/build-and-deploy-modelcar-container-openshift-ai)
+article for guidance on building OCI image with a model from Hugging Face suitable for an vLLM model server.
+
+```Dockerfile
+FROM quay.io/microshift/busybox:1.37
```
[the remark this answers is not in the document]
ggiguash: Yeah, I'm not sure we can use microshift quay.io repo anywhere, but CI tests.

docs/user/ai_model_serving.md
ggiguash: I'm not sure we need to emphasize what we're testing or not in the docs - it may change quickly.
If models are there, let's just mention them? If we specifically support (not test) selected models, that's important distinction.

pmtk: These are tested and supported by RHOAI, so just like we don't retest every possible feature of other optional components, I think we can say these are supported even without tests (especially for "upstream" doc)

docs/user/ai_model_serving.md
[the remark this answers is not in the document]
agullon: Yes, OTEL collector can expose metrics in a Prometheus format. Manually tried this yesterday.

docs/user/ai_model_serving.md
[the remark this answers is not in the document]
agullon: Great, thanks.

docs/user/ai_model_serving.md
```diff
@@ -0,0 +1,478 @@
...
+For RHEL For Edge and RHEL Image Mode systems, create a new manifest making sure it's applied after `/usr/lib/microshift/manifests.d/001-microshift-ai-model-serving`.
+
+### Limitations
+- AI Model Serving on MicroShift is only available on the x86_64 platform.
+- AI Model Serving on MicroShift supports a very specific subset of the RHOAI Operator components.
+- You must secure the exposed model server's endpoint (e.g. OAUTH2).
+- Not all model servers support IPv6.
+- Only OCI (ModelCar) model delivery system is tested and supported
+
+### Known issues
```
agullon, line 469: @pmtk should we add here OVMS is not supported on IPv6? https://issues.redhat.com/browse/RHOAIENG-21756

agullon: I agree, I missed that comment. Thanks.

PR #4697, opened 2025-03-18
OCPBUGS-51365: Move Multus manifests from RPM to embedded assets & run kustomizer standalone

pmtk:


packaging/microshift/config.yaml
```diff
@@ -521,6 +521,11 @@ network:
     # assumes an empty string to mean the OVN-K should be deployed.
     # Allowed values are: unset or one of ["", "ovnk", "none"]
     cniPlugin: ""
+    multus:
+        # Status controls the deployment of the Multus CNI.
+        # Changing from "Enabled" to "Disabled" will not cause Multus CNI to be deleted.
```
pacevedom, line 526: Just as a note, we need to ensure this is highlighted in the docs, as its the first time that a configuration option is not reversible?

pacevedom (approved):
/lgtm

packaging/microshift/config.yaml
pmtk: I think storage (lvms) also follows that route.
[ref](https://docs.redhat.com/en/documentation/red_hat_build_of_microshift/4.18/html/storage/microshift-storage-plugin-overview#microshift-disabling-uninstalling-lvms-csi-snapshot_microshift-storage-plugin-overview):

> Automated uninstallation is not supported as this can cause orphaning of the provisioned volumes. Without the LVMS CSI driver, the cluster does not have knowledge of the underlying storage interface and cannot perform provisioning and deprovisioning or mounting and unmounting operations

pmtk: Also, microshift-multus.rpm will continue existing, except it'll just contain the drop in, so from user's perspective it's the same (they could remove the rpm with the manifests, but the objects in api server would continue living).

PR #4703, opened 2025-03-19
USHIFT-4882: Bootc published image standard tests can now use public registry when available

ggiguash:


test/scenarios-bootc/periodics/el96-crel@published-images-standard2.sh
```diff
@@ -9,26 +9,24 @@ IMAGE_SIGSTORE_ENABLED=true
 
 scenario_create_vms() {
     if [[ "${CURRENT_RELEASE_REPO}" == "" ]] ; then
-        # TODO: While 4.19-ec is not available, it needs to exit without an error.
+        # Empty string means there's no EC build yet, so the test needs to be skipped.
```
pmtk, line 12: I think that eventually it would be nice to have some report of skipped tests so they're not hidden away.

PR #4726, opened 2025-03-24
USHIFT-5531: Auto-rebase procedure with flannel and kube-proxy image update

praveenkumar:

Closes #<Issue Number>

### DIFF
--- a/scripts/auto-rebase/rebase.sh
+++ b/scripts/auto-rebase/rebase.sh
@@ -816,6 +816,7 @@ EOF
 
     update_olm_images
     update_multus_images
+    update_kubeproxy_images
 
     popd >/dev/null
 }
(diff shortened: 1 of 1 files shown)


scripts/auto-rebase/rebase.sh
```diff
@@ -1118,6 +1119,44 @@ EOF
     done  # for goarch
 }
 
+update_kubeproxy_images() {
```
ggiguash, line 1122: If kubeproxy is only used with flannel, should we move this code to the dedicated script too?

praveenkumar: I didn't move this in separate script because kube-proxy is core component of release artifacts so doesn't make sense to pull that artifacts again and parse it. It is optional like olm but part of release artifacts so better to be here ?

ggiguash: OK, if there's an analogy with other existing components - that's fine.

praveenkumar:
> /test ?

@ggiguash Do we need any other test than default running ones?

praveenkumar:
> > > /test ?
> > 
> > 
> > @ggiguash Do we need any other test than default running ones?
> 
> No, the most important one is `test-rebase`. Can you check [this log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_microshift/4726/pull-ci-openshift-microshift-main-test-rebase/1904916886445887488/artifacts/test-rebase/openshift-microshift-rebase/build-log.txt) output?

Yes I was checking that and looks like still ignored so may be https://github.com/openshift/microshift/blob/main/scripts/auto-rebase/assets.yaml#L268-L296 also need to updated?

scripts/auto-rebase/rebase.py
```diff
@@ -102,6 +102,23 @@ def run_rebase_ai_model_serving_sh(release):
     return RebaseScriptResult(success=result.returncode == 0, output=result.stdout)
 
 
+def run_rebase_flannel():
```
pmtk, line 105: Should this function be executed?
I recall that newest flannel requires a kernel module, should we even proceed with this PR?

scripts/auto-rebase/rebase.sh
pmtk, line 1122: Do you know what's the difference between `kube-proxy` and `kube-rbac-proxy`? Can they be used interchangeably?

scripts/auto-rebase/rebase.py
praveenkumar: @pmtk I just removed that commit, so whatever we choose for CNI which doesn't require any kernel module then will check how that component need to be rebased.

PR #4735, opened 2025-03-26
NO-ISSUE: Revert "Update flannel to v0.26.5 and plugin to v1.6.2-flannel1"

praveenkumar:
This reverts commit 177391dda436e7cc82adfc146afcc4d37e7a95de.

So looks like flannel now require br_netfilter module to start as per https://github.com/flannel-io/flannel?tab=readme-ov-file#deploying-flannel-with-helm because https://github.com/kubernetes/kubernetes/pull/123464, which means it would problem to run microshift in container. I might going to check much simpler CNI like what kind use https://github.com/kubernetes-sigs/kind/tree/main/images/kindnetd

I am reverting this to make sure main branch have working flannel plugin.

### DIFF
--- a/assets/optional/flannel/kustomization.aarch64.yaml
+++ b/assets/optional/flannel/kustomization.aarch64.yaml
@@ -2,7 +2,7 @@
 images:
   - name: flannel
     newName: docker.io/flannel/flannel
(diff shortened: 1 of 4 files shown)


praveenkumar:
/hold

I am testing this again on the container side and will remove hold once tested successfully

praveenkumar:
/unhold

tested and it works again with older version of flannel.

PR #4748, opened 2025-03-31
USHIFT-5541: Increase timeout for CatalogSource & add extra logging

pmtk:


test/bin/scenario.sh
```diff
@@ -553,7 +553,7 @@ launch_vm() {
     local network="default"
     local vm_memory=4096
     local vm_vcpus=2
-    local vm_disksize=20
+    local vm_disksize=25
```
ggiguash, line 556: I think we should not increase the total disk size because we should be OK with 5GB leftover for unallocated partition for LVMS.
Can we update the test/kickstart-templates/includes/main-prologue.cfg comment with the new partitioning? We can also consider removing the example if that would better support future changes.

pmtk:
/override ci/prow/e2e-aws-tests-bootc-periodic ci/prow/e2e-aws-tests-bootc-periodic-arm  ci/prow/e2e-aws-tests-periodic

PR #4751, opened 2025-04-01
OCPBUGS-54239: Add hostname package dependency to microshift RPM

ggiguash:


ggiguash:
The failures of GatewayAPI and OLM tests are not related to this change. 
Overriding to save CI cycles.
/override ci/prow/e2e-aws-tests-bootc-arm

PR #4754, opened 2025-04-01
USHIFT-5519: Add microshift_version to telemetry

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/telemetry/gather.go
+++ b/pkg/telemetry/gather.go
@@ -16,6 +16,7 @@ limitations under the License.
 package telemetry
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/tls"
@@ -24,10 +25,13 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
 
 	routev1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"github.com/openshift/microshift/pkg/config"
+	"github.com/openshift/microshift/pkg/util"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
 	io_prometheus_client "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
@@ -42,6 +46,10 @@ const (
 	kubeletMetricsCAdvisor = "https://%s:10250/metrics/cadvisor"
 	kubeletMetricsResource = "https://%s:10250/metrics/resource"
 	kubeletMetrics         = "https://%s:10250/metrics"
+
+	DeploymentTypeBootC  = "bootc"
+	DeploymentTypeOSTree = "ostree"
+	DeploymentTypeRPM    = "rpm"
 )
 
 func makeHTTPClient() (*http.Client, error) {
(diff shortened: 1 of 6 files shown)


pkg/telemetry/gather.go
```diff
@@ -233,3 +240,38 @@ func fetchKubernetesResources(ctx context.Context, cfg *config.Config) (map[stri
 	}
 	return metrics, nil
 }
+func getOSVersion() (string, error) {
```
ggiguash, line 243: Is this code more "appropriate" for the util package?

pkg/telemetry/gather.go
```diff
@@ -233,3 +240,38 @@ func fetchKubernetesResources(ctx context.Context, cfg *config.Config) (map[stri
...
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("error reading /etc/os-release: %w", err)
+	}
+	if version == "" {
+		return "", fmt.Errorf("VERSION_ID not found in /etc/os-release")
+	}
+	return version, nil
+}
+
+func getDeploymentType() string {
```
ggiguash, line 269: Same here: should we move this to utils? Others might benefit from this function.

pkg/telemetry/gather.go
pacevedom: Makes sense, will move it.

pkg/telemetry/gather.go
pacevedom: This was intentionally left out of `util` because of the constants for the metrics. It is true though that they are generic enough to be in a common place. The functions called in the body are in `util` already though.

pkg/telemetry/gather.go
```diff
@@ -233,3 +240,38 @@ func fetchKubernetesResources(ctx context.Context, cfg *config.Config) (map[stri
...
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "VERSION_ID=") {
+			version = strings.TrimPrefix(line, "VERSION_ID=")
+			version = strings.Trim(version, `"`)
+			break
+		}
+	}
+	if err := scanner.Err(); err != nil {
```
pmtk, line 260: What if by that point we already have `version`? I.e. error occurs after VERSION_ID=. Not sure if that even possible without some solar storms

pkg/telemetry/telemetry.go
```diff
@@ -271,6 +281,24 @@ func computeUsageMetrics(kubeletMetrics map[string]*io_prometheus_client.MetricF
...
+func computeMicroShiftVersionMetric() (Metric, error) {
+	osVersion, err := getOSVersion()
+	if err != nil {
+		return Metric{}, fmt.Errorf("failed to get OS version: %v", err)
+	}
+	currentTimestamp := time.Now().UnixNano() / time.Millisecond.Nanoseconds()
+	return Metric{
+		Name: MetricNameMicroShiftVersion,
+		Labels: []MetricLabel{
+			{Name: LabelNameVersion, Value: prerun.GetVersionStringOfData()},
```
pmtk, line 293: Is this the right function call? `GetVersionStringOfData()` returns `string(version)` contained in `/var/lib/microshift/version`. 
Shouldn't we use the version of the executable?

pkg/util/util.go
```diff
@@ -164,3 +166,32 @@ func CreateTempDir(path string) (string, error) {
 	dir, pattern := getTempPathArgs(path)
 	return os.MkdirTemp(dir, pattern)
 }
+
+func IsOSTree() bool {
+	exists, err := PathExists("/run/ostree-booted")
```
pmtk, line 171: We have couple places in `pkg/admin/{prerun,autorecovery}/` that do the same thing, except they honor the `err`. Should we unify this a bit more?

pkg/util/util.go
```diff
@@ -164,3 +166,32 @@ func CreateTempDir(path string) (string, error) {
...
+	if err != nil {
+		return false
+	}
+	return exists
+}
+
+func IsBootC() bool {
```
pmtk, line 178: They're using `bootc`, not `BootC` or `bootC` (https://bootc-dev.github.io/bootc/), so I think this is more accurate 
```suggestion
func IsBootc() bool {
```

PR #4782, opened 2025-04-09
USHIFT-5568: Split AI Model Serving manifest

pmtk:
Split AIMS manifest into two parts to enable user to add a manifest in between that overrides the kserve settings.


agullon:
it looks good to me

PR #4791, opened 2025-04-11
USHIFT-5580: Use CentOS base image for RHEL 9.6 ISO creation until GA

ggiguash:
Note that [USHIFT-5584](https://issues.redhat.com/browse/USHIFT-5584) issue was opened to revert the current workaround when RHEL 9.6 is in GA.


pacevedom (approved):
/lgtm

PR #4813, opened 2025-04-17
USHIFT-5528: Add new option to scenario.sh to target any given VM

kasturinarra:

This PR address to add a new option to scenario.sh to target any given VM.

For example: ./test/bin/scenario.sh run ./test/scenarios/periodics/el94-src@rpm-standard1.sh <vm_ip>

Closes https://issues.redhat.com/browse/USHIFT-5528

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -33,6 +33,7 @@ VNC_CONSOLE=${VNC_CONSOLE:-false}  # may be overridden in global settings file
 TEST_RANDOMIZATION="all"  # may be overridden in scenario file
 TEST_EXECUTION_TIMEOUT="30m" # may be overriden in scenario file
 SUBSCRIPTION_MANAGER_PLUGIN="${SUBSCRIPTION_MANAGER_PLUGIN:-${SCRIPTDIR}/subscription_manager_register.sh}"  # may be overridden in global settings file
+RUN_HOST_OVERRIDE=""  # target any given VM for running scenarios
 
 declare -i TESTCASES=0
 declare -i FAILURES=0
@@ -398,21 +399,25 @@ function get_vm_ip {
     local -r vmname="${1}"
     local -r start=$(date +%s)
     local ip
-    ip=$("${ROOTDIR}/scripts/devenv-builder/manage-vm.sh" ip -n "${vmname}" | head -1)
-    while true; do
-        now=$(date +%s)
-        if [ $(( now - start )) -ge ${VM_BOOT_TIMEOUT} ]; then
-            echo "Timed out while waiting for IP retrieval"
-            exit 1
-        fi
-        sleep 1
-        # Try pinging the IP address to avoid stale DHCP leases that would falsely
-        # return as the current IP for the VM.
-        ip=$("${ROOTDIR}/scripts/devenv-builder/manage-vm.sh" ip -n "${vmname}" | head -1)
-        if ping -c 1 -W 1 "${ip}" &> /dev/null; then
-          break
-        fi
-    done
+    if [[ "${vmname}" =~ ([0-9]{1,3}\.){3}[0-9]{1,3} ]]; then
+            ip="${BASH_REMATCH[0]}"
+    else
+       ip=$("${ROOTDIR}/scripts/devenv-builder/manage-vm.sh" ip -n "${vmname}" | head -1)
+       while true; do
+           now=$(date +%s)
+           if [ $(( now - start )) -ge ${VM_BOOT_TIMEOUT} ]; then
+               echo "Timed out while waiting for IP retrieval"
+               exit 1
+           fi
+           sleep 1
+           # Try pinging the IP address to avoid stale DHCP leases that would falsely
+           # return as the current IP for the VM.
+           ip=$("${ROOTDIR}/scripts/devenv-builder/manage-vm.sh" ip -n "${vmname}" | head -1)
+           if ping -c 1 -W 1 "${ip}" &> /dev/null; then
+             break
+           fi
+       done
+    fi
     echo "${ip}"
 }
 
@@ -933,8 +938,19 @@ stress_testing() {
 
 # Run the tests for the current scenario
 run_tests() {
-    local -r vmname="${1}"
-    local -r full_vmname="$(full_vm_name "${vmname}")"
+    local vmname="${1}"
+    local full_vmname
+    full_vmname="$(full_vm_name "${vmname}")"
+    if [[ -n "${RUN_HOST_OVERRIDE}" ]]; then
+	vmname="${RUN_HOST_OVERRIDE}"
+	full_vmname="$(full_vm_name "${vmname}")"
+        ip=$(get_vm_ip "${full_vmname}")
+	set_vm_property "${vmname}" "ip" "${ip}"
+        set_vm_property "${vmname}" "ssh_port" "22"
+        set_vm_property "${vmname}" "api_port" "6443"
+        set_vm_property "${vmname}" "lb_port" "5678"
+    fi
+
     shift
 
     echo "Running tests with $# args" "$@"
@@ -994,7 +1010,7 @@ VM_IP: ${vm_ip}
 API_PORT: ${api_port}
 LB_PORT: ${lb_port}
 USHIFT_HOST: ${vm_ip}
-USHIFT_USER: redhat
+USHIFT_USER: "${USHIFT_USER:-redhat}"
 SSH_PRIV_KEY: "${SSH_PRIVATE_KEY:-}"
 SSH_PORT: ${ssh_port}
 EOF
@@ -1187,7 +1203,12 @@ action_run() {
         sos_report true || rc=1 ; \
         close_junit ; exit "${rc}"' EXIT
 
-    check_dependencies
+    if [ $# -eq 0 ]; then
+        RUN_HOST_OVERRIDE=""
+	check_dependencies
+    else
+        RUN_HOST_OVERRIDE="$1"
+    fi
 
     scenario_run_tests
     record_junit "run" "scenario_run_tests" "OK"
@@ -1216,6 +1237,9 @@ Settings
 Login
 
   scenario.sh login <scenario-script> [<host>]
+Run
+
+  scenario.sh run <scenario-script> [<host | ip>]
 EOF
 }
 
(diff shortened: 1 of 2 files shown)


kasturinarra:
@agullon @jogeo could you please help review the PR ? thanks !!

agullon:
> @agullon @jogeo could you please help review the PR ? thanks !!

I can see you are duplicating code. For example, to create `variables.yaml` file and execute RF tests. Can you reuse existing code, please?

test/bin/scenario.sh
```diff
@@ -1096,6 +1063,74 @@ check_dependencies() {
     fi
 }
 
+create_input_variables_file() {
+    local -r vmname="$1"
+    local variable_file="${SCENARIO_INFO_DIR}/${SCENARIO}/variables.yaml"
+    echo "[INFO] Writing variables to ${variable_file}" >&2
+    mkdir -p "$(dirname "${variable_file}")"
+
+    cat - <<EOF > "${variable_file}"
```
ggiguash, line 1072:
```suggestion
    cat > "${variable_file}" <<EOF 
```

test/bin/scenario.sh
```diff
@@ -1096,6 +1063,74 @@ check_dependencies() {
...
+VM_IP: ${vm_ip}
+API_PORT: ${api_port}
+LB_PORT: ${lb_port}
+USHIFT_HOST: ${vm_ip}
+USHIFT_USER: redhat
+SSH_PRIV_KEY: "${SSH_PRIVATE_KEY:-}"
+SSH_PORT: ${ssh_port}
+EOF
+    echo "${variable_file}"  # return path
```
ggiguash, line 1081: This would produce an extra message in the regular flow. 
Should we move the variable to global scope so that it can be accessed outside of the function?

test/bin/scenario.sh
```diff
@@ -1192,9 +1227,64 @@ action_run() {
     record_junit "run" "scenario_run_tests" "OK"
 }
 
+action_run_on_vm() {
```
ggiguash, line 1230: I wonder if it's possible to avoid this duplication of the run logic for this specific case. 
If we only need to override the VM IP, can we provide an optional argument (e.g. `--ip_override 1.2.3.4`) and set it un the current `run_tests` function if specified.

test/bin/scenario.sh
```diff
@@ -1223,6 +1245,9 @@ if [ $# -lt 2 ]; then
     exit 1
 fi
 
+# Change added for USHIFT-5528
```
ggiguash, line 1248: When adding comment, it is often preferrable to explain in place rather than refer to external sources.

test/bin/scenario.sh
```diff
@@ -1223,6 +1245,9 @@ if [ $# -lt 2 ]; then
...
+SCENARIO_ARGS=("$@")
```
ggiguash, line 1249: Looking at the code, action_run already gets the command line, so we should check and assign that variable in `action_run`

ggiguash: Also, let's use a concrete variable name rather than saving the whole command line. 
Something like RUN_HOST_OVERRIDE, which can be defined as empty in the beginning of the script and reset of necessary in `action_run`

test/bin/scenario.sh
```diff
@@ -1188,6 +1202,13 @@ action_run() {
 
     check_dependencies
 
+    for hname in "$@"; do
+	    if [[ ! "${hname}" == "host1" ]]; then
```
ggiguash, line 1206: We cannot assume `host1` by default. There are tests that are using different host names.

I would simply check the number of the command line arguments. If there's 3 - assume the third one is the override host name or IP.

test/bin/scenario.sh
```diff
@@ -224,7 +224,8 @@ sos_report_for_vm() {
 }
 
 invoke_qemu_script() {
-    "${ROOTDIR}/_output/robotenv/bin/python" "${ROOTDIR}/test/resources/qemu-guest-agent.py" "$@"
+    timeout --verbose --foreground 2m \
```
ggiguash, line 227: This does not seem to be related to the current PR.
Please, rebase the code.

kasturinarra: sorry, my bad !! Let me do that.

test/bin/scenario.sh
kasturinarra: when i try to print $@ it just shows only 1 argument which is either host1 or the ip, am i missing something here ?

ggiguash: It depends on how you run the scenario.sh script. We're always removing action and script arguments [here](https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L1227-L1230).

Next, we're passing the remaining command line to `action_xyz` handers. 

So, if you run `scenario.sh run myscenario.sh my_ip` command, the `my_ip` will be passed to `action_run`.

ggiguash: Most of the scenario scripts use host1, but look at [this one](https://github.com/openshift/microshift/blob/main/test/scenarios-bootc/periodics/el96-src%40cncf-conformance.sh#L172-L173) for example - it creates more than one host.

kasturinarra: this is how i am trying to run `scenario.sh run myscenario.sh my_ip` and to handle the other case that you are talking about here would using a regex instead of `host1` work ?

ggiguash: It's hard for me to comment without seeing the new code. Maybe, you can upload changes and I will take another look?

kasturinarra: sure, will add in sometime, thanks !!

test/bin/scenario.sh
```diff
@@ -1224,6 +1247,9 @@ if [ $# -lt 2 ]; then
     exit 1
 fi
 
+# Change added to target any given VM for running scenario.sh
```
ggiguash, line 1250: Please, move the definition to the top of the file, together with others.

```suggestion
# Target any given VM for running scenarios
```

test/bin/scenario.sh
```diff
@@ -1189,6 +1202,13 @@ action_run() {
 
     check_dependencies
 
+    for hname in "$@"; do
+	    if [[ ! "$hname" =~ ^host[0-9]+$ ]]; then
+	       RUN_HOST_OVERRIDE=( "$@" )
+	       break
+	    fi
+    done
```
ggiguash, line 1210: I'm not sure I understand why we need this check. This function is getting argv from the command line, so it's either empty or contains a single host or IP.

ggiguash: Can't we just check if $1 exists and not empty and assign it?

kasturinarra: one question, somewhere in either action_run or run_tests i need to check if  the new variable matches host1 or not right ? if not host1, i will need to run set_vm_property to populate the variables.yaml file right ?

ggiguash: I'm not sure host name validation is necessary.
What we need to do is to override the location where **real** VM host IP is derived. 

Could it be that it's sufficient to do this in the get_vm_ip function?

kasturinarra:
@ggiguash could you please help give another round of review ? This time i see all other tests have passed except for the ones that are failing above and i see that these failed with error Some tests or VM boots failed

test/bin/scenario.sh
```diff
@@ -33,6 +33,7 @@ VNC_CONSOLE=${VNC_CONSOLE:-false}  # may be overridden in global settings file
 TEST_RANDOMIZATION="all"  # may be overridden in scenario file
 TEST_EXECUTION_TIMEOUT="30m" # may be overriden in scenario file
 SUBSCRIPTION_MANAGER_PLUGIN="${SUBSCRIPTION_MANAGER_PLUGIN:-${SCRIPTDIR}/subscription_manager_register.sh}"  # may be overridden in global settings file
+RUN_HOST_OVERRIDE=""  #Target any given VM for running scenarios
```
ggiguash, line 36: Nit: consistency with the comment above
```suggestion
RUN_HOST_OVERRIDE=""  # target any given VM for running scenarios
```

test/bin/scenario.sh
```diff
@@ -1189,6 +1204,12 @@ action_run() {
 
     check_dependencies
 
+    if [ $# -eq 0 ]; then
+        RUN_HOST_OVERRIDE="host1"
```
ggiguash, line 1208: We should not be setting artificial values in this variable. 
Let's keep it empty when not specified?

test/bin/scenario.sh
```diff
@@ -933,8 +938,18 @@ stress_testing() {
 
 # Run the tests for the current scenario
 run_tests() {
-    local -r vmname="${1}"
-    local -r full_vmname="$(full_vm_name "${vmname}")"
+    local vmname="${1}"
+    local full_vmname
```
ggiguash, line 942: Related to my comment on not setting `host1` default for the override, we should initialize `full_vmname` like it was before and override when necessary.

test/bin/scenario.sh
kasturinarra: Done !!

test/bin/scenario.sh
kasturinarra: Done.

test/bin/scenario.sh
kasturinarra: Done.

kasturinarra:
@agullon could you please help review as well ? thanks !!

test/bin/scenario.sh
```diff
@@ -1189,6 +1205,12 @@ action_run() {
 
     check_dependencies
 
+    if [ $# -eq 0 ]; then
+        RUN_HOST_OVERRIDE=""
```
agullon, line 1209:
```suggestion

    if [ $# -eq 0 ]; then
        RUN_HOST_OVERRIDE=""
        check_dependencies
```

Move check_suggestion here to avoid starting the ngingx server when we only want to run RF tests

agullon:
> @agullon could you please help review as well ? thanks !!

I try to run it from QE VMs and find out we need to set `USHIFT_USER=microshift` to make it work.

To add this change, I suggest you take the same approach used for `SSH_PRIV_KEY`, so you'll need to do 2 small changes:

- https://github.com/openshift/microshift/pull/4813/files#diff-fb1c182cf6f43fe577ebbbbb69aa2ca710cb8832142f53ed5c2c21caa836a253R1013-L998
- Add `USHIFT_USER` here: https://github.com/openshift/microshift/blob/main/test/scenario_settings.sh.example

keep redhat user as default user value

PR #4823, opened 2025-04-23
Configure internal RHEL 9.6 repos for osbuild.

jogeo:
**Which issue(s) this PR addresses**: [USHIFT-5575](https://issues.redhat.com//browse/USHIFT-5575)

This PR checks to see if an internal rpm repo is configured for the host and if so configures osbuild repos as well.

### DIFF
--- a/scripts/devenv-builder/configure-composer.sh
+++ b/scripts/devenv-builder/configure-composer.sh
@@ -23,7 +23,7 @@ install_and_configure_composer() {
     # Necessary for embedding container images
     if [ ! -e /etc/osbuild-worker/pull-secret.json ] ; then
         sudo mkdir -p /etc/osbuild-worker
-        sudo ln -s /etc/crio/openshift-pull-secret /etc/osbuild-worker/pull-secret.json
+        sudo ln -sf /etc/crio/openshift-pull-secret /etc/osbuild-worker/pull-secret.json
         sudo tee /etc/osbuild-worker/osbuild-worker.toml &>/dev/null <<EOF
 [containers]
 auth_file_path = "/etc/osbuild-worker/pull-secret.json"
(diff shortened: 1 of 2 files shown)


scripts/devenv-builder/configure-composer.sh
```diff
@@ -95,6 +95,25 @@ enable_beta_or_eus_repositories() {
     # The configuration will remain unchanged for non-beta and non-EUS operating systems.
     if grep -qE "Red Hat Enterprise Linux.*Beta" /etc/redhat-release; then
         sudo sed -i "s,dist/rhel${version_id_major}/${version_id},beta/rhel${version_id_major}/${version_id_major},g" "${composer_config}"
+    elif dnf repolist | grep -q download.eng.brq.redhat.com; then
```
ggiguash, line 98: I'm not sure this should be an `elif`. Should we better put it as its own clause?

jogeo: I made it an elif because I think we would only want the composer system source to use either a Beta url or this internal download.eng.brq.redhat.com url but not both. It's also fine to put it in it's own clause, since composer_config is being over written for this special case.

scripts/devenv-builder/configure-composer.sh
```diff
@@ -99,6 +99,36 @@ enable_beta_or_eus_repositories() {
...
...D6RLfi59B9lmAbekecnsMIFMx7qR7ZKyQb3GOuZwNYOaYFevuxusSwCHv\n4FtLDIhk+Fge+EbPdEva+VLJeMOb02gC4V/cX/oFoPkxM1A5LHjkuAM+aFLAiIRd\nNp/tAPWk1k6yc+FqkcDqOttbP4ciiXb9JPtmzTCbJD8lgH0rGp8ufyMXC9x7/dqX\nTjsiGzyvlMnrkKB4GL4DqRFl8LAR02A3846DD8CAcaxoXggL2bJCU2rgUQARAQAB\ntDVSZWQgSGF0LCBJbmMuIChhdXhpbGlhcnkga2V5IDMpIDxzZWN1cml0eUByZWRo\nYXQuY29tPokCUgQTAQgAPBYhBH5GJCWMQGU11W1vE1BU5KRaY0CzBQJiKSKeAhsD\nBQsJCAcCAyICAQYVCgkICwIEFgIDAQIeBwIXgAAKCRBQVOSkWmNAsyBfEACuTN/X\nYR+QyzeRw0pXcTvMqzNE4DKKr97hSQEwZH1/v1PEPs5O3psuVUm2iam7bqYwG+ry\nEskAgMHi8AJmY0lioQD5/LTSLTrM8UyQnU3g17DHau1NHIFTGyaW4a7xviU4C2+k\nc6X0u1CPHI1U4Q8prpNcfLsldaNYlsVZtUtYSHKPAUcswXWliW7QYjZ5tMSbu8jR\nOMOc3mZuf0fcVFNu8+XSpN7qLhRNcPv+FCNmk/wkaQfH4Pv+jVsOgHqkV3aLqJeN\nkNUnpyEKYkNqo7mNfNVWOcl+Z1KKKwSkIi3vg8maC7rODsy6IX+Y96M93sqYDQom\naaWue2gvw6thEoH4SaCrCL78mj2YFpeg1Oew4QwVcBnt68KOPfL9YyoOicNs4Vuu\nfb/vjU2ONPZAeepIKA8QxCETiryCcP43daqThvIgdbUIiWne3gae6eSj0EuUPoYe\nH5g2Lw0qdwbHIOxqp2kvN96Ii7s1DK3VyhMt/GSPCxRnDRJ8oQKJ2W/I1IT5VtiU\nzMjjq5JcYzRPzHDxfVzT9CLeU/0XQ+2OOUAiZKZ0dzSyyVn8xbpviT7iadvjlQX3\nCINaPB+d2Kxa6uFWh+ZYOLLAgZ9B8NKutUHpXN66YSfe79xFBSFWKkJ8cSIMk13/\nIfs7ApKlKCCRDpwoDqx/sjIaj1cpOfLHYjnefg==\n=UZd/\n-----END PGP PUBLIC KEY BLOCK-----",
```
agullon, line 110: I'd suggest setting this key in a var and use the var instead of duplicating the key value three times.

jogeo: Replaced the static gpgkey value with a jq query to pull the value from /usr/share/osbuild-composer/repositories/rhel-9.6.json. Verified the key used for the internal rpm repos is the same as the one being installed by the rpm. This can be expected since the ISO used to install the host also came from the same internal location.

scripts/devenv-builder/configure-composer.sh
```diff
@@ -99,6 +99,38 @@ enable_beta_or_eus_repositories() {
         sudo sed -i "s,dist/rhel${version_id_major}/${version_id}/$(uname -m)/baseos/,${version_id_eus}/rhel${version_id_major}/${version_id}/$(uname -m)/baseos/,g" "${composer_config}"
         sudo sed -i "s,dist/rhel${version_id_major}/${version_id}/$(uname -m)/appstream/,${version_id_eus}/rhel${version_id_major}/${version_id}/$(uname -m)/appstream/,g" "${composer_config}"
     fi
+    # If the host OS is configured to use the internal repo, overwrite the composer_config to match
```
ggiguash, line 102: Nit
```suggestion
    # If the host OS is configured to use the internal repo, overwrite the composer configuration
```

scripts/devenv-builder/configure-composer.sh
```diff
@@ -99,6 +99,38 @@ enable_beta_or_eus_repositories() {
...
+    if dnf repolist | grep -q download.eng.brq.redhat.com; then
+        # The gpgkey from /usr/share/osbuild-composer/repositories is valid and common for all repos
+        local -r GPGKEY=$(ARCH=$(uname -m) jq '.[env.ARCH][] | select(.name=="baseos") | .gpgkey' /usr/share/osbuild-composer/repositories/rhel-9.6.json)
```
ggiguash, line 105: The file version is hard-coded in the json file name. Should we use variables?

Also a nit: GPGKEY is a local variable, so we use lower case.

jogeo: updated.

PR #4835, opened 2025-04-24
Update go mod vendor

ashwindasr:
Fix go mod vendor discrepancies for Konflux


ggiguash:
To properly address the problem, we need a fix in the rebase procedure.
It will be worked on at https://issues.redhat.com/browse/USHIFT-5658

/close

PR #4884, opened 2025-05-08
OCPBUGS-55955: Include OpenTelemetry configuration examples in the RPM

pmtk:


pmtk:
/hold


it might require a paths change in the tests

PR #4889, opened 2025-05-09
USHIFT-5660: Ingress access logging implementation and tests

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/assets/components/openshift-router/deployment.yaml
+++ b/assets/components/openshift-router/deployment.yaml
@@ -110,19 +110,24 @@ spec:
             {{- if .HttpLogFormat }}
             - name: ROUTER_SYSLOG_FORMAT
               value: '{{ .HttpLogFormat }}'
+            {{- /* HttpLogFormat */}}
             {{- end }}
             {{- if .HttpCaptureHeadersRequest }}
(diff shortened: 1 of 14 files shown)


assets/components/openshift-router/deployment.yaml
```diff
@@ -102,6 +102,34 @@ spec:
...
+            - name: ROUTER_CAPTURE_HTTP_COOKIE
+              value: '{{ .HttpCaptureCookies }}'
+            {{- end }}
+            {{- end }}
+            {{- if .HttpErrorCodePages }}
+            - name: ROUTER_ERRORFILE_503
+              value: /var/lib/haproxy/errorfiles/error-page-503.http
+            - name: ROUTER_ERRORFILE_404
+              value: /var/lib/haproxy/errorfiles/error-page-404.http
+            {{- end }}
```
pmtk, line 132: Should we add a comment to `end`s? Everything at the same indent makes it a bit hard to see what ends where.

```suggestion
            {{- end /* HttpErrorCodePages */ }}
```

pacevedom: Done

PR #4900, opened 2025-05-13
USHIFT-5397: [QE] Replace MicroShift build images phase with dev scripts

agullon:


test/bin/build_rpms.sh
```diff
@@ -119,13 +120,22 @@ download_brew_rpms() {
     mkdir -p "${BREW_RPM_SOURCE}"
     if "${SCRIPTDIR}/manage_brew_rpms.sh" access ; then
```
agullon: @ggiguash  This block is not accessible from CI on cache-nightly runs, [see this log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-microshift-release-4.19-periodics-e2e-aws-tests-cache-nightly/1933320950179172352/artifacts/e2e-aws-tests-cache-nightly/openshift-microshift-infra-iso-build/build-log.txt):
```
+ 00:55:44.906969816 ./bin/build_rpms.sh:130 	echo 'WARNING: The Brew Hub site is not accessible, skipping the download'
```

For that reason, I also implemented this logic in the CI script, see https://github.com/openshift/release/pull/66024 PR

agullon: I suggest we should extract the block to download RPMs from brew into a separate function and use it here and on [release repo](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh#L44-L59) instead of duplicating the code.

ggiguash: You are right. Sorry for misleading you.

ggiguash: There is no need to extract it into separate function. 
Just add the necessary version downloads in the [release repo](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh#L44-L59)

agullon: I'm sorry, I don't understand what you mean by `necessary version downloads`

ggiguash: I meant adding the following commands [here](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh#L56) - pseudo code:
```
 bash -x ./test/bin/manage_brew_rpms.sh download "${ocpversion}" "${out_path}"
 bash -x ./test/bin/manage_brew_rpms.sh download "${ocpversion}" - 1 "${out_path}"
 bash -x ./test/bin/manage_brew_rpms.sh download "${ocpversion}" - 2 "${out_path}"
```

test/bin/manage_brew_rpms.sh
```diff
@@ -6,7 +6,7 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # its execution in a containerized environment with limited set of tools.
 
 usage() {
-    echo "Usage: $(basename "$0") [access | download <version> <path>]"
+    echo "Usage: $(basename "$0") [access | download <version> <path> <version_type>]"
```
ggiguash: Version type seems to be optional.
```suggestion
    echo "Usage: $(basename "$0") [access | download <version> <path> [version_type]]"
```

test/bin/manage_brew_rpms.sh
```diff
@@ -38,10 +39,17 @@ action_download() {
...
+        package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}.0~${ver_type}." | tail -1) || true
+    else
+        package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}.*${ver_type}" | tail -1) || true
+    fi
+
     if [ -z "${package}" ] ; then
-        echo "ERROR: Cannot find MicroShift '${ver}' packages in brew"
-        exit 1
+        echo "WARNING: Cannot find MicroShift '${ver}' packages in brew"
+        exit 0
```
ggiguash: Why do we need this to be a non-fatal exit?

test/bin/manage_brew_rpms.sh
```diff
@@ -51,7 +59,12 @@ action_download() {
     # cannot be identified easily when running in a CI job
     for arch in x86_64 aarch64 ; do
         local adir
-        adir="${dir}/${arch}"
+        # shellcheck disable=SC2001
+        if [ -z "${ver_type}" ] ; then
+            adir="${dir}/${arch}"
+        else
+            adir="${dir}/$(echo "${package}" | sed 's/.*microshift-\([^-]*\).*/\1/')/${arch}"
```
ggiguash: Can we split this expression into another var? 
Please, explain in the comment what we're trying to extract.

test/bin/manage_brew_rpms.sh
```diff
@@ -75,7 +88,7 @@ case "${action}" in
         "action_${action}"
         ;;
     download)
-        [ $# -ne 3 ] && usage && exit 1
+        [ $# -gt 4 ] && usage && exit 1
```
ggiguash:
```suggestion
        [ $# -ne 3 ] && [ $# -ne 4 ] && usage && exit 1
```

test/bin/manage_brew_rpms.sh
```diff
@@ -38,10 +39,17 @@ action_download() {
 
     # Attempt downloading the specified build version
     local package
-    package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}" | sort | tail -1) || true
+    if [ "${ver_type}" = "zstream" ]; then
```
ggiguash: Nit: This might be more readable if we use `case "${ver_type}"`

test/bin/manage_brew_rpms.sh
ggiguash: We also need to add a message about supported `version_type` values.

test/bin/build_rpms.sh
```diff
@@ -126,6 +126,14 @@ download_brew_rpms() {
         # Run the download procedure
         bash -x "${SCRIPTDIR}/../../scripts/fetch_tools.sh" brew
         bash -x "${SCRIPTDIR}/manage_brew_rpms.sh" download "4.${MINOR_VERSION}" "${BREW_RPM_SOURCE}"
+
+        out_path="${IMAGEDIR}/released-brew-rpms"
```
ggiguash: Do we have to put those in a separate directory?
Nit: if we have to, can we use `brew-rpms-released` name so that they are sorted together in the output?

test/bin/manage_brew_rpms.sh
agullon: done in https://github.com/openshift/microshift/pull/4900/commits/45975296e2817f9b8f7be0b23f0202fad1ab29e1 commit

test/bin/manage_brew_rpms.sh
agullon: I agree, added it in https://github.com/openshift/microshift/pull/4900/commits/6c39fc71cba995ec14b8c546708adc2de63e9b6a commit

test/bin/manage_brew_rpms.sh
agullon: Because of the [automation from CI job](https://github.com/openshift/release/pull/66160/files#diff-82319ba702bdaee1b14321c74a639884e72f183e986dafb2680a19eeac248460R64-R69) when running for latest non released versions like 4.20. For example, when ER or RC don't exist, the script must continue running.

test/bin/build_rpms.sh
agullon: We do not have to, no needed it. But I did it this way to change as less as possible the rest of the code on `manage_build_cache.sh` script when uploading the RPMs into S3 bucket.
I agree to rename it to `brew-rpms-released`, done in https://github.com/openshift/microshift/pull/4900/commits/615fc4e3f53022689a40d7b8282e39bfdb6eebb7 commit.

test/bin/manage_brew_rpms.sh
agullon: Split into an extra var in https://github.com/openshift/microshift/pull/4900/commits/c05d860367f21b2418986b46a99c05baa1ddfa33 commit.
The reason I extract it is to have this dir structure:
```
$ tree -L 2 _output/test-images/brew-rpms-released
_output/test-images/brew-rpms-released
├── 4.16.41
│   ├── aarch64
│   └── x86_64
├── 4.17.27
│   ├── aarch64
│   └── x86_64
├── 4.18.0~0.nightly_2025_06_16_132618
│   ├── aarch64
│   └── x86_64
├── 4.18.0~ec.4
│   ├── aarch64
│   └── x86_64
├── 4.18.0~rc.10
│   ├── aarch64
│   └── x86_64
└── 4.18.11
    ├── aarch64
    └── x86_64
```
 
 instead of this:

 ```
 $ tree -L 2 _output/test-images/brew-rpms-released
_output/test-images/brew-rpms-released
├── microshift-4.16.41-202505260636.p0.g789ff62.assembly.4.16.41.el9
│   ├── aarch64
│   └── x86_64
├── microshift-4.17.27-202504251637.p0.ge4822bf.assembly.4.17.27.el9
│   ├── aarch64
│   └── x86_64
├── microshift-4.18.0~0.nightly_2025_06_16_132618-202506170128.p0.gbc96f1d.assembly.microshift.el9
│   ├── aarch64
│   └── x86_64
├── microshift-4.18.0~ec.4-202411221636.p0.gbfd3ba1.assembly.ec.4.el9
│   ├── aarch64
│   └── x86_64
├── microshift-4.18.0~rc.10-202502170426.p0.g84f67ec.assembly.rc.10.el9
│   ├── aarch64
│   └── x86_64
└── microshift-4.18.11-202504281317.p0.g48a03d6.assembly.4.18.11.el9
    ├── aarch64
    └── x86_64
```
 
 What do you think? to be honest not sure if it's worth it.

File: test/bin/manage_brew_rpms.sh [hunk 8]
```diff
@@ -6,8 +6,11 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # its execution in a containerized environment with limited set of tools.
 
 usage() {
-    echo "Usage: $(basename "$0") [access | download <version> <path>]"
+    echo "Usage: $(basename "$0") [access | download <version> <path> [version_type]]"
     echo "  download:   Download the RPM version to the path as specified"
+    echo "    - version: the X.Y version. Example: 4.19"
+    echo "    - path: the output directory. Example: /_output/test-images/brew-rpms"
+    echo "    - version_type: this is an optional parameter to specify the version type. Valid values: rc, ec, zstream and nightly."
```
ggiguash:
```suggestion
    echo "    - version_type: optional version type. Valid values: rc, ec, zstream and nightly."
```

test/bin/manage_brew_rpms.sh
```diff
@@ -38,10 +42,15 @@ action_download() {
 
     # Attempt downloading the specified build version
     local package
-    package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}" | sort | tail -1) || true
+    case ${ver_type} in 
+        zstream) package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}" | grep -v "~" | uniq | tail -n1) || true;;
+        rc|ec)   package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}.0~${ver_type}." | tail -1) || true;;
+        *)       package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}.*${ver_type}" | tail -1) || true;;
+    esac
```
ggiguash: Nit: we usually use this type of formatting, which is a bit more readable.
```suggestion
    case ${ver_type} in 
    zstream)
        package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}" | grep -v "~" | uniq | tail -n1) || true
        ;;
    rc|ec)
        package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}.0~${ver_type}." | tail -1) || true
        ;;
    *)
    package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}.*${ver_type}" | tail -1) || true
    ;;
    esac
```

test/bin/manage_brew_rpms.sh
ggiguash: I would like to make this conditional then. We should not allow failure to fetch main packages.
Let's issue a warning and skip for specific package types (exception), or fail as a rule?

test/bin/manage_brew_rpms.sh
```diff
@@ -51,7 +60,14 @@ action_download() {
...
     for arch in x86_64 aarch64 ; do
         local adir
-        adir="${dir}/${arch}"
+        # shellcheck disable=SC2001
+        if [ -z "${ver_type}" ] ; then
+            adir="${dir}/${arch}"
+        else
+            # remove date and commit id from package name and use it in dir name
+            version=$(echo "${package}" | sed 's/.*microshift-\([^-]*\).*/\1/')
+            adir="${dir}/${version}/${arch}"
```
ggiguash: Can we simplify this and just use `${version_type}` instead of `${version}`?

test/bin/manage_build_cache.sh
```diff
@@ -117,6 +117,26 @@ action_upload() {
         run_aws_cli s3 cp --quiet "${pkg_src}" "${pkg_dst}"
         rm -f "${pkg_src}"
     done
+
+    # Upload released RPM from brew
+    pushd "${src_base}"
+    released_brew_rpms_dirs="$(find "brew-rpms-released" -maxdepth 1 -type d | tail -n +2)"
```
ggiguash: 1. Why can't we add brew-rpms-released to the loop above?
2. We also need to implement download operation

test/bin/manage_brew_rpms.sh
agullon: I think it's not a good idea. I prefer a directory explicitly pointing to the version:

```
$ ls -1d *
4.16.41
4.17.27
4.18.0~0.nightly_2025_06_18_163729
4.18.0~ec.4
4.18.0~rc.10
4.18.11
```

instead of:

```
$ ls -1d *
ec
nightly
rc
zstream
```

or

```
$ ls -1d *
4.16-zstream
4.17-zstream
4.18-ec
4.18-nightly
4.18-rc
4.18-zstream
```

ggiguash: Can you explain why? 
Having floating top-level names also makes is more difficult to automate the usage of those.

With the `<version>-<version_type>` naming, you can just point to a version and its type in the automation and run with whatever is there. Note that the actual versions are still present, but inside the directory.

test/bin/manage_brew_rpms.sh
agullon: done in https://github.com/openshift/microshift/pull/4900/commits/45975296e2817f9b8f7be0b23f0202fad1ab29e1 commit

test/bin/manage_brew_rpms.sh
agullon: done it in https://github.com/openshift/microshift/pull/4900/commits/2de69c57cf1fbd092d2ec3d9f4079ee1910ae022 commit
Also, added skip for released brew version here: https://github.com/openshift/microshift/pull/4900/commits/1b9f044a5827401bcdcc6074eb497a492bcb35cd

test/bin/manage_build_cache.sh
agullon: 1. Because I think it's better to put the artifacts (RPMs, images and refs) in a separate S3 bucket directory. With this approach, we won't have them duplicated for every directory branch, and we save resources by uploading them for every directory branch.

2. Yes, my idea was to do it on another PR. Because this PR should only focus on the build phase.

test/bin/manage_brew_rpms.sh
agullon: Not sure if we are both in the same page here. I'll try to explain myself better.

For me, the final goal of this PR is to have a database/repository (S3 bucket in this case, which is perfect) from where I can fetch any given MicroShift version (RPMs, images or refs) to deploy a MicroShift cluster and run release regression testing.

So my idea/proposal is to put all those artifacts (RPMs, images and refs) in a S3 bucket directory to point to them when triggering the release regression testing. By using the naming approach, `4.18-zstream` I find it difficult to know beforehand which version is that: `4.18.2`? `4.18.11`?

test/bin/build_rpms.sh
agullon: done here: https://github.com/openshift/microshift/pull/4900/files#diff-c5558625fd237b08c74b6c9a3a55f7d994792bebf3c4a9ee2548a96d18810db4R130-R136

test/bin/manage_build_cache.sh
ggiguash: This requires a discussion. We do not store version collection of data without a release tag at this time. I do not think we should change this approach. 

The disk space (duplication) is not a problem. We need to manage release-specific information in the buckets. This script is release specific and it should only manipulate release-specific data.

test/bin/manage_brew_rpms.sh
ggiguash: For what you were trying to achieve, the versioning makes sense. I missed that point, so thank you for explaining.

As I commented in another thread, changing release-specific storage of artifacts is not what we do and I'm not sure it's a way to go. 

I would suggest to continue with storing the artifacts under a given release because all the system is written to work like this. Then, I suppose, we can use more generic names, like `4.18-zstream`, etc.

test/bin/manage_build_cache.sh
agullon: I see, no problem then.

test/bin/manage_brew_rpms.sh
agullon: I'm thinking about this regarding the images, how to build them pointing to the right parent version for upgrades, and I convinced myself we should follow the current approach. Otherwise, I foresee another complete and different approach for parent images, which I think it's not a good idea at all.

So, I change it to follow current approach.

test/bin/manage_build_cache.sh
```diff
@@ -102,7 +102,10 @@ action_upload() {
     run_aws_cli s3 sync --quiet --include '*.iso' "${iso_base}" "${iso_dest}"
 
     # Upload mirror-registry, brew-rpms and repo archives
-    for dir in mirror-registry brew-rpms repo ; do
+    pushd "${src_base}"
+    brew_rpms_dirs="$(find "brew-rpms" -maxdepth 1 -type d | grep '4..[0-9]-')"
```
agullon: Output example from these lines to understand dir structure:
```
$ pushd _output/test-images/
~/microshift/_output/test-images ~/microshift

$ find "brew-rpms" -maxdepth 1 -type d | grep '4..[0-9]-'
brew-rpms/4.18-nightly
brew-rpms/4.18-zstream
brew-rpms/4.17-zstream
brew-rpms/4.16-zstream
brew-rpms/4.18-rc
brew-rpms/4.18-ec

File: test/bin/manage_brew_rpms.sh [hunk 13]
```diff
@@ -38,7 +47,18 @@ action_download() {
...
     local package
-    package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}" | sort | tail -1) || true
+    case ${ver_type} in 
+        zstream)
+            package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}" | grep -v "~" | uniq | tail -n1) || true
+            ;;
+        rc|ec)
+            package=$(brew list-builds --quiet --package=microshift --state=COMPLETE | grep "^microshift-${ver}.0~${ver_type}." | tail -1) || true
+            ;;
+        *)
```
ggiguash, line 60: If version_type is not specified, let's use nightly by default.
We also need to produce an error in case ver_type is not one of the supported strings.

test/bin/manage_build_cache.sh
ggiguash: The download should be relatively fast for RPMs. Let's try not to optimize upfront. We can start from a simple on tar and optimize if necessary in the future.

ggiguash: This way, we should also get upload and download without changes in this script.

test/bin/manage_brew_rpms.sh
agullon: done in https://github.com/openshift/microshift/pull/4900/commits/c54d4833784fadc62714ce96dd090ff53d52cc42 commit

test/bin/manage_build_cache.sh
agullon: I agree, remove changes from `manage_build_cache.sh` on this PR.

ggiguash:
The failures have nothing to do with the changes. Saving on CI cycles.
/override ci/prow/e2e-aws-tests-bootc-periodic ci/prow/test-unit ci/prow/verify

PR #4903, opened 2025-05-14
OCPBUGS-56157: Use large OTEL config example as default + descriptions tweaks

pmtk:
Removed `packaging/observability/opentelemetry-collector.yaml` as it should be equivalent to the large example (see https://github.com/openshift/openshift-docs/pull/92758#discussion_r2087172237).

`packaging/observability/opentelemetry-collector-large.yaml` is copied as default config to `/etc/microshift/observability/opentelemetry-collector.yaml`. Not symlinked to avoid problematic scenario like:
- User edits `opentelemetry-collector.yaml` which is symlink to `opentelemetry-collector-large.yaml`.
- User upgrades MicroShift, `opentelemetry-collector-large.yaml` is overwritten and so is user's configuration.

Presets/examples description updates:
- Changed `Container, Pod, Volume, and Node metrics` to avoid interpretation that the Pod's `/metrics` are also included
- Removed `(Warnings only)` regarding events - could not find in the receivers source code any configuration or logic related to filtering events types
- Added explicit `priority: info` (which is the default value) to journald receiver to match the description

Changed the OTEL backend vars to OTEL_BACKEND to avoid confusion when modifying the configuration. Kubelet-stats receiver's endpoint should not be changed, but if it's named the same as the backend vars, it's easy to bulk replace or misinterpret reconfiguration.


pmtk:
/hold

Thinking about changing variables names to avoid confusion

pmtk:
standard1 test taking too long due to recent router test suite size increase
/retest

packaging/observability/opentelemetry-collector-large.yaml
```diff
@@ -73,10 +74,11 @@ exporters:
   otlp:
     sending_queue:
       storage: file_storage
-    # Endpoint must point to an ip or hostname and port of an otlp service. Here, the K8S_NODE_NAME is used because it
-    # will be resolved to the local node's hostname automatically. An unreachable endpoint will reported in the logs
-    # of the microshift-observability service.
-    endpoint: ${env:K8S_NODE_NAME}:4317
+    # Endpoint must point an IP or hostname, and port of an OTLP backend service.
+    # Here, the OTEL_BACKEND env var is used. It should be changed to point to the backend.
+    # If left unchanged and undefined, it will be resolved to localhost.
```
agullon, line 79: I'm curious, where the default value of `${OTEL_BACKEND}` is set?

pmtk: Nowhere actually. But it doesn't cause an error, just evaluates to empty string which is later replaced with localhost:
```
2025-05-16T09:08:10.316+0200        warn        envprovider/provider.go:51        Configuration references unset environment variable        {"name": "OTEL_BACKEND"}
...
2025-05-16T09:08:10.359+0200        warn        zapgrpc/zapgrpc.go:193        [core] [Channel #1 SubChannel #3]grpc: addrConn.createTransport failed to connect to {Addr: "[::1]:4317", ServerName: "localhost:4317", }. Err: connection error: desc = "transport: Error while dialing: dial tcp [::1]:4317: connect: connection refused"
```

So user can just replace it directly in config or add an `Environment=` to microshift-observability.service, but I prefer the former.


Alternative is to use work like `backend` and that would be expected to be actual resolvable hostname, but again this doesn't cause otel-collector to exit with error:
```
2025-05-16T10:58:52.371+0200        info        exporterhelper/retry_sender.go:118        Exporting failed. Will retry the request after interval.        {"kind": "exporter", "data_type": "logs", "name": "otlp", "error": "rpc error: code = Unavailable desc = name resolver error: produced zero addresses", "interval": "2.644063999s"}
```

I'm open to other alternatives, anything beside reusing K8S_NODE_NAME as it's confusing

agullon:
this PR looks good to me

packaging/observability/opentelemetry-collector-large.yaml
pmtk: Changed the comment line from
```
    # If left unchanged and undefined, it will be resolved to localhost.
```
to
```
    # Unless replaced in config or defined in service file, it'll be empty and OTEL will use 'localhost' instead.
```

WDYT?

pmtk: When I think of resolve I think of DNS, there's no DNS here. Maybe someone else will chime in with ideas

packaging/observability/opentelemetry-collector-large.yaml
```diff
@@ -73,10 +74,11 @@ exporters:
...
     sending_queue:
       storage: file_storage
-    # Endpoint must point to an ip or hostname and port of an otlp service. Here, the K8S_NODE_NAME is used because it
-    # will be resolved to the local node's hostname automatically. An unreachable endpoint will reported in the logs
-    # of the microshift-observability service.
-    endpoint: ${env:K8S_NODE_NAME}:4317
+    # Endpoint must point an IP or hostname, and port of an OTLP backend service.
+    # Here, the OTEL_BACKEND env var is used. It should be changed to point to the backend.
+    # Unless replaced in config or defined in service file, it'll be empty and OTEL will use 'localhost' instead.
+    # An unreachable endpoint will reported in the logs of the microshift-observability service.
```
copejon, line 80:
```suggestion
    # An unreachable endpoint will be reported in the logs of the microshift-observability service.
```

packaging/observability/opentelemetry-collector-large.yaml
```diff
@@ -117,8 +119,9 @@ service:
         - periodic:
             exporter:
               otlp:
-                # Endpoint must point to an ip or hostname and port of an otlp service. Here, the K8S_NODE_NAME is used
-                # because it will be resolved to the local node's hostname automatically. An unreachable endpoint will
-                # reported in the logs of the microshift-observability service.
+                # Endpoint must point an IP or hostname, and port of an OTLP backend service.
+                # Here, the OTEL_BACKEND env var is used. It should be changed to point to the backend.
+                # Unless replaced in config or defined in service file, it'll be empty and OTEL will use 'localhost' instead.
+                # An unreachable endpoint will reported in the logs of the microshift-observability service.
```
copejon, line 125:
```suggestion
                # An unreachable endpoint will be reported in the logs of the microshift-observability service.
```

packaging/rpm/microshift.spec
```diff
@@ -570,6 +570,8 @@ install -p -m644 assets/optional/ai-model-serving/release-ai-model-serving-x86_6
 # observability
 install -d -m755 %{buildroot}/%{_sysconfdir}/microshift/observability
 install -p -m644 packaging/observability/*.yaml -D %{buildroot}%{_sysconfdir}/microshift/observability/
+# Explicit copy of large config as default. Not using symlink to avoid accidental package upgrade overwriting user config if the user edits the config without copying (i.e. edits the target of symlink).
+install -p -m644 packaging/observability/opentelemetry-collector-large.yaml -D %{buildroot}%{_sysconfdir}/microshift/observability/opentelemetry-collector.yaml
```
copejon, line 574: 👍🏻

copejon:
Looks like the observability rewrite of the otel config is tripping the `rpm -V --nomtime` checks run by other tests in the suite.

agullon:
> Looks like the observability rewrite of the otel config is tripping the `rpm -V --nomtime` checks run by other tests in the suite.

I also saw these error/issue in QE regression tests for 4.19 RC. Wondering if we should exclude configuration files from the `rpm -V` checks. wdyt?

pmtk:
> > Looks like the observability rewrite of the otel config is tripping the `rpm -V --nomtime` checks run by other tests in the suite.
> 
> I also saw these error/issue in QE regression tests for 4.19 RC. Wondering if we should exclude configuration files from the `rpm -V` checks. wdyt?

Looks like it looses eof newline, I'll just copy back from the large preset.

```
$ diff opentelemetry-collector.yaml opentelemetry-collector-large.yaml
127c127
<                 protocol: http/protobuf
\ No newline at end of file
---
>                 protocol: http/protobuf
```

copejon:
> > Looks like the observability rewrite of the otel config is tripping the `rpm -V --nomtime` checks run by other tests in the suite.
> 
> I also saw these error/issue in QE regression tests for 4.19 RC. Wondering if we should exclude configuration files from the `rpm -V` checks. wdyt?

Yeah, it's probably a good idea.

pmtk:
> > > Looks like the observability rewrite of the otel config is tripping the `rpm -V --nomtime` checks run by other tests in the suite.
> > 
> > 
> > I also saw these error/issue in QE regression tests for 4.19 RC. Wondering if we should exclude configuration files from the `rpm -V` checks. wdyt?
> 
> Yeah, it's probably a good idea. @pmtk @pacevedom what do y'all think?

I fixed the test to copy back the correct config example, so it should be good now - no need.

PR #4904, opened 2025-05-14
USHIFT-5626: Update test framework to use 4.19 RC packages from the mirror

ggiguash:


pacevedom (commented):
I think we can also remove this exclude tag now: https://github.com/openshift/microshift/blob/main/test/suites/standard1/router.robot#L191

pacevedom (commented):
/lgtm

pacevedom (approved):
/lgtm

PR #4916, opened 2025-05-16
OCPBUGS-49917 [WIP] 

copejon:
This is this initial, WIP code, and is likely to drastically change. The design goals for this PR are:
1. Get rid of embedded storage class asset. LVM-O storage classes should only be generated based on lvmd and vgs
2. Dynamically generate the storage class with the deviceClass.Name appended to the storage class name.
3. Integrate this into lvmd hot swapping logic.
4. Do not cleanup StorageClasses just because the device class has vanished. If a user removes a device class, that will orphan child lvm volumes. If the storageClasses are deleted, that would have the same effect, which the system should never do.

### DIFF
--- a/packaging/observability/opentelemetry-collector.yaml
+++ b/packaging/observability/opentelemetry-collector.yaml
@@ -8,6 +8,18 @@
 # - System journals for selected MicroShift services and dependencies, priority < Info
 
 receivers:
+  prometheus:
+    config:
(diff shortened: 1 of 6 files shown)


copejon:
Closing for now. I think the work is going to produce some large changes, and I just want to solve the bug as it is right now. Will follow up with the team

PR #4932, opened 2025-05-19
OCPBUGS-49917 LVM config device-class name does not apply to storageClass name

copejon:


copejon:
OOT today, still need to:

- [ ] Update test assets / RF suite to handle dynamic naming

pmtk:
Also, what about the upgrade? If we render a new StorageClass and the formerly named still exists, and both have `is-default-class: true` annotation. I think we need to remove the old one on upgrade.

copejon:
@pmtk Big agree. Throwing the brakes now and bring this to the team soon. There's a significant amount of gotchas that need to be considered: yours, and also the complexities of the lvm-o hot-reloading the lvmd.yaml.

/home

copejon:
@pmtk Big agree. Throwing the brakes now and bring this to the team soon. There's a significant amount of gotchas that need to be considered: yours, and also the complexities of the lvm-o hot-reloading the lvmd.yaml.

/hold

copejon:
@pmtk  For posterity, I'll copy my sidebar response to this thread (with some edits):

There's even more complicating factors. The lvm-o runtime permits live config changes. To enable this at the cluster layer, Microshift monitors the lvmd.yaml for changes. When a change is detected, the new config is pushed into the lvm-o container. This creates a whole new set of lifecycle management questions:


1. Should the controlplane be responsible for deleting SCs? How do you guarantee ushift doesn't orphan PVs and the underlying volumes. The answer is "no." 
2. What if the "default" DC changes? Should the controlplane change the default SC? (No again)
3. What if a batch of PVs are created? They will reconcile concurrently with the live lvmd change. It's possible you end up with part of the batch being provisioned via one SC, and the remainder provisioned via the new SC. (Big no)
4. Renaming a DC implies renaming a SC. (_NO_)
5. Should controlplane maintain a map of DC updates to determine if just DC metadata is changed, or it's a whole new DC? (moot, given my prior answers)

PR #4939, opened 2025-05-20
NO-ISSUE: skip cipher check if not available in ssl list

agullon:


test/suites/standard2/tls-configuration.robot
```diff
@@ -172,22 +177,32 @@ Check TLS Endpoints
...
     [Documentation]    Verify system logs contain expected error messages for configuration errors
     [Arguments]    ${error_type}    ${error_case}
     ${output}    ${rc}=    Get Log Output With Pattern    ${CURSOR}    apiServer.tls
     @{expected_lines}=    Get Expected Fault Messages    ${error_type}    ${error_case}
     Compare Output Logs    ${output}    @{expected_lines}
+
+Is Cipher Available
+    [Documentation]    Check if openssl keywork is available from ssl list
+    [Arguments]    ${cipher}
+    ${rc}=    Execute Command    openssl ciphers ${cipher}
```
pacevedom, line 206: You are returning both return code and stdout and storing in a single variable. This variable `rc` is now an array that holds the stdout first and the rc second. This can be seen here: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_microshift/4939/pull-ci-openshift-microshift-main-e2e-aws-tests/1924851897060036608/artifacts/e2e-aws-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el94-src@standard-suite2/log.html#s1-s5
All checks are failing because of comparing the array to 0.
Adding the logic to keep going if the cipher is not available makes it silently fail every check.

agullon: thanks, addresed in https://github.com/openshift/microshift/pull/4939/commits/ff5a733aad9282a42965be7a1d0a1c7fd49d19bc commit

pacevedom (approved):
/lgtm
/hold
Just one nit, held in case you want to do it. If not we can always do it later as its not critical and we are a bit tight on schedule.

test/suites/standard2/tls-configuration.robot
```diff
@@ -155,15 +155,20 @@ Restore Valid TLS Configuration
 
 Openssl Connect Command
     [Documentation]    Run Openssl Connect Command in the remote server
-    [Arguments]    ${host_and_port}    ${args}
+    [Arguments]    ${host_and_port}    ${args}    ${expected_rc}    ${expected_str}
```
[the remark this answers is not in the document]
agullon: We shouldn't remove the IF because it's needed to validate the cipher under tests is in the stdout of the openssl command. See this line: https://github.com/openshift/microshift/pull/4939/files#diff-af253f15d1596775b3ca9cf4154438c05a0007533c49259a222eb7f9d4adc9aaR184 

Furthermore, I'm happy to set the default value of `${expected_str}` to an empty string, but I prefer to be not use default values on testing functions to force user to configure everything making sure what is being tested.

PR #4946, opened 2025-05-21
[release-4.19] USHIFT-5775: TLS config RF test: skip cipher if not available

agullon:
Manual backport from https://github.com/openshift/microshift/pull/4939 PR

Added a condition to skip checking a cipher when it's not available, for example, when FIPS mode is enabled.


pacevedom (approved):
/lgtm
/label backport-risk-assessed
/label cherry-pick-approved
/label jira/valid-bug

PR #4978, opened 2025-05-28
NO-ISSUE:  add RestartSec=10s for systemd service

lance5890:

As the default RestartSec is 100ms，The microshift may not startup in default StartLimitBurst(5 times) ; we should expand the systemd RestartSec to 10s，just like the etcd serivce does as follows:
https://github.com/openshift/microshift/blob/032d259f3c000e68758e7f2e207ddfce2397a19e/ansible/roles/install-microshift/files/etcd.service#L20

### DIFF
--- a/packaging/flannel/microshift-flannel.service
+++ b/packaging/flannel/microshift-flannel.service
@@ -14,6 +14,7 @@ Before=kubepods.slice
 WorkingDirectory=/usr/bin/
 ExecStart=microshift run
 Restart=always
+RestartSec=10s
 User=root
 Type=notify
 Delegate=yes
--- a/packaging/observability/microshift-observability.service
+++ b/packaging/observability/microshift-observability.service
@@ -10,6 +10,7 @@ Environment=K8S_NODE_NAME="%l"
 ExecStartPre=/usr/bin/mkdir -p /var/lib/microshift-observability
 ExecStart=/usr/bin/opentelemetry-collector --config=/etc/microshift/observability/opentelemetry-collector.yaml
 Restart=always
+RestartSec=10s
 User=root
 
 [Install]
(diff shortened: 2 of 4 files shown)


pmtk:
Hey @lance5890, thanks for your contribution.

Could you explain how increasing time between restarts will help you?
What problem are you observing?

/label ok-to-test

lance5890:
> Hey @lance5890, thanks for your contribution.
> 
> Could you explain how increasing time between restarts will help you? What problem are you observing?
> 
> /label ok-to-test

We deployed MicroShift on a host with insufficient performance. During startup, we found that the network was not fully configured before the systemd service started, which caused MicroShift to fail to start. After failing more than five times, it stopped attempting to start. The issue was resolved after we manually adjusted the RestartSec=10s.
image

pmtk:
I noticed you've taken steps to address the problem with StartLimitIntervalSec.
However, I'm afraid that this change needlessly increases the timeout that some feature depend on and I'm hesitant to apply it by default.

I would suggest instead that you use a systemd drop-in to configure required settings to match your environment.

lance5890:
> I noticed you've taken steps to address the problem with StartLimitIntervalSec. However, I'm afraid that this change needlessly increases the timeout that some feature depend on and I'm hesitant to apply it by default.
> 
> I would suggest instead that you use a systemd drop-in to configure required settings to match your environment.

directly extending the RestartSec parameter to 10s might be a bit risky, but I still suggest starting by modifying this value to 1s (compared to the default of 100ms) would resolve most issues

pmtk:
Since the problem isn't plaguing neither the users setups nor our tests except for one place which we accounted for with systemd drop-in (which we think is valid way of tweaking the behavior depending on the hardware) I think we'll not include this change at current time.

/close

PR #4997, opened 2025-06-02
USHIFT-5811: Initiate dir with patched dependencies to keep vendor clean

pmtk:
PR adds new dir tree `deps/` that contains clones of `github.com/openshift/kubernetes`,  `github.com/openshift/route-controller-manager`, and `github.com/kubernetes/klog` for patching purposes.

Goal is to get rid of patching the `vendor/` and instead patch the `deps/` which means that `go mod vendor` will not produce changes (because it would revert the patches).

New "command" was added to the rebase.sh: `deps` in 3 flavors:
- `deps copy` - copy dependency from _output/staging to deps/
- `deps clone github.com/kubernetes/klog from kubernetes` - clone repo to deps/ with reference from another component (from staging)
- `deps kubernetes-version` - update module's required version in go.mod (for information purposes only as they're replaced anyway).

Changes in `go.mod`:
- Updated rebase.sh commands
- `replace()`d `k8s.io/*` modules: version in `require()` are now matching Kubernetes version (for information only)
- Previously replaced k8s staging modules are now pointing to `./deps/github.com/openshift/kubernetes/...`

Changes in `etcd/go.mod`:
- Previously replaced k8s staging modules are now pointing to `../deps/github.com/openshift/kubernetes/...`


pacevedom (approved):
/lgtm

PR #5005, opened 2025-06-04
NO-ISSUE: Fix ami list filtering in manage_aws_stack.sh script

ggiguash:
Relax the filtering to accomodate for RHEL 9.6 image names.


ggiguash:
Will be included in the https://github.com/openshift/microshift/pull/5008 PR.
/close

PR #5016, opened 2025-06-05
USHIFT-5534: Use kindnet instead flannel as CNI

praveenkumar:
Since flannel now require br_netfilter module to start as per https://github.com/flannel-io/flannel?tab=readme-ov-file#deploying-flannel-with-helm because https://github.com/kubernetes/kubernetes/pull/123464, which means it would problem to run microshift in container for WSL environment(For podman-desktop usecase).

This PR make switch from flannel to kindnet which is used by kind/minikube and doesn't depend on any kernel module.

- https://github.com/aojea/kindnet

Closes #<Issue Number>


assets/optional/kindnet/release-kindnet-aarch64.json
```diff
@@ -0,0 +1,8 @@
+{
+  "release": {
+    "base": "4.18.0-0.nightly-arm64-2024-08-29-120159"
```
pmtk: I think we could bump the version

assets/optional/kindnet/02-cluster-role.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: flannel
+  name: kindnet
```
pmtk, line 4: I think there are superfluous rules here, like `k8s.ovn.org`, not sure about `security.openshift.io` and `cloud.network.openshift.io`. Can we remove unnecessary rules?

packaging/crio.conf.d/13-microshift-kindnet.conf
```diff
@@ -0,0 +1,4 @@
+[crio.network]
+# kindnet is the name configured by kindnet in /etc/cni/net.d/ config file
+# by declaring this CRI-O will wait until that network is configured.
+cni_default_network = "kindnet"
```
pmtk, line 4: Do you know how did the flannel work without this?

packaging/rpm/microshift.spec
```diff
@@ -788,7 +794,7 @@ fi
 * Wed Apr 16 2025 vanhalenar <thalenar@redhat.com> 4.19.0
 - Remove observability preset that enables the service automatically
 
-* Wed Apr 09 2025 Patryk Matuszak <pmatusza@redhat.com> 4.19.0
+* Wed Apr 09 2025 Patryk Matuszak <pmatusza@redhat.com> 4.19.0s
```
pmtk:
```suggestion
* Wed Apr 09 2025 Patryk Matuszak <pmatusza@redhat.com> 4.19.0
```

pmtk (commented):
Minor comments only.

CI job failures are unrelated to the change (we're in the middle of bumping version of different things)

packaging/crio.conf.d/13-microshift-kindnet.conf
praveenkumar: @pmtk for flannel I used the name as `ovn-kubernetes` as part of configmap which is passed as argument for flannel command but kindnet doesn't provide something like this so need to use drop-in files. I think config drop-in file is much cleaner should've used for flannel also.

assets/optional/kindnet/release-kindnet-aarch64.json
praveenkumar: Done

assets/optional/kindnet/02-cluster-role.yaml
praveenkumar: I think scc is required but removed `k8s.ovn.org` and `cloud.network.openshift.io` groups.

packaging/crio.conf.d/13-microshift-kindnet.conf
pmtk: That's definitely the right approach.

pmtk:
Taking away lgtm to stop bot from retesting job we can simply override
/lgtm-cancel

PR #5018, opened 2025-06-05
NO-ISSUE: Add teardown for "Router Disabled" test

pmtk:
PR aims to fix recent issues where `Router Verify Security Configuration` test would fail because the namespace does not exist.


pmtk:
image

PR #5036, opened 2025-06-09
generic device plugin

pmtk:


pkg/config/gdp.go
```diff
@@ -0,0 +1,165 @@
...
+
+	if gdp.Domain != "" {
+		c.GenericDevicePlugin.Domain = gdp.Domain
+	}
+
+	if len(gdp.Devices) > 0 {
+		c.GenericDevicePlugin.Devices = gdp.Devices
+	}
+
+	klog.Info("gdp.incorporateUserSettings - end")
```
pacevedom: Is this a debug leftover or should we keep it? We log the whole config at the start of microshift.

pkg/config/gdp.go
```diff
@@ -0,0 +1,165 @@
...
+	if domainErrs := validation.IsDNS1123Subdomain(gdp.Domain); len(domainErrs) > 0 {
+		errs = append(errs, fmt.Errorf("failed to parse domain %q: %s", gdp.Domain, strings.Join(domainErrs, ", ")))
+	}
+
+	if len(gdp.Devices) == 0 {
+		errs = append(errs, fmt.Errorf("genericDevicePlugin.Devices is empty - at least one device must be specified"))
+	}
+
+	deviceTypeFmt := "[a-z0-9][-a-z0-9]*[a-z0-9]"
+	deviceTypeRegexp := regexp.MustCompile("^" + deviceTypeFmt + "$")
```
pacevedom: Move to the top as a var?

pacevedom (commented):
Just two nits, lgtm otherwise

PR #5045, opened 2025-06-11
USHIFT-5819: Update klog patch to adjust to go1.24 changes

pmtk:


pacevedom:
/close
All commits included in https://github.com/openshift/microshift/pull/5095

PR #5121, opened 2025-06-30
USHIFT-5818: Generic Device Plugin - Sanity Smoke Test

pmtk:


test/image-blueprints-bootc/layer2-presubmit/group2/rhel96-bootc-source-optionals.containerfile
```diff
@@ -26,3 +26,13 @@ RUN dnf repoinfo --enabled && \
     rm -vf /etc/yum.repos.d/microshift-*.repo && \
     rm -rvf $USHIFT_RPM_REPO_PATH && \
     dnf clean all
+
+# Prepare system for testing Generic Device Plugin.
+RUN KERNEL_VER=$(rpm -q --qf "%{VERSION}-%{RELEASE}" kernel); \
+    KERNEL_VER_ARCH="${KERNEL_VER}.$(uname -i)"; \
+    dnf install -y git make "kernel-devel-${KERNEL_VER}" python3-pyserial && \
```
kasturinarra, line 34: Do you think we should use uname -r / pass one single kernel if there is more than one being listed ? if the system has more than one kernel i see this command is failing.

pmtk: We cannot use `uname -r` because it returns the kernel of the host, bootc container can (and had when I tested it) contain different kernel.
It's true that `rpm -q --qf` will return all matching pkgs, but the problem is that what you're running might not be the newest one.

Because these commands are inside containerfile, not really reusable automatically (by importing a file) anywhere else, I would recommend that for QE testing you can copy and tweak them to your needs.

test/assets/generic-device-plugin/job.yaml
```diff
@@ -0,0 +1,31 @@
...
+kind: Job
+metadata:
+  name: gdp-test
+spec:
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: gdp-test
+        image: quay.io/fedora/python-311
```
ggiguash: Can we use a RHEL 9 UBI image here?

test/image-blueprints-bootc/layer2-presubmit/group2/rhel96-bootc-source-optionals.containerfile
ggiguash: This is not clear for me. Containers are using kernel of the host, aren't they?
What would happen if the kernel in the container is different from the kernel on the host (majority of cases)?

Should we install both the kernel and the kernel-devel of the host?

test/suites/optional/generic-device-plugin.robot
```diff
@@ -0,0 +1,111 @@
...
+Sanity Test
+    [Documentation]    Performs a simple test of Generic Device Plugin
+    [Setup]    Run Keywords
+    ...    Enable And Configure GDP
+    ...    Enable Serialsim
+    ...    Copy Script To Host
+
+    Wait Until Device Is Allocatable
+
+    Command Should Work    crictl pull quay.io/fedora/python-311
```
ggiguash: Can we use RHEL UBI images, please?

test/assets/generic-device-plugin/job.yaml
pmtk: Done

test/suites/optional/generic-device-plugin.robot
pmtk: Done

test/image-blueprints-bootc/layer2-presubmit/group2/rhel96-bootc-source-optionals.containerfile
```diff
@@ -26,3 +26,14 @@ RUN dnf repoinfo --enabled && \
     rm -vf /etc/yum.repos.d/microshift-*.repo && \
     rm -rvf $USHIFT_RPM_REPO_PATH && \
     dnf clean all
+
+# Prepare system for testing Generic Device Plugin.
+# hadolint ignore=DL3003
+RUN KERNEL_VER=$(rpm -q --qf "%{VERSION}-%{RELEASE}" kernel); \
+    KERNEL_VER_ARCH="${KERNEL_VER}.$(uname -i)"; \
```
ggiguash:
```suggestion
    KERNEL_VER_ARCH="${KERNEL_VER}.$(uname -m)"; \
```

test/image-blueprints-bootc/layer2-presubmit/group2/rhel96-bootc-source-optionals.containerfile
pmtk: Yes, containers are using kernel from the host and that's why there's a difference between `uname -m` and version returned by `rpm`.

There's already a kernel installed in the base bootc image, so I decided to install matching kernel-devel (rather to upgrade kernel and install kernel-devel). The kernel in the image is not active when building an image, but it will be when the image is installed onto the disk.

Personally, I wouldn't attempt to install the kernel of the host because there's no guarantee what someone has on their hypervisor (maybe it's older one). I feel like keeping the kernel might be easier if there's some kind of bug to report (I mean: image tag is pointing to kernel we're not changing).
Though maybe there's value in upgrading the kernel, so we can do it.
WDYT?

ggiguash: Let's think what users would do in this case. Wouldn't they likely try to sync the kernel version between the host and the container?

pmtk: What would they get by syncing the, for examples, CI's host kernel with the kernel of the image that will be installed on another host? At most, I would expect them to run `dnf upgrade` to get latest packages (including kernel).

I feel like this discussion is out of scope for this PR - current solution works and is completely valid: it installs kernel-devel for kernel already included in the bootc image, then compiles a driver for that particular kernel, there's no runtime involved, so there's no need to involve host's kernel.
Whether we should upgrade all packages or match kernel versions if more of a testing approach question and is, imho, applicable to all images we build (like, let's take rhel96-bootc-source image for example, question of upgrading the pkgs and/or kernel is just as valid).

ggiguash: OK, we can continue this interesting discussion elsewhere.
I'll approve the PR for now.

PR #5160, opened 2025-07-07
USHIFT-5620: Create scripts for testing NVIDIA Device Plugin 

pmtk:
- Most of the existing ai-model-serving scripts are moved to new directory. They are now symlinks to ai-model-serving.
- Bumped NVIDIA Device Plugin from v0.17.0 to v0.17.2.
- Made some changes to original scripts to make them work for older MicroShift versions:
  - For 4.14 and 4.15, CRI-O dropins are renamed to include numeric prefix so the CRIO dropin from NVIDIA Container Toolkit overwrites MicroShift's settings.
  - For MicroShift version not having `healthcheck` command, use `oc rollout status` for waiting on the NVIDIA Device Plugin

Scripts are not planned to be backported to older branches. Scripts on release repository will get the scripts from the main branch. This is because the guide is generic, not dependent on any particular version of MicroShift or RHDE.


scripts/ci-nvidia-device-plugin/tests/04-verify-device-plugin.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -xeuo pipefail
+
+if microshift --help | grep -q "healthcheck"; then
+    sudo microshift healthcheck \
+        --namespace nvidia-device-plugin \
+        --daemonsets nvidia-device-plugin-daemonset
+else
```
ggiguash, line 13: Nit: feel free to fix in another PR. 

I think we need to add a comment here or in a README why we're checking it. The fact that these scripts will be used in multiple releases is not something we'd usually do

PR #5186, opened 2025-07-11
OCPNODE-3467: Remove containernetworking-plugins dependencies from crio.conf

bitoku:

Closes #OCPNODE-3467

Since RHEL has dropped the containernetworking-plugins package because podman stopped using it, cri-o is also going to install it as a dependency.
cri-o already stopped using containernetworking-plugins, so we remove dependencies from crio.conf as well.

### DIFF
--- a/packaging/crio.conf.d/10-microshift_amd64.conf
+++ b/packaging/crio.conf.d/10-microshift_amd64.conf
@@ -16,7 +16,6 @@ absent_mount_sources_to_reject = [
 [crio.network]
 # rhel8 crio is configured to only look at /usr/libexec/cni, we override that here
 plugin_dirs = [
-        "/usr/libexec/cni",
         "/run/cni/bin"
 ]
 
--- a/packaging/crio.conf.d/10-microshift_arm64.conf
+++ b/packaging/crio.conf.d/10-microshift_arm64.conf
@@ -16,7 +16,6 @@ absent_mount_sources_to_reject = [
 [crio.network]
 # rhel8 crio is configured to only look at /usr/libexec/cni, we override that here
 plugin_dirs = [
-        "/usr/libexec/cni",
         "/run/cni/bin"
 ]
 
--- a/packaging/crio.conf.d/12-microshift-multus.conf
+++ b/packaging/crio.conf.d/12-microshift-multus.conf
@@ -2,9 +2,6 @@
 # Override default network so CRI-O waits for and calls Multus CNI instead of ovn-kubernetes.
 cni_default_network = "multus-cni-network"
 
-# Change the order, so the CNIs from the container-networking-plugins (copied to host's /run/cni/bin)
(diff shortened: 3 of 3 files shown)


packaging/crio.conf.d/10-microshift_amd64.conf
```diff
@@ -16,7 +16,6 @@ absent_mount_sources_to_reject = [
 [crio.network]
 # rhel8 crio is configured to only look at /usr/libexec/cni, we override that here
```
ggiguash: How should the comment change if we're removing /usr/libexec/cni

packaging/crio.conf.d/10-microshift_arm64.conf
```diff
@@ -16,7 +16,6 @@ absent_mount_sources_to_reject = [
 [crio.network]
 # rhel8 crio is configured to only look at /usr/libexec/cni, we override that here
 plugin_dirs = [
```
ggiguash, line 18: Do we need the override at all? Should we be removing the whole `plugin_dirs` block?

packaging/crio.conf.d/10-microshift_amd64.conf
bitoku: I think no change is needed, because the comment is true though we can remove it.

packaging/crio.conf.d/10-microshift_arm64.conf
bitoku: I don't know how microshift installs cni, but as long as microshift uses `/run/cni/bin`, it needs to be specified.

ggiguash: OK, let's agree how to fix the comments then.

packaging/crio.conf.d/10-microshift_amd64.conf
ggiguash: We no longer support rhel8, so it's a good opportunity to fix the comments too.
Can you suggest the right comment? Why are we specifying this if it's the default setting in `crio` now?

```
$ crio config  | grep -A4 plugin_dirs
plugin_dirs = [
        "/usr/libexec/cni",
]
```

bitoku: I heard it had been required to work cri-o (sorry I was not there at that time).
Now that cri-o is no longer dependent on the plugins, I'm going to change the install script and the default will be changed as well.

packaging/crio.conf.d/10-microshift_arm64.conf
bitoku: how about `ovn-kube installs plugins in /run/cni/bin, and microshift uses them.`?

ggiguash: Suggested comment: `Configure crio to use the ovn-k plugins from /run/cni/bin`

bitoku: Updated!

PR #5203, opened 2025-07-16
NO-ISSUE: kindnet: Add plugin dir for crio config

praveenkumar:
With https://github.com/openshift/microshift/pull/5186 "/usr/libexec/cni" is removed from plugin dir which makes kindnet to failed since it checks the plugins from that directory. This pr should add it back only for kindnet usecase.


ggiguash:
/retitle NO-ISSUE: kindnet: Add plugin dir for crio config
/lgtm

PR #5230, opened 2025-07-21
NO-ISSUE: Rebase 4.20 (Kube bump to 1.33)

pmtk:


pacevedom:
IPv6 failures should be ok by switching to [EndpointSlice](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) instead of plain `Endpoint`.

PR #5261, opened 2025-07-28
USHIFT-4528: microshift-gitops smoke test

vanhalenar:
A simple smoke test to check ArgoCD pods are up and running. Next step would be to check with a simple workload.


test/package-sources-bootc/microshift-gitops.repo
```diff
@@ -0,0 +1,10 @@
+[microshift-gitops]
+name = Red Hat OpenShift GitOps 1.16 for RHEL 9 x86_64 (RPMs)
+baseurl = https://cdn.redhat.com/content/dist/layered/rhel9/{{ .Env.UNAME_M }}/gitops/1.16/os
```
agullon: Can you please move the gitops version `1.16` into the `common_versions.sh` script?
It would be easier to update the version in the future, as soon as a new gitops version is available.

vanhalenar: Done

test/suites/gitops/gitops.robot
```diff
@@ -0,0 +1,57 @@
...
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+    Remove Kubeconfig
+    Remove Directory    ${TMPDIR}    recursive=${True}
+
+Deploy Guestbook
+    [Documentation]    Deploys Guestbook app as test workload
+    VAR    ${file_path}=    ${TMPDIR}/guestbook-ui-deployment.yaml
+    Download File    ${URL}    ${file_path}
+    Oc Apply    -f ${file_path}
```
agullon: afaik you can set the url directly in the command: `oc apply -f https://example.com/manifest.yaml` so we can simplify the code, no need to download it first

vanhalenar: done

test/suites/gitops/gitops.robot
```diff
@@ -0,0 +1,43 @@
...
+Teardown
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+    Remove Kubeconfig
+
+Deploy Guestbook And Verify
+    [Documentation]    Deploys Guestbook app as test workload
+    Oc Apply    -f ${URL}
+    Wait Until Keyword Succeeds    5min    10s
+    ...    Named Deployment Should Be Available    guestbook-ui    default
```
pmtk: Let's use different namespace than `default`.
1. It's better to create a NS for the test/suite. Deleting a namespace will remove all the resources from it, so the cluster should be ready for another test.
2. Recently added "sos on keyword failure" does not detect usage of `default` namespace (because it listens for `VAR` with `ns` or `namespace`). If we'd create the namespace (there are keyword for setting up suite and creating ns in one go), then that `sos` could include some information about why this failed.

test/suites/gitops/gitops.robot
```diff
@@ -0,0 +1,43 @@
...
+    Login MicroShift Host
+    Setup Kubeconfig
+    Restart MicroShift
+
+Teardown
+    [Documentation]    Test suite teardown
+    Logout MicroShift Host
+    Remove Kubeconfig
+
+Deploy Guestbook And Verify
```
pmtk: TBH This keyword could be removed and its contents be just part of the `Verify Workload Deployed Correctly` directly

test/suites/gitops/gitops.robot
```diff
@@ -0,0 +1,43 @@
...
+    [Documentation]    Deploys workload and waits for ready status
+    Deploy Guestbook And Verify
+
+
+*** Keywords ***
+Setup
+    [Documentation]    Test suite setup
+    Login MicroShift Host
+    Setup Kubeconfig
+    Restart MicroShift
```
pmtk: Why do we restart MicroShift before testing? We are not changing configuration and the GitOps manifests are on disk from the very beginning, so there shouldn't be a need to restart ushift

test/suites/gitops/gitops.robot
```diff
@@ -0,0 +1,43 @@
+*** Settings ***
+Documentation       MicroShift GitOps tests
+
+Resource            ../../resources/microshift-process.resource
+Library             RequestsLibrary
```
pmtk: Not sure if we need that import

test/suites/gitops/gitops.robot
```diff
@@ -0,0 +1,43 @@
...
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+
+*** Variables ***
+${URL}      https://raw.githubusercontent.com/argoproj/argocd-example-apps/refs/heads/master/guestbook/guestbook-ui-deployment.yaml
```
pmtk: I'd change `URL` to something more descriptive

test/suites/gitops/gitops.robot
```diff
@@ -0,0 +1,43 @@
...
+
+
+*** Test Cases ***
+Verify GitOps Pods Start Correctly
+    [Documentation]    Waits for pods to enter a running state
+
+    Wait Until Keyword Succeeds    2min    10s
+    ...    All Pods Should Be Running    openshift-gitops
+
+Verify Workload Deployed Correctly
```
pmtk, line 17: This test could use a `[Teardown]` that would remove the workload (guestbook)

test/suites/gitops/gitops.robot
vanhalenar: done

test/suites/gitops/gitops.robot
vanhalenar: done

test/suites/gitops/gitops.robot
vanhalenar: Fixed

test/suites/gitops/gitops.robot
vanhalenar: Fixed

test/suites/gitops/gitops.robot
vanhalenar: Changed

test/suites/gitops/gitops.robot
vanhalenar: By adding a custom namespace I believe this is taken care of

agullon:
good work @vanhalenar!
I do have one question, why do you finally go with `http-24` test app instead of `guestbook-ui` app? It's also valid, of course, I'm just curious.

vanhalenar:
I went with `http-24` because I was unable to deploy `guestbook-ui` in a custom namespace, only the default one, I think because of security policy, the workload was always trying to bind to port 80 and was being denied. With `http-24`, I could change the port to 8080 and get it to work a lot easier.

PR #5283, opened 2025-08-05
USHIFT-5809: Upgrade Robot Framework environment

vanhalenar:


test/requirements.txt
```diff
@@ -1,8 +1,8 @@
...
+robotframework-requests==0.9.7
 robotframework-scplibrary==1.2.0
 robotframework-sshlibrary==3.8.0
-pyyaml==6.0.1
-robotframework-robocop==3.1.1
-robotframework-tidy==4.3.0
-packaging==23.2
+pyyaml==6.0.2
+robotframework-robocop==6.3.0
+robotframework-tidy==4.17.0
```
pmtk: Since robocop now includes the robotidy, I think we can get rid of it?

test/requirements.txt
```diff
@@ -1,8 +1,8 @@
...
+robotframework==7.3
+robotframework-requests==0.9.7
 robotframework-scplibrary==1.2.0
 robotframework-sshlibrary==3.8.0
-pyyaml==6.0.1
-robotframework-robocop==3.1.1
-robotframework-tidy==4.3.0
-packaging==23.2
+pyyaml==6.0.2
+robotframework-robocop==6.3.0
```
pmtk: I see it's already 6.5.2 - https://robocop.readthedocs.io/en/stable/
Shall we bump it? Should be minimal effort now

scripts/verify/verify-rf.sh
```diff
@@ -13,11 +13,6 @@ cd "${ROOTDIR}/test"
...
-"${RF_VENV}/bin/robocop" \
-    --exclude 1015 \
-    --configure 0504:max_len:40 \
-    --configure 0505:max_calls:20 \
-    --configure 0508:line_length:200 \
-    --configure 0506:max_lines:1000
+"${RF_VENV}/bin/robocop" check
 
-"${RF_VENV}/bin/robotidy" --check --diff .
+"${RF_VENV}/bin/robocop" format --diff --no-overwrite
```
pmtk: Does it fail on non-formatted file?
https://robocop.readthedocs.io/en/stable/overview/formatter.html#status-code mentions `--check`

test/resources/common.resource
```diff
@@ -20,7 +20,7 @@ Setup Suite With Namespace
     Login MicroShift Host
     Setup Kubeconfig
     ${ns}=    Create Unique Namespace
-    Set Suite Variable    \${NAMESPACE}    ${ns}
+    VAR    ${NAMESPACE}=    ${ns}    scope=SUITE
```
pmtk, line 23: Looking at [the release notes](https://github.com/robotframework/robotframework/blob/master/doc/releasenotes/rf-7.0.rst#native-var-syntax) I think we don't need the `=`
```suggestion
    VAR    ${NAMESPACE}    ${ns}    scope=SUITE
```

pmtk (commented):
🔥

test/requirements.txt
vanhalenar: Removed

test/requirements.txt
vanhalenar: Bumped, it didn't require any additional changes

scripts/verify/verify-rf.sh
vanhalenar: You're right, it didn't fail, fixed

test/resources/common.resource
vanhalenar: I thought so as well, but robotidy wants the `=` there 🤷

pmtk: Wait, but it's not linter, it's the formatter.
`robocop docs NormalizeAssignments` says:
>It can change all assignment signs to either the most commonly used in a given file or a configured fixed one.
Default behaviour is autodetect for assignments from Keyword Calls and removing assignment signs in
*** Variables *** section. It can be freely configured.

So it happens because we already have bunch of
```
${ns}=    Keyword
```

So I guess this is fine...

pmtk:
I'm wondering if we should add a bit of code to scenario.sh to upgrade the RF, like: if robotenv exists, check the version, if outdated -> upgrade or remove+install.
It's still worth to inform others about the change (I think `VAR` is breaking change), but at least the process will be smoother.

PR #5291, opened 2025-08-06
OCPBUGS-55818: Remove redundant settings from the sample blueprint files

ggiguash:


packaging/blueprint/blueprint.toml.template
```diff
@@ -9,27 +9,18 @@ name = "microshift"
...
 [customizations.services]
 enabled = ["microshift"]
 
 [customizations.firewall]
-ports = ["22:tcp", "80:tcp", "443:tcp", "5353:udp", "6443:tcp", "30000-32767:tcp", "30000-32767:udp"]
+ports = ["22:tcp"]
 
 [customizations.firewall.services]
-enabled = ["mdns", "ssh", "http", "https"]
+enabled = ["ssh"]
```
pmtk: Do we even need both?
https://osbuild.org/docs/user-guide/blueprint-reference/#firewall
> If the distribution uses firewalld you can specify services listed by firewall-cmd --get-services in a customizations.firewall.services section:

```
$ cat /usr/lib/firewalld/services/ssh.xml

<?xml version="1.0" encoding="utf-8"?>
<service>
  <short>SSH</short>
  <description>Secure Shell (SSH) is a protocol for logging into and executing commands on remote machines. It provides secure encrypted communications. If you plan on accessing your machine remotely via SSH over a firewalled interface, enable this option. You need the openssh-server package installed for this option to be useful.</description>
  <port protocol="tcp" port="22"/>
</service>
```

ggiguash: Do you mean we do not even need to specify any ports?

I did not to be consistent with https://github.com/openshift/microshift/blob/main/packaging/imagemode/Containerfile.repobase#L19. I can remove it if you think it's better - please let me know.

pmtk: I mean that 
```
[customizations.firewall]
ports = ["22:tcp"]
```
and
```
[customizations.firewall.services]
enabled = ["ssh"]
```
do exactly the same - just the latter spares us remembering (yeah, `22` very hard 😉) ports (see other files in `/usr/lib/firewalld/services/`).

We can do one or the other. The decision is yours. If we want to educate on how the blueprint can look, we can leave both.

I don't have any preference - more of a fun fact.

pmtk:
If you wish to keep both options, just unhold
/hold
/lgtm

PR #5293, opened 2025-08-06
USHIFT-5723: Add test for GDP device allocation

kasturinarra:
**Which issue(s) this PR addresses**:  This PR tests configuring some device to be allocatable N times and then comparing it to the oc describe node

Closes #<Issue Number>

### DIFF
--- a/test/suites/optional/generic-device-plugin.robot
+++ b/test/suites/optional/generic-device-plugin.robot
@@ -14,7 +14,8 @@ Test Tags           generic-device-plugin
 
 
 *** Variables ***
-${NAMESPACE}    ${EMPTY}
+${NAMESPACE}        ${EMPTY}
+${DEVICE_COUNT}     5
 
 
 *** Test Cases ***
@@ -35,6 +36,18 @@ Verify that mountPath correctly renames the device within the container
     Wait For Job Completion And Check Logs
 
     [Teardown]    GDP Test Teardown
+
+Verify ttyUSB glob pattern device discovery and allocation
+    [Documentation]    Tests GDP with glob pattern to discover multiple ttyUSB devices and verify device allocation across multiple pods
+    [Tags]    glob-pattern
+    [Setup]    TtyUSB Glob Test Setup
+
+    # Create and verify pods with device allocation
+    Create Pod And Verify Allocation    serial-test-pod    2    2
+    Create Pod And Verify Allocation    serial-test-pod1    1    3
+    Create Pod And Verify Allocation    serial-test-pod2    2    5
+
+    [Teardown]    TtyUSB Glob Test Teardown
 
 
 *** Keywords ***
@@ -119,20 +132,21 @@ Create Test Job With Modified Script
 
 Wait Until Device Is Allocatable
     [Documentation]    Waits until device device.microshift.io/fakeserial is allocatable
+    [Arguments]    ${expected_count}=1
     ${node}=    Run With Kubeconfig    oc get node -o=name
     ${node_name}=    Remove String    ${node}    node/
     Wait Until Keyword Succeeds    60s    5s
-    ...    Device Should Be Allocatable    ${node_name}
+    ...    Device Should Be Allocatable    ${node_name}    ${expected_count}
 
 Device Should Be Allocatable
     [Documentation]    Checks if device device.microshift.io/fakeserial is allocatable
-    [Arguments]    ${node_name}
+    [Arguments]    ${node_name}    ${expected_count}=1
     ${device_amount}=    Oc Get JsonPath
     ...    node
     ...    ${EMPTY}
     ...    ${node_name}
     ...    .status.allocatable.device\\.microshift\\.io/fakeserial
-    Should Be Equal As Integers    ${device_amount}    1
+    Should Be Equal As Integers    ${device_amount}    ${expected_count}
 
 Wait For Job Completion And Check Logs
     [Documentation]    Waits for Job completion and checks Pod logs looking for 'Test successful' message
@@ -146,6 +160,64 @@ Wait For Job Completion And Check Logs
     ${logs}=    Oc Logs    ${pod}    ${NAMESPACE}
     Should Contain    ${logs}    Test successful
 
+Create TtyUSB Devices
+    [Documentation]    Creates dummy ttyUSB devices
+    FOR    ${i}    IN RANGE    ${DEVICE_COUNT}
+        Command Execution    sudo mknod --mode=666 /dev/ttyUSB${i} c 166 ${i}
+    END
+
+Remove Dummy TtyUSB Devices
+    [Documentation]    Removes dummy ttyUSB devices
+    FOR    ${i}    IN RANGE    ${DEVICE_COUNT}
+        Command Execution    sudo rm -f /dev/ttyUSB${i}
+    END
+
+TtyUSB Glob Test Setup
+    [Documentation]    Setup for ttyUSB glob pattern test - creates devices, configures GDP, waits for allocation
+    Create TtyUSB Devices
+    Drop In MicroShift Config    ${GDP_CONFIG_SERIAL_GLOB}    10-gdp
+    Restart MicroShift
+    Wait Until Device Is Allocatable    ${DEVICE_COUNT}
+
+TtyUSB Glob Test Teardown
+    [Documentation]    Cleanup for ttyUSB glob pattern test
+    Oc Delete    pod --all -n ${NAMESPACE}
+    Remove Dummy TtyUSB Devices
+    Disable GDP
+
+Verify Node Device Allocation
+    [Documentation]    Verifies device allocation on the node matches expected total
+    [Arguments]    ${expected_total_allocated}
+    ${node}=    Run With Kubeconfig    oc get node -o=name
+    ${node_name}=    Remove String    ${node}    node/
+    ${describe_output}=    Run With Kubeconfig    oc describe node ${node_name}
+    ${allocated_line}=    Get Lines Containing String    ${describe_output}    device.microshift.io/fakeserial
+    ${allocated_count}=    Get Regexp Matches
+    ...    ${allocated_line}
+    ...    device\\.microshift\\.io/fakeserial\\s+(\\d+)
+    ...    1
+    VAR    ${allocated_count}=    ${allocated_count[0]}
+    Should Be Equal As Integers    ${allocated_count}    ${expected_total_allocated}
+
+Create Pod And Verify Allocation
+    [Documentation]    Creates a pod using dynamic spec generation and verifies device allocation
+    [Arguments]    ${pod_name}    ${requested_devices}    ${expected_total_allocated}
+
+    # Generate pod spec dynamically
+    ${pod_spec}=    Get Ttyusb Pod Definition    ${pod_name}    ${requested_devices}
+
+    # Create and wait for pod
+    ${path}=    Create Random Temp File    ${pod_spec}
+    Oc Create    -f ${path} -n ${NAMESPACE}
+    Oc Wait    -n ${NAMESPACE} pod/${pod_name}    --for=condition=Ready --timeout=120s
+
+    # Verify correct number of devices allocated to pod
+    ${devices_in_pod}=    Oc Exec    ${pod_name}    ls /dev/ | grep ttyUSB | wc -l    ${NAMESPACE}
+    Should Be Equal As Integers    ${devices_in_pod}    ${requested_devices}
+
+    # Verify node shows correct allocation
+    Verify Node Device Allocation    ${expected_total_allocated}
+
 Teardown Suite With GDP Cleanup
     [Documentation]    Suite teardown that cleans up GDP configuration and restarts MicroShift
     # Clean up any remaining GDP configuration
@@ -153,5 +225,4 @@ Teardown Suite With GDP Cleanup
     # Restart MicroShift to clean state for next suite
     Restart MicroShift
     Restart Greenboot And Wait For Success
-    # Call original suite teardown
     Teardown Suite With Namespace
--- a/test/suites/optional/strings.py
+++ b/test/suites/optional/strings.py
@@ -19,6 +19,16 @@ genericDevicePlugin:
                mountPath: /dev/myrenamedserial
 '''
 
+GDP_CONFIG_SERIAL_GLOB = '''
+genericDevicePlugin:
+  status: Enabled
+  devices:
+     - name: fakeserial
+       groups:
+         - paths:
+             - path: /dev/ttyUSB*
+'''
+
 CONFIGMAP_PREAMBLE = '''
 apiVersion: v1
 kind: ConfigMap
@@ -34,6 +44,30 @@ data:
 
   fake-serial-communication.py: |
 '''
+
+
+def get_ttyusb_pod_definition(name: str, num_devices: int):
+    return f"""
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {name}
+spec:
+  containers:
+  - name: serialdevice-app-container
+    image: registry.access.redhat.com/ubi9/ubi:9.6
+    command: ["sleep", "infinity"]
+    resources:
+      limits:
+        device.microshift.io/fakeserial: "{num_devices}"
+    securityContext:
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop: ["ALL"]
+      runAsNonRoot: true
+      seccompProfile:
+        type: "RuntimeDefault"
+"""
 
 
 def append_to_preamble(content: str) -> str:


test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +137,62 @@ Wait For Job Completion And Check Logs
...
+    ${result}=    Run Process    oc    create    -f    ${path}    -n    ${NAMESPACE}
+    ...    shell=False    stderr=STDOUT    timeout=30s
+    Log    Pod creation error output: ${result.stdout}
+
+    # Verify the error contains the expected Kubernetes validation messages
+    Should Contain    ${result.stdout}    The Pod "fractional-test-pod" is invalid
+    Should Contain    ${result.stdout}    spec.containers[0].resources.limits[device.microshift.io/fakeserial]
+    Should Contain    ${result.stdout}    spec.containers[0].resources.requests[device.microshift.io/fakeserial]
+    Should Contain    ${result.stdout}    Invalid value: resource.Quantity
+    Log    SUCCESS: Pod creation correctly failed with Kubernetes fractional device validation error
```
pmtk: I think this isn't testing GDP but rather K8s itself.
I would recommend on not testing this.

https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/#Quantity: 
> Quantity is a fixed-point representation of a number
> ...
> 1.5 will be serialized as "1500m" - 1.5Gi will be serialized as "1536Mi"


https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/#discussion:
> Extended resources are opaque to Kubernetes; Kubernetes does not know anything about what they are. Kubernetes knows only that a Node has a certain number of them. Extended resources must be advertised in integer amounts. For example, a Node can advertise four dongles, but not 4.5 dongles

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +137,62 @@ Wait For Job Completion And Check Logs
...
+    ...    pod
+    ...    ${NAMESPACE}
+    ...    ${pod_name}
+    ...    .status.phase
+    Should Be Equal    ${status}    Pending
+
+    # Check for scheduling failure event
+    ${events}=    Run With Kubeconfig    oc describe pod ${pod_name} -n ${NAMESPACE}
+    Should Contain    ${events}    FailedScheduling
+    Should Contain    ${events}    Insufficient device.microshift.io/fuse
```
pmtk: I think we don't have to use fuse for these. We could just reuse existing config with fakeserial and also attempt to request more than available.

TBH this is, again, not really testing GDP but kubernetes.
To me, testing GDP would be: configuring some device to be allocatable N times and then comparing it to the `oc describe node`.
Whether K8s allows to over-request a device is most likely tested upstream (extended resources cannot be overallocated).

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +137,62 @@ Wait For Job Completion And Check Logs
...
+    Log    Pod creation error output: ${result.stdout}
+
+    # Verify the error contains the expected Kubernetes validation messages
+    Should Contain    ${result.stdout}    The Pod "fractional-test-pod" is invalid
+    Should Contain    ${result.stdout}    spec.containers[0].resources.limits[device.microshift.io/fakeserial]
+    Should Contain    ${result.stdout}    spec.containers[0].resources.requests[device.microshift.io/fakeserial]
+    Should Contain    ${result.stdout}    Invalid value: resource.Quantity
+    Log    SUCCESS: Pod creation correctly failed with Kubernetes fractional device validation error
+
+Verify GDP Config Fails Microshift Restart
```
pmtk:
```suggestion
Verify Invalid GDP Config Causes Microshift To Restart
```

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +137,62 @@ Wait For Job Completion And Check Logs
...
+    [Documentation]    Verifies that invalid GDP config causes microshift restart to fail
+    [Arguments]    ${invalid_config}
+    # Apply invalid config
+    Drop In MicroShift Config    ${invalid_config}    10-gdp
+    # Attempt restart and expect it to fail
+    ${result}=    Run Keyword And Expect Error    *
+    ...    Restart MicroShift
```
pmtk: I think we could drop comments like that. Keywords and args are self-explanatory in this case
```suggestion
    Drop In MicroShift Config    ${invalid_config}    10-gdp
    ${result}=    Run Keyword And Expect Error    *
    ...    Restart MicroShift
```

test/suites/optional/generic-device-plugin.robot
```diff
@@ -37,6 +38,28 @@ Sanity Test
     ...    Stop Script On Host
     ...    Disable GDP
 
+Negative Scenarios Test
```
pmtk: This test case lacks [Teardown] to properly clear up the machine for other optional suites.

test/suites/optional/strings.py
```diff
@@ -8,6 +8,110 @@
...
+  devices:
+     - name: fakeserial
+       groups:
+         - paths:
+             - path: /dev/ttyPipeB0
+     - name: fakeserial1
+       groups:
+         - paths:
+             - path: /dev/ttyPipeB0
+'''
```
pmtk: This is already verified in the unit tests: https://github.com/openshift/microshift/blob/632b914aad1556baf01ac095d460301a4b51d255/pkg/config/gdp_test.go#L125-L163

test/suites/optional/generic-device-plugin.robot
```diff
@@ -37,6 +38,28 @@ Sanity Test
...
+    [Tags]    negative
+
+    # Test 1: Fractional device limits should fail
+    Verify Pod Creation Fails With Fractional Limits
+
+    # Test 2: Invalid permissions should fail microshift restart
+    Verify GDP Config Fails Microshift Restart    ${GDP_CONFIG_INVALID_PERMISSIONS}
+
+    # Test 3: Invalid YAML structure should fail microshift restart
+    Verify GDP Config Fails Microshift Restart    ${GDP_CONFIG_INVALID_YAML}
```
pmtk: I don't think this is testing GDP. It's testing yaml deserialization

test/suites/optional/strings.py
```diff
@@ -8,6 +8,110 @@
...
+GDP_CONFIG_INVALID_PERMISSIONS = '''
+genericDevicePlugin:
+  status: Enabled
+  devices:
+     - name: fakeserial
+       groups:
+         - paths:
+             - path: /dev/ttyPipeB0
+               permissions: xyz
+'''
```
pmtk: This is tested in unit tests: https://github.com/openshift/microshift/blob/632b914aad1556baf01ac095d460301a4b51d255/pkg/config/gdp_test.go#L188-L215

test/suites/optional/generic-device-plugin.robot
kasturinarra: Done, removed.

test/suites/optional/generic-device-plugin.robot
kasturinarra: >To me, testing GDP would be: configuring some device to be allocatable N times and then comparing it to the oc describe node

Automating this as part of a new testcase

kasturinarra: Removed the count based allocation testing.

test/suites/optional/generic-device-plugin.robot
kasturinarra: Done.

test/suites/optional/generic-device-plugin.robot
kasturinarra: Done

test/suites/optional/generic-device-plugin.robot
kasturinarra: Done.

test/suites/optional/strings.py
kasturinarra: Done.

test/suites/optional/generic-device-plugin.robot
kasturinarra: okay. Removed.

test/suites/optional/strings.py
kasturinarra: Removed.

kasturinarra:
@pmtk Based on the above it looks like we do not need this test at all, is that right ?

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +128,71 @@ Wait For Job Completion And Check Logs
...
+    FOR    ${i}    IN RANGE    ${DEVICE_COUNT}
+        Command Execution    sudo mknod /dev/ttyUSB${i} c 166 ${i}
+        Command Execution    sudo chmod 666 /dev/ttyUSB${i}
+    END
+
+Remove Dummy TtyUSB Devices
+    [Documentation]    Removes dummy ttyUSB devices
+    FOR    ${i}    IN RANGE    ${DEVICE_COUNT}
+        Run Keyword And Ignore Error
+        ...    Command Execution    sudo rm -f /dev/ttyUSB${i}
```
pmtk: Errors are ignored by default in the Teardown (https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#test-setup-and-teardown)

```suggestion
        Command Execution    sudo rm -f /dev/ttyUSB${i}
```

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +128,71 @@ Wait For Job Completion And Check Logs
     ...    .items[*].metadata.name
     ${logs}=    Oc Logs    ${pod}    ${NAMESPACE}
     Should Contain    ${logs}    Test successful
+
+Create And Verify TtyUSB Devices
+    [Documentation]    Creates dummy ttyUSB devices
+    FOR    ${i}    IN RANGE    ${DEVICE_COUNT}
+        Command Execution    sudo mknod /dev/ttyUSB${i} c 166 ${i}
+        Command Execution    sudo chmod 666 /dev/ttyUSB${i}
```
pmtk:
```suggestion
        Command Execution    sudo mknod --mode=666 /dev/ttyUSB${i} c 166 ${i}
```

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +128,71 @@ Wait For Job Completion And Check Logs
     ...    .items[*].metadata.name
     ${logs}=    Oc Logs    ${pod}    ${NAMESPACE}
     Should Contain    ${logs}    Test successful
+
+Create And Verify TtyUSB Devices
```
pmtk: Keyword does not verify the devices

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +128,71 @@ Wait For Job Completion And Check Logs
...
+    [Documentation]    Setup for ttyUSB glob pattern test - creates devices, configures GDP, waits for allocation
+    Create And Verify TtyUSB Devices
+    Drop In MicroShift Config    ${GDP_CONFIG_SERIAL_GLOB}    10-gdp
+    Restart MicroShift
+    Wait Until Device Is Allocatable    ${DEVICE_COUNT}
+
+TtyUSB Glob Test Teardown
+    [Documentation]    Cleanup for ttyUSB glob pattern test
+    Run Keyword And Ignore Error
+    ...    Oc Delete    pod --all -n ${NAMESPACE}
```
pmtk:
```suggestion
    Oc Delete    pod --all -n ${NAMESPACE}
```

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +128,71 @@ Wait For Job Completion And Check Logs
...
+    Remove Dummy TtyUSB Devices
+    Disable GDP
+
+Verify Node Device Allocation
+    [Documentation]    Verifies device allocation on the node matches expected total
+    [Arguments]    ${expected_total_allocated}
+    ${node}=    Run With Kubeconfig    oc get node -o=name
+    ${node_name}=    Remove String    ${node}    node/
+    ${describe_output}=    Run With Kubeconfig    oc describe node ${node_name}
```
pmtk, line 193: Instead of using `oc describe`, you could use `Oc Get JsonPath` keyword.
Example: https://github.com/openshift/microshift/blob/3eec5d7da55f0e85eff32e9be0bd325d48601450/test/suites/tuned/profile.robot#L99

(I'm a bit ashamed I used `Run With Kubeconfig    oc get node -o=name` in this file, haha)

kasturinarra:
@pmtk updated the test as per our discussion, could you please help review again ? thanks !!

test/suites/optional/generic-device-plugin.robot
```diff
@@ -5,16 +5,18 @@ Resource            ../../resources/microshift-config.resource
 Resource            ../../resources/microshift-process.resource
 Resource            ../../resources/ostree-health.resource
 Variables           strings.py
+Library             Process
```
pmtk: Why do we need to add this?

test/suites/optional/generic-device-plugin.robot
kasturinarra: Done.

test/suites/optional/generic-device-plugin.robot
kasturinarra: Done.

test/suites/optional/generic-device-plugin.robot
kasturinarra: Updated the keyword accordingly.

test/suites/optional/generic-device-plugin.robot
kasturinarra: Done.

test/suites/optional/generic-device-plugin.robot
kasturinarra: @pmtk For some reason i am not seeing allocated_devices under `oc get jsonpath` so used `oc describe` actually. oc get JsonPath only gives capacity and allocatable but not allocated. https://privatebin.corp.redhat.com/?397bae5c0633b449#4tneRUszxyV7fe4N3Jcc7cQ9zm8YUbMfzRpNt1WhYE2m

test/suites/optional/generic-device-plugin.robot
kasturinarra: I can remove. some left over i believe.

test/suites/optional/generic-device-plugin.robot
pmtk: Yes, you're right. Upon some investigation it seems that this part of Describe is calculated at runtime by the CLI (it basically gets Node and Pods running on that Node and creates the tables, there's not one single API to get that info).

test/suites/optional/generic-device-plugin.robot
```diff
@@ -113,3 +127,68 @@ Wait For Job Completion And Check Logs
...
+    Create TtyUSB Devices
+    Drop In MicroShift Config    ${GDP_CONFIG_SERIAL_GLOB}    10-gdp
+    Restart MicroShift
+    Wait Until Device Is Allocatable    ${DEVICE_COUNT}
+
+TtyUSB Glob Test Teardown
+    [Documentation]    Cleanup for ttyUSB glob pattern test
+    Oc Delete    pod --all -n ${NAMESPACE}
+    Remove Dummy TtyUSB Devices
+    Disable GDP
```
pmtk, line 186: This keyword also restarts microshift - since every test in this suite is setting dropin and restarting MicroShift, we could just remove the dropin.

In other words go from:
> - setup test: set up dropin, restart ushift
> - run test
> - teardown: remove dropin, restart ushift
> - setup test: set up dropin, restart ushift
> ...

we could do
> - setup test: set up dropin, restart ushift
> - run test
> - teardown: remove dropin
> - setup test: set up dropin, restart ushift
> ...

pmtk (commented):
One comment about general suite/test logic, but we can fix it in another PR

test/suites/optional/generic-device-plugin.robot
kasturinarra: okay, once these two gets merged, i will check and fix the same.

pmtk: Surely you'll get conflicts after merging one or the other. So whichever goes first, when fixing conflicts we can fix this

kasturinarra:
@pmtk could you please help review again ? thanks !!

test/suites/optional/strings.py
```diff
@@ -19,6 +19,82 @@
...
+      limits:
+        device.microshift.io/fakeserial: "2"
+    securityContext:
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop: ["ALL"]
+      runAsNonRoot: true
+      seccompProfile:
+        type: "RuntimeDefault"
+'''
```
pmtk: These three are basically identical. May I propose changing them to a function?

```suggestion
def get_ttyusb_pod_definition(name: str, num_devices: int):
    return f"""
apiVersion: v1
kind: Pod
metadata:
  name: {name}
spec:
  containers:
  - name: serialdevice-app-container
    image: registry.access.redhat.com/ubi9/ubi:9.6
    command: ["sleep", "infinity"]
    resources:
      limits:
        device.microshift.io/fakeserial: "{num_devices}"
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop: ["ALL"]
      runAsNonRoot: true
      seccompProfile:
        type: "RuntimeDefault"
"""
```

Now, to obtain specific combination of name + amount of devices:

```
Get Ttyusb Pod Definition    serial-test-pod    2
Get Ttyusb Pod Definition    serial-test-pod1    1
Get Ttyusb Pod Definition    serial-test-pod2    2
```

kasturinarra: Done, addressed. @pmtk could you please help review again ?

test/suites/optional/generic-device-plugin.robot
```diff
@@ -146,12 +160,70 @@ Wait For Job Completion And Check Logs
...
+    ${pod_spec}=    Get Ttyusb Pod Definition    ${pod_name}    ${requested_devices}
+
+    # Create and wait for pod
+    ${path}=    Create Random Temp File    ${pod_spec}
+    Oc Create    -f ${path} -n ${NAMESPACE}
+    Oc Wait    -n ${NAMESPACE} pod/${pod_name}    --for=condition=Ready --timeout=120s
+
+    # Verify correct number of devices allocated to pod
+    ${devices_in_pod}=    Run With Kubeconfig
+    ...    oc exec -n ${NAMESPACE} ${pod_name} -- sh -c "ls /dev/ | grep ttyUSB | wc -l"
```
pmtk: Sorry, i just noticed this now. We have `Oc Exec` keyword in oc.robot, would it be enough for what we need here? If not, we can adjust the `Oc Exec`

kasturinarra: Let me check.

kasturinarra: Done, updated

PR #5299, opened 2025-08-07
NO-ISSUE: Automatically upgrade RF if installed & required versions are different

pmtk:


test/bin/scenario.sh
```diff
@@ -973,6 +974,17 @@ run_tests() {
     fi
     record_junit "${vmname}" "robot_framework_installed" "OK"
 
+    local -r rf_version_installed=$("${RF_VENV}/bin/pip" list --format json | jq -r '.[] | select(.name == "robotframework") | .version')
```
agullon, line 977: hi @pmtk this command does not work on my local environment (only tested on macOS):
```
venv ❯ pip list --format json
[{"name": "pip", "version": "25.0"}]
```
I suggest another approach:
```
venv ❯ robot --version || true
Robot Framework 6.0.2 (Python 3.9.19 on darwin)
```

agullon: tested it again, working fine:
```
❯ _output/robotenv/bin/pip list --format json | jq -r '.[] | select(.name == "robotframework") | .version'
6.0.2
```

PR #5311, opened 2025-08-11
USHIFT-5353: Collect SOS Report on test failure

pmtk:


scripts/microshift-sos-report.sh
```diff
@@ -13,22 +13,41 @@ function usage() {
     echo "  profiles: ${PROFILES}"
     echo "  plugins: ${PLUGINS}"
     echo ""
-    echo "Usage: ${SCRIPT_NAME} [--tmp-dir TMP-DIR]"
-    echo "   --tmp-dir TMP-DIR  Temporary directory for saving the report. Defaults to ${TEMPDIR}."
+    echo "Usage: ${SCRIPT_NAME} [--tmp-dir TMP-DIR] [--help] [sos-report-args...]"
+    echo "   --tmp-dir TMP-DIR     Temporary directory for saving the report. Defaults to ${TEMPDIR}."
+    echo "   --help                Show this help message and exit."
+    echo "   --profiles PROFILES   Profiles to include in the report. Defaults to ${PROFILES}."
+    echo "   --plugins PLUGINS     Plugins to include in the report. Defaults to ${PLUGINS}."
```
pacevedom: Maybe a note that this is overrideable with env var?

test/resources/sos-on-failure-listener.py
```diff
@@ -0,0 +1,72 @@
...
+        value = os.getenv("SKIP_SOS")
+        if value == "true":
+            BuiltIn().log("sos-on-failure-listener.py: SKIP_SOS is set to true, skipping SOS report collection")
+            return
+        BuiltIn().log("sos-on-failure-listener.py: Failure was detected, collecting SOS report")
+        BuiltIn().import_resource('microshift-host.resource')
+        cmd = "microshift-sos-report --profiles microshift --plugins ''"
+        if len(suite_namespaces) > 0 or len(test_namespaces) > 0:
+            cmd += f" --namespaces {','.join(suite_namespaces + test_namespaces)}"
+        stdout, _, _ = BuiltIn().run_keyword("Command Execution", cmd)
```
pacevedom, line 54: This will leave the report in /tmp iirc, can we copy it to the test dir? The functions in scenario.sh grab the journal too, but thats a nit.

pmtk: It's left for scenario.sh to copy it since the mechanism is already in place.

I think the logic to get the paths right might be a bit convoluted so for me it's good enough for now.
We could implement it when start getting problems where host crashes after sos report but before scenario ends and sos is copied to hypervisor.

scripts/microshift-sos-report.sh
pmtk: done

PR #5329, opened 2025-08-14
NO-ISSUE Add lifecycle hook test for pod creation with sleep action

copejon:


pacevedom:
/hold
This test is passing consistently in the non-optional job and we would be skipping it there too. We are working on finding a way to fix the optionals failures, this is tracked here https://issues.redhat.com/browse/USHIFT-5589.

PR #5342, opened 2025-08-19
USHIFT-6027: Add metrics-server user docs

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/docs/user/howto_metrics_server.md
+++ b/docs/user/howto_metrics_server.md
@@ -2,7 +2,7 @@
 This document describes the basic workflow and changes to deploy [metrics-server](https://github.com/kubernetes-sigs/metrics-server) in MicroShift.
 
 ## Create MicroShift cluster
-Use the instructions in the [Install MicroShift on RHEL for Edge](../contributor/rhel4edge_iso.md) document to configure a virtual machine running MicroShift.
+Use the instructions in the [Getting Started with MicroShift](getting_started.md) document to configure a virtual machine running MicroShift.
 
 Log into the virtual machine and run the following commands to configure the MicroShift access and check if the PODs are up and running.
 


docs/user/howto_metrics_server.md
```diff
@@ -0,0 +1,113 @@
+# Deploying metrics-server in MicroShift
+This document describes the basic workflow and changes to deploy [metrics-server](https://github.com/kubernetes-sigs/metrics-server) in MicroShift.
+
+## Create MicroShift cluster
+Use the instructions in the [Install MicroShift on RHEL for Edge](../contributor/rhel4edge_iso.md) document to configure a virtual machine running MicroShift.
```
ggiguash: Should we refer to getting started here since the current doc is in users' section?

docs/user/howto_metrics_server.md
```diff
@@ -0,0 +1,113 @@
...
+
+In order to use them, we need to mount it as a volume for the metrics-server and use the CA as an argument for the Deployment's command.
+```bash
+curl -sL https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml | \
+yq 'select(.kind == "Deployment") |= (
+.spec.template.spec.volumes += [{"name": "ca-bundle", "configMap": {"name": "kubelet-client-ca"}}] |
+.spec.template.spec.containers[0].volumeMounts += [{"name": "ca-bundle", "mountPath": "/var/run/secrets/kubernetes.io/certs/ca.crt", "subPath": "ca.crt", "readOnly": true}] |
+.spec.template.spec.containers[0].args += "--kubelet-certificate-authority=/var/run/secrets/kubernetes.io/certs/ca.crt"
+)' | \
+oc apply -f -
```
ggiguash, line 42: A nit: is there a value in "simplifying" this command by splitting it?

docs/user/howto_metrics_server.md
pacevedom: Done

docs/user/howto_metrics_server.md
pacevedom: The alternative is to manually edit the manifest, but I thought a one-liner you could copy would be more convenient. Cant really split it into different commands because it needs to be in its final form before running `oc apply -f`. Or do you mean split into different lines?

PR #5354, opened 2025-08-21
USHIFT-5998: Give warning to user if no rpm advisory is present

kasturinarra:
**Which issue(s) this PR addresses**: RPM advisories are still shipped through erratas, so in case script does not find rpm advisory in konflux release data, a warning is shown asking the user to run advisory_publication_report.sh that supports errata advisories from 4.19 branch.

Closes #<Issue Number>


kasturinarra:
@agullon could you please help review ? thanks !!

PR #5389, opened 2025-09-02
USHIFT-6077: missing workload annotation for cert-manager-operator pods

agullon:
Add `cert-manager-operator` namespace to `All Pods Should Be Annotated As Management` RF test case to prevent test case from failing.


agullon:
@eslutsky can you please review it?

PR #5394, opened 2025-09-03
OCPBUGS-61198: Verify http proxy environment variables absense in standard suite

ggiguash:


test/suites/standard2/configuration.robot
```diff
@@ -208,3 +216,16 @@ LVMS Is Deployed
...
+    ...    in the bootc image used to install the system.
+    # Obtain the current bootc image reference
+    ${bootc_image}=    Command Should Work    bootc status --json | jq -r .spec.image.image
+    # Inspect the bootc image environment variables
+    ${env_vars}=    Command Should Work
+    ...    skopeo inspect --authfile /etc/crio/openshift-pull-secret --config docker://${bootc_image} | jq -r '.config.Env'
+    # Verify that the environment variables are not defined
+    Should Not Contain    ${env_vars}    HTTP_PROXY
+    Should Not Contain    ${env_vars}    HTTPS_PROXY
+    Should Not Contain    ${env_vars}    NO_PROXY
```
pacevedom: Can these variables be defined in lower case too? Should we use the `ignore_case` option?

pacevedom:
/lgtm cancel
Oops, miscommented before

test/suites/standard2/configuration.robot
ggiguash: These are environment, so I think we're looking for something very specific.
I can add ignore_case option, just in case :).

pacevedom (approved):
/lgtm

PR #5423, opened 2025-09-08
USHIFT-6051: Add support to filter microshift tests

kasturinarra:
**Which issue(s) this PR addresses**: This PR address filtering out microshift tests based on components and destructive test cases.

Closes #[USHIFT-6051](https://issues.redhat.com//browse/USHIFT-6051)

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -1128,8 +1128,32 @@ run_gingko_tests() {
     export KUBECONFIG="${kubeconfig}"
     record_junit "${vmname}" "setup_kubeconfig" "OK"
 
+    # Create case selection file
+    local case_selected="${test_results_dir}/case_selected"
+
+    # Get all MicroShift tests using dry-run
+    "${GINKGO_TEST_BINARY}" run all --dry-run | grep -E "MicroShift" > "${case_selected}"
+
+    # Apply filters if provided
+    if [[ $# -gt 0 ]]; then
+        local -r filter_pattern="$1"
+        echo "Applying filter: ${filter_pattern}"
+        if [[ "${filter_pattern}" == ~* ]]; then
+            sed -i "/${filter_pattern#~}/d" "${case_selected}"
+        else
+            sed -i -n "/${filter_pattern}/p" "${case_selected}"
+        fi
+        echo "Running Ginkgo tests with MicroShift filter and additional filter: ${filter_pattern}..."
+    else
+        echo "Running Ginkgo tests with MicroShift filter..."
+    fi
+
+    echo "------------------Selected test cases------------------"
+    cat "${case_selected}"
+    echo "-----------------------------------------------------"
+
     # Run the tests and capture output
-    if ! "${GINKGO_TEST_BINARY}" run all --dry-run | grep "MicroShift" | "${GINKGO_TEST_BINARY}" run -f - --timeout 60m 2>&1 | tee "${test_results_dir}/test-output.log"; then
+    if ! "${GINKGO_TEST_BINARY}" run --timeout 60m --junit-dir="${test_results_dir}" -f "${case_selected}" 2>&1 | tee "${test_results_dir}/test-output.log"; then
         record_junit "${vmname}" "run_gingko_tests" "FAILED"
         return 1
     fi


kasturinarra:
@ggiguash @agullon could you please help review ? thanks !!

test/bin/scenario.sh
```diff
@@ -1128,12 +1135,45 @@ run_gingko_tests() {
...
+        echo "Running Gingko tests with MicroShift filter..."
+    fi
+
+    echo "------------------Selected test cases------------------"
+    cat "${case_selected}"
+    echo "-----------------------------------------------------"
+
     # Run the tests and capture output
-    if ! "${GINKGO_TEST_BINARY}" run all --dry-run | grep "MicroShift" | "${GINKGO_TEST_BINARY}" run -f - --timeout 60m 2>&1 | tee "${test_results_dir}/test-output.log"; then
+    if ! "${GINKGO_TEST_BINARY}" run --timeout 60m --junit-dir="${test_results_dir}" -f "${case_selected}" "$@" 2>&1 | tee "${test_results_dir}/test-output.log"; then
```
ggiguash: Why do we need "$@" here?

test/bin/scenario.sh
```diff
@@ -1128,12 +1135,45 @@ run_gingko_tests() {
...
+
+    # Get all MicroShift tests using dry-run
+    "${GINKGO_TEST_BINARY}" run all --dry-run | grep -E "MicroShift" > "${case_selected}"
+
+    # Apply filter if provided
+    if [[ -n "${filter}" ]]; then
+        echo "Applying filter: ${filter}"
+        local temp_file="${case_selected}.tmp"
+
+        if [[ "${filter}" == ~* ]]; then
```
ggiguash: Instead of this pseudo parsing of filter expression, can we provide filters in grep -E format? 
This should allow us to simplify the logic.

test/bin/scenario.sh
```diff
@@ -1087,6 +1087,13 @@ run_gingko_tests() {
     local vmname="${1}"
     shift
 
+    # Parse optional filter argument (can be ~Disruptive or Multus)
+    local filter=""
```
ggiguash: We could probably assume that "$@" is the filter here, right? We do not have any other positional arguments, do we?

test/bin/scenario.sh
kasturinarra: I would want to support both inclusion and exclusion patterns in the same filter list and using it this way i feel it would be much simpler and intuitive and matches the current QE pipelines.

test/bin/scenario.sh
kasturinarra: We do not. good catch, removed.

test/bin/scenario.sh
```diff
@@ -1128,12 +1128,45 @@ run_gingko_tests() {
...
+            echo "Applying filter: ${filter_pattern}"
+            if [[ "${filter_pattern}" == ~* ]]; then
+                # Handle exclude filter (starts with ~)
+                local exclude_pattern="${filter_pattern#~}"
+                grep -v -E "${exclude_pattern}" "${case_selected}" > "${temp_file}"
+            else
+                # Handle include filter
+                grep -E "${filter_pattern}" "${case_selected}" > "${temp_file}"
+            fi
+            mv "${temp_file}" "${case_selected}"
```
ggiguash: I'm not sure how this logic would work with multiple filters. Say you want "t1 t2" filtering.
- line 1135: case_selected contains all MicroShift tests to start with
- line 1148: temp_file contains all t1 tests
- line 1150: temp_file contents are copied into case_selected
- line 1148: temp_file would be empty for t2 because case_selected only contains t1 tests

So, the current logic seems to work for expressions like "~t1 t2", but not for multiple inclusive statements like "t1 t2".

What am I missing?

ggiguash: To make it work like you want it, probably, we should first process all not (~) filters and then the rest of them.

kasturinarra: @ggiguash so basially here is the idea, I looked at the current job_definitions that QE has and i see we are specifying one filter at a time i.e either ~Disruptive or Multus or Storage, so the code works for either of these. Secondly, if we would like to run any component but not include disruptive cases that works too. But did see the job_definitions having to run multiple components at once like the one you said Storage Multus. I can check with john to see if we need as you have suggested above and i will try doing the same.

Example QE job definition file: https://privatebin.corp.redhat.com/?c6904b3c0d58c666#H6wgjSWodMiJCpAf8ZQnhQKeeRp4f5ZD3dfKyPkPSCxs   you can look at the gingko filters.

ggiguash: I think we need to fix the code. The current version allows for multiple filters, but it only works in certain filter type combination.

kasturinarra: I fixed the code to handle just a single filter after discussing with john and it aligns with the current QE pipelines. Let me know if this looks good to you .

https://redhat-internal.slack.com/archives/C042LATR98R/p1757525464695179

test/bin/scenario.sh
```diff
@@ -1128,12 +1128,44 @@ run_gingko_tests() {
...
+            # Handle exclude filter (starts with ~)
+            local exclude_pattern="${filter_pattern#~}"
+            grep -v -E "${exclude_pattern}" "${case_selected}" > "${temp_file}"
+        else
+            # Handle include filter
+            grep -E "${filter_pattern}" "${case_selected}" > "${temp_file}"
+        fi
+        mv "${temp_file}" "${case_selected}"
+        echo "Running Ginkgo tests with MicroShift filter and additional filter: ${filter_pattern}..."
+    else
```
ggiguash, line 1147: This code can be optimized without temporary files when using sed. Can you consider this instead?

```
if [[ $# -gt 0 ]]; then
    local -r filter_pattern=$1
    echo "Applying filter: $filter_pattern"
    if [[ $filter_pattern == ~* ]]; then
        sed -i "/${filter_pattern#~}/d" "${case_selected}"
    else
        sed -i -n "/$filter_pattern/p" "${case_selected}"
    fi
    echo "Running Ginkgo tests with MicroShift filter and additional filter: $filter_pattern..."
else
    echo "Running Ginkgo tests with MicroShift filter..."
fi
```

test/bin/scenario.sh
```diff
@@ -1128,12 +1128,44 @@ run_gingko_tests() {
...
+
     # Run the tests and capture output
-    if ! "${GINKGO_TEST_BINARY}" run all --dry-run | grep "MicroShift" | "${GINKGO_TEST_BINARY}" run -f - --timeout 60m 2>&1 | tee "${test_results_dir}/test-output.log"; then
+    if ! "${GINKGO_TEST_BINARY}" run --timeout 60m --junit-dir="${test_results_dir}" -f "${case_selected}" 2>&1 | tee "${test_results_dir}/test-output.log"; then
         record_junit "${vmname}" "run_gingko_tests" "FAILED"
         return 1
     fi
 
+    # Clean up temporary file
+    rm -f "${case_selected}"
```
ggiguash: I'm not sure we should delete this file. Let's keep it for post-run examination?
It might be useful for local debugging when the terminal output is lost.

test/bin/scenario.sh
kasturinarra: Addressed.

PR #5445, opened 2025-09-10
USHIFT-6035: Switch to 4.20-rc packages in common_versions.sh

ggiguash:
Also addresses [USHIFT-5819](https://issues.redhat.com//browse/USHIFT-5819) by reenabling brew tests


pacevedom (approved):
/lgtm

PR #5446, opened 2025-09-10
USHIFT-6098: create ingress-class during microshift startup

eslutsky:

Closes #<Issue Number>

### DIFF
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -17,6 +17,8 @@ import (
 	"github.com/openshift/microshift/pkg/assets"
 	"github.com/openshift/microshift/pkg/config"
 	"github.com/openshift/microshift/pkg/util/cryptomaterial"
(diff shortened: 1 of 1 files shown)


pkg/components/controllers.go
```diff
@@ -167,6 +172,21 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 			klog.Warningf("Failed to delete cluster roles %v: %v", clusterRole, err)
 			return err
 		}
+		// Delete ingress class (cluster-scoped resource)
```
pacevedom, line 175: This is just my ignorance, but why delete when we are creating it again a few lines below?

eslutsky: thats only when the default router is disabled

pacevedom: Wait, missed the if block for the router disable logic, my bad, sorry!

PR #5475, opened 2025-09-16
NO-ISSUE: Fix repository configuration in the main branch

ggiguash:


scripts/devenv-builder/configure-vm.sh
```diff
@@ -320,9 +304,9 @@ function build_and_install_microshift_rpms() {
         else
             SKIPPED_RPMS="gateway-api"
         fi
-
+
```
pacevedom: blanks

scripts/devenv-builder/configure-vm.sh
```diff
@@ -212,26 +212,10 @@ EOF
         if [[ "${PREVIOUS_RHOCP}" =~ ^[0-9]{2} ]]; then
             sudo subscription-manager repos --enable "rhocp-4.${PREVIOUS_RHOCP}-for-rhel-9-$(uname -m)-rpms"
         else
-            # When its early in the new release cycle the previous release may not have a repo yet, use the beta.
```
pacevedom, line 215: This is removed to make it fail on purpose when cri-tools/cri-o have not been updated, right?

test/bin/common_versions.sh
pacevedom, line 1: I am missing some of the repo adjustments, check https://github.com/openshift/microshift/pull/5461/commits/fb2a5e83dd6580dfac851d7f4ebc87b0a058d922 (from a closed PR, but this commit was valid)

scripts/devenv-builder/configure-vm.sh
```diff
@@ -320,9 +304,9 @@ function build_and_install_microshift_rpms() {
         else
             SKIPPED_RPMS="gateway-api"
         fi
-
+        
         # shellcheck disable=SC2046,SC2086
-        "${DNF_RETRY}" "localinstall" "$(find ~/microshift/_output/rpmbuild/RPMS -type f -name "*.rpm" $(printf ' -not -name *%s* ' ${SKIPPED_RPMS}))"
+        "${DNF_RETRY}" "localinstall" "$(find ~/microshift/_output/rpmbuild/RPMS -type f -name "*.rpm" $(printf ' -not -name *%s* ' ${SKIPPED_RPMS}))"
```
pacevedom: space at the end

scripts/devenv-builder/configure-vm.sh
ggiguash: Yes, the cri-o update unfortunately always comes in parts: first cri-tools and then cri-o.
We've been updating cri-o versions in makefile.spec rather than adding repo fallbacks.

So, if it's possible to resolve the problem with the version update, it would be pity to complicate repository setup even more, wouldn't it?

PR #5487, opened 2025-09-18
USHIFT-6147: Observability RPM package must be built by default

ggiguash:


ggiguash:
We should remove ccf7d5c4c1a16a4953d649f63f60bb18e87cee0a commit as it disables the "wrong" telemetry.

/close

PR #5491, opened 2025-09-18
introduce hosts plugin into coreDNS

eslutsky:

Closes #<Issue Number>


pkg/controllers/hostswatcher.go
pacevedom, line 120: Should it signal the removal/rename of the file? At that point it will fail in the next restart but remain working until then. Maybe just a log line.

pacevedom, line 195: I am missing RBAC for this specific config map, only CoreDNS pods should be able to read it and only MicroShift should be able to write it.

pkg/config/dns.go
pacevedom, line 61: My only concern here is the exposure of sensitive files in the system. Other that that looking good!

pkg/controllers/hostswatcher.go
pacevedom: Wait, just saw the files. Disregard this comment, my bad.

pkg/config/dns.go
pacevedom: This has been solved now. Ensure we document it in the enhancement and in the docs.

PR #5509, opened 2025-09-22
USHIFT-5548 Disable telemetry for community builds

copejon:

Closes #<Issue Number>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,11 @@ SRC_ROOT :=$(shell pwd)
 
 WITH_KINDNET ?= 0
 WITH_TOPOLVM ?= 0
+# VARIANT: specify this make variable to build microshift with community feature set.
+# If VARIANT is unset or != community, build MicroShift with enterprise feature set, which requires openshift pull-secret.
+# OCP telemetry is disabled in community version.
+VARIANT ?= enterprise
+
 OUTPUT_DIR :=_output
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -32,6 +32,11 @@ SRC_ROOT :=$(shell pwd)
 
 WITH_KINDNET ?= 0
 WITH_TOPOLVM ?= 0
+# VARIANT: specify this make variable to build microshift with community feature set.
+# If VARIANT is unset or != community, build MicroShift with enterprise feature set, which requires openshift pull-secret.
+# OCP telemetry is disabled in community version.
+VARIANT ?= enterprise
```
ggiguash: A small request about the naming of the var - let's make it more specific, please.
E.g. MICROSHIFT_VARIANT

pkg/controllers/telemetry.go
```diff
@@ -59,7 +60,11 @@ func (t *TelemetryManager) Run(ctx context.Context, ready chan<- struct{}, stopp
 	// to stop gracefully.
 	defer close(stopped)
 	if t.config.Telemetry.Status == config.StatusDisabled {
-		klog.Info("Telemetry is disabled")
+		if info := version.Get(); info.Variant == version.VariantCommunity {
```
pacevedom: Should we move this outside the if statement? That way we dont even bother checking what was configured.

PR #5518, opened 2025-09-24
USHIFT-6175: Ginkgo releases tests are failing in CI

agullon:
- Filter out Ginkgo disruptive tests adding `~Disruptive` filter
 - Add global timeout for all the Ginkgo tests suite
 - Change `go build` custom command to create the binary to run Ginkgo test to `make build`
 - Add python script to remove 'Monitor cluster while tests execute' test case from ginkgo output junit xml file


test/bin/scenario.sh
```diff
@@ -1152,8 +1152,21 @@ run_gingko_tests() {
...
+    # exit after 5m, it sends the KILL signal to terminate the process.
+    local timeout_ginkgo="timeout -v --kill-after=5m ${TEST_EXECUTION_TIMEOUT} ${GINKGO_TEST_BINARY}"
+    if [ -t 0 ]; then
+        # Disable timeout for interactive mode when stdin is a terminal.
+        # This is necessary for proper handling of test interruption by user.
+        timeout_ginkgo="${GINKGO_TEST_BINARY}"
+    fi
+
+    # Run the tests and capture output with 10m timeout for every test case
+    if ! "${timeout_ginkgo}" run --timeout 10m --junit-dir="${test_results_dir}" -f "${case_selected}" 2>&1 | tee "${test_results_dir}/test-output.log"; then
```
kasturinarra: Is not 10m too less ?

test/bin/scenario.sh
```diff
@@ -1152,8 +1152,21 @@ run_gingko_tests() {
     cat "${case_selected}"
     echo "-----------------------------------------------------"
 
-    # Run the tests and capture output
-    if ! "${GINKGO_TEST_BINARY}" run --timeout 60m --junit-dir="${test_results_dir}" -f "${case_selected}" 2>&1 | tee "${test_results_dir}/test-output.log"; then
+    # Make sure the test execution times out after a predefined period.
+    # The 'timeout' command sends the HUP signal and, if the test does not
+    # exit after 5m, it sends the KILL signal to terminate the process.
```
kasturinarra, line 1157: also should we remove some of these comments ? may be lot of comments  ?

test/bin/scenario.sh
agullon: for every test case, I think is more than enough. For example, on Jenkins, for release-4.20 branch 59 tests take less than 23 mins to run.

test/bin/scenario.sh
agullon: I found them very useful to understand what does `--kill-after=5m` mean. So I think we should keep them.

kasturinarra:
/lgtm
From the code perspective.

test/bin/scenario.sh
```diff
@@ -1139,7 +1139,7 @@ run_gingko_tests() {
         local -r filter_pattern="$1"
         echo "Applying filter: ${filter_pattern}"
         if [[ "${filter_pattern}" == ~* ]]; then
-            sed -i "/${filter_pattern#~}/d" "${case_selected}"
+            sed -i "/${filter_pattern#\~}/d" "${case_selected}"
```
kasturinarra, line 1142: why do we need this ?

test/scenarios-bootc/releases/el96-lrel@ginkgo-tests.sh
```diff
@@ -12,5 +12,5 @@ scenario_remove_vms() {
 }
 
 scenario_run_tests() {
-    run_gingko_tests host1
+    run_gingko_tests host1 "~Disruptive"
```
kasturinarra, line 32: Here as well i think you could give with out doublequotes.

test/bin/scenario.sh
agullon: to escape the `~` char, otherwise it doesn't work, [you can check this log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_microshift/5518/pull-ci-openshift-microshift-main-e2e-aws-tests-bootc-release-arm/1970929886965010432/artifacts/e2e-aws-tests-bootc-release-arm/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el96-lrel@ginkgo-tests/boot_and_run.log).

I also check it manually:
```
$ echo "${pattern#~}"
~disruptive
$ echo "${pattern#\~}"
disruptive

test/scenarios-bootc/releases/el96-lrel@ginkgo-tests.sh
agullon: yes, I agree, because of the special `~` I prefer to add quotes because it's a string.

kasturinarra: okay

PR #5570, opened 2025-10-07
OCPEDGE-2037: add-node command

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/pkg/controllers/etcd.go
+++ b/pkg/controllers/etcd.go
@@ -169,29 +169,58 @@ func stopMicroshiftEtcdScopeIfExists() error {
 }
 
 func checkIfEtcdIsReady(ctx context.Context) error {
-	client, err := getEtcdClient(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to obtain etcd client: %v", err)
-	}
-	defer func() { _ = client.Close() }()
+	var client *clientv3.Client
+	defer func() {
+		if client != nil {
+			_ = client.Close()
+		}
+	}()
 
-	s, err := client.Status(ctx, "localhost:2379")
-	if err != nil {
-		return fmt.Errorf("failed to get etcd status: %v", err)
-	}
-	if s.IsLearner {
-		return nil
-	}
+	for attempt := 0; attempt < HealthCheckRetries; attempt++ {
+		if client == nil {
+			var err error
+			client, err = getEtcdClient(ctx)
+			if err != nil {
+				klog.Infof("failed to obtain etcd client: %v", err)
+				if attempt < HealthCheckRetries-1 {
+					time.Sleep(HealthCheckWait)
+					continue
+				}
+				return fmt.Errorf("failed to obtain etcd client after %d attempts: %v", HealthCheckRetries, err)
+			}
+		}
 
-	for i := 0; i < HealthCheckRetries; i++ {
-		time.Sleep(HealthCheckWait)
+		s, err := client.Status(ctx, "localhost:2379")
+		if err != nil {
+			_ = client.Close()
+			client = nil
+			klog.Infof("failed to get etcd status: %v", err)
+			if attempt < HealthCheckRetries-1 {
+				time.Sleep(HealthCheckWait)
+				continue
+			}
+			return fmt.Errorf("failed to get etcd status after %d attempts: %v", HealthCheckRetries, err)
+		}
+
+		// If my own instance is a learner I dont need to check readiness because apiserver is going
+		// to connect to other voting members in the cluster.
+		if s.IsLearner {
+			return nil
+		}
+
 		if _, err = client.Get(ctx, "health"); err == nil {
 			return nil
 		} else {
+			_ = client.Close()
+			client = nil
 			klog.Infof("etcd not ready yet: %v", err)
 			if err == context.Canceled {
 				return err
 			}
+		}
+
+		if attempt < HealthCheckRetries-1 {
+			time.Sleep(HealthCheckWait)
 		}
 	}
 	return fmt.Errorf("etcd still not healthy after checking %d times", HealthCheckRetries)
--- a/pkg/controllers/kube-apiserver.go
+++ b/pkg/controllers/kube-apiserver.go
@@ -164,7 +164,7 @@ func (s *KubeAPIServer) configure(ctx context.Context, cfg *config.Config) error
 		}
 	}
 
-	etcdServers, err := discoverEtcdServers(ctx, s.configuration.Node.HostnameOverride, s.configuration.BootstrapKubeConfigPath())
+	etcdServers, err := discoverEtcdServers(ctx, s.configuration.BootstrapKubeConfigPath())
 	if err != nil {
 		return fmt.Errorf("failed to discover etcd servers: %w", err)
 	}
@@ -412,7 +412,7 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped 
 	}
 }
 
-func discoverEtcdServers(ctx context.Context, hostname, kubeconfigPath string) ([]string, error) {
+func discoverEtcdServers(ctx context.Context, kubeconfigPath string) ([]string, error) {
 	certsDir := cryptomaterial.CertsDirectory(config.DataDir)
 	etcdPeerCertDir := cryptomaterial.EtcdPeerCertDir(certsDir)
 
@@ -441,39 +441,49 @@ func discoverEtcdServers(ctx context.Context, hostname, kubeconfigPath string) (
 	if err != nil {
 		return nil, fmt.Errorf("failed to get etcd status: %w", err)
 	}
-	if st.IsLearner {
-		//TODO if its a learner I need to take the server from the current non-learner members. Use the bootstrap for that.
-		kubeconfig, err := clientcmd.LoadFromFile(kubeconfigPath)
-		if err != nil {
-			return nil, fmt.Errorf("failed to load bootstrap kubeconfig: %w", err)
-		}
 
-		if kubeconfig == nil || kubeconfig.Clusters == nil || len(kubeconfig.Clusters) == 0 {
-			return nil, fmt.Errorf("invalid bootstrap kubeconfig: no clusters found")
-		}
+	// If I am not a learner it means I am a voting member, so connecting to my own etcd instance
+	// is fine because everything is synced.
+	if !st.IsLearner {
+		return []string{"https://localhost:2379"}, nil
+	}
 
-		var etcdHost string
-		for _, cluster := range kubeconfig.Clusters {
-			etcdHost = cluster.Server
-			break
-		}
+	// If I am a learner I need to connect to a member, retrieve the list of voting
+	// members and connect to all of them.
+	kubeconfig, err := clientcmd.LoadFromFile(kubeconfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load bootstrap kubeconfig: %w", err)
+	}
 
-		if etcdHost == "" {
-			return nil, fmt.Errorf("failed to extract etcd hostname from bootstrap kubeconfig")
-		}
+	if kubeconfig == nil || kubeconfig.Clusters == nil || len(kubeconfig.Clusters) == 0 {
+		return nil, fmt.Errorf("invalid bootstrap kubeconfig: no clusters found")
+	}
 
-		etcdHost = strings.TrimPrefix(etcdHost, "https://")
-		etcdHost, _, _ = net.SplitHostPort(etcdHost)
-		etcdHost = fmt.Sprintf("https://%s", net.JoinHostPort(etcdHost, "2379"))
-		client, err = clientv3.New(clientv3.Config{
-			DialTimeout: 5 * time.Second,
-			Endpoints:   []string{etcdHost},
-			TLS:         tlsConfig,
-			Context:     ctx,
-		})
-		if err != nil {
-			return nil, fmt.Errorf("failed to create etcd client: %w", err)
-		}
+	if len(kubeconfig.Clusters) > 1 {
+		return nil, fmt.Errorf("invalid bootstrap kubeconfig: multiple clusters found")
+	}
+
+	var etcdHost string
+	for _, cluster := range kubeconfig.Clusters {
+		etcdHost = cluster.Server
+		break
+	}
+
+	if etcdHost == "" {
+		return nil, fmt.Errorf("failed to extract etcd hostname from bootstrap kubeconfig")
+	}
+
+	etcdHost = strings.TrimPrefix(etcdHost, "https://")
+	etcdHost, _, _ = net.SplitHostPort(etcdHost)
+	etcdHost = fmt.Sprintf("https://%s", net.JoinHostPort(etcdHost, "2379"))
+	client, err = clientv3.New(clientv3.Config{
+		DialTimeout: 5 * time.Second,
+		Endpoints:   []string{etcdHost},
+		TLS:         tlsConfig,
+		Context:     ctx,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create etcd client: %w", err)
 	}
 
 	resp, err := client.MemberList(ctx)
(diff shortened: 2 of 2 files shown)


pkg/cmd/addnode.go
```diff
@@ -0,0 +1,555 @@
...
+
+type JoinClusterOptions struct {
+	KubeconfigPath string
+	Timeout        time.Duration
+	Learner        bool
+}
+
+func NewJoinClusterCommand() *cobra.Command {
+	opts := &JoinClusterOptions{
+		KubeconfigPath: "/var/lib/microshift/resources/kubeadmin/bootstrap/kubeconfig",
```
pmtk: It's not the same as the one stored by `configureBootstrapKubeconfig()` and used by kubelet, and it's different dir from the ones cleaned in the `cleanupMicroShiftData()`, right?
I find this value a bit confusing, it's very similar and initially I thought it's the same one and we're removing it mid process...

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,555 @@
...
+	nodeName := cfg.CanonicalNodeName()
+	if isNodeAlreadyInCluster(ctx, client, nodeName) {
+		klog.Infof("Node %s is already part of the cluster. Skipping join process.", nodeName)
+		return nil
+	}
+
+	if err := cleanupMicroShiftData(cfg); err != nil {
+		return fmt.Errorf("failed to cleanup MicroShift data directories: %w", err)
+	}
+	klog.Info("MicroShift data directories cleaned up successfully")
```
pmtk, line 110: Should there be a prompt before clearing the data? Or maybe an option like `--erase`. Or maybe we don't need to worry because that's community only and thus "unsupported"?

etcd/cmd/microshift-etcd/run.go
```diff
@@ -217,3 +217,35 @@ func checkFragmentationPercentage(ondisk, inuse int64) float64 {
...
+	return filepath.Join(config.DataDir, "etcd", "config")
+}
+
+func updateConfigFromFile(etcdCfg *etcd.Config, configPath string) {
+	data, err := os.ReadFile(configPath)
+	if err != nil {
+		klog.Errorf("failed to read config file: %v", err)
+		return
+	}
+	lines := strings.Split(string(data), "\n")
```
pmtk, line 231: If we control the file, why not use json?

pkg/components/certificateauthority.go
```diff
@@ -0,0 +1,167 @@
...
+	// Resource namespace
+	caResourceNamespace = "kube-system"
+)
+
+var (
+	CertificateAuthorityResources = []struct {
+		Name string
+		Dir  string
+	}{
+		{Name: "kube-control-plane-signer", Dir: cryptomaterial.KubeControlPlaneSignerCertDir(cryptomaterial.CertsDirectory(config.DataDir))},
```
pmtk, line 30: How this would compare to asking 1st microshift to sign certs? Instead of publishing CAs

pacevedom:
/hold
Found a bug with the prerun checks in the second node.

pacevedom:
/hold
Investigating CNCF conformance failures

pkg/config/kubeconfig.go
```diff
@@ -27,3 +30,12 @@ func (cfg *Config) KubeConfigAdminPath(id string) string {
 func (cfg *Config) KubeConfigRootAdminPath() string {
 	return filepath.Join(DataDir, "resources", string(KubeAdmin))
 }
+
+func (cfg *Config) BootstrapKubeConfigPath() string {
+	return filepath.Join(DataDir, "bootstrap", "kubeconfig")
+}
+
+func (cfg *Config) BootstrapKubeConfigExists() bool {
+	_, err := os.Stat(cfg.BootstrapKubeConfigPath())
```
pmtk: Check `util` pkg for `PathExists` or `PathExistsAndIsNotEmpty`

pkg/controllers/etcd.go
```diff
@@ -175,6 +175,14 @@ func checkIfEtcdIsReady(ctx context.Context) error {
 	}
 	defer func() { _ = client.Close() }()
 
+	s, err := client.Status(ctx, "localhost:2379")
```
pmtk: Could this call be too soon? Or it will always make it to the destination?

pkg/controllers/etcd.go
```diff
@@ -175,6 +175,14 @@ func checkIfEtcdIsReady(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get etcd status: %v", err)
+	}
+	if s.IsLearner {
+		return nil
```
pmtk: Perhaps a comment why we don't check health for a learner?

pkg/controllers/kube-apiserver.go
```diff
@@ -407,3 +411,91 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+		return nil, fmt.Errorf("failed to create etcd client: %w", err)
+	}
+	defer func() { _ = client.Close() }()
+
+	st, err := client.Status(ctx, "localhost:2379")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get etcd status: %w", err)
+	}
+	if st.IsLearner {
+		//TODO if its a learner I need to take the server from the current non-learner members. Use the bootstrap for that.
```
pmtk: What about this TODO?

pkg/controllers/kube-apiserver.go
```diff
@@ -407,3 +411,91 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+
+		if kubeconfig == nil || kubeconfig.Clusters == nil || len(kubeconfig.Clusters) == 0 {
+			return nil, fmt.Errorf("invalid bootstrap kubeconfig: no clusters found")
+		}
+
+		var etcdHost string
+		for _, cluster := range kubeconfig.Clusters {
+			etcdHost = cluster.Server
+			break
+		}
```
pmtk: Given the check `len(kubeconfig.Clusters) == 0`, could we just do following?

```suggestion
		if kubeconfig == nil || kubeconfig.Clusters == nil || len(kubeconfig.Clusters) == 0 {
			return nil, fmt.Errorf("invalid bootstrap kubeconfig: no clusters found")
		}

		etcdHost := kubeconfig.Clusters[0].Server
```

or maybe it should be
```
		var etcdHost string
		for _, cluster := range kubeconfig.Clusters {
            if cluster.Server != "" {
		    	etcdHost = cluster.Server
		    	break
            }
		}
```

pkg/controllers/kube-apiserver.go
```diff
@@ -407,3 +411,91 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+			return nil, fmt.Errorf("failed to create etcd client: %w", err)
+		}
+	}
+
+	resp, err := client.MemberList(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve etcd member list: %w", err)
+	}
+
+	//TODO I already know if I am a learner, I had to do this before.
```
pmtk: TODO?

pkg/controllers/kube-apiserver.go
```diff
@@ -407,3 +411,91 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+		}
+		if !member.IsLearner {
+			members = append(members, member.ClientURLs...)
+		}
+	}
+	if iAmLearner {
+		return members, nil
+	}
+
+	return []string{"https://localhost:2379"}, nil
```
pmtk: So basically, if my etcd is a learner, instruct KAS to use other etcds. If it's not a learner, just use my etcd, right?

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,591 @@
...
+func runAddNode(ctx context.Context, opts *AddNodeOptions) error {
+	ctx, cancel := context.WithTimeout(ctx, opts.Timeout)
+	defer cancel()
+
+	klog.Info("Starting cluster join process...")
+	if opts.Learner {
+		klog.Info("Will add etcd node as learner")
+	}
+
+	cfg, err := config.ActiveConfig()
```
pmtk, line 90: In a way I'm wondering if we should get the config from the original node?

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,591 @@
...
+	if err := waitForNodeReady(ctx, client, cfg.CanonicalNodeName()); err != nil {
+		return fmt.Errorf("node failed to become ready: %w", err)
+	}
+
+	klog.Info("Node successfully joined the cluster")
+	return nil
+}
+
+func createKubernetesClient(kubeconfigPath string) (*kubernetes.Clientset, error) {
+	if _, err := os.Stat(kubeconfigPath); os.IsNotExist(err) {
```
pmtk: See `util.PathExists`.

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,591 @@
...
+		return nil, fmt.Errorf("kubeconfig file does not exist at %s", kubeconfigPath)
+	}
+
+	_, err := clientcmd.LoadFromFile(kubeconfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("invalid kubeconfig file: %w", err)
+	}
+
+	restConfig, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
```
pmtk: Won't `BuildConfigFromFlags` do the necessary validation?

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,591 @@
...
+func restartMicroShift() error {
+	cmd := exec.Command("systemctl", "restart", "microshift")
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to restart microshift service: %w", err)
+	}
+	return nil
+}
+
+func isNodeAlreadyInCluster(ctx context.Context, client kubernetes.Interface, nodeName string) bool {
+	nodes, err := client.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
pmtk: Can't just GET node with particular name?

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,591 @@
...
+			if nodeIP != "" {
+				members = append(members, fmt.Sprintf("%s=https://%s:2380", node.Name, nodeIP))
+			}
+		}
+	}
+
+	return members, nil
+}
+
+func isJoinNodeReady(node *corev1.Node) bool {
```
pmtk: It's checking ANY given node
```suggestion
func isNodeReady(node *corev1.Node) bool {
```

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,591 @@
...
+	clientKeyPath := cryptomaterial.ClientKeyPath(clientCertDir)
+	if err := clientTLS.WriteCertConfigFile(clientCertPath, clientKeyPath); err != nil {
+		return fmt.Errorf("failed to write client certificate: %w", err)
+	}
+
+	klog.Info("All etcd certificates generated successfully with proper signatures and SAN entries")
+	return nil
+}
+
+func getClusterInfo(ctx context.Context, client kubernetes.Interface) ([]string, error) {
```
pmtk: getNonLearnerEtcdNodes?

pkg/cmd/addnode.go
```diff
@@ -0,0 +1,591 @@
...
+			return runAddNode(cmd.Context(), opts)
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.KubeconfigPath, "kubeconfig", opts.KubeconfigPath,
+		"Path to kubeconfig file for connecting to the cluster")
+	cmd.Flags().DurationVar(&opts.Timeout, "timeout", opts.Timeout,
+		"Timeout for cluster join operations")
+	cmd.Flags().BoolVar(&opts.Learner, "learner", true,
+		"Join the cluster as a learner node")
```
pmtk, line 72: How do you move on from being a learner to non learner?

etcd/cmd/microshift-etcd/run.go
pacevedom: Basically similarity with regular etcd files you can feed it, there is no other reason but using the same names to make it more understandable. I can swap it with json without trouble too.

pkg/cmd/addnode.go
pacevedom: you are right, will leave this one empty instead because the contents are copied to a MicroShift controlled dir and cleaned up prior to that. Done

pkg/cmd/addnode.go
pacevedom: That is another command, which will come after this one (its also part of another ticket in the epic).

pkg/cmd/addnode.go
pacevedom: Yes, this is something I intend to do in future PRs. In this stage the configuration is used only to get IP address and hostname. When MicroShift restarts later on it should be synced with the node already in the cluster, but this is not included yet.

pkg/cmd/addnode.go
pacevedom: The command already skips all of this logic if the node is already part of the cluster (this is just a few lines above). If someone wants to add the node again to a cluster then first you have to delete it (future command in a different PR), and then the add is explicitly on purpose and all the data is cleared, which is something the user should know by now. Also unsupported, as you mentioned.
Some extra protection like also checking etcd members may be desirable.

pkg/cmd/addnode.go
pacevedom: Done

pkg/cmd/addnode.go
pacevedom: Done

pkg/cmd/addnode.go
pacevedom: Done

pkg/config/kubeconfig.go
pacevedom: Done

pkg/controllers/etcd.go
pacevedom: Done

pkg/controllers/kube-apiserver.go
pacevedom: Correct. If you are not a learner then your instance is part of a bigger cluster where every node is synced. If not, use the ones you know are synced.

pkg/cmd/addnode.go
pacevedom: Done

pkg/cmd/addnode.go
pacevedom: Strictly speaking is any node in the cluster, so just using all ready nodes is enough.

pkg/components/certificateauthority.go
pacevedom: This has a few different sides. So the most important one maybe is that the root CA should be shared across the whole cluster, in every node. MicroShift does not have that, but instead has dedicated CAs for different purposes. Changing that now would require some analysis as there are many things you can break. MicroShift also generates its certificates in a special way, if they are not present they get created and they would differ, breaking internal API endpoints across nodes.

When joining the cluster you could very well produce CSRs and let the original node sign them all and then new nodes grab them using the api. This, however, does not prevent the fact that you need the CAs in each node in order for things like for example, retrieving logs across different nodes or rotating certs.

pkg/controllers/etcd.go
pacevedom: Done

pkg/controllers/kube-apiserver.go
pacevedom: Leftover. Done.

pkg/controllers/kube-apiserver.go
pacevedom: I had the same idea but not possible because `clusters` is a map using cluster name as key. Since this may be whatever we have to iterate. Also, the `server` is a required parameter so its guaranteed to exist.
I added another check so that there is only 1 cluster, as that would introduce ambiguity.

pkg/controllers/kube-apiserver.go
pacevedom: This was a dated comment. Updated the code to simplify it a bit.

pkg/cmd/addnode.go
pmtk: What if it's single-node microshift instance? And someone attempts to join it?

PR #5586, opened 2025-10-09
USHIFT-5539: Use add-node in cncf conformance scenarios

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/scripts/multinode/configure-node.sh
+++ b/scripts/multinode/configure-node.sh
@@ -103,8 +103,8 @@ function run_healthcheck() {
     fi
 
     greenboot_status=$(systemctl show -p Result --value greenboot-healthcheck)
-    if [ "$greenboot_status" != "success" ]; then
-        echo "Error: greenboot-healthcheck did not complete successfully (Result: $greenboot_status)"
+    if [ "${greenboot_status}" != "success" ]; then
+        echo "Error: greenboot-healthcheck did not complete successfully (Result: ${greenboot_status})"
         exit 1
     fi
 }
@@ -149,7 +149,7 @@ else
 fi
 echo
 echo "Node configuration completed"
-if [ ! -n "${BOOTSTRAP_KUBECONFIG}" ]; then
+if [ -z "${BOOTSTRAP_KUBECONFIG}" ]; then
     copy_bootstrap_kubeconfig
     echo
     echo "To add other nodes to this cluster, copy the following kubeconfig file to other nodes:"


pacevedom:
Built on top of https://github.com/openshift/microshift/pull/5570. This change is blocked until that PR merges and we can rebase and mark as ready for review.

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,175 @@
...
+OC_CMD="sudo -i oc --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig"
+
+# Variables for node configuration
+NODE_ADDR=""
+BOOTSTRAP_KUBECONFIG="${BOOTSTRAP_KUBECONFIG:-}"
+
+function usage() {
+    echo "Usage: $(basename "$0")"
+    echo "Environment variables:"
+    echo "  BOOTSTRAP_KUBECONFIG    - Path to kubeconfig file for joining existing cluster (optional)"
```
ggiguash: Can we convert this to an optional command line?

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,175 @@
...
+    exit 1
+}
+
+function configure_system() {
+    # Disable selinux
+    # TODO: remove once selinux is working properly again
+    sudo setenforce 0 || true
```
ggiguash: I think we should be able to remove this with the new system.

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,175 @@
...
+
+    # TODO: Edit firewall rules instead of stopping firewall
+    sudo systemctl stop firewalld
+    sudo systemctl disable firewalld
+
+    # Greenboot checks are tuned for a single node
+    sudo systemctl stop greenboot-healthcheck
+    sudo systemctl reset-failed greenboot-healthcheck
+    sudo systemctl disable greenboot-healthcheck
```
ggiguash, line 25: What's the status of greenboot with the new system? 
Do we still need to disable it, or we can just use it on each node?

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,175 @@
...
+}
+
+function configure_microshift() {
+    # Clean the current MicroShift configuration and stop the service
+    echo 1 | sudo microshift-cleanup-data --all --keep-images
+
+    get_node_ip_from_config
+
+    # Configure MicroShift to disable telemetry
+    cat <<EOF | sudo tee /etc/microshift/config.yaml &>/dev/null
```
ggiguash: Let's use a drop-in here?

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,175 @@
...
+    fi
+    echo "Successfully added node using kubeconfig: ${BOOTSTRAP_KUBECONFIG}"
+}
+
+function get_node_ip_from_config() {
+    # Extract nodeIP from MicroShift running configuration and store in global variable
+    local node_ip=""
+
+    # Use microshift show-config to get the actual running configuration
+    if command -v microshift >/dev/null 2>&1; then
```
ggiguash: Why do we need this check. We do not have similar checks in other places we use microshift exe.

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,175 @@
...
+        "deployments": ["lvms-operator"],
+        "daemonsets": ["vg-manager"]
+    },
+    "kube-system": {
+        "deployments": ["csi-snapshot-controller"]
+    }
+}
+EOF
+)
+    if ! sudo microshift healthcheck -v=2 --timeout="${wait_secs}s" --custom "${custom_json}"; then
```
ggiguash: Hopefully, we should just wait for the greenboot service to exit with success here.

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,175 @@
...
+        echo "Error: MicroShift healthcheck failed"
+        exit 1
+    fi
+}
+
+#
+# Main function
+#
+
+KUBELET_HOME="${HOME}"
```
ggiguash: We should probably rename this, now that we're not using kubelet.

scripts/multinode/configure-node.sh
pacevedom: Done.

scripts/multinode/configure-node.sh
pacevedom: Done.

scripts/multinode/configure-node.sh
pacevedom: Done.

scripts/multinode/configure-node.sh
pacevedom: Felt like abusing greenboot because it was geared towards single node and I only needed the healthcheck command, but swapped it for simplicity and it works.

scripts/multinode/configure-node.sh
pacevedom: Leftover, removed.

scripts/multinode/configure-node.sh
pacevedom: Done.

scripts/multinode/configure-node.sh
pacevedom: Turns out it works, so removed these lines.

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,162 @@
...
+
+function configure_system() {
+    # TODO: Edit firewall rules instead of stopping firewall
+    sudo systemctl stop firewalld
+    sudo systemctl disable firewalld
+
+    # Greenboot checks are tuned for a single node
+    sudo systemctl stop greenboot-healthcheck
+    # sudo systemctl reset-failed greenboot-healthcheck
+    # sudo systemctl disable greenboot-healthcheck
```
ggiguash: Why are these commented out? We probably need them to make sure greenboot does not come back.

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,162 @@
...
+    fi
+    echo "Successfully added node using kubeconfig: ${BOOTSTRAP_KUBECONFIG}"
+}
+
+function get_node_ip_from_config() {
+    # Extract nodeIP from MicroShift running configuration and store in global variable
+    local node_ip=""
+
+    # Use microshift show-config to get the IP address of the node
+    node_ip=$(sudo microshift show-config 2>/dev/null | grep -E "^\s*nodeIP\s*:" | head -1 | sed 's/.*nodeIP\s*:\s*//' | tr -d ' \t\r\n')
```
ggiguash:
```suggestion
    node_ip=$(sudo microshift show-config | awk '/nodeIP:/ {print $2}' | tr -d ' \t\r\n')
```

scripts/multinode/configure-node.sh
ggiguash: Looks like this change is half-baked somehow. Still commented out.

scripts/multinode/configure-node.sh
pacevedom: done

scripts/multinode/configure-node.sh
pacevedom: done

scripts/multinode/configure-node.sh
pacevedom: Done only with awk in the end.

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
+set -euo pipefail
+
+OC_CMD="sudo -i oc --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig"
```
pmtk: Should we expect to run this script as root and drop all the sudos from the file?

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,161 @@
...
+if [ -n "${BOOTSTRAP_KUBECONFIG}" ]; then
+    run_add_node_commands
+    run_healthcheck
+else
+    start_microshift
+fi
+echo
+echo "Node configuration completed"
+if [ ! -n "${BOOTSTRAP_KUBECONFIG}" ]; then
+    copy_kubeconfig
```
pmtk: prepare_bootstrap_kubeconfig?

scripts/multinode/configure-node.sh
```diff
@@ -0,0 +1,161 @@
...
+OC_CMD="sudo -i oc --kubeconfig /var/lib/microshift/resources/kubeadmin/kubeconfig"
+
+# Variables for node configuration
+NODE_ADDR=""
+BOOTSTRAP_KUBECONFIG=""
+
+function usage() {
+    echo "Usage: $(basename "$0") [OPTIONS]"
+    echo "Options:"
+    echo "  --bootstrap-kubeconfig PATH    Path to kubeconfig file for joining existing cluster (optional)"
```
pmtk, line 13:
```suggestion
    echo "  --join-to-cluster BOOTSTRAP_KUBECONFIG_PATH    Path to kubeconfig file for joining existing cluster (optional)"
```

I'm wondering if this would be clear. Looking at the invocs I'm missing a bit info whether I'm setting up a primary or adding (secondary) node - aka invocs are not clear at first glance, but maybe that's just me :) 

Like
```
./configure-node.sh
./configure-node.sh --bootstrap-kubeconfig ~/path
```
compared to something like (just thinking out loud, I don't expect it to be change to this)
```
./configure-node.sh primary
./configure-node.sh add-node ~/path
```

scripts/multinode/configure-node.sh
pacevedom: You just made me realize that `oc` is no longer used. About root, the more commands we have with least privileges the better. Sonobuoy for example does not need root. There is a lot of interaction with systemctl/protected directories from MicroShift so no escape from there.

scripts/multinode/configure-node.sh
pacevedom: What about `copy_bootstrap_kubeconfig`? Its just a copy, not so much preparation.

scripts/multinode/configure-node.sh
pacevedom: Added some help to the command.

Just calling the script without options means "configure this node to run MicroShift", while also allowing it to be prepared to add more nodes. This is implied by the location of the script (`scripts/multinode`).
Similar semantics with the `--bootstrap-kubeconfig`, which implies there is another cluster (the kubeconfig) that you will be using to bootstrap this node into it. Also the help shows it.
Keep in mind that this is all unsupported, so quite niche for now.

PR #5613, opened 2025-10-14
USHIFT-6210: Fetch git repo using SSH

pmtk:


pacevedom (approved):
/lgtm

pmtk:
/verified by @pmtk 

Locally with the flag. It's an opt-in behavior so CI is unaffected.

PR #5637, opened 2025-10-20
NO-ISSUE: fix lvms and tlsv13 RF tests

agullon:
- fix `SCENARIOS_SOURCES` var from `test/bin/ci_phase_boot_and_test.sh` script
   - previously `SCENARIO_TYPE` var was always empty, with this change ginkgo binary is build before the scenarios start
 - split the code into 2 functions to be able to use them from `lvms` scenarios:
    - `setup_oc_and_kubeconfig()` and `wait_for_microshift_to_be_ready()`
- fix start images for `lvms` and `tlsv13` tests
- add `setup_oc_and_kubeconfig()` in all `lvms` scenarios because `kubeconfig` file must be set before running `lvms-helpers`scripts


test/bin/scenario.sh
```diff
@@ -1118,18 +1134,30 @@ run_gingko_tests() {
...
+    pushd . &>/dev/null
+
+    # Setup oc client and kubeconfig for gingko tests
+    setup_oc_and_kubeconfig_tests "${vmname}"
+
+    # Wait for MicroShift to be ready
+    wait_for_microshift_to_be_ready "${vmname}"
 
     # Create case selection file
     local case_selected="${test_results_dir}/case_selected"
```
kasturinarra, line 1158:
```suggestion
local -r    test_results_dir="${SCENARIO_INFO_DIR}/${SCENARIO}/gingko-results"
mkdir -p "${test_results_dir}"
local case_selected="${test_results_dir}/case_selected"
```

agullon: thanks, add it in cf5c041eba412280d9ea4cac80e07dc3f0a8946d commit

test/bin/scenario.sh
```diff
@@ -1102,36 +1118,46 @@ run_gingko_tests() {
     fi
     record_junit "${vmname}" "oc_installed" "OK"
 
-    # Check/get openshift-tests-binary
-    if ! "${ROOTDIR}/scripts/fetch_tools.sh" "ginkgo"; then
-        record_junit "${vmname}" "build_test_binary" "FAILED"
+    # Wait for MicroShift to be ready
+    wait_for_microshift_to_be_ready "${vmname}"
```
pacevedom: This is not just setting up oc and kubeconfig, right? Kubeconfig is available as soon as MicroShift starts, you dont need to wait until its ready. It can be done earlier.

test/scenarios-bootc/releases/el96-y1@el96-lrel@lvms.sh
```diff
@@ -44,15 +44,8 @@ scenario_run_tests() {
...
-    local vmname="host1"
-    local -r full_vmname="$(full_vm_name "${vmname}")"
-    local -r vm_ip="$(get_vm_property "${vmname}" ip)"
-    if ! wait_for_greenboot "${full_vmname}" "${vm_ip}"; then
-        record_junit "${vmname}" "pre_test_greenboot_check" "FAILED"
-        return 1
-    fi
-    record_junit "${vmname}" "pre_test_greenboot_check" "OK"
+    # Setup oc client and kubeconfig for gingko tests
+    setup_oc_and_kubeconfig host1
```
pacevedom, line 51: Losing the explicit wait_for_greenboot looks non-obvious to me. Also this introduces an implicit waiting in places where it was not present.

test/bin/scenario.sh
agullon: yes, I though about this... I agree with you. I move out this func call in last 4acded66438302a4914e8727c63e0cb8b54a4935 commit

test/scenarios-bootc/releases/el96-y1@el96-lrel@lvms.sh
agullon: I agree, yes, in last 4acded66438302a4914e8727c63e0cb8b54a4935 commit I explicity wait for microshift before this line

PR #5657, opened 2025-10-23
USHIFT-6276: introduce hosts plugin coreDNS RF tests

eslutsky:

Closes #<Issue Number>

### DIFF
--- a/test/suites/standard1/dns.robot
+++ b/test/suites/standard1/dns.robot
@@ -47,30 +47,16 @@ Resolve Host from Non-Default Hosts File
 
 
 *** Keywords ***
-Wait For DNS Updated With Hostname
-    [Documentation]    Wait for the file to be updated
-    [Arguments]    ${hostname}
-    Wait Until Keyword Succeeds    2m    2s
-    ...    DNS Pod Contain Hosts    ${hostname}
-
-DNS Pod Contain Hosts
-    [Documentation]    Check if the hosts file contains the given hostname
-    [Arguments]    ${hostname}
-
-    # Find the dns-default-.* pod name in openshift-dns namespace, e.g., dns-default-dqp9n
-    ${dns_pod}=    Run With Kubeconfig
-    ...    oc get pod -n openshift-dns -o jsonpath='{.items[*].metadata.name}' | grep '^dns-default-'
-    ${fuse_device}=    Oc Exec    ${dns_pod}    cat ${PODS_HOSTS_FILE}    openshift-dns
-    Should Contain    ${fuse_device}    ${hostname}
-
 Resolve Host From Pod
     [Documentation]    Resolve host from pod
     [Arguments]    ${hostname}
-    Create Hello MicroShift Pod
-    Expose Hello MicroShift
-    Wait For DNS Updated With Hostname    ${hostname}
-    Sleep    5 seconds
-    ${fuse_device}=    Oc Exec    hello-microshift    nslookup ${hostname}    ${NAMESPACE}    /bin/sh
+    Wait Until Keyword Succeeds    40x    2s
+    ...    Router Should Resolve Hostname    ${hostname}
+
+Router Should Resolve Hostname
+    [Documentation]    Check if the router pod resolves the given hostname
+    [Arguments]    ${hostname}
+    ${fuse_device}=    Oc Exec    router-default    nslookup ${hostname}    openshift-ingress    deployment
     Should Contain    ${fuse_device}    Name:    ${hostname}
 
 Setup With Custom Config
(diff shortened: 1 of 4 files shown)


pkg/components/controllers.go
pacevedom: Why not using https://github.com/openshift/microshift/pull/5657/files#diff-73e9e9d7371412e8d5d73dc0c82539659a29dfa51c724fc441357398a087bec2R8 constant instead?

pkg/config/dns.go
pacevedom: Few comments for this function:
* Use a switch case, maybe?
* If `Enabled` we should validate the file exists and we can read it.
* We are allowing the exposure of critical files (such as `/etc/passwd`) because we are not checking the actual file path. Same with relative paths, we should sanitize and require absolute paths here.
* File size. This will go to a config map, meaning we are limited to [1Mi in total size](https://kubernetes.io/docs/concepts/configuration/configmap/).


Just an idea: Should we restrict the files we can load to a handful of paths? For example, either `/etc/hosts` or something outside system paths?

pkg/controllers/hostswatcher.go
pacevedom, line 194: I am missing RBAC for this specific config map, only CoreDNS pods should be able to read it and only MicroShift should be able to write it.

pacevedom: This is constant, should it be a member of the struct?

pacevedom: This is constant, should it be a member of the struct?

pacevedom, line 22: Should these be constants?

pacevedom, line 84: Should it signal readiness here instead? One it reaches this point you know you have at least one update to the configmap plus the file being watched.

pacevedom, line 113: Should it close the watcher?

pacevedom, line 120: Should it signal the removal/rename of the file? At that point it will fail in the next restart but remain working until then.

pacevedom: Remove the V(2) so that it always show up?

eslutsky:
@pacevedom this PR was opened for the tests but its based on the https://github.com/openshift/microshift/pull/5491 commits , so i will address the  code changed there.

test/suites/standard2/validate-custom-certificates.robot
pacevedom: Should these be in the new `hosts.resource`?

test/suites/standard1/dns.robot
pacevedom, line 50: Is it possible to use dns pod instead? something line `oc exec -n openshift-dns ds/dns-default -c dns nslookup $hostname`.
This way we save some time and make the test simpler.

pacevedom: Can we instead try if it resolved until success or timeout?

pacevedom: You can also exec into ds/dns-default instead of specifying the name of the pod.

eslutsky: DNS system pods do not resolve using CoreDNS; they can only use upstream DNS because they inherit the host’s /etc/resolv.conf

eslutsky: but we can use the router pod   `oc exec -ti deployment/router-default -n openshift-ingress -- nslookup test.api.com`

test/suites/standard1/dns.robot
```diff
@@ -0,0 +1,111 @@
...
+DNS Pod Contain Hosts
+    [Documentation]    Check if the hosts file contains the given hostname
+    [Arguments]    ${hostname}
+    ${fuse_device}=    Oc Exec    dns-default    cat ${PODS_HOSTS_FILE}    openshift-dns    ds
+    Should Contain    ${fuse_device}    ${hostname}
+
+Resolve Host From Pod
+    [Documentation]    Resolve host from pod
+    [Arguments]    ${hostname}
+    Wait For DNS Updated With Hostname    ${hostname}
```
pacevedom: Do we need to wait for this explicitly or could we rely on the next keyword? It wont succeed if the file has not been updated.

test/suites/standard2/validate-custom-certificates.robot
pacevedom: Do we want to move this one and the next to `hosts.resource`?

pacevedom (commented):
Only 2 minor nits.

test/suites/standard1/dns.robot
eslutsky: we can remove it , just leave the nslookup check

test/suites/standard2/validate-custom-certificates.robot
eslutsky: yea its moved already

pacevedom: I mean the `Local Hosts` keywords, sounds to me like they belong in the hosts library instead of a test?

PR #5679, opened 2025-10-30
USHIFT-6282: 'brew download-build' command fails

agullon:
Add a fallback mechanism to download the RPMs using `curl` command with `-k` option to ignore invalid and self-signed certificate errors.

### DIFF
--- a/test/bin/manage_brew_rpms.sh
+++ b/test/bin/manage_brew_rpms.sh
@@ -77,10 +77,44 @@ action_download() {
         mkdir -p "${adir}"
         pushd "${adir}" &>/dev/null
         if ! brew download-build --arch="${arch}" --arch="noarch" "${package}" ; then
-            echo "ERROR: Failed to download '${package}' packages from brew"
-            exit 1
+            echo "WARNING: Failed to download '${package}' packages using brew download-build command, using curl as a fallback mechanism"
+            if ! brew_curl_download "${package}" "${arch}" ; then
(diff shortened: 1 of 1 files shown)


test/bin/manage_brew_rpms.sh
```diff
@@ -77,13 +77,47 @@ action_download() {
...
+    local package=$1
+    local arch=$2
+
+    # Parse package to extract version and build release
+    local version_and_release="${package#microshift-}"
+    local pkg_version="${version_and_release%%-*}"
+    local pkg_release="${version_and_release#*-}"
+
+    for current_arch in ${arch} noarch; do
+        local base_url="https://download-01.beak-001.prod.iad2.dc.redhat.com/rhel-9/brew/packages/microshift/${pkg_version}/${pkg_release}/${current_arch}/"
```
pacevedom, line 101: Wondering if this URL is fixed or there are any other instances that might serve the same files?

kasturinarra, line 101: can we make the `rhel-9` configurable  so that it would be easy later when new rhel versions are introduced ?

agullon: yes, I missed that, I'll do it

agullon: There are no references in the repo pointing to different RHEL 9 and RHEL 10. Neither there are MicroShift RPMs published for RHEL 10.
I think this should be addressed in a separate PR at the same time as other RHEL 9 references.

PR #5719, opened 2025-11-11
USHIFT-6306: Do not attempt to start greenboot-healthcheck service on RPM install

ggiguash:
See https://github.com/microshift-io/microshift/issues/113


ggiguash:
The original problem was introduced due to RHEL 10 tests, which only run in bootc-periodics and bootc-upstream.

/test e2e-aws-tests-bootc-periodic
/test e2e-aws-tests-bootc-periodic-arm
/test e2e-aws-tests-bootc-upstream
/test e2e-aws-tests-bootc-upstream-arm

PR #5723, opened 2025-11-11
USHIFT-6178 Enable Custom FeatureGates

copejon:


pkg/config/apiserver.go
```diff
@@ -127,3 +132,76 @@ func (t *TLSConfig) Validate() error {
...
+		return nil
+	}
+	// Must use a recognized feature set, or else empty
+	if fg.FeatureSet != "" && fg.FeatureSet != FeatureSetCustomNoUpgrade && fg.FeatureSet != FeatureSetTechPreviewNoUpgrade && fg.FeatureSet != FeatureSetDevPreviewNoUpgrade {
+		return fmt.Errorf("invalid feature set: %s", fg.FeatureSet)
+	}
+	// Must set FeatureSet to CustomNoUpgrade to use custom feature gates
+	if fg.FeatureSet != FeatureSetCustomNoUpgrade && (len(fg.CustomNoUpgrade.Enabled) > 0 || len(fg.CustomNoUpgrade.Disabled) > 0) {
+		return fmt.Errorf("CustomNoUpgrade must be empty when FeatureSet is empty")
+	}
```
agullon, line 200:
```suggestion
	}
	// Must set CustomNoUpgrade enabled or disabled lists when FeatureSet is CustomNoUpgrade
	if fg.FeatureSet == FeatureSetCustomNoUpgrade && len(fg.CustomNoUpgrade.Enabled) == 0 && len(fg.CustomNoUpgrade.Disabled) == 0 {
		return fmt.Errorf("CustomNoUpgrade enabled or disabled lists must be set when FeatureSet is CustomNoUpgrade")
	}
```
suggesting an extra check to trigger an error when FeatureSet is CustomNoUpgrade and not features in enabled and disabled lists

copejon: I don't think this is necessary because an empty list is just as valid as a populated list.

copejon: Rethinking my last response: if microshift treated empty lists with `CustomNoUpgrade` as valid, that would mean that after microshift is started, the logic would not permit any more changes to the config or upgrades of the cluster. That would create a situation where a supported cluster would be treated as unsupported.

So I take back what I initially said

agullon: no problem, anyway, I think probably this check I'm suggesting it's not really needed because this is already enforced by the API definition, not sure, tbh.

cmd/generate-config/config/config-openapi-spec.json
```diff
@@ -59,6 +60,39 @@
             }
           }
         },
+        "featureGates": {
+          "type": "object",
+          "required": [
+            "customNoUpgrade",
+            "featureSet"
+          ],
```
pacevedom, line 68: These two fields are marked as required but default values and validations can handle empty values.

pkg/controllers/kube-apiserver.go
```diff
@@ -169,6 +168,15 @@ func (s *KubeAPIServer) configure(ctx context.Context, cfg *config.Config) error
...
 	}
 
+	featureGateArgs, err := cfg.ApiServer.FeatureGates.ConvertToCLIFlags()
+	if err != nil {
+		return fmt.Errorf("failed to convert feature gates to CLI flags: %w", err)
+	}
+
+	// Build the final feature gates list
+	hardcodedFeatureGates := []string{"UserNamespacesSupport=true", "UserNamespacesPodSecurityStandards=true"}
+	featureGateArgs = append(featureGateArgs, hardcodedFeatureGates...)
```
pacevedom: Could we have a conflict here? Configuration allows disabling `UserNamespacesSupport` and `UserNamespacesPodSecurityStandards`.

cmd/generate-config/config/config-openapi-spec.json
copejon: I think this may not be an issue because existing elements of the config also include 
a populated `required` array: 
https://github.com/openshift/microshift/blob/ad6b2f2e8b4543350095ccdc7116e19e3514ed7f/cmd/generate-config/config/config-openapi-spec.json#L3-L16

pkg/config/config_test.go
```diff
@@ -800,6 +800,56 @@ func TestValidate(t *testing.T) {
...
+			config: func() *Config {
+				c := mkDefaultConfig()
+				c.ApiServer.FeatureGates.FeatureSet = ""
+				c.ApiServer.FeatureGates.CustomNoUpgrade.Enabled = []string{"feature1"}
+				c.ApiServer.FeatureGates.CustomNoUpgrade.Disabled = []string{"feature2"}
+				return c
+			}(),
+			expectErr: true,
+		},
+		{
```
agullon, line 842:
```suggestion
		},
		{
			name: "feature-gates-custom-no-upgrade-with-empty-enabled-and-disabled-lists",
			config: func() *Config {
				c := mkDefaultConfig()
				c.ApiServer.FeatureGates.FeatureSet = "CustomNoUpgrade"
				c.ApiServer.FeatureGates.CustomNoUpgrade.Enabled = []string{}
				c.ApiServer.FeatureGates.CustomNoUpgrade.Disabled = []string{}
				return c
			}(),
			expectErr: true,
		},
		{
```
adding a test for the suggestion check I added last week

pkg/admin/prerun/featuregate_lock.go
```diff
@@ -0,0 +1,213 @@
...
+		// First time configuring custom feature gates - create lock file
+		return createFeatureGateLockFile(cfg)
+	}
+
+	// Lock file exists - validate configuration hasn't changed
+	return validateFeatureGateLockFile(cfg)
+}
+
+// isCustomFeatureGatesConfigured checks if any custom feature gates are configured
+func isCustomFeatureGatesConfigured(fg config.FeatureGates) bool {
```
pmtk: This _could_ be a method on `config.FeatureGates` object

packaging/microshift/config.yaml
```diff
@@ -14,6 +14,11 @@ apiServer:
         maxFiles: 10
         # profile is the OpenShift profile specifying a specific logging policy
         profile: Default
+    featureGates:
```
pmtk, line 17: I started wondering - if FGs are propagated, but someone would want to apply kubelet FG, should we move it out of `apiServer` and make it global to avoid confusion?

pkg/admin/prerun/featuregate_lock.go
```diff
@@ -0,0 +1,213 @@
...
+		klog.ErrorS(err, "FAIL feature gate lock management")
+		return err
+	}
+	klog.InfoS("END feature gate lock management")
+	return nil
+}
+
+func featureGateLockManagement(cfg *config.Config) error {
+	// Check if custom feature gates are configured
+	if !isCustomFeatureGatesConfigured(cfg.ApiServer.FeatureGates) {
```
pmtk: THis only checks the config - so if:
1. I set the FGs, restart microshift, lock is created
2. Unset the FGs, this function decided: no FG configure -> return

What's missing is that unconditional check for the FG lock file: if the lock file exists, but cfg.ApiServer.FGs is empty, we still should refuse to run.

pkg/config/apiserver.go
```diff
@@ -127,3 +132,80 @@ func (t *TLSConfig) Validate() error {
...
+	if fg.FeatureSet == FeatureSetCustomNoUpgrade && len(fg.CustomNoUpgrade.Enabled) == 0 && len(fg.CustomNoUpgrade.Disabled) == 0 {
+		return fmt.Errorf("CustomNoUpgrade enabled or disabled lists must be set when FeatureSet is CustomNoUpgrade")
+	}
+	// Must not have any feature gates that are enabled and disabled at the same time
+	var illegalFeatures []string
+	for _, enabledFeature := range fg.CustomNoUpgrade.Enabled {
+		if slices.Contains(fg.CustomNoUpgrade.Disabled, enabledFeature) {
+			illegalFeatures = append(illegalFeatures, enabledFeature)
+		}
+	}
```
pmtk: Nit: we could try using [Sets](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set)

```go
enabledSet := Set(fg.CustomNoUpgrade.Enabled)
disabledSet := Set(fg.CustomNoUpgrade.Disabled)
inBothSets := enabledSet.Intersect(disabledSet)
if inBothSets.Len() > 0 {
    return fmt.Errorf(...)
}
```

apiserver-config-flow.md
```diff
@@ -0,0 +1,397 @@
...
+    // ... many more
+}
+```
+
+## Configuration Flow Steps
+
+### Step 1: Config Loading (`pkg/config/files.go`)
+
+**Function**: `ActiveConfig()`  
+**Line**: [120-127](file:///home/microshift/microshift/pkg/config/files.go#L120-L127)
```
pmtk, line 131: I could imagine line numbers getting out of date rather quickly. Also the filepath is hardcoded and local (would not work viewing it via github) and not everyone has microshift in that dir.
Do we need this and sequenceDiagram.mdd files? Could it be easily regenerated using LLM when needed? I'd rather avoid merging things that would only rot (not maintained)

copejon (commented):
Darn it, I thought I'd made the comment as a one off, not a review.

packaging/microshift/config.yaml
copejon: That's a very good point. I had shifted FGs under kube-apiserver to smooth the passthrough logic, but it doesn't solve the problem you rightly pointe out. I'll give this a second look and get back to you

apiserver-config-flow.md
copejon: Oops! This entire file was added by accident. I generated it for my own understanding. It is removed now.

pkg/config/apiserver.go
copejon: Thanks! I've used your suggestion to enhance the code

pkg/admin/prerun/featuregate_lock.go
copejon: Great point. I've addressed this in the latest commit

pkg/admin/prerun/featuregate_lock.go
```diff
@@ -39,27 +39,29 @@ func FeatureGateLockManagement(cfg *config.Config) error {
...
+	// If a lock file exists, it must be validated regardless of current config
+	// This prevents users from removing feature gates from config in order to block upgrades and configuration changes
 	lockExists, err := util.PathExists(featureGateLockFilePath)
 	if err != nil {
 		return fmt.Errorf("failed to check if lock file exists: %w", err)
 	}
 
-	if !lockExists {
-		// First time configuring custom feature gates - create lock file
+	hasCustomFeatureGates := isCustomFeatureGatesConfigured(cfg.ApiServer.FeatureGates)
```
pmtk: I'd move it couple lines lower, before it is actually used

pkg/admin/prerun/featuregate_lock.go
```diff
@@ -161,14 +163,20 @@ func validateFeatureGateLockFile(cfg *config.Config) error {
...
 
 	if !reflect.DeepEqual(lockFile.CustomNoUpgrade, current.CustomNoUpgrade) {
-		return fmt.Errorf("custom feature gates changed: locked config has %#v, current config has %#v",
-			lockFile.CustomNoUpgrade, current.CustomNoUpgrade)
+		errs = append(errs, fmt.Errorf("custom feature gates changed: locked config has %#v, current config has %#v",
+			lockFile.CustomNoUpgrade, current.CustomNoUpgrade))
+	}
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
```
pmtk: I think errors.Join works with empty slice

pkg/config/apiserver.go
```diff
@@ -197,15 +202,30 @@ func (fg *FeatureGates) validateFeatureGates() error {
...
-	// Must not have any feature gates that are enabled and disabled at the same time
-	var illegalFeatures []string
-	for _, enabledFeature := range fg.CustomNoUpgrade.Enabled {
-		if slices.Contains(fg.CustomNoUpgrade.Disabled, enabledFeature) {
-			illegalFeatures = append(illegalFeatures, enabledFeature)
+
+	var errs = make(sets.Set[error], 0)
+	for _, requiredFG := range RequiredFeatureGates {
+		// Edge case: Users must not be allowed to explicitly disable required feature gates.
+		if sets.NewString(fg.CustomNoUpgrade.Disabled...).Has(requiredFG) {
```
pmtk, line 209: Maybe instead of prevent disabling them, we could not interfere if they're specified by user? I.e. if user wants them disabled, we don't enable them; if the don't mention them or enable them, we enable them.

Lack of UserNS FG doesn't break the cluster, just the origin tests would not pass, but it's up to the users to disable or keep it. WDYT?

pkg/admin/prerun/featuregate_lock.go
copejon: Done!

pkg/admin/prerun/featuregate_lock.go
copejon: Let's leave it as is for now. It's not anymore useful as a method

pkg/admin/prerun/featuregate_lock.go
copejon: Nice catch!

pkg/config/apiserver.go
copejon: Ty I was missing that context. I'm okay with dropping those as hardcoded FGs. Obv it will require an update to pre-test steps to explicitly enable it.

Also, we should think about whether enabling via the config.yaml would qualify as "custom" FGs because that will dictate the FG rollback / cluster upgrade blocking.

Maybe for now we keep it as is and follow up in another PR?

pacevedom: The user namespace feature gates are required to pass origin tests, which in turn means blocking jobs. Blocking upgrades wont be an issue with that as we do not perform upgrade tests in those jobs.
The suggested approach to allow disable but enable them anyway would show no trouble with the way we test it and would require no changes to the MicroShift's configuration for them.
Can we include them if not explicitly disabled? It would also be good to do it as defaults so that it shows in `show-config` command. If there is an explicit disable, remove them and add a note about this.

copejon: Capturing our sidebar agreement:

The work required in microshift and openshift/release should not block this PR from merging. This can be treated as a bug and fixed after merge.

packaging/microshift/config.yaml
copejon: Enabling a kubelet FG on the kube-apiserver and disabling it in the kubelet config results in the FG being disabled. This is b/c the kubelet goes against convention by prioritizing the config file over all other config vectors (CLI flags, KAS propagation). 

This is only an issue if the user were to set contradictory featureGate values in the microshift config.

pkg/config/config_test.go
copejon: @agullon Thanks! For the sake of time, would you mind submitting this as a follow-up PR?  Else CI will have to test the change, and then retest again for merge, which would take a long time.

pkg/controllers/kube-apiserver.go
copejon: This is now protected against by preventing the user to specify these values in the ushift config.

https://github.com/copejon/microshift/blob/e108194ba7642c410cfe7c95281d173682e55f94/pkg/config/apiserver.go#L206-L218

pkg/config/config_test.go
agullon: add it https://github.com/openshift/microshift/pull/5795
please, review it

PR #5746, opened 2025-11-14
USHIFT-6201: Add sriov rpm

pacevedom:

Closes #<Issue Number>

### DIFF
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -253,21 +253,21 @@ release. These files contain the list of container image references used by Cert
 and can be used to embed those images into osbuilder blueprints or bootc containerfiles.
 
 %package sriov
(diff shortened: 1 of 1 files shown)


pacevedom:
/hold
Until rebase PR is ready.

packaging/rpm/microshift.spec
ggiguash: Looks like it should be spelled `SR-IOV`, right?

ggiguash (changes requested):

packaging/rpm/microshift.spec
pacevedom: I was following the names of the components in the package (also the operator spells as sriov without hyphens). `microshift-sr-iov` sounds like separate words with the format we use.

ggiguash: Oh, I only meant in the description, not in the RPM names.

PR #5764, opened 2025-11-19
USHIFT-6133: Updating cert-manager to use coredns feature

kasturinarra:
**Which issue(s) this PR addresses**: To have local acme test working for cert-manager, we added a workaround to the configmap in openshift-dns namespace, now with coredns feature implemented, we do not need that change any more, so reverting the same.

Closes #<Issue Number>

### DIFF
--- a/test/suites/optional/cert-manager.robot
+++ b/test/suites/optional/cert-manager.robot
@@ -7,8 +7,10 @@ Library             Process
 Library             String
 Library             ../../resources/journalctl.py
 Resource            ../../resources/common.resource
+Resource            ../../resources/hosts.resource
 Resource            ../../resources/kubeconfig.resource
 Resource            ../../resources/oc.resource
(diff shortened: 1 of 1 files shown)


kasturinarra:
@eslutsky could you please help review when you get sometime ?

ggiguash:
@kasturinarra , it looks like we cannot unconditionally remove this workaround because CoreDNS feature is not available in all releases we're testing.
See https://github.com/openshift/microshift/pull/5766

kasturinarra:
> @kasturinarra , it looks like we cannot unconditionally remove this workaround because CoreDNS feature is not available in all releases we're testing. See #5766

yes, just realized this yesterday when you gave me the bug.. Let me see what workaround i can put here.

kasturinarra:
@ggiguash @eslutsky could you please help review and merge ? thanks !!

PR #5776, opened 2025-11-20
USHIFT-6133: Add more robot framework test cases

kasturinarra:
**Which issue(s) this PR addresses**: Adding three more robot framework test cases that check if hosts-file is dynamically updated with out restarting microshift, disabling microshift is deleting the hosts-file configmap in openshift-dns namespace, validating cluster DNS takes precedence over host file.

Closes #[USHIFT-6133](https://issues.redhat.com//browse/USHIFT-6133)

### DIFF
--- a/test/suites/standard1/dns.robot
+++ b/test/suites/standard1/dns.robot
@@ -1,5 +1,5 @@
 *** Settings ***
-Documentation       Networking smoke tests
+Documentation       Core DNS smoke tests
 
 Resource            ../../resources/common.resource
 Resource            ../../resources/oc.resource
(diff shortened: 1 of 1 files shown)


test/suites/standard1/dns.robot
```diff
@@ -57,9 +81,13 @@ Resolve Host From Pod
 
 Router Should Resolve Hostname
     [Documentation]    Check if the router pod resolves the given hostname
-    [Arguments]    ${hostname}
+    [Arguments]    ${hostname}    ${expected_ip}=${EMPTY}
```
eslutsky: why do we need ${expected_ip}=${EMPTY}, where is that being called?

kasturinarra: I have used it for a different test case which i removed, but forgot to remove this. Now removed it. Could you please help review again ? thanks !!

PR #5695, opened 2025-11-05
USHIFT-6106: Add RHEL 10 support to bootc test scenarios

kasturinarra:
**Which issue(s) this PR addresses**:  Add RHEL10 support for bootc scenarios in releases directory

Closes #<Issue Number>

PR #5785, opened 2025-11-21
NO-ISSUE: Remove test case 62738 from ginkgo tests

agullon:
- This PR will unblock https://github.com/openshift/microshift/pull/5695 by fixing the Ginkgo failures.
- This PR adds a line to remove the ETCD test: 62738

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -1192,6 +1192,9 @@ run_ginkgo_tests() {
         echo "Running Ginkgo tests with MicroShift filter..."
     fi
 
+    echo "----------------Remove 62738 test case---------------"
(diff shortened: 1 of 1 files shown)


ggiguash:
@agullon , we need to document this removal:
- Add a comment in the code on why we are doing it
- Potentially create a JIRA issue with the problem description (especially if we plan to backport the change)

/hold

agullon:
> @agullon , we need to document this removal:
> 
> * Add a comment in the code on why we are doing it
> * Potentially create a JIRA issue with the problem description (especially if we plan to backport the change)
> 
> /hold

Jira ticket created to remove this change and addressed properly: https://issues.redhat.com/browse/USHIFT-6340

PR #5805, opened 2025-11-25
USHIFT-6334: delete configmap when hostswatcher service is disabled

eslutsky:


pkg/controllers/hostswatcher.go
```diff
@@ -44,6 +45,14 @@ func (s *HostsWatcherManager) Run(ctx context.Context, ready chan<- struct{}, st
 
 	if s.status != config.HostsStatusEnabled {
 		klog.Infof("%s is disabled (not configured)", s.Name())
+		// Delete ConfigMap if it exists when service is disabled
+		if kubeClient, err := s.createKubeClient(); err == nil {
+			if err := s.deleteConfigMap(ctx, kubeClient); err != nil {
+				klog.Warningf("%s failed to delete ConfigMap when disabled: %v", s.Name(), err)
+			}
+		} else {
+			klog.V(2).Infof("%s could not create Kubernetes client to delete ConfigMap: %v", s.Name(), err)
```
pacevedom: Should we warn here too?

PR #5807, opened 2025-11-26
NO-ISSUE: Increase timeout to wait for cert-manager kustomization log

kasturinarra:
Increased wait time as cert-manager is always  the last to be applied.


ggiguash:
/retitle NO-ISSUE: Increase timeout to wait for cert-manager kustomization log

PR #5820, opened 2025-11-27
USHIFT-6349: Reduce VMs in CI

pacevedom:


test/scenarios/presubmits/el96-src@standard-configuration-1.sh
ggiguash, line 1: Nit: we currently use `@standard-suite1.sh` suffix, so can we use the same naming here?
Something like: `@standard-config1`?

test/scenarios/presubmits/el96-src@standard-configuration-1.sh
```diff
@@ -0,0 +1,29 @@
...
+
+scenario_run_tests() {
+    run_tests host1 \
+          --variable "TARGET_REF:rhel-9.6-microshift-source-fake-next-minor" \
+          suites/backup/backups.robot \
+          suites/backup/backup-and-restore-on-reboot.robot \
+          suites/logscan \
+          suites/storage/ \
+          suites/upgrades/upgrade-fails.robot \
+          suites/upgrades/upgrade-fails-then-recovers.robot
```
ggiguash: Let's watch the timing these tests take. We should try not to prolong the job runtime with this change.

test/scenarios/presubmits/el96-src@standard-configuration-1.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+
+# NOTE: Logscan tests rely on being run IN ORDER to ensure the host
+# is in a good state at the start of each test. We could have
+# separated them and run them as separate scenarios, but did not
+#want to spend the resources on a new VM.
+export TEST_RANDOMIZATION=none
```
ggiguash: Note that this only applies to `logscan`, but we're removing randomization from all the other tests too. Can we call `run_tests` twice? Once with and once without randomization?

PR #5837, opened 2025-12-01
NO-ISSUE: add extra check to not delete cache if only 1 exists

agullon:
Adding an extra check to never delete the cache if there's only 1 cache available.


ggiguash:
Included in https://github.com/openshift/microshift/pull/5839.
Closing the current PR to save on CI cycles.
/close

PR #5851, opened 2025-12-03
USHIFT-6358: Fix CoreDNS flaky test

pacevedom:


copejon:
`el96-src@ipv6` are failing because the kubelet isn't able to pull the bb container image from `quay.io/microshift/busybox`.

The dual-stack scenarios seem to be flaking. In some cases, the scenario times out while waiting for the VM's ip addr. In others, the scenario creates but the test fails because there isn't an IPv6 addr.

test/scenarios-bootc/presubmits/el96-src@standard-suite1.sh
```diff
@@ -14,5 +14,6 @@ scenario_remove_vms() {
 scenario_run_tests() {
     run_tests host1 \
         --variable "EXPECTED_OS_VERSION:9.6" \
+        --variable "SYNC_FREQUENCY:5s" \
```
ggiguash, line 19: We need to add a comment on why we pass sync frequency in this specific test.

test/scenarios/presubmits/el96-src@standard-suite1.sh
```diff
@@ -14,5 +14,6 @@ scenario_remove_vms() {
 scenario_run_tests() {
     run_tests host1 \
         --variable "EXPECTED_OS_VERSION:9.6" \
+        --variable "SYNC_FREQUENCY:5s" \
```
ggiguash, line 19: We need to add a comment on why we pass sync frequency in this specific test.

test/scenarios-bootc/presubmits/el96-src@standard-suite1.sh
pacevedom: Done

test/scenarios/presubmits/el96-src@standard-suite1.sh
pacevedom: Done

ggiguash:
Reapplying LGTM after a minor fix in comments
/lgtm

PR #5796, opened 2025-11-24
USHIFT-6141: SR-IOV RF smoke test

vanhalenar:
This test is deploying a NetworkNodePolicy and checking if the VFs are configured correctly.

PR #5858, opened 2025-12-04
USHIFT-6141: SR-IOV e2e tests

pacevedom:
Built on top of the same commits as https://github.com/openshift/microshift/pull/5796, it supersedes that PR.


test/assets/sriov/sriov-network-policy-template.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: policy-1
+  namespace: sriov-network-operator
```
agullon: `namespace` can be removed from here because it's also specified when applying it. But I don't see a problem to be redundant about this.

pacevedom: Done

PR #5892, opened 2025-12-11
USHIFT-6404: Persist microshift_variant macro in specfile

pmtk:
If not persisted, there's no way to override it when rebuilding the RPMs from SRPM (see https://github.com/microshift-io/microshift/pull/164)


ggiguash:
/retitle [USHIFT-6404](https://issues.redhat.com//browse/USHIFT-6404): Persist microshift_variant macro in specfile

PR #5930, opened 2025-12-17
USHIFT-6402: Test should wait 1 min for gateway route to be created

copejon:


ggiguash:
/retitle [USHIFT-6042](https://issues.redhat.com//browse/USHIFT-6042): Test should wait 1 min for gateway route to be created

ggiguash:
/retitle [USHIFT-6402](https://issues.redhat.com//browse/USHIFT-6402): Test should wait 1 min for gateway route to be created

PR #5933, opened 2025-12-17
USHIFT-6380: Fix OLM test flakes

pacevedom:


pacevedom:
Just found a different issue, will push a different fix.

scripts/mirror-images.sh
```diff
@@ -74,6 +74,20 @@ function mirror_registry() {
...
+        local success=0
+        if skopeo_retry copy ${skopeo_opts} --quiet \
+            --preserve-digests \
+            --authfile "${img_pull_file}" \
+            docker://"${dst_img}" dir://"${tmp_dir}/img" ; then
+        else
+            echo "ERROR: Could not successfully verify mirrored image '${dst_img}'" >&2
+            return 1
+        fi
+        rm -rf "${tmp_dir}"
```
ggiguash: Did we measure how much longer it takes to run mirroring with this verification?
Especially with the images already present on the disk, the second pass performance might be hit significantly.

IMO, I'm not sure we need this type of verification because we never found an evidence that skopeo does not report download errors.

pacevedom: This was a leftover from previous investigation, which is wrong. Now it should be ok, going to run all tests couple of times to see if the flake manifests itself.

copejon:
**e2e-aws-tests-periodics**: greenboot timeout while waiting for cert-manager to start

ggiguash:
Closing in favor of https://github.com/openshift/microshift/pull/5942, which also adds more scenarios
/close

PR #5941, opened 2025-12-18
USHIFT-6411: Scenarios must report junit status when image/commit to start the VM not found

agullon:


ggiguash:
That's going to create a lot of errors because current setup assumes skipping is normal. Many scenarios are conditional on image existence. I'm not sure we will be able to merge the PR with so many errors

/lgtm cancel

PR #5949, opened 2025-12-19
USHIFT-6400: Fix sriov operator to work on RHEL 9 and 10

ggiguash:


ggiguash:
This solution does not work
/close

PR #5958, opened 2025-12-21
USHIFT-6386: Update common_versions.sh file when 4.21 rc is released

ggiguash:


ggiguash:
OLM and Observability errors will be handled elsewhere.
/verified by ci

PR #6006, opened 2026-01-02
USHIFT-6437: Add a delay when restarting MicroShift service

ggiguash:


test/suites/standard1/tls-configuration.robot
```diff
@@ -134,6 +134,7 @@ Setup Invalid TLS Configuration
     [Arguments]    ${config}
     Drop In MicroShift Config    ${config}    10-tls
     Save Journal Cursor
+    Sleep    10s    # To avoid systemctl start error: Start request repeated too quickly
```
pacevedom: Should we just run the restart keyword and expect it to fail instead?

PR #6028, opened 2026-01-06
OCPBUGS-66414: Fix cert-manager rebase to use per-platform images instead of multi-platform manifest lists

ggiguash:


ggiguash:
Failed prometheus image pull with retries.
/test e2e-aws-tests-bootc

pmtk:
Have you tested the changes in the offline env?

There's `assets/optional/cert-manager/manager/images.yaml` file which contains references to images I think are manifest lists.
I think we might need to create versions of images.yaml for each arch and include correct one when building an RPM - something like we have for OLM (and other optional pkgs): https://github.com/openshift/microshift/blob/ce1759e0c658564f59a2a1d04c0a279d0797263e/packaging/rpm/microshift.spec#L413-L419



Edit: what I mean by all this is, even if we package correct refs for mirroring/embedding for disconnected env, if the yaml has multiarch manifest digest, it'll attempt to pull that in order to get ref for its specific architecture.

assets/optional/cert-manager/manager/manager-x86_64.yaml
```diff
@@ -0,0 +1,126 @@
...
+                fieldRef:
+                  fieldPath: metadata.annotations['olm.targetNamespaces']
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: OPERATOR_NAME
+              value: cert-manager-operator
+            - name: RELATED_IMAGE_CERT_MANAGER_WEBHOOK
+              value: registry.redhat.io/cert-manager/jetstack-cert-manager-rhel9@sha256:40b47054e42363a365b491ccfe0d86f109e19fc317058ba834ffe0f5c733880b
```
pmtk: I'm not sure if we need two manager files. See `assets/optional/cert-manager/manager/kustomization.yaml` - it replaces these values using the configmap, so I hope the two images configmap files are enough.
After we resolve this discussion I'll review rest of the PR

ggiguash: That practically means that the write_cert_manager_images_for_arch function should not touch this file.

PR #6078, opened 2026-01-14
WIP: Use legacy servers in subscription-manager

pacevedom:
/hold


pacevedom:
Testing in https://github.com/openshift/release/pull/73482 instead

PR #6082, opened 2026-01-14
OCPBUGS-73946: Ignore (Dev|Tech)PreviewNoUpgrade FeatureSets

copejon:
If microshift config contains `apiServer.featureGates.featureSet: DevPreviewNoUpgrade`, microshift will refuse to start.

After the prohibited featureset is removed from the config, microshift will once again start.

### DIFF
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -6,7 +6,6 @@ import (
 	"slices"
 
 	configv1 "github.com/openshift/api/config/v1"
-	featuresUtils "github.com/openshift/api/features"
 	"github.com/openshift/library-go/pkg/crypto"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -154,28 +153,15 @@ type FeatureGates struct {
 	CustomNoUpgrade CustomNoUpgrade `json:"customNoUpgrade"`
 }
 
+// ToApiserverArgs converts the FeatureGates struct to a list of feature-gates arguments for the kube-apiserver.
+// Validation checks should be performed before calling this function to ensure the FeatureGates struct is valid.
 func (fg FeatureGates) ToApiserverArgs() ([]string, error) {
 	ret := sets.NewString()
-
-	switch fg.FeatureSet {
-	case FeatureSetCustomNoUpgrade:
-		for _, feature := range fg.CustomNoUpgrade.Enabled {
-			ret.Insert(fmt.Sprintf("%s=true", feature))
-		}
-		for _, feature := range fg.CustomNoUpgrade.Disabled {
-			ret.Insert(fmt.Sprintf("%s=false", feature))
-		}
-	case FeatureSetDevPreviewNoUpgrade, FeatureSetTechPreviewNoUpgrade:
-		fgEnabledDisabled, err := featuresUtils.FeatureSets(featuresUtils.SelfManaged, configv1.FeatureSet(fg.FeatureSet))
-		if err != nil {
-			return nil, fmt.Errorf("failed to get feature set gates: %w", err)
-		}
-		for _, f := range fgEnabledDisabled.Enabled {
-			ret.Insert(fmt.Sprintf("%s=true", f.FeatureGateAttributes.Name))
-		}
-		for _, f := range fgEnabledDisabled.Disabled {
-			ret.Insert(fmt.Sprintf("%s=false", f.FeatureGateAttributes.Name))
-		}
+	for _, feature := range fg.CustomNoUpgrade.Enabled {
+		ret.Insert(fmt.Sprintf("%s=true", feature))
+	}
+	for _, feature := range fg.CustomNoUpgrade.Disabled {
+		ret.Insert(fmt.Sprintf("%s=false", feature))
 	}
 	return ret.List(), nil
 }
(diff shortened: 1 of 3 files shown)


copejon:
Unrelated RF test failure from TLS drop-in config suite.

/retest-required

agullon:
@copejon I think we should treat `TechPreviewNoUpgrade` same way as `DevPreviewNoUpgrade` because any FeatureGate in `TechPreviewNoUpgrade` and `DevPreviewNoUpgrade` Feature sets can break MicroShift.

So I suggest expanding the change in this ticket and PR to also `TechPreviewNoUpgrade`.

copejon:
Pausing until we have a decision on MicroShift supporting specific tech-preview features

/hold

pkg/config/apiserver.go
```diff
@@ -190,18 +176,21 @@ func (fg *FeatureGates) validateFeatureGates() error {
...
 		return fmt.Errorf("invalid feature set: %s", fg.FeatureSet)
 	}
-	// Must set FeatureSet to CustomNoUpgrade to use custom feature gates
-	if fg.FeatureSet != FeatureSetCustomNoUpgrade && (len(fg.CustomNoUpgrade.Enabled) > 0 || len(fg.CustomNoUpgrade.Disabled) > 0) {
-		return fmt.Errorf("CustomNoUpgrade must be empty when FeatureSet is empty")
-	}
-	// Must set CustomNoUpgrade enabled or disabled lists when FeatureSet is CustomNoUpgrade
-	if fg.FeatureSet == FeatureSetCustomNoUpgrade && len(fg.CustomNoUpgrade.Enabled) == 0 && len(fg.CustomNoUpgrade.Disabled) == 0 {
-		return fmt.Errorf("CustomNoUpgrade enabled or disabled lists must be set when FeatureSet is CustomNoUpgrade")
-	}
```
agullon, line 204: why do you remove this check? afaiu this is still relevant

pkg/config/apiserver.go
```diff
@@ -190,18 +176,21 @@ func (fg *FeatureGates) validateFeatureGates() error {
...
+
+	// FeatureSet must be empty or CustomNoUpgrade. If empty, CustomNoUpgrade.Enabled/Disabled lists must be empty.
+	switch fg.FeatureSet {
+	case "":
+		if len(fg.CustomNoUpgrade.Enabled) > 0 || len(fg.CustomNoUpgrade.Disabled) > 0 {
+			return fmt.Errorf("CustomNoUpgrade enabled/disabled lists must be empty when FeatureSet is empty")
+		}
+		return nil
+	case FeatureSetCustomNoUpgrade:
+		// Valid - continue to validate enabled/disabled lists below
```
agullon, line 188: what lists below do you refer to?

pkg/config/config_test.go
```diff
@@ -840,23 +841,43 @@ func TestValidate(t *testing.T) {
...
+			name: "feature-gates-required-feature-gate-cannot-be-explicitly-enabled",
+			config: func() *Config {
+				c := mkDefaultConfig()
+				c.ApiServer.FeatureGates.FeatureSet = "CustomNoUpgrade"
+				c.ApiServer.FeatureGates.CustomNoUpgrade.Enabled = []string{"UserNamespacesSupport"}
+				c.ApiServer.FeatureGates.CustomNoUpgrade.Disabled = []string{"feature2"}
+				return c
+			}(),
+			expectErr: true,
+		},
```
agullon, line 875:
```suggestion
		},
		{
			name: "feature-gates-custom-no-upgrade-with-empty-enabled-and-disabled-lists",
			config: func() *Config {
				c := mkDefaultConfig()
				c.ApiServer.FeatureGates.FeatureSet = "CustomNoUpgrade"
				c.ApiServer.FeatureGates.CustomNoUpgrade.Enabled = []string{}
				c.ApiServer.FeatureGates.CustomNoUpgrade.Disabled = []string{}
				return c
			}(),
			expectErr: false,
		},
```

pkg/config/apiserver.go
copejon: Looking at OpenShift kubebuilder validation, empty lists are accepted for CustomNoUpgrade. This makes sense because OpenShift also doesn't restrict changes to the enable/disable FG lists.

pkg/config/apiserver.go
copejon: "lists" just refers to additional validations further down (L195-L217)

pkg/config/apiserver.go
agullon: understood, thanks

pkg/config/apiserver.go
agullon: understood, thanks

pkg/config/config_test.go
copejon: This change would not align with openshift's validation. OpenShift only verifies that the feature set did not change from to CustomNoUpgrade to anything else, and ignores changes to the enabled/disabled list.

agullon: I see, then, should we change the expected error from `true` to `false`?

copejon: Yes please :D

agullon: done, suggestion changed

agullon:
it looks good to me

pmtk (commented):
/lgtm

PR #6083, opened 2026-01-14
OCPBUGS-71234: Permit z-stream upgrades when custom featureGates are enabled

copejon:


pkg/admin/prerun/featuregate_lock.go
```diff
@@ -142,15 +142,14 @@ func validateFeatureGateLockFile(cfg *config.Config) error {
...
+			"Upgrades (major or minor) are not supported when custom feature gates are configured.\n"+
 			"Custom feature gates (%s) were configured in version %s.\n"+
 			"To restore MicroShift to a supported state, you must:\n"+
-			"1. Roll back to version %s, OR\n"+
-			"2. Run: sudo microshift-cleanup-data --all\n"+
-			"3. Remove custom feature gates from /etc/microshift/config.yaml\n"+
-			"4. Restart MicroShift: sudo systemctl restart microshift",
+			"- Roll back MicroShift to version %s, OR\n"+
+			"- Run: sudo microshift-cleanup-data --all\n"+
+			"  and remove custom feature gates from /etc/microshift/config.yaml before restarting MicroShift.",
```
agullon: I prefer previous version of instruction with number points. From my point of view it's more readable.

pkg/admin/prerun/featuregate_lock_test.go
```diff
@@ -393,59 +393,114 @@ func TestFeatureGateLockManagement_ConfigChange(t *testing.T) {
...
+			currentVer:  versionMetadata{Major: 4, Minor: 18, Patch: 0},
+			wantErr:     true,
+			description: "Minor version downgrade (4.19.0 -> 4.18.0) should be blocked",
+		},
+		{
+			name:        "version downgrade major should fail",
+			lockFileVer: versionMetadata{Major: 5, Minor: 0, Patch: 0},
+			currentVer:  versionMetadata{Major: 4, Minor: 18, Patch: 0},
+			wantErr:     true,
+			description: "Major version downgrade (5.0.0 -> 4.18.0) should be blocked",
```
agullon: I don't know if I've understood this change correctly but these unit tests seems to be wrong or at least misleading because FeatureSet feature does not exist before `4.21` so I wouldn't use versions previous like `4.18` or `4.19`. Also, I don't think we should test downgrade paths because they are not supported.

copejon:
Pausing until we have a decision on MicroShift supporting specific tech-preview features

/hold

pkg/admin/prerun/featuregate_lock_test.go
copejon: Makes sense. The values themselves are unimportant so long as they represent different upgrade scenarios.

pkg/admin/prerun/featuregate_lock.go
copejon: reverted

pkg/admin/prerun/featuregate_lock_test.go
```diff
@@ -393,59 +393,114 @@ func TestFeatureGateLockManagement_ConfigChange(t *testing.T) {
...
+		},
+		{
+			name:        "same version should succeed",
+			lockFileVer: versionMetadata{Major: 4, Minor: 21, Patch: 0},
+			currentVer:  versionMetadata{Major: 4, Minor: 21, Patch: 0},
+			wantErr:     false,
+			description: "Same version (4.21.0 -> 4.21.0) should be allowed",
+		},
+		{
+			name:        "version downgrade minor should fail",
```
agullon:
```suggestion
			name:        "minor version downgrade should fail",
```
format consistency with previous unit tests consistency

pkg/admin/prerun/featuregate_lock_test.go
```diff
@@ -393,59 +393,114 @@ func TestFeatureGateLockManagement_ConfigChange(t *testing.T) {
...
+			lockFileVer: versionMetadata{Major: 4, Minor: 22, Patch: 0},
+			currentVer:  versionMetadata{Major: 4, Minor: 21, Patch: 0},
+			wantErr:     true,
+			description: "Minor version downgrade (4.22.0 -> 4.21.0) should be blocked",
+		},
+		{
+			name:        "version downgrade major should fail",
```
agullon:
```suggestion
			name:        "major version downgrade should fail",
```
format consistency with previous unit tests consistency

copejon (commented):
/jira refresh

agullon:
it looks good to me

PR #6097, opened 2026-01-16
dummy PR: 4.21.0-rc.2 Release Testing

agullon:


ggiguash:
The test situation is much better. The only failing test is with DNS Host Config in previous releases.
I would recommend adding a skip for that test in the .robot code if the configmap does not exist.

kasturinarra:
> The test situation is much better. The only failing test is with DNS Host Config in previous releases. I would recommend adding a skip for that test in the .robot code if the configmap does not exist.

If this test is failing i will need to reopen the bug that has been fixed.

PR #6138, opened 2026-01-23
OCPBUGS-71236: Align featureGate validations with openshift

copejon:


copejon:
TODO: I need to update rf featuregate tests

agullon:
it looks good to me

Makefile
```diff
@@ -97,6 +97,7 @@ GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi
 
 # Set variables for test-unit target
 GO_TEST_FLAGS=$(GO_BUILD_FLAGS)
+GO_TEST_ARGS=$(GO_LD_FLAGS)
```
pmtk: I can't find it being used anywhere either in our code or deps/ or vendor/. Why do we need it?

pkg/admin/prerun/featuregate_lock.go
```diff
@@ -127,22 +95,21 @@ func validateFeatureGateLockFile(cfg *config.Config) error {
...
+		return fmt.Errorf("detected invalid changes in feature gate configuration: %w\n\n"+
 			"To restore MicroShift to a supported state, you must:\n"+
 			"1. Run: sudo microshift-cleanup-data --all\n"+
 			"2. Remove custom feature gates from /etc/microshift/config.yaml\n"+
 			"3. Restart MicroShift: sudo systemctl restart microshift", err)
 	}
 
 	// Check if version has changed (upgrade attempted)
-	currentVersion, err := getVersionOfData()
+	currentExecutableVersion, err := getVersionOfData()
```
pmtk: so... version of executable or data?
```
currentExecutableVersion, err := getVersionOfData()
       ^^^^^^^^^^                            ^^^^
```

copejon: good catch!

Makefile
copejon: This is to enable unit testing featuregate prerun functions that handle version skew validation for upgrade/downgrades.

copejon: I'm glad you pointed this out. Relying on the buildflag version values wasn't entirely in line with how the unit tests were checking the validation functions.  The cleaner solution was to mock the function that validation used to get the compiled version values and override it in tests to manipulate the version values as needed.

test/suites/standard2/feature-gates.robot
```diff
@@ -39,7 +33,7 @@ Custom Feature Gates Are Passed To Kube APIServer
     ...    kube-apiserver. This test verifies that arbitrary feature gate values are correctly propagated from the
     ...    MicroShift configuration to the kube-apiserver, regardless of whether the feature gates are valid or have any effect.
     ...    Also verify that feature gate lock file is created when custom feature gates are configured.
-    ...    The lock file prevents upgrades and configuration changes when CustomNoUpgrade feature set is used.
+    ...    The lock file prevents upgrades and configuration changes when CustomNoUpgrade feature set is used.j
```
agullon:
```suggestion
    ...    The lock file prevents upgrades and configuration changes when CustomNoUpgrade feature set is used.
```

pkg/config/config_test.go
```diff
@@ -829,17 +829,6 @@ func TestValidate(t *testing.T) {
...
-			name: "feature-gates-custom-no-upgrade-with-feature-set-empty",
-			config: func() *Config {
-				c := mkDefaultConfig()
-				c.ApiServer.FeatureGates.FeatureSet = ""
-				c.ApiServer.FeatureGates.CustomNoUpgrade.Enabled = []string{"feature1"}
-				c.ApiServer.FeatureGates.CustomNoUpgrade.Disabled = []string{"feature2"}
-				return c
-			}(),
-			expectErr: true,
-		},
```
agullon, line 842: because you are removing this test, I understand the following configuration is a valid one, right? If so, I'd prefer to change the `expectErr:` to `false`
```
apiServer:
  featureGates:
    featureSet: ""
    customNoUpgrade:
      enabled:
      - "CPUManagerPolicyBetaOptions"

File: pkg/config/config_test.go [hunk 5]
```diff
@@ -862,17 +851,6 @@ func TestValidate(t *testing.T) {
...
-		{
-			name: "feature-gates-required-feature-gate-cannot-be-explicitly-enabled",
-			config: func() *Config {
-				c := mkDefaultConfig()
-				c.ApiServer.FeatureGates.FeatureSet = "CustomNoUpgrade"
-				c.ApiServer.FeatureGates.CustomNoUpgrade.Enabled = []string{"UserNamespacesSupport"}
-				c.ApiServer.FeatureGates.CustomNoUpgrade.Disabled = []string{"feature2"}
-				return c
-			}(),
-			expectErr: true,
```
agullon, line 874: Wouldn't be better to change the `expectErr` instead of removing the whole test?

pkg/config/config_test.go
copejon: Right, the case is not validated. I'll re-add and set `expectErr` to `false`.

pkg/config/config_test.go
copejon: Will revert and flip `expectErr`

pkg/config/apiserver.go
```diff
@@ -171,49 +173,53 @@ func (fg FeatureGates) GoString() string {
 	return fmt.Sprintf("FeatureGates{FeatureSet: %q, CustomNoUpgrade: %#v}", fg.FeatureSet, fg.CustomNoUpgrade)
 }
 
+// checkFeatureGateConflict checks if two sets of feature gates have any intersection and returns an error if they do.
+func checkFeatureGateConflict(a, b sets.Set[string], errorMsg string) error {
+	if intersect := a.Intersection(b); intersect.Len() > 0 {
+		return fmt.Errorf("%s: %s", errorMsg, intersect.UnsortedList())
+	}
+	return nil
+}
```
pmtk: nit: since this is only used within the `validateFeatureGates()`, it **could** be defined as variable - similar to `addFeatures := func(features []string, enabled bool)` in `ToApiserverArgs()`

pkg/config/config_test.go
agullon: thanks

pkg/config/config_test.go
agullon: great, thanks

pkg/config/apiserver.go
copejon: That's a great observation! Will make the change

copejon: +1!

PR #6150, opened 2026-01-26
USHIFT-6469: Propagate service failures for correct exit code

pmtk:
Changes the way of restarting MicroShift in case of service failure. Previously ServiceManager would send a SIGTERM to MicroShift process to trigger shutdown, however that carries no context whatsoever.

When service was failing to start and MicroShift quit before sending READY message, systemd would report MicroShift as failing with (readiness notification) protocol error which might be confusing.

To avoid protocol error in case of quiting before readiness, MicroShift needs to exit with non-0 code which will happen if the RunMicroShift() function returns an error.

### DIFF
--- a/pkg/servicemanager/manager.go
+++ b/pkg/servicemanager/manager.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"runtime/debug"
-	"syscall"
 	"time"
 
 	"github.com/openshift/microshift/pkg/servicemanager/startuprecorder"
@@ -20,6 +19,8 @@ type ServiceManager struct {
 	services   []Service
 	serviceMap map[string]Service
 	startRec   *startuprecorder.StartupRecorder
+
+	errChan chan error
 }
 
 func NewServiceManager(startRec *startuprecorder.StartupRecorder) *ServiceManager {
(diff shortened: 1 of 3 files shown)


pkg/servicemanager/manager.go
```diff
@@ -30,8 +31,10 @@ func NewServiceManager(startRec *startuprecorder.StartupRecorder) *ServiceManage
 		services:   []Service{},
 		serviceMap: make(map[string]Service),
 		startRec:   startRec,
+		ErrChan:    make(chan error, 1),
```
pacevedom: Can multiple services fail at the same time? This would only pick up one.

pkg/servicemanager/manager.go
```diff
@@ -20,6 +19,8 @@ type ServiceManager struct {
 	services   []Service
 	serviceMap map[string]Service
 	startRec   *startuprecorder.StartupRecorder
+
+	ErrChan chan error
```
pacevedom: Should we include `func (s *ServiceManager) ErrChan() <-chan error` instead? That way you protect the channel from being written by whoever, as its not exported for writing

pmtk: I had that initially, but considered it unnecessary abstraction :P 
Can add back

pkg/servicemanager/manager.go
pmtk: But how do you wait for many on the other side if one is enough to kill the process?
And that could pollute an image - when first svc fail, the rest is canceled so they could fail with something unrelated

pacevedom (approved):
/lgtm

PR #6155, opened 2026-01-27
USHIFT-6514: Fix version of planned removal of greenboot functions

pmtk:


agullon:
@pmtk do we agree to cherrypick it into release-4.21, right?

PR #6211, opened 2026-02-11
USHIFT-6517: Use host network for bootc image builds for proxy access

ggiguash:
Since we do not delete RHEL 9.8 repo from the images, it must be reusable from cache. 
Previously, we embedded `WEB_SERVER_URL`, which might be pointing to a specific host name or IP.

To make the `.repo` configuration reusable, we can refer to `localhost` and use `--network host` argument **only** when building images.

### DIFF
--- a/test/bin/common.sh
+++ b/test/bin/common.sh
@@ -187,7 +187,7 @@ get_vm_bridge_ip() {
 VM_BRIDGE_IP="$(get_vm_bridge_ip "default")"
 
 # Web server port number
-WEB_SERVER_PORT=8080
+export WEB_SERVER_PORT=8080
 
 # Web server URL using VM bridge IP with fallback to host name
 # shellcheck disable=SC2034  # used elsewhere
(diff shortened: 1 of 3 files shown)


ggiguash:
@coderabbitai perform full review

ggiguash:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > test/bin/pyutils/build_bootc_images.py (1)> `389-397`: _⚠️ Potential issue_ | _🟡 Minor_
> > > **Remove `--network host` flag; bootc-image-builder doesn't require network access.**
> > > The BIB container is given an already-pulled local image reference (`bf_imgref` at line 373) and operates only on local images to produce an ISO. The `--network host` flag is unnecessary here—unlike the earlier `podman build` step (which needs network access to the localhost RPM proxy), bootc-image-builder doesn't perform any network operations. The `--pull=newer` flag won't trigger a fetch since the image reference is local.

@coderabbitai
bootc-image-builder requires localhost access implicitly because it builds images from `localhost/rhel98-test-agent:latest`. That image is configured with repositories referencing `localhost:8080` proxy server.

ggiguash:
Not related to the changes. Saving on the CI cycles.

PR #6455, opened 2026-04-02
USHIFT-6948: Verify local host name in standard suite

ggiguash:
Verify that anaconda correctly sets the host name during kickstart installation.

This PR adds a new test case to the `standard1` suite that checks:
- The hostname is not "localhost" or any variant containing "local"
- The hostname contains the expected suite name (e.g., "standard"), confirming that the testing framework's VM naming convention was applied by anaconda during kickstart

The existing "Verify local name resolution" test is renamed to "Verify Local Host Name Resolution" for clarity.

Jira: https://redhat.atlassian.net/browse/USHIFT-6948

### DIFF
--- a/test/suites/standard1/hostname.robot
+++ b/test/suites/standard1/hostname.robot
@@ -1,5 +1,5 @@
 *** Settings ***
-Documentation       Tests verifying hostname resolution
+Documentation       Tests verifying hostname
 
 Resource            ../../resources/microshift-process.resource
 Resource            ../../resources/microshift-host.resource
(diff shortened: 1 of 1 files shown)


suleymanakbas91:
@ggiguash could you please add a description to the PR explaining briefly why we need this change? As there is no Jira linked, it is difficult to understand the intent.

ggiguash:
> @ggiguash could you please add a description to the PR explaining briefly why we need this change? As there is no Jira linked, it is difficult to understand the intent.

Created a JIRA story in the CI enhancements epic and linked to the PR.

suleymanakbas91:
Thank you!

/lgtm
/retest

PR #6471, opened 2026-04-07
USHIFT-6401, USHIFT-6788: Add fail-fast RBAC bootstrap hook deadlock detection

copejon:
## Summary

- Add parallel RBAC bootstrap hook deadlock detector to kube-apiserver
- Detect deadlock condition in ~15 seconds instead of waiting 60 seconds
- Restart microshift-etcd.scope to recover from deadlock
- **[USHIFT-6788](https://redhat.atlassian.net/browse/USHIFT-6788)**: Add wall-clock deadline (`rbacHookMaxWaitDuration = 30s`) to prevent etcd health flapping from extending detection indefinitely

## Problem

MicroShift enters a permanent crash loop after first restart due to the RBAC bootstrap hook deadlocking when etcd contains existing data. The `rbac/bootstrap-roles` PostStartHook uses `context.TODO()` for API calls with no timeout, causing a circular dependency where the hook waits for API server readiness while the API server waits for the hook to complete.

Related upstream issues: kubernetes/kubernetes#86715, kubernetes/kubernetes#97119

## Root Cause

In `vendor/k8s.io/kubernetes/pkg/registry/rbac/rest/storage_rbac.go`, the `ensureRBACPolicy` function calls:
```go
client.RbacV1().ClusterRoles().List(context.TODO(), metav1.ListOptions{})
```

The loopback client blocks because KAS isn't ready. KAS isn't ready because the hook hasn't completed. `context.TODO()` means no timeout, so it hangs forever. The outer `wait.Poll` has a 30s timeout, but it's ineffective because each poll iteration can hang indefinitely.

## Solution

Add a fail-fast deadlock detector that:
1. Monitors `/readyz/poststarthook/rbac/bootstrap-roles` specifically
2. Checks if etcd is healthy while the hook is stuck
3. If etcd is healthy but hook not completing for 15 seconds, declares deadlock
4. **Enforces absolute 30s wall-clock deadline** regardless of etcd state changes (prevents flapping from extending detection)
5. Restarts `microshift-etcd.scope` to force clean recovery

This approach was chosen because it requires no vendor patches that must be carried through rebases, and keeps all MicroShift-specific logic in MicroShift code rather than modifying upstream Kubernetes sources.

### Patch the vendored hook

**Approach:** Replace `context.TODO()` with the hook's `context.Context` (from `PostStartHookContext`). Three-line change to `storage_rbac.go`.

**Why rejected:** Requires patching both `deps/` and `vendor/` copies of upstream Kubernetes code. This patch must be carried through every rebase, adding maintenance burden and risk of the fix being lost during upstream syncs.

### In-process KAS restart

**Approach:** Bypass `NewAPIServerCommand()` (which has a one-shot signal handler that panics on second call), replicate the RunE setup with pflag, call `kubeapiserver.Run()` directly in a retry loop.

**Why rejected:** Fragile across rebases — must track internal KAS initialization changes. The signal handler workaround is brittle and could break silently with upstream changes. High maintenance burden for marginal benefit over the simpler etcd restart approach.

## Test plan

- [ ] Start MicroShift fresh - verify normal startup
- [ ] Stop and restart MicroShift - verify no crash loop
- [ ] If deadlock occurs, verify detection in ~15 seconds
- [ ] Verify "RBAC bootstrap hook deadlock detected" log message on recovery
- [ ] Verify wall-clock deadline prevents indefinite flapping extension
- [ ] Verify `make verify-go` passes
- [ ] Verify `make test-unit` passes

### DIFF
--- a/pkg/controllers/kube-apiserver.go
+++ b/pkg/controllers/kube-apiserver.go
@@ -62,6 +62,10 @@ const (
 	rbacHookDeadlockTimeout = 15
 	// rbacHookCheckInterval is how often to check the RBAC hook status
 	rbacHookCheckInterval = 2
+	// rbacHookMaxWaitDuration is the absolute maximum time to wait for the RBAC hook
+	// regardless of etcd health state changes. This prevents flapping from extending
+	// detection indefinitely.
+	rbacHookMaxWaitDuration = 30 * time.Second
 )
 
 var (
@@ -456,10 +460,18 @@ func (s *KubeAPIServer) detectRBACHookDeadlock(ctx context.Context, restClient r
 	case <-time.After(5 * time.Second):
 	}
 
+	// Track wall-clock deadline to prevent flapping from extending detection indefinitely
+	startTime := time.Now()
 	checkCount := 0
 	maxChecks := (rbacHookDeadlockTimeout - 5) / rbacHookCheckInterval // Account for initial delay
 
 	for checkCount < maxChecks {
+		// Check absolute deadline first - this cannot be reset by etcd state changes
+		if time.Since(startTime) >= rbacHookMaxWaitDuration {
+			klog.Errorf("RBAC bootstrap hook exceeded maximum wait duration of %v", rbacHookMaxWaitDuration)
+			break
+		}
+
 		select {
 		case <-ctx.Done():
 			return
@@ -486,18 +498,20 @@ func (s *KubeAPIServer) detectRBACHookDeadlock(ctx context.Context, restClient r
 		}
 
 		if etcdHealthy {
-			klog.Warningf("RBAC bootstrap hook not ready (check %d/%d), but etcd is healthy - potential deadlock",
-				checkCount, maxChecks)
+			klog.Warningf("RBAC bootstrap hook not ready (check %d/%d, elapsed %v), but etcd is healthy - potential deadlock",
+				checkCount, maxChecks, time.Since(startTime).Round(time.Second))
 		} else {
 			// etcd not healthy - not a deadlock, just waiting for etcd
 			klog.V(4).Infof("RBAC hook waiting, etcd not yet healthy (check %d/%d)", checkCount, maxChecks)
 			// Reset counter since this isn't a deadlock condition
+			// Note: wall-clock deadline (startTime) is NOT reset - flapping cannot extend indefinitely
 			checkCount = 0
 		}
 	}
 
 	// Reached max checks with etcd healthy but hook not completing - deadlock detected
-	klog.Error("RBAC bootstrap hook deadlock confirmed: etcd healthy but hook not completing")
+	klog.Errorf("RBAC bootstrap hook deadlock confirmed after %v: etcd healthy but hook not completing",
+		time.Since(startTime).Round(time.Second))
 	close(deadlockDetected)
 }
 
--- a/pkg/controllers/kube-apiserver.go
+++ b/pkg/controllers/kube-apiserver.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -55,6 +56,12 @@ import (
 
 const (
 	kubeAPIStartupTimeout = 60
+	// rbacHookDeadlockTimeout is the time to wait for the RBAC bootstrap hook
+	// before declaring a deadlock. This is shorter than kubeAPIStartupTimeout
+	// to allow for faster recovery.
+	rbacHookDeadlockTimeout = 15
+	// rbacHookCheckInterval is how often to check the RBAC hook status
+	rbacHookCheckInterval = 2
 )
 
 var (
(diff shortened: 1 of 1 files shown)


copejon:
@coderabbitai help

pkg/controllers/kube-apiserver.go
```diff
@@ -420,7 +433,127 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+
+		if etcdHealthy {
+			klog.Warningf("RBAC bootstrap hook not ready (check %d/%d), but etcd is healthy - potential deadlock",
+				checkCount, maxChecks)
+		} else {
+			// etcd not healthy - not a deadlock, just waiting for etcd
+			klog.V(4).Infof("RBAC hook waiting, etcd not yet healthy (check %d/%d)", checkCount, maxChecks)
+			// Reset counter since this isn't a deadlock condition
+			checkCount = 0
+		}
```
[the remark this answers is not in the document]
copejon: This is already implemented. Won't Do.

pkg/controllers/kube-apiserver.go
```diff
@@ -420,7 +438,168 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+	_, err = client.Status(checkCtx, "localhost:2379")
+	if err != nil {
+		return false, nil // etcd not healthy, but not an error condition
+	}
+
+	return true, nil
+}
+
+// restartMicroshiftEtcdScope restarts the microshift-etcd.scope to recover from deadlock.
+// This forces a clean restart of etcd which can help break the circular dependency.
```
pmtk, line 585: Restarting MicroShift from within doesn't cause that?

pkg/controllers/kube-apiserver.go
```diff
@@ -420,7 +438,168 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+// detectRBACHookDeadlock monitors the RBAC bootstrap hook status and detects deadlock conditions.
+// A deadlock is detected when:
+// 1. The RBAC hook is not completing (stuck in "not finished" state)
+// 2. etcd is healthy and responsive
+// This indicates the circular dependency where the hook waits for API server
+// while API server waits for the hook.
+//
+// Closed upstream Kubernetes issues:
+// https://github.com/kubernetes/kubernetes/issues/86715
+// https://github.com/kubernetes/kubernetes/issues/97119
```
pmtk, line 459: This ticket suggest that if we restart microshift, we just start from the beginning? How robust is this solution because if the API server has trouble reconciling RBACs, we'll just restart endlessly.
Or is it a best effort fix?

pkg/controllers/kube-apiserver.go
```diff
@@ -420,7 +438,168 @@ func (s *KubeAPIServer) Run(ctx context.Context, ready chan<- struct{}, stopped
...
+
+// detectRBACHookDeadlock monitors the RBAC bootstrap hook status and detects deadlock conditions.
+// A deadlock is detected when:
+// 1. The RBAC hook is not completing (stuck in "not finished" state)
+// 2. etcd is healthy and responsive
+// This indicates the circular dependency where the hook waits for API server
+// while API server waits for the hook.
+//
+// Closed upstream Kubernetes issues:
+// https://github.com/kubernetes/kubernetes/issues/86715
```
pmtk, line 458: What about the last comment in that issue? https://github.com/kubernetes/kubernetes/issues/86715#issuecomment-967048646
If this problems comes from not-fast-enough system, then we're just putting a bandaid that might blow up in different, unexpected place

pmtk (commented):
This seems to be very tailored fix for a problem that is very unique. I think we need consensus from wider team on this

pkg/controllers/kube-apiserver.go
copejon: It would, but it would take longer.

pkg/controllers/kube-apiserver.go
copejon: Right. I'm rethinking the approach on this, but the solution pattern is still the same in that we should restart the KAS internally.

pkg/controllers/kube-apiserver.go
copejon: Maybe but it's impossible to predict what that will be. Regardless, the bug is expressed when the system becomes resource-starved, which can happen no matter how much hardware you have.

pkg/controllers/kube-apiserver.go
copejon: Rebooting the whole system is a crap-shoot because the resource-starvation is given another opportunity to occur. This way, microshift service start-up pauses until this passes or the timeout is reached. That means the affected service has more headroom to boot up, hopefully reducing the race window significantly.

pmtk:
@copejon have you considered patching deps/ (vendor/) to introduce some context deadline for RBAC? Might more compact solution

copejon:
> @copejon have you considered patching deps/ (vendor/) to introduce some context deadline for RBAC? Might more compact solution

I've tried to avoid patching deps directly, but it would be a much simpler approach. The upstream community isn't interested in fixing it, so the bug isn't going anywhere.

copejon:
closing in favor of simpler solution in #6635 

/close

PR #6489, opened 2026-04-10
USHIFT-6715: Enable building multiple bootc blueprint layers in parallel

vanhalenar:
## Summary by CodeRabbit

* **New Features**
  * Per-OS layer selection for combined el9/el10 image builds.
  * Release-specific layer build handling in CI pipelines.
  * Support processing multiple layer directories in a single invocation for combined OS builds.

* **Chores**
  * Improved validation of provided layer directories with explicit existence checks.


test/bin/pyutils/build_bootc_images.py
```diff
@@ -581,7 +581,7 @@ def main():
                         choices=["image-bootc", "containerfile", "container-encapsulate"],
                         help="Only build images of the specified type.")
     dirgroup = parser.add_mutually_exclusive_group(required=False)
-    dirgroup.add_argument("-l", "--layer-dir", type=str, help="Path to the layer directory to process.")
+    dirgroup.add_argument("-l", "--layer-dir", type=str, help="Path to the layer directory to process. Accepts comma-separated list of directories.")
```
pmtk: How about using `action="append"`?

From the [docs](https://docs.python.org/3/library/argparse.html#action):
```
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--foo', action='append', default=['0'])
>>> parser.parse_args('--foo 1 --foo 2'.split())
Namespace(foo=['0', '1', '2'])
```

```suggestion
    dirgroup.add_argument("-l", "--layer-dir", action="append", default=[], help="Path to the layer directory to process. Can be specified multiple times.")
```
and then:
```
$ ./bin/build_bootc_images.sh -l ./image-blueprints-bootc/el9/layer1-base -l ./image-blueprints-bootc/el10/layer1-base
```

vanhalenar: i like this better than comma separating, reworked the script to use `action="append"`

PR #6505, opened 2026-04-15
USHIFT-6859: Update generate_common_versions.py to take "X.Y" as argument for 5.0 handling

pmtk:
## Summary by CodeRabbit

* **Refactor**
  * CLI now accepts a single required version string in X.Y format instead of separate major/minor arguments.
  * Corrected a cross-major example so the prior-release minor for 5.0 is now 4.22.

* **Chore**
  * Updated how the verification/regeneration process derives the version string, which changes the expected regenerated version output.

### DIFF
--- a/test/bin/pyutils/generate_common_versions.py
+++ b/test/bin/pyutils/generate_common_versions.py
@@ -19,7 +19,7 @@ import ghutils   # noqa: E402
 ARCH = os.uname().machine
 
 # Version map defining the last minor version for each major version.
-# Used for cross-major Y-1/Y-2 calculations (e.g., 5.0's Y-1 is 4.18).
+# Used for cross-major Y-1/Y-2 calculations (e.g., 5.0's Y-1 is 4.22).
 VERSION_MAP = {
     4: {'last_minor': 22}
 }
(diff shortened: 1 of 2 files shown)


test/bin/pyutils/generate_common_versions.py
pacevedom: This is still using args.major, it will break when creating a PR

test/bin/pyutils/generate_common_versions.py
```diff
@@ -420,8 +426,8 @@ def main():
             exit(0)
 
         base_branch = g.git_repo.active_branch.name
-        if not base_branch.startswith(f"release-{args.major}"):
-            logging.error(f"Script is expected to be executed on branch starting with 'release-{args.major}', but it's {base_branch}")
+        if not base_branch.startswith(f"release-{args.version}"):
```
pacevedom, line 429: I think there is another bug here. `args.version` resolves to a tuple after calling parse_major_minor, and when using format strings this prints the parentheses around it. With 4.22 this would look like `release-(4,22)`

pmtk: good catch

pacevedom (approved):
/lgtm

PR #6545, opened 2026-04-20
USHIFT-6794: Config for Cluster 2 Cluster Communication

pmtk:
## Summary by CodeRabbit

* **New Features**
  * Added remote-cluster connectivity config: define remote cluster/service CIDRs, next-hop routes, and optional DNS domains; feature enables when remotes are provided and validates conflicts, routing loops, host-interface overlap, and IP-family compatibility.

* **Tests**
  * Added comprehensive validation and enablement tests covering parsing, overlap detection, uniqueness, mask/cardinality rules, host-interface checks, and DNS validation.

* **Documentation**
  * Added user config examples showing the new c2cc section.

* **Chores**
  * Updated configuration schema and packaging defaults to include c2cc.

### DIFF
--- a/packaging/microshift/config.yaml
+++ b/packaging/microshift/config.yaml
@@ -41,7 +41,7 @@ apiServer:
         # to serve from the API server. Allowed values: VersionTLS12, VersionTLS13.
         # Defaults to VersionTLS12.
         minVersion: VersionTLS12
-c2cc:
+clusterToCluster:
     # List of remote clusters to establish connectivity with.
     # C2CC is disabled when this list is empty.
     remoteClusters:
--- a/packaging/microshift/config.yaml
+++ b/packaging/microshift/config.yaml
@@ -41,6 +41,20 @@ apiServer:
         # to serve from the API server. Allowed values: VersionTLS12, VersionTLS13.
         # Defaults to VersionTLS12.
         minVersion: VersionTLS12
+c2cc:
+    # List of remote clusters to establish connectivity with.
+    # C2CC is disabled when this list is empty.
+    remoteClusters:
+        - # Pod CIDRs of the remote cluster. Must not overlap with local cluster or other remotes.
+          clusterNetwork: []
+          # DNS domain suffix for the remote cluster (e.g., "cluster-b.remote").
+          # Services are reachable as <svc>.<ns>.svc.<domain>.
+          # Optional — if empty, no DNS forwarding is configured for this remote.
+          domain: ""
+          # IP address of the remote cluster's node, used as next-hop for routing.
+          nextHop: ""
+          # Service CIDRs of the remote cluster. Must not overlap with local cluster or other remotes.
+          serviceNetwork: []
 debugging:
     # Valid values are: "Normal", "Debug", "Trace", "TraceAll".
     # Defaults to "Normal".
--- a/pkg/config/c2cc.go
+++ b/pkg/config/c2cc.go
@@ -15,6 +15,9 @@ type C2CC struct {
 	// List of remote clusters to establish connectivity with.
 	// C2CC is disabled when this list is empty.
 	RemoteClusters []RemoteCluster `json:"remoteClusters,omitempty"`
+
+	// Populated during validation with parsed network objects.
+	Resolved []ResolvedRemoteCluster `json:"-"`
 }
 
 type RemoteCluster struct {
@@ -28,6 +31,18 @@ type RemoteCluster struct {
 	// Services are reachable as <svc>.<ns>.svc.<domain>.
 	// Optional — if empty, no DNS forwarding is configured for this remote.
 	Domain string `json:"domain,omitempty"`
+}
+
+type ResolvedRemoteCluster struct {
+	NextHop        net.IP
+	ClusterNetwork []*net.IPNet
+	ServiceNetwork []*net.IPNet
+	Domain         string
+}
+
+type labeledCIDR struct {
+	net *net.IPNet
+	str string
 }
 
 func (c *C2CC) IsEnabled() bool {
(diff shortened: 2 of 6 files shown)


pkg/config/c2cc.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+		}
+	}
+	return errs
+}
+
+func cidrsOverlap(a, b string) bool {
+	_, netA, errA := net.ParseCIDR(a)
+	_, netB, errB := net.ParseCIDR(b)
+	if errA != nil || errB != nil {
```
agullon: should we add a check before this one to report an error if errA or errB are `nil`?
for example, we can silently return `false` when errA is `nil`

pkg/config/c2cc_test.go
```diff
@@ -0,0 +1,519 @@
...
+						NextHop:        "10.100.0.2",
+						ClusterNetwork: []string{"10.45.0.0/16"},
+						ServiceNetwork: []string{"10.46.0.0/16"},
+					}},
+				})
+				cfg.Network.CNIPlugin = CniPluginNone
+				return cfg
+			}(),
+			expectErr: true,
+			errMsg:    "requires OVN-Kubernetes",
```
agullon, line 300: why this is valid error if CNI can be unset? [see this error message](https://github.com/openshift/microshift/pull/6545/changes#diff-33f015a27c6f16fd92c4faf87ea3650f37bfaeccaded03df44642adce44dad1dR73)

pkg/config/c2cc_test.go
```diff
@@ -0,0 +1,519 @@
...
+					{
+						NextHop:        "10.100.0.3",
+						ClusterNetwork: []string{"10.55.0.0/16"},
+						ServiceNetwork: []string{"10.45.0.0/16"},
+					},
+				},
+			}),
+			expectErr: true,
+			errMsg:    "overlaps with",
+		},
```
agullon, line 230: should we add this test? not sure about the `errMsg` though
```suggestion
		},
	    {
			name: "clusterNetwork and serviceNetwork overlap",
			cfg: mkC2CCConfig(C2CC{
				RemoteClusters: []RemoteCluster{
					{
						NextHop:        "10.100.0.2",
						ClusterNetwork: []string{"10.45.0.0/16"},
						ServiceNetwork: []string{"10.45.0.0/16"},
					},
				},
			}),
			expectErr: true,
			errMsg:    "overlaps with",
		},
```

pkg/config/c2cc_test.go
```diff
@@ -0,0 +1,519 @@
...
+			cfg: mkC2CCConfig(C2CC{
+				RemoteClusters: []RemoteCluster{{
+					NextHop:        "10.100.0.2",
+					ClusterNetwork: []string{"10.45.0.0/16", "10.47.0.0/16"},
+					ServiceNetwork: []string{"10.46.0.0/16"},
+				}},
+			}),
+			expectErr: true,
+			errMsg:    "multiple IPv4 entries",
+		},
```
agullon, line 325: This check is missing on IPv6
```suggestion
		},
		{
			name: "multiple IPv6 entries in clusterNetwork",
			cfg: mkC2CCConfig(C2CC{
				RemoteClusters: []RemoteCluster{{
					NextHop:        "10.100.0.2",
                     ClusterNetwork: []string{"fd03::/48", "fd05::/48"},
                     ServiceNetwork: []string{"fd04::/112"},
				}},
			}),
			expectErr: true,
			errMsg:    "multiple IPv6 entries",
		},
```

pkg/config/c2cc.go
pmtk: those CIDRs are validated in `validateRemoteCIDR`:
```
	_, ipNet, err := net.ParseCIDR(cidr)
	if err != nil {
		return fmt.Errorf("%s %q is not a valid CIDR: %w", field, cidr, err)
	}
```

pkg/config/c2cc_test.go
pmtk: But this is `CniPluginNone` not `Unset`, right?
(I think) we use `None` in upstream to deploy Kindnet

agullon: I see, so if we use `CniPluginNone` and `Unset` I suggest we add a unit test to validate the `unset` as a valid option.

pkg/config/c2cc_test.go
pmtk: added

pkg/config/c2cc_test.go
pmtk: added

packaging/microshift/config.yaml
```diff
@@ -41,6 +41,20 @@ apiServer:
         # to serve from the API server. Allowed values: VersionTLS12, VersionTLS13.
         # Defaults to VersionTLS12.
         minVersion: VersionTLS12
+c2cc:
```
ggiguash: The `c2cc` name in the configuration file may seem a bit cryptic for some users.
Can we make it more explicit, like `clusterToCluster`, etc.?

pmtk: That's a good point. Changed to `clusterToCluster`

pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type Config struct {
 
 	GenericDevicePlugin GenericDevicePlugin `json:"genericDevicePlugin"`
 
+	C2CC C2CC `json:"clusterToCluster"`
```
pacevedom, line 65: Should this be optional instead of remoteClusters inside c2cc? It is required now because it doesnt have an `omitempty`. Then the remoteClusters inside is optional, which creates a confusing setup?

packaging/microshift/config.yaml
```diff
@@ -41,6 +41,20 @@ apiServer:
...
+        - # Pod CIDRs of the remote cluster. Must not overlap with local cluster or other remotes.
+          clusterNetwork: []
+          # DNS domain suffix for the remote cluster (e.g., "cluster-b.remote").
+          # Services are reachable as <svc>.<ns>.svc.<domain>.
+          # Optional — if empty, no DNS forwarding is configured for this remote.
+          domain: ""
+          # IP address of the remote cluster's node, used as next-hop for routing.
+          nextHop: ""
+          # Service CIDRs of the remote cluster. Must not overlap with local cluster or other remotes.
+          serviceNetwork: []
```
pacevedom, line 57: Are these defaults ok? Some might fail to validate?

pmtk: There's a code to remove empty objects like this to make sure default config works (which we have test for)

pkg/config/config.go
pmtk: None of fields in `Config` has omitempty and they still work as expected.
As for `remoteClusters` - I guess it only matters for `microshift config --effective` or whatever the command is?

pacevedom (approved):
/lgtm

PR #6550, opened 2026-04-21
USHIFT-6868: Increase greenboot timeout for release scenarios

agullon:
## Summary
- Set `export GREENBOOT_TIMEOUT=1200` in all release scenario files (bootc and non-bootc)
- Release scenarios use images with all optional operators (OLM, cert-manager, gateway-api, SR-IOV, LVMS) that need more time to pull images and start
- The default `GREENBOOT_TIMEOUT=600` (10 min) is insufficient for the full optional stack, especially on ARM
- Presubmit scenarios already set `GREENBOOT_TIMEOUT=1200` explicitly but release scenarios were missed
- Excludes low-latency scenarios which have `SKIP_GREENBOOT=true`

## Test plan
- [x] Monitor next runs of periodic bootc release jobs (ARM and x86)
- [x] Verify release scenarios with optional images pass greenboot healthcheck

Jira: https://issues.redhat.com/browse/USHIFT-6868

## Tests
* Enhanced release test reliability by increasing the greenboot timeout to 1200 seconds across all release validation scenarios. This configuration change allocates more time for system boot verification during testing phases, reducing timeout-related failures and improving overall release quality across various platform configurations and upgrade scenarios.

### DIFF
--- a/test/scenarios-bootc/el10/releases/el102-lrel@ai-model-serving-online.sh
+++ b/test/scenarios-bootc/el10/releases/el102-lrel@ai-model-serving-online.sh
@@ -2,6 +2,10 @@
 
 # Sourced from scenario.sh and uses functions defined there.
 
+# Increase greenboot timeout for release scenarios (optional operators need more time)
+# shellcheck disable=SC2034  # used elsewhere
+export GREENBOOT_TIMEOUT=1200
+
 start_image="rhel102-bootc-brew-lrel-optional"
 
 # Currently, RHOAI is only available for x86_64
--- a/test/scenarios-bootc/el10/releases/el102-lrel@configuration.sh
+++ b/test/scenarios-bootc/el10/releases/el102-lrel@configuration.sh
@@ -2,6 +2,10 @@
 
 # Sourced from scenario.sh and uses functions defined there.
 
+# Increase greenboot timeout for release scenarios (optional operators need more time)
+# shellcheck disable=SC2034  # used elsewhere
+export GREENBOOT_TIMEOUT=1200
+
 start_image="rhel102-bootc-brew-lrel-optional"
 
 scenario_create_vms() {
--- a/test/scenarios-bootc/el10/releases/el102-lrel@dual-stack.sh
+++ b/test/scenarios-bootc/el10/releases/el102-lrel@dual-stack.sh
@@ -2,6 +2,10 @@
 
 # Sourced from scenario.sh and uses functions defined there.
 
+# Increase greenboot timeout for release scenarios (optional operators need more time)
+# shellcheck disable=SC2034  # used elsewhere
+export GREENBOOT_TIMEOUT=1200
+
 start_image="rhel102-bootc-brew-lrel-optional"
 
 scenario_create_vms() {
--- a/test/scenarios-bootc/el10/releases/el102-lrel@fips.sh
+++ b/test/scenarios-bootc/el10/releases/el102-lrel@fips.sh
@@ -2,6 +2,10 @@
 
 # Sourced from scenario.sh and uses functions defined there.
 
+# Increase greenboot timeout for release scenarios (optional operators need more time)
+# shellcheck disable=SC2034  # used elsewhere
+export GREENBOOT_TIMEOUT=1200
+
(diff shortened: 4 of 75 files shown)


test/bin/scenario.sh
```diff
@@ -1448,6 +1448,13 @@ load_scenario_script() {
 
     # shellcheck source=/dev/null
     source "${SCENARIO_SCRIPT}"
+
+    # Auto-increase greenboot timeout for release scenarios
+    # because all optional RPMs are installed and operators need more time to pull images
+    if [[ "${SCENARIO_SCRIPT}" == */releases/* && "${GREENBOOT_TIMEOUT}" -le 600 ]]; then
```
pacevedom, line 1454: Do we have only optionals in the release tests?

agullon: @pacevedom In all release tests we are installing always all the RPM packages, [see this as an example](https://github.com/openshift/microshift/blob/main/test/image-blueprints-bootc/el10/layer4-release/group1/rhel102-bootc-brew-lrel-optional.containerfile#L21-L34).

pacevedom, line 1454: The flow of how scenarios are sourced is a bit different when running locally than CI. This is the CI flow:
- [`get_source_dir $SCENARIO_TYPE`](https://github.com/openshift/release/blob/main/ci-operator/step-registry/openshift/microshift/e2e/metal-tests/openshift-microshift-e2e-metal-tests-commands.sh#L23) will get the right directory, in this case its `[scenarios-bootc/releases](https://github.com/openshift/release/blob/main/ci-operator/step-registry/openshift/microshift/includes/openshift-microshift-includes-commands.sh#L745)`. This is the value for `SCENARIO_SOURCES` now.
- The `SCENARIO_SOURCES` gets transformed into [`SCENARIOS_TO_RUN`](https://github.com/openshift/microshift/blob/main/test/bin/ci_phase_boot_and_test.sh#L16). Note that this is a different directory, now off from the original location. Its in the OUTPUT_DIR and the name is built from `scenarios-` and the result of [this function](https://github.com/openshift/microshift/blob/main/test/bin/common.sh#L263). In this case the name is `scenarios-bootc`.
- [`prepare_scenario_sources`](https://github.com/openshift/microshift/blob/main/test/bin/ci_phase_boot_and_test.sh#L21) will copy all the scenarios defined in `SCENARIO_SOURCES` into `SCENARIOS_TO_RUN` directory.
- Tests are launched reading from [`SCENARIOS_TO_RUN`](https://github.com/openshift/microshift/blob/main/test/bin/ci_phase_boot_and_test.sh#L94) directory.

So this approach works locally but not in CI. The one that works everywhere is to define the variable per scenario, even if its all of them (which is also the pattern in the rest of the scenarios, presubmits or periodics).

agullon: I see, but if when I continue the execution flow the changes I added always run.

1. `scenario.sh create-and-run` execution from https://github.com/openshift/microshift/blob/main/test/bin/ci_phase_boot_and_test.sh#L94
2. Triggers action_create() https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L1647
3. Trigger load_scenario_script() https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L1502

`load_scenario_script() ` contains the change I proposed in this PR.
 Am I missing something?
 
In this PR CI jobs logs I can see the new code exercised, [look for scenario.sh:1454](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_microshift/6550/pull-ci-openshift-microshift-main-e2e-aws-tests/2046524980115542016/artifacts/e2e-aws-tests/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el98-src@standard-suite1/boot_and_run.log)

pacevedom: What I mean is that the code within the if statement is never going to run because there is no `release` in the path in CI, only if you do it manually/locally. In other words, the greenboot timeout is not being modified, you have to do this in the scenario itself.

agullon: understood, thanks, I'll propose a different approach

pacevedom (approved):
/lgtm

PR #6568, opened 2026-04-23
USHIFT-6886: Comment out tuned boot-wait in ginkgo-multi-config scenario

agullon:
## Summary
- Comment out the boot-wait loop in `el98-lrel@ginkgo-multi-config` that expects a `microshift-tuned` reboot
- The scenario uses `start_image=rhel98-brew-lrel-optional` which has no tuned configuration, so the reboot never happens and the 60s timeout always fires
- The tuned image blueprint is disabled because `kernel-rt` is unavailable on RHEL 9.8

## Test plan
- [ ] Verify `el98-lrel@ginkgo-multi-config` scenario passes in periodic CI (no longer times out waiting for 2 boots)
- [ ] Verify the scenario still validates TLSv1.3, LVMS, IPv6, and ginkgo tests

### DIFF
--- a/test/scenarios/releases/el98-lrel@ginkgo-multi-config.sh
+++ b/test/scenarios/releases/el98-lrel@ginkgo-multi-config.sh
@@ -48,20 +48,22 @@ scenario_remove_vms() {
 scenario_run_tests() {
     exit_if_commit_not_found "${start_image}"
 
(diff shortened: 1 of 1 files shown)


agullon:
Closing: retargeting to main branch. Replaced by new PR.

PR #6621, opened 2026-05-04
USHIFT-6912: Add DNS deployment resource configuration support

Neilhamza:
Allow users to configure CPU and memory resources for the dns container in the dns-default DaemonSet via dns.resources in config.yaml. Defaults to the current hardcoded values (cpu=50m, memory=70Mi requests, no limits) when not configured.

## Summary by CodeRabbit

* **New Features**
  * DNS container CPU/memory can be configured via requests and optional limits; defaults apply when requests omitted.

* **Documentation**
  * Configuration examples and default settings updated to show dns.resources.requests and dns.resources.limits.

* **Bug Fixes**
  * Validation now rejects invalid quantities and limits smaller than requests; invalid configs prevent startup with clear errors.

* **Tests**
  * Added unit and integration tests plus a Robot Framework suite covering parsing, validation, overrides, and runtime behavior.

### DIFF
--- a/assets/components/openshift-dns/dns/daemonset.yaml
+++ b/assets/components/openshift-dns/dns/daemonset.yaml
@@ -57,8 +57,15 @@ spec:
             failureThreshold: 5
           resources:
             requests:
-              cpu: 50m
-              memory: 70Mi
+              {{- range $key, $value := .DNSRequests }}
+              {{ $key }}: {{ $value }}
+              {{- end }}
+            {{- if .DNSLimits }}
+            limits:
+              {{- range $key, $value := .DNSLimits }}
+              {{ $key }}: {{ $value }}
+              {{- end }}
+            {{- end }}
           securityContext:
             readOnlyRootFilesystem: true
           image: '{{ .ReleaseImage.coredns }}'
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -444,6 +444,24 @@ func (c *Config) incorporateUserSettings(u *Config) {
 			c.DNS.Hosts.File = u.DNS.Hosts.File
 		}
 	}
+
+	// DNS resource configuration - merge key-by-key to preserve defaults
+	if u.DNS.Resources.Requests != nil {
+		if c.DNS.Resources.Requests == nil {
+			c.DNS.Resources.Requests = make(map[string]string)
+		}
+		for k, v := range u.DNS.Resources.Requests {
+			c.DNS.Resources.Requests[k] = v
+		}
+	}
+	if u.DNS.Resources.Limits != nil {
+		if c.DNS.Resources.Limits == nil {
+			c.DNS.Resources.Limits = make(map[string]string)
+		}
+		for k, v := range u.DNS.Resources.Limits {
+			c.DNS.Resources.Limits[k] = v
+		}
+	}
 	if u.ApiServer.FeatureGates.FeatureSet != "" {
 		c.ApiServer.FeatureGates.FeatureSet = u.ApiServer.FeatureGates.FeatureSet
 	}
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -80,6 +80,81 @@ func TestGetActiveConfigFromYAML(t *testing.T) {
 			}(),
 		},
 		{
+			name: "dns-resources-requests",
+			config: dedent(`
+            dns:
+              resources:
+                requests:
+                  cpu: "100m"
+                  memory: "150Mi"
+            `),
+			expected: func() *Config {
+				c := mkDefaultConfig()
+				c.DNS.Resources.Requests = map[string]string{
+					"cpu":    "100m",
+					"memory": "150Mi",
+				}
+				return c
+			}(),
+		},
+		{
+			name: "dns-resources-partial-request",
+			config: dedent(`
+            dns:
+              resources:
+                requests:
+                  cpu: "100m"
+            `),
+			expected: func() *Config {
+				c := mkDefaultConfig()
+				c.DNS.Resources.Requests["cpu"] = "100m"
+				return c
+			}(),
+		},
+		{
+			name: "dns-resources-with-limits",
+			config: dedent(`
+            dns:
+              resources:
+                requests:
+                  cpu: "100m"
+                  memory: "150Mi"
+                limits:
+                  cpu: "200m"
+                  memory: "256Mi"
+            `),
+			expected: func() *Config {
+				c := mkDefaultConfig()
+				c.DNS.Resources.Requests = map[string]string{
+					"cpu":    "100m",
+					"memory": "150Mi",
+				}
+				c.DNS.Resources.Limits = map[string]string{
+					"cpu":    "200m",
+					"memory": "256Mi",
+				}
+				return c
+			}(),
+		},
+		{
+			name: "dns-resources-limits-only",
+			config: dedent(`
+            dns:
+              resources:
+                limits:
+                  cpu: "200m"
+                  memory: "256Mi"
+            `),
+			expected: func() *Config {
+				c := mkDefaultConfig()
+				c.DNS.Resources.Limits = map[string]string{
+					"cpu":    "200m",
+					"memory": "256Mi",
+				}
+				return c
+			}(),
+		},
+		{
 			name: "network",
 			config: dedent(`
             network:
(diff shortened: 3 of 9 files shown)


Neilhamza:
@coderabbitai resolve

agullon:
hi @Neilhamza thank you very much for creating the PR early in the dev cycle.
Only one question from QE POV, are you planning to add RF tests in current PR or in a separate one?
I think best practise is to added them also in this PR to validate changes are working as expected. Let me know if you have any questions or I can help you with RF tests.

Neilhamza:
Thanks for the feedback @agullon !! i was thinking about a seperate PR so i wont bloat this one, but if adding the tests here is the accepted approach i can do it, what do you think? :)

agullon:
> Thanks for the feedback @agullon !! i was thinking about a seperate PR so i wont bloat this one, but if adding the tests here is the accepted approach i can do it, what do you think? :)

I don't have a strong opinion about when/where add the RF, it's up to you. As I said I think it's better to add them along with product change. This PR is not that big, no problem.

Neilhamza:
@coderabbitai perform full review

Neilhamza:
@agullon added RF testscoverage - : default resources, custom requests+limits, requests only, partial requests, invalid quantity rejection, and  DNS resolution after change

assets/components/openshift-dns/dns/daemonset.yaml
```diff
@@ -57,8 +57,17 @@ spec:
...
+              memory: {{ .DNSMemoryRequest }}
+            {{- if .DNSHasLimits }}
+            limits:
+              {{- if .DNSCPULimit }}
+              cpu: {{ .DNSCPULimit }}
+              {{- end }}
+              {{- if .DNSMemoryLimit }}
+              memory: {{ .DNSMemoryLimit }}
+              {{- end }}
+            {{- end }}
```
pmtk: Why not just give customers whole `resources: {}`? If they define anything, we take it whatever they provide. Would simplify templating greatly

eslutsky: I agree . since resource is something  generic we might be able to reuse it for different components besides dns. (ie ingress)

pacevedom:
/hold
Wait for [enhancement](https://github.com/openshift/enhancements/pull/1989) to merge first to avoid subsequent PRs.

assets/components/openshift-dns/dns/daemonset.yaml
Neilhamza: Nice! simplified the template to pass resource map directly with range loops of 2 params instead of 7 and is reusable for other components

test/suites/configuration1/dns-resource-configuration.robot
```diff
@@ -0,0 +1,189 @@
...
+...                                 \ \ resources:
+...                                 \ \ \ requests:
+...                                 \ \ \ \ cpu: "200m"
+...                                 \ \ \ limits:
+...                                 \ \ \ \ cpu: "50m"
+
+
+*** Test Cases ***
+Default DNS Resources
+    [Documentation]    Verify default DNS resources when no custom config is applied
```
agullon, line 68:
```suggestion
    [Documentation]    Verify default DNS resources when no custom config is applied
    [Setup]    Remove DNS Resource Config
```
Tests must be indempotent, meaning regardless the previous status of the system they must always pass. For that reason the test must enforce default DNS config before checking it.

test/suites/configuration1/dns-resource-configuration.robot
```diff
@@ -0,0 +1,189 @@
...
+    ${limits}=    Oc Get JsonPath    daemonset    openshift-dns    dns-default
+    ...    .spec.template.spec.containers[0].resources.limits
+    Should Be Equal As Strings    ${cpu}    100m
+    Should Be Equal As Strings    ${memory}    150Mi
+    Should Be Empty    ${limits}
+    [Teardown]    Remove DNS Resource Config
+
+Partial Requests Preserves Defaults
+    [Documentation]    Configure only CPU request and verify memory default is preserved
+    [Setup]    Apply DNS Resource Config    ${DNS_PARTIAL_REQUESTS}
```
agullon, line 101: the setup of this test only set requests.cpuin the config.
What would happen if requests.memory was already set to a value different than 70Mi (default) before this test starts? I think it would failed because the different value will still be present when the check is done.
To avoid this test must remove all configurations and then apply any config before doing any check.

Something not great about this suggestion is that MicroShift will restart many times (I think 3) on every test, which will lead into long test executions. Checks take less than a second. Successful microshift restart at least 1 min. We must have this in mind. An alternative may be running all the tests in a predefined order to reduce the number of times microshift restarts.

test/suites/configuration/dns-resource-configuration.robot
```diff
@@ -0,0 +1,189 @@
...
+...                                 \ \ \ requests:
+...                                 \ \ \ \ cpu: "200m"
+...                                 \ \ \ limits:
+...                                 \ \ \ \ cpu: "50m"
+
+
+*** Test Cases ***
+Default DNS Resources
+    [Documentation]    Verify default DNS resources when no custom config is applied
+    ${cpu}=    Oc Get JsonPath    daemonset    openshift-dns    dns-default
```
agullon: `Oc Get JsonPath    daemonset    openshift-dns    dns-default` is duplicated 15 times. Can you move it into a Keyword

test/suites/configuration/dns-resource-configuration.robot
```diff
@@ -0,0 +1,189 @@
...
+    ...    .spec.template.spec.containers[0].resources.requests.cpu
+    ${memory}=    Oc Get JsonPath    daemonset    openshift-dns    dns-default
+    ...    .spec.template.spec.containers[0].resources.requests.memory
```
agullon: You may also want to extract `.spec.template.spec.containers[0].resources` into a var `${DNS_RESOURCE_PATH}    .spec.template.spec.containers[0].resources` and then use it `${DNS_RESOURCE_PATH}.requests.cpu`

test/suites/configuration1/dns-resource-configuration.robot
Neilhamza: Thanks for the detailed review @agullon , i committed this suggested change

test/suites/configuration1/dns-resource-configuration.robot
Neilhamza: fixed - now we remove any existing drop-in before applying the new one

test/suites/configuration/dns-resource-configuration.robot
Neilhamza: done created  Get DNS Resource Value keyword that wraps it

test/suites/configuration/dns-resource-configuration.robot
Neilhamza: fixed

test/suites/configuration/dns-resource-configuration.robot
```diff
@@ -0,0 +1,184 @@
...
+    [Documentation]    Get a resource value from the dns-default DaemonSet
+    [Arguments]    ${jsonpath}
+    ${value}=    Oc Get JsonPath    daemonset    openshift-dns    dns-default    ${jsonpath}
+    RETURN    ${value}
+
+Apply DNS Resource Config
+    [Documentation]    Remove any existing drop-in, apply a new DNS resource config and restart MicroShift
+    [Arguments]    ${config}
+    Remove Drop In MicroShift Config    ${DNS_DROPIN}
+    Drop In MicroShift Config    ${config}    ${DNS_DROPIN}
```
agullon: Inside `Drop In MicroShift Config` MicroShift config is being removed.
By only adding `Remove Drop In MicroShift Config` before `Drop In MicroShift Config` you are not changing any MicroShift config, only some files. This is because MicroShift config is only update it when microshift restarts. MicroShift must always restart after any config change.


```suggestion
    Remove DNS Resource Config
    Drop In MicroShift Config    ${config}    ${DNS_DROPIN}
```

test/suites/configuration/dns-resource-configuration.robot
```diff
@@ -0,0 +1,184 @@
...
+    Restart MicroShift
+
+Apply Invalid DNS Resource Config
+    [Documentation]    Apply an invalid DNS resource config that should prevent MicroShift from starting
+    [Arguments]    ${config}
+    Remove Drop In MicroShift Config    ${DNS_DROPIN}
```
agullon:
```suggestion
    Remove DNS Resource Config
```

test/suites/configuration/dns-resource-configuration.robot
```diff
@@ -0,0 +1,184 @@
...
+
+Remove DNS Resource Config
+    [Documentation]    Remove the DNS resource drop-in config and restart MicroShift
+    Remove Drop In MicroShift Config    ${DNS_DROPIN}
+    Restart MicroShift
+
+Restore Valid DNS Config
+    [Documentation]    Remove invalid config and restore MicroShift to a working state
+    Remove Drop In MicroShift Config    ${DNS_DROPIN}
+    Restart MicroShift
```
agullon: These 2 are the same, let's use only one.

Neilhamza: fixed

agullon:
it looks good to me from QE point of view
great work @Neilhamza !!!

pkg/config/dns.go
```diff
@@ -13,6 +15,20 @@ const (
...
+type DNSResources struct {
+	// Requests specifies the minimum resources required for the dns container.
+	// Valid keys are "cpu" and "memory". Values must be valid Kubernetes resource quantities.
+	// When not set, defaults to cpu=50m, memory=70Mi.
+	Requests map[string]string `json:"requests,omitempty"`
+
+	// Limits specifies the maximum resources the dns container can use.
+	// Valid keys are "cpu" and "memory". Values must be valid Kubernetes resource quantities.
+	// When not set, no limits are applied.
+	Limits map[string]string `json:"limits,omitempty"`
```
pacevedom, line 29: Is there any way to specify the defaults here? So that the config.yaml and the openapi spec have them too.

test/suites/configuration1/dns-resource-configuration.robot
```diff
@@ -0,0 +1,179 @@
...
+    Restart MicroShift
+
+Apply Invalid DNS Resource Config
+    [Documentation]    Apply an invalid DNS resource config that should prevent MicroShift from starting
+    [Arguments]    ${config}
+    Remove DNS Resource Config
+    Drop In MicroShift Config    ${config}    ${DNS_DROPIN}
+    ${cursor}=    Get Journal Cursor
+    VAR    ${CURSOR}=    ${cursor}    scope=TEST
+    Run Keyword And Expect Error    0 != 1    Restart MicroShift
```
pacevedom, line 174: Maybe use `Keyword Should Fail`?

test/suites/configuration1/dns-resource-configuration.robot
```diff
@@ -0,0 +1,179 @@
...
+
+Requests Only Without Limits
+    [Documentation]    Configure only requests without limits and verify no limits are injected
+    [Setup]    Apply DNS Resource Config    ${DNS_REQUESTS_ONLY}
+    ${cpu}=    Get DNS Resource Value    ${DNS_RESOURCE_PATH}.requests.cpu
+    ${memory}=    Get DNS Resource Value    ${DNS_RESOURCE_PATH}.requests.memory
+    ${limits}=    Get DNS Resource Value    ${DNS_RESOURCE_PATH}.limits
+    Should Be Equal As Strings    ${cpu}    100m
+    Should Be Equal As Strings    ${memory}    150Mi
+    Should Be Empty    ${limits}
```
pacevedom, line 96: Using limits as an empty map would bypass the check?

pkg/config/dns.go
```diff
@@ -90,14 +118,62 @@ func (t *DNS) validateHosts() error {
...
+			return fmt.Errorf("invalid dns resource limit %s=%q: %v", key, val, err)
+		}
+	}
+	for key, limitVal := range t.Resources.Limits {
+		reqVal, ok := t.Resources.Requests[key]
+		if !ok {
+			continue
+		}
+		limit := resource.MustParse(limitVal)
+		req := resource.MustParse(reqVal)
```
pacevedom, line 172: These could be constants?

pkg/config/config.go
```diff
@@ -442,6 +442,24 @@ func (c *Config) incorporateUserSettings(u *Config) {
 			c.DNS.Hosts.File = u.DNS.Hosts.File
 		}
 	}
+
+	// DNS resource configuration - merge key-by-key to preserve defaults
+	if u.DNS.Resources.Requests != nil {
+		if c.DNS.Resources.Requests == nil {
+			c.DNS.Resources.Requests = make(map[string]string)
+		}
```
pacevedom, line 452: This is correctly handled, but is there a test for this case?

pkg/config/dns.go
Neilhamza: i would need to modify  cmd/generate-config/parser.go to handle object defaults i'm not sure if that should be a scope of this PR - what do you think?

test/suites/configuration1/dns-resource-configuration.robot
Neilhamza: Run Keyword And Expect Error is the existing pattern used in tls-configuration.robot and configuration.robot. Keyword Should Fail isn't used anywhere in the test suite.  should i keep the existing pattern for consistency? or you would prefer to use Should Fail anyway?

pkg/config/dns.go
Neilhamza: The kubebuilder annotation +kubebuilder:default={"cpu":"50m","memory":"70Mi"} adds the default to the OpenAPI spec, but the config generator (cmd/generate-config/parser.go)  doesn't render map-type defaults

test/suites/configuration1/dns-resource-configuration.robot
Neilhamza: empty limits map (limits: {}) just means "no limits"  same as not setting it. The jsonpath returns empty, and "Should Be Empty" passes :dancers:

pkg/config/dns.go
Neilhamza: limitVal and reqVal arent' costant values they come from user config and change each loop iteration so MustParse is safe here

pkg/config/config.go
Neilhamza: Added dns-resources-limits-only test case

pkg/config/dns.go
pacevedom: Well , fillDefaults will do then. Thanks!

test/suites/configuration1/dns-resource-configuration.robot
pacevedom: Was asking because if the error changes this is brittle, but if thats the pattern lets address it in a different change.

test/suites/configuration1/dns-resource-configuration.robot
pacevedom: Thanks!

pkg/config/dns.go
pacevedom: Oh wait, I meant the constants in dnsMinimumRequests, but that might be too convoluted because of types.

pkg/config/config.go
pacevedom: Thanks

pkg/config/config_test.go
```diff
@@ -904,6 +979,113 @@ func TestValidate(t *testing.T) {
 			}(),
 			expectErr: true,
 		},
+		{
```
pacevedom, line 982: Missing tests with 0 and negative quantities.

test/suites/configuration1/dns-resource-configuration.robot
pacevedom, line 1: Keep in mind the directory is now `test/suites/configuration1`

pacevedom (commented):
/hold cancel

Minor nits, looking good overall!

pacevedom (approved):
/lgtm

PR #6622, opened 2026-05-04
USHIFT-6925 USHIFT-6851: Introduce Post-Quantum Curves to Ingress defaults and FIPs detection

eslutsky:
## Summary by CodeRabbit

* **New Features**
  * Post‑quantum curve (X25519MLKEM768) included in the default TLS curve list; router exposes selected TLS curves via a new environment variable.
  * Automatic FIPS detection toggles FIPS‑compliant TLS cipher and curve restrictions when enabled.

* **Tests**
  * Added integration test for ML‑KEM post‑quantum curve negotiation and unit tests validating FIPS cipher/curve behavior.

### DIFF
--- a/assets/components/openshift-router/deployment.yaml
+++ b/assets/components/openshift-router/deployment.yaml
@@ -58,6 +58,8 @@ spec:
               value: '{{ .ThreadCount }}'
             - name: SSL_MIN_VERSION
               value: '{{ .RouterSSLMinVersion }}'
+            - name: ROUTER_CURVES
+              value: '{{ .RouterTLSCurves }}'
             - name: ROUTER_USE_PROXY_PROTOCOL
               value: "false"
             - name: GRACEFUL_SHUTDOWN_DELAY
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -551,6 +551,14 @@ func generateIngressParams(cfg *config.Config) (assets.RenderParams, error) {
 		RouterCiphersSuites = strings.Join(tls13Ciphers, ":")
 	}
 
+	// Default TLS supportedGroups (curves) include X25519MLKEM768 for
+	// post-quantum readiness. In FIPS mode, ML-KEM and X25519 are not
+	// supported by OpenSSL FIPS 140-3.
+	tlsCurves := "X25519MLKEM768:X25519:P-256:P-384:P-521"
+	if isFIPSEnabled {
+		tlsCurves = "P-256:P-384:P-521"
+	}
+
 	var RouterSSLMinVersion string
 	switch tlsProfileSpec.MinTLSVersion {
 	// TLS 1.0 is not supported, convert to TLS 1.1.
@@ -642,6 +650,7 @@ func generateIngressParams(cfg *config.Config) (assets.RenderParams, error) {
 		"RouterCiphers":               RouterCiphers,
 		"RouterCiphersSuites":         RouterCiphersSuites,
 		"RouterSSLMinVersion":         RouterSSLMinVersion,
+		"RouterTLSCurves":             tlsCurves,
 		"RouterAllowWildcardRoutes":   RouterAllowWildcardRoutes,
 		"ClientCAMapName":             clientCAMapName,
 		"ClientAuthPolicy":            clientAuthPolicy,
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -27,9 +27,6 @@ const (
 	haproxyMaxTimeoutMilliseconds = 2147483647 * time.Millisecond
 )
 
-// isFIPSEnabled reports whether the cluster has FIPS enabled.
-var isFIPSEnabled = detectFIPS()
-
 var (
 	tlsVersion13Ciphers = sets.NewString(
 		"TLS_AES_128_GCM_SHA256",
@@ -266,7 +263,8 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		return err
 	}
 
-	extraParams, err := generateIngressParams(cfg, isFIPSEnabled)
+	fipsEnabled := detectFIPS()
+	extraParams, err := generateIngressParams(cfg, fipsEnabled)
 	if err != nil {
 		return err
 	}
--- a/pkg/components/controllers.go
+++ b/pkg/components/controllers.go
@@ -36,6 +36,10 @@ var (
 		"TLS_AES_128_CCM_8_SHA256",
 	)
 
+	// fipsApprovedTLS13Ciphers contains only GCM-based ciphers.
+	// CCM ciphers (TLS_AES_128_CCM_SHA256, TLS_AES_128_CCM_8_SHA256) are excluded
+	// because OpenSSL's FIPS provider does not enable them and HAProxy does not
+	// support CCM mode cipher suites.
 	fipsApprovedTLS13Ciphers = sets.NewString(
 		"TLS_AES_128_GCM_SHA256",
 		"TLS_AES_256_GCM_SHA384",
@@ -45,6 +49,8 @@ var (
 // detectFIPS reports whether the cluster is operating in FIPS
 // mode by checking the FIPS_ENABLED environment variable if set or
 // the /proc/sys/crypto/fips_enabled file otherwise.
+// FIPS_ENABLED is an optional override for testing environments where
+// /proc/sys/crypto/fips_enabled may not reflect the desired FIPS state.
 func detectFIPS() bool {
 	if v, ok := os.LookupEnv("FIPS_ENABLED"); ok {
 		if result, err := strconv.ParseBool(v); err != nil {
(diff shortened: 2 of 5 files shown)


test/suites/optional/tls-scanner.robot
```diff
@@ -137,3 +127,19 @@ Cleanup TLS Scanner Job
     IF    '${TLS_SCANNER_DIR}' != ''
         Run Keyword And Ignore Error    Remove Directory    ${TLS_SCANNER_DIR}    recursive=True
     END
+
+Verify ML-KEM Post Quantum Curve Negotiation
+    [Documentation]    Verify X25519MLKEM768 post-quantum hybrid key exchange
+    ...    negotiates successfully via oc exec into the router pod, which
+    ...    has OpenSSL 3.5+ (the host OpenSSL may be too old for ML-KEM).
+    ...    Skipped on FIPS clusters where ML-KEM is not configured.
```
pacevedom, line 149: How does it get skipped? Should we check the value for the ENV var in the deployment and skip it based on that? I see in the controller this situation is handled like that.

pkg/components/controllers.go
```diff
@@ -27,6 +27,9 @@ const (
 	haproxyMaxTimeoutMilliseconds = 2147483647 * time.Millisecond
 )
 
+// isFIPSEnabled reports whether the cluster has FIPS enabled.
+var isFIPSEnabled = detectFIPS()
```
pacevedom: Should we have this injectable in the function below (`generateConfigParams`) so that we can unit test it?

pkg/components/controllers.go
```diff
@@ -35,8 +38,41 @@ var (
...
+	data, err := os.ReadFile("/proc/sys/crypto/fips_enabled")
+	if err != nil {
+		klog.Warningf("Failed to read /proc/sys/crypto/fips_enabled: %v; assuming FIPS is not enabled", err)
+		return result
+	}
+	if len(data) == 0 {
+		klog.Warningf("Got empty /proc/sys/crypto/fips_enabled; assuming FIPS is not enabled")
+		return result
+	}
+	result = data[0] == '1'
```
pacevedom: Can we make this a bit more robust/clear with
```go
strings.TrimSpace(string(data)) == "1"
```

pacevedom (commented):
/retitle USHIFT-6925 USHIFT-6851: Introduce Post-Quantum Curves to Ingress defaults and FIPs detection

pkg/components/controllers.go
```diff
@@ -230,7 +266,7 @@ func startIngressController(ctx context.Context, cfg *config.Config, kubeconfigP
 		return err
 	}
 
-	extraParams, err := generateIngressParams(cfg)
+	extraParams, err := generateIngressParams(cfg, isFIPSEnabled)
```
pacevedom: Now wondering if we should just call the detectFIPS() here, when starting the actual ingress controller manifests. Having it in a variable at the top means that importing this package will be using I/O immediately without using a function.
Not a showstopper but a nit worth mentioning.

test/suites/optional/tls-scanner.robot
```diff
@@ -137,3 +141,23 @@ Cleanup TLS Scanner Job
...
+    Skip If    "X25519MLKEM768" not in """${curves}"""
+    ...    ROUTER_CURVES does not include X25519MLKEM768 (FIPS mode); skipping ML-KEM test
+    ${router_ip}=    Oc Get JsonPath    svc    openshift-ingress    router-default
+    ...    .spec.clusterIP
+    ${pod_name}=    Oc Get JsonPath    pod    openshift-ingress    ${EMPTY}
+    ...    .items[0].metadata.name
+    ${output}=    Oc Exec    ${pod_name}
+    ...    echo Q | openssl s_client -connect ${router_ip}:443 -groups X25519MLKEM768 2>&1 || true
+    ...    ns=openshift-ingress
+    Should Contain    ${output}    Negotiated TLS1.3 group: X25519MLKEM768
```
pacevedom: So if it doesnt exist, what is the message? Is there any way to make this a bit more loose on the string matching? In the end this is like a log line, which could change anytime?

test/suites/optional/tls-scanner.robot
```diff
@@ -51,6 +51,10 @@ TLS Scanner Host Scan Completes And Produces Artifacts
     ...    Cleanup TLS Scanner Job
     ...    Ensure Cluster Reader Role Deleted
 
+Ingress Router TLS Curves supports ML-KEM Post Quantum Curves
```
pacevedom, line 54: Now wondering if the router part fits in the router.robot file?

assets/components/openshift-router/deployment.yaml
```diff
@@ -58,6 +58,8 @@ spec:
               value: '{{ .ThreadCount }}'
             - name: SSL_MIN_VERSION
               value: '{{ .RouterSSLMinVersion }}'
+            - name: ROUTER_CURVES
+              value: '{{ .RouterTLSCurves }}'
```
pacevedom, line 62: This should be in the rebase.sh mutations, right?

pkg/components/controllers.go
```diff
@@ -495,12 +529,34 @@ func generateIngressParams(cfg *config.Config) (assets.RenderParams, error) {
 		}
 	}
 
+	// On FIPS-enabled clusters, remove non-FIPS-compliant TLS 1.3 cipher
+	// suites (e.g. TLS_CHACHA20_POLY1305_SHA256). HAProxy would fail TLS
+	// handshakes when a client offers a non-FIPS cipher first if that cipher
+	// is listed in ssl-default-bind-ciphersuites but excluded by the OS FIPS policy.
+	if fipsEnabled {
+		fipsCiphers := tls13Ciphers[:0]
```
pacevedom: It would be simpler and more clear to maybe allocate a new empty slice?

pkg/components/controllers.go
```diff
@@ -35,8 +35,41 @@ var (
 		"TLS_AES_128_CCM_SHA256",
 		"TLS_AES_128_CCM_8_SHA256",
 	)
+
+	fipsApprovedTLS13Ciphers = sets.NewString(
+		"TLS_AES_128_GCM_SHA256",
+		"TLS_AES_256_GCM_SHA384",
```
pacevedom, line 45: Are CCM algorithms not fips compliant? They are part of TLS 1.3 too.

pkg/components/controllers.go
```diff
@@ -35,8 +35,41 @@ var (
...
+	)
 )
 
+// detectFIPS reports whether the cluster is operating in FIPS
+// mode by checking the FIPS_ENABLED environment variable if set or
+// the /proc/sys/crypto/fips_enabled file otherwise.
+func detectFIPS() bool {
+	if v, ok := os.LookupEnv("FIPS_ENABLED"); ok {
```
pacevedom, line 55: Is this env var documented somewhere?

test/suites/optional/tls-scanner.robot
```diff
@@ -137,3 +141,23 @@ Cleanup TLS Scanner Job
...
         Run Keyword And Ignore Error    Remove Directory    ${TLS_SCANNER_DIR}    recursive=True
     END
+
+Verify ML-KEM Post Quantum Curve Negotiation
+    [Documentation]    Verify X25519MLKEM768 post-quantum hybrid key exchange
+    ...    negotiates successfully via oc exec into the router pod, which
+    ...    has OpenSSL 3.5+ (the host OpenSSL may be too old for ML-KEM).
+    ...    Skipped on FIPS clusters where ML-KEM is not configured.
+    ${curves}=    Oc Get JsonPath    deployment    openshift-ingress    router-default
+    ...    .spec.template.spec.containers[0].env[?(@.name=="ROUTER_CURVES")].value
```
pacevedom: Should we drop the container indexing? we do not own the deployment yaml and it may change.

pkg/components/controllers.go
eslutsky: those are the ciphers which used in the ingress controller https://github.com/openshift/cluster-ingress-operator/blob/7b7406ee0d4bf03de360d53c9cc4a83ee14332cc/pkg/operator/controller/ingress/controller.go#L91  so we need to be aligned with them.

test/suites/optional/tls-scanner.robot
eslutsky: well , I think it's better to have the compliance things in one common place.

pkg/components/controllers.go
eslutsky: Well, not really, it's for manual overrides and testing, currently its not documented.

pacevedom: I think we should document it then, as you can force fips ciphers without the system being strictly configured for fips.

pacevedom (approved):
/lgtm

Minor nit on documentation, but thats outside of this PR. We need to ensure the env var makes it into the docs.

PR #6638, opened 2026-05-08
USHIFT-6796: C2CC: DNS forwarding between clusters

pmtk:
## Summary by CodeRabbit

* **New Features**
  * Added configurable DNS cache TTL and negative TTL settings for cluster-to-cluster remote cluster DNS resolution, enabling fine-tuned control over DNS caching behavior for improved cross-cluster connectivity performance.

* **Documentation**
  * Updated configuration documentation examples to include new DNS cache settings for cluster-to-cluster connectivity configuration.


test/suites/c2cc/dns.robot
```diff
@@ -0,0 +1,123 @@
+*** Settings ***
```
agullon, line 1: @pmtk this robot file is missing from el10 scenarios, also I suggest to target all scenarios from c2cc/ directory and not robot files individually

test/suites/c2cc/dns.robot
```diff
@@ -0,0 +1,123 @@
...
+        ...    ${alias}
+        ...    oc wait pod/hello-microshift pod/curl-pod -n ${NAMESPACES}[${alias}] --for=condition=Ready --timeout=120s
+    END
+
+Cleanup DNS Test Workloads
+    [Documentation]    Delete test namespace on both clusters. Ignores errors.
+    FOR    ${alias}    IN    cluster-a    cluster-b
+        Run Keyword And Ignore Error
+        ...    Oc On Cluster    ${alias}    oc delete namespace ${NAMESPACES}[${alias}] --timeout=60s
+    END
```
agullon: These 3 Keywords are almost the same as the ones in connectivity.robot. I'd suggest to move all of them into `c2cc.resource`

test/suites/c2cc/dns.robot
pmtk: Cannot add dir without renaming the files to keep the order - what would you prefer?

test/suites/c2cc/dns.robot
agullon: the order is not really important, because it's mainly for human. The tests does not depend on each other, so it's ok.

test/suites/c2cc/dns.robot
agullon: I like current approach with numbers for order, thanks!

packaging/microshift/config.yaml
```diff
@@ -42,6 +42,14 @@ apiServer:
         # Defaults to VersionTLS12.
         minVersion: VersionTLS12
 clusterToCluster:
+    # DNS cache settings for CoreDNS server blocks generated for remote clusters.
+    dns:
```
agullon, line 46: I'd add these 2 TTL values are set to 10 seconds by default (when not defined). Should it be also add it in cmd/generate-config/config/config-openapi-spec.json ?

pkg/config/c2cc.go
```diff
@@ -11,7 +11,20 @@ import (
 	netutils "k8s.io/utils/net"
 )
 
+type C2CCDNS struct {
+	// Maximum TTL (seconds) for positive DNS cache entries in CoreDNS server blocks
+	// generated for remote clusters. Must be >= 0. Setting to 0 disables positive caching.
+	// +kubebuilder:validation:Minimum=0
+	CacheTTL *int `json:"cacheTTL,omitempty"`
```
agullon, line 19: do you think it's necessary to add a check and an error log message if the user set a non integer as a TTL value. Or is it enough with the type enforced in this struct?

agullon:
lgtm from QE

pkg/config/c2cc.go
pmtk: I think that might be too defensive: I think the parser will throw an error

pkg/config/c2cc.go
```diff
@@ -164,6 +180,13 @@ func (c *C2CC) validate(cfg *Config) error {
 		return fmt.Errorf("cluster to cluster requires OVN-Kubernetes CNI (network.cniPlugin must be \"\" or \"ovnk\", got %q)", cfg.Network.CNIPlugin)
 	}
 
+	if *c.DNS.CacheTTL < 0 {
+		return fmt.Errorf("dns.cacheTTL must be >= 0, got %d", *c.DNS.CacheTTL)
+	}
+	if *c.DNS.CacheNegativeTTL < 0 {
+		return fmt.Errorf("dns.cacheNegativeTTL must be >= 0, got %d", *c.DNS.CacheNegativeTTL)
+	}
```
pacevedom: These can panic in tests if not calling fillDefaults first, should we check for non-nil too?

pkg/config/c2cc.go
```diff
@@ -345,3 +377,33 @@ func checkCIDRConflicts(cidr *net.IPNet, cidrStr, label string, seenCIDRs []labe
...
+		blocks = append(blocks, formatDNSBlock(rc.Domain, rc.DNSIP, cacheTTL, cacheNegativeTTL))
+	}
+	if len(blocks) == 0 {
+		return ""
+	}
+	return "\n" + strings.Join(blocks, "\n")
+}
+
+func formatDNSBlock(domain, dnsIP string, cacheTTL, cacheNegativeTTL int) string {
+	return fmt.Sprintf(`    %s:5353 {
```
pacevedom, line 400: Should we make this a constant at the top of the file?

test/resources/c2cc.resource
```diff
@@ -222,3 +222,54 @@ Remove Foreign Subnet From SNAT Annotation
...
+    FOR    ${alias}    IN    cluster-a    cluster-b
+        Oc On Cluster
+        ...    ${alias}
+        ...    oc wait pod/hello-microshift pod/curl-pod -n ${NAMESPACES}[${alias}] --for=condition=Ready --timeout=120s
+    END
+
+Cleanup Test Workloads
+    [Documentation]    Delete test namespace on both clusters. Ignores errors.
+    FOR    ${alias}    IN    cluster-a    cluster-b
+        Run Keyword And Ignore Error
```
pacevedom: Should we ignore the error? Do you foresee problems with finalizers or timeouts hitting?

assets/components/openshift-dns/dns/configmap.yaml
```diff
@@ -1,6 +1,7 @@
 apiVersion: v1
 data:
   Corefile: |
+    {{- .C2CCDNSBlocks }}
```
pacevedom, line 4: Is it possible to mask out the next server block by setting up a special domain? 
Say you configure `cluster.local` as domain. That would render
```yaml
cluster.local:5353 {
...
}
.:5353 {
...
}
```
Which is more specific than the catch all, and would route all the `cluster.local` requests to remote cluster.
This is a bit of a stretch because no real use case configures a remote cluster with the same domain than the local one, but should we validate it?

pkg/config/c2cc.go
pmtk: It's only used in single place so I'd rather not move it to a place almost 400 lines away.

test/resources/c2cc.resource
pmtk: I don't recall reason for this, I'll drop ignoring errors

assets/components/openshift-dns/dns/configmap.yaml
pmtk: True, good catch

pacevedom (approved):
/lgtm

PR #6657, opened 2026-05-12
USHIFT-6978: remove initramfs FIPS module check from bootc validation

agullon:
## Summary
- Remove the `lsinitrd -m | grep fips` check from `Fips Should Be Enabled Bootc` in `validate-fips.robot`
- This check tests a RHEL packaging detail (whether `dracut-fips` is installed), not actual FIPS enablement
- The remaining two checks — `/proc/sys/crypto/fips_enabled` and `update-crypto-policies --show` — are sufficient to verify FIPS mode is active
- On RHEL 9, the `fips` dracut module lives in a separate `dracut-fips` package not present in bootc images; on RHEL 10+ it's in base `dracut`

## Test plan
- [x] Verify `el98-lrel@ai-model-serving-online-fips` scenario passes on RHEL 9 bootc
- [x] Verify FIPS test still passes on RHEL 10 bootc

Jira: https://issues.redhat.com/browse/USHIFT-6978

### DIFF
--- a/test/suites/fips/validate-fips.robot
+++ b/test/suites/fips/validate-fips.robot
@@ -132,12 +132,6 @@ Fips Should Be Enabled Bootc
     Should Be Equal As Integers    0    ${rc}
     Should Be Equal As Strings    ${stdout.strip()}    FIPS
 
-    # Verify initramfs FIPS module presence
-    ${stdout}    ${stderr}    ${rc}=    Execute Command
-    ...    bash -c 'lsinitrd -m 2>/dev/null | grep -Fxq fips'
-    ...    sudo=False    return_rc=True    return_stdout=True    return_stderr=True
-    Should Be Equal As Integers    0    ${rc}
-
 Get Images From Release File
     [Documentation]    Obtains list of Images from Release.
     ${stdout}    ${stderr}    ${rc}=    Execute Command


test/suites/fips/validate-fips.robot
```diff
@@ -133,10 +133,15 @@ Fips Should Be Enabled Bootc
...
-    Should Be Equal As Integers    0    ${rc}
+    # On RHEL 9, the fips dracut module is in the separate dracut-fips package
+    # which is not installed in bootc images, so this check only applies to RHEL 10+.
+    ${os_version}=    Get Host OS Version
+    IF    not "${os_version}".startswith("9")
+        ${stdout}    ${stderr}    ${rc}=    Execute Command
+        ...    bash -c 'lsinitrd -m 2>/dev/null | grep -Fxq fips'
+        ...    sudo=False    return_rc=True    return_stdout=True    return_stderr=True
+        Should Be Equal As Integers    0    ${rc}
+    END
```
ggiguash: Can we install the missing package instead of disabling the test?

agullon: yes, we can, it's another possible approach, I thought about it
I prefer to keep the test prerequisites as small and simple as possible. I don't see a benefict of adding a new package for a small check. Also, this new package is not needed and tested for RHEL10.

Also, whit this extra if in the test we have documented in the test code the differences between RHEL9 and RHEL10 which may be useful as extra info in the future.

ggiguash: We should either remove the test completely, or make it work on both operating systems.
Adding OS-specific conditional code should only be a last resort.

agullon: I'll remove this check, I tried these two ideas ([add an RPM](https://github.com/openshift/microshift/pull/6662/changes/0b6d253dba2248c66206c16319736636f6a52309) and [add a fips config](https://github.com/openshift/microshift/pull/6662/changes/2a8820d550b38b1cc9f0e6ba0f2b6ab0c0f137d6)) ideas but they didn't work.

agullon: @ggiguash all changes done, please review the PR again

PR #6658, opened 2026-05-12
USHIFT-6979: Add PCP metrics collection to test scenarios

pacevedom:
Install and run Performance Co-Pilot on all online VMs during test execution, then collect the archives as artifacts alongside SOS reports. Controlled via SKIP_PCP environment variable (defaults to false).

## Summary by CodeRabbit

* **New Features**
  * Test images now include PCP packages and enable PCP services so performance collection runs inside test agents.

* **Tests**
  * Scenarios collect PCP data by default (can be disabled via configuration); collection works for both online and offline VMs and archives are gathered on test exit, even on failures.

* **Chores**
  * CI image build invocation updated to use an explicit build flag.

### DIFF
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -298,7 +298,7 @@ collect_pcp_reports_for_vm() {
     local -r vmname="${2}"
 
     echo "Collecting PCP data from ${vmname}"
-    run_command_on_vm "${vmname}" "sudo systemctl stop pmlogger" || true
+    run_command_on_vm "${vmname}" "sudo systemctl stop --now pmlogger" || true
 
     if ! run_command_on_vm "${vmname}" "test -d /var/log/pcp/pmlogger" ; then
         echo "WARNING: No PCP data directory on ${vmname}, skipping collection"
@@ -328,7 +328,7 @@ collect_pcp_reports_for_vm_offline() {
 
     invoke_qemu_script "bash" \
         "--vm" "${full_vmname}" \
-        "--args" "sudo systemctl stop pmlogger"
+        "--args" "sudo systemctl stop --now pmlogger"
 
     invoke_qemu_script "bash" \
         "--vm" "${full_vmname}" \
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -293,32 +293,54 @@ sos_report_for_vm_offline() {
         "--filename" "*.log"
 }
 
-start_pcp_on_all_vms() {
-    if "${SKIP_PCP}"; then
-        echo "Skipping PCP collection"
+collect_pcp_reports_for_vm() {
+    local -r vmdir="${1}"
+    local -r vmname="${2}"
+
+    echo "Collecting PCP data from ${vmname}"
+    run_command_on_vm "${vmname}" "sudo systemctl stop pmlogger" || true
+
+    if ! run_command_on_vm "${vmname}" "test -d /var/log/pcp/pmlogger" ; then
+        echo "WARNING: No PCP data directory on ${vmname}, skipping collection"
         return 0
     fi
 
-    for vmdir in "${SCENARIO_INFO_DIR}"/"${SCENARIO}"/vms/*; do
-        if [ ! -d "${vmdir}" ]; then
-            continue
-        fi
+    run_command_on_vm "${vmname}" \
+        "sudo tar czf /tmp/pcp-archives.tar.gz -C /var/log/pcp/pmlogger ." || true
 
-        local vmname
-        vmname=$(basename "${vmdir}")
-        local ip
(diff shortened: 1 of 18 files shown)


test/bin/scenario.sh
```diff
@@ -292,6 +293,73 @@ sos_report_for_vm_offline() {
         "--filename" "*.log"
 }
 
+start_pcp_on_all_vms() {
```
ggiguash: I think we should enable services in the containerfile / toml as we install them.

pacevedom: Will do before publishing final version. Let me leave this comment open to remember. Thanks!

pacevedom: Ready

test/bin/scenario.sh
```diff
@@ -1526,13 +1606,14 @@ action_create() {
...
-    # Set the exit handler to attempt the sos report collection and error logging
+    # Set the exit handler to attempt PCP and SOS report collection and error logging
     # - Preserve the original exit code
     # - Log junit message on failure
     # - Override the exit code if sos report collection fails
     # shellcheck disable=SC2154
     trap 'rc=$? ; \
         [ "${rc}" -ne 0 ] && record_junit "setup" "scenario_create_vms" "FAILED" ; \
+        collect_pcp_reports || true ; \
         sos_report true || rc=1 ; \
```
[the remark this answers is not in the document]
pacevedom: Thats ok, this is intended to run only in CI

test/bin/scenario.sh
```diff
@@ -22,6 +22,7 @@ PULL_SECRET_CONTENT="$(jq -c . "${PULL_SECRET}")"
 VM_BOOT_TIMEOUT=1200 # Overall total boot times are around 15m
 VM_GREENBOOT_TIMEOUT=1800 # Greenboot readiness may take up to 15-30m depending on the load
 SKIP_SOS=${SKIP_SOS:-false}  # may be overridden in global settings file
+SKIP_PCP=${SKIP_PCP:-false}  # may be overridden in global settings file
```
ggiguash, line 25: This should probably also go to scenario_settings.sh.example?

test/bin/scenario.sh
```diff
@@ -292,6 +293,85 @@ sos_report_for_vm_offline() {
         "--filename" "*.log"
 }
 
+collect_pcp_reports_for_vm() {
+    local -r vmdir="${1}"
+    local -r vmname="${2}"
+
+    echo "Collecting PCP data from ${vmname}"
+    run_command_on_vm "${vmname}" "sudo systemctl stop pmlogger" || true
```
ggiguash: We should probably make this call synchronous by using `--now`?

test/bin/scenario.sh
```diff
@@ -292,6 +293,85 @@ sos_report_for_vm_offline() {
...
+    echo "Collecting PCP data offline from ${vmname}"
+
+    "${ROOTDIR}/scripts/fetch_tools.sh" "robotframework"
+
+    invoke_qemu_script "wait" \
+        "--vm" "${full_vmname}"
+
+    invoke_qemu_script "bash" \
+        "--vm" "${full_vmname}" \
+        "--args" "sudo systemctl stop pmlogger"
```
ggiguash: Add `--now`?

test/bin/scenario.sh
```diff
@@ -292,6 +293,85 @@ sos_report_for_vm_offline() {
...
+}
+
+collect_pcp_reports_for_vm_offline() {
+    local -r vmdir="${1}"
+    local -r vmname="${2}"
+    local -r full_vmname="$(full_vm_name "${vmname}")"
+
+    echo "Collecting PCP data offline from ${vmname}"
+
+    "${ROOTDIR}/scripts/fetch_tools.sh" "robotframework"
```
ggiguash, line 324: Why do we need it here?

ggiguash (changes requested):

test/bin/scenario.sh
pacevedom: Done

test/bin/scenario.sh
pacevedom: Done

test/bin/scenario.sh
pacevedom: Done

test/bin/scenario.sh
pacevedom: This is mirroring the [sos_report_for_vm_offline](https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L236) pattern. [invoke_qemu_script](https://github.com/openshift/microshift/blob/main/test/bin/scenario.sh#L228) runs python from the robot framework venv.

pacevedom:
/override ci/prow/e2e-aws-tests-bootc-periodic-el10

Unrelated failure, being addressed in a [different PR](https://github.com/openshift/microshift/pull/6667).

PR #6698, opened 2026-05-19
USHIFT-6986: Add retry and error handling to cfssl downloads in CNCF …

pacevedom:
…conformance

The cfssl binary downloads in configure-pri.sh intermittently fail due to transient network issues, causing silent checksum mismatches because curl saves error responses as the output file.

Add --fail to reject HTTP errors, --retry with --retry-all-errors for resilience, and -S to surface errors in logs for debugging.

### DIFF
--- a/scripts/multinode/configure-pri.sh
+++ b/scripts/multinode/configure-pri.sh
@@ -112,8 +112,10 @@ function generate_service_certs() {
             exit 1
     esac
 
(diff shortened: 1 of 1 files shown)


pacevedom:
/override ci/prow/e2e-aws-tests-bootc
/override ci/prow/e2e-aws-tests-bootc-arm

The changes in this PR only impact cncf conformance tests, which are not present in the failures seen above, therefore its an unrelated error.

PR #6727, opened 2026-05-25
USHIFT-6798: C2CC: RemoteClusters.microshift.io/v1alpha1 Custom Resource

pmtk:
## Summary by CodeRabbit

* **New Features**
  * Added a RemoteCluster CRD and a controller to manage per-remote-cluster healthcheck resources.
  * Introduced clusterToCluster.dns.probeInterval setting (default 10s, range 1s–5m).

* **Tests**
  * Added unit and integration tests covering probe interval validation, RemoteCluster reconciliation, and end-to-end CR lifecycle.

* **Documentation**
  * Updated configuration docs and default config examples to include the new probeInterval setting.

* **Chores**
  * Added tooling and a make target to generate CRDs and client code.


test/suites/c2cc/07-healthcheck.robot
```diff
@@ -0,0 +1,107 @@
...
+Verify RemoteCluster CR Spec
+    [Documentation]    Verify the RemoteCluster CR spec has the correct probe target
+    ...    (11th IP in the remote service CIDR on port 8080)
+    ...    and a non-empty probe interval duration string.
+    [Arguments]    ${alias}    ${remote_svc_cidr}
+    ${expected_ip}=    Compute 11th IP    ${remote_svc_cidr}
+    ${stdout}=    Oc On Cluster
+    ...    ${alias}
+    ...    oc get remoteclusters.microshift.io -l app.kubernetes.io/managed-by=c2cc-route-manager -o jsonpath='{.items[0].spec.probeTarget}'
+    Should Be Equal As Strings    ${stdout}    ${expected_ip}:8080    strip_spaces=True
```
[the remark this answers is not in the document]
pmtk: Applied — IPv6 fix for probe target assertion.

cmd/generate-config/config/config-openapi-spec.json
```diff
@@ -203,6 +203,10 @@
             }
           }
         },
+        "probeInterval": {
+          "description": "Interval between healthcheck probe attempts to each remote cluster.\nParsed as a Go duration string (e.g. \"10s\", \"1m\"). Must be between 1s and 5m.",
+          "type": "string"
+        },
```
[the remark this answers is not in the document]
pmtk: Won't fix — no other duration field in this spec uses `pattern`. Runtime Go validation already enforces 1s–5m bounds; adding OpenAPI regex adds no real safety.

pkg/controllers/c2cc/healthcheck.go
```diff
@@ -0,0 +1,107 @@
...
+	for i := range existing.Items {
+		existingByName[existing.Items[i].Name] = &existing.Items[i]
+	}
+
+	for name, want := range desired {
+		got, ok := existingByName[name]
+		if !ok {
+			if _, err := h.client.RemoteClusters().Create(ctx, want, metav1.CreateOptions{}); err != nil {
+				return fmt.Errorf("creating RemoteCluster %q: %w", name, err)
+			}
```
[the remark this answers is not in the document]
pmtk: Won't fix — the managed-by label is set at creation time in `buildDesiredCRs()`. This CRD is brand new, so no label-less versions exist. The only path to AlreadyExists is manual operator error with a name matching the `c2cc-<ip>` convention, which should surface as a hard failure rather than be silently adopted.

test/suites/c2cc/07-healthcheck.robot
```diff
@@ -0,0 +1,112 @@
...
+        ${expected_target}=    Set Variable    [${expected_ip}]:8080
+    ELSE
+        ${expected_target}=    Set Variable    ${expected_ip}:8080
+    END
+    Should Be Equal As Strings    ${stdout}    ${expected_target}    strip_spaces=True
+    ${interval}=    Oc On Cluster
+    ...    ${alias}
+    ...    oc get remoteclusters.microshift.io -l app.kubernetes.io/managed-by=c2cc-route-manager -o jsonpath='{.items[0].spec.probeInterval}'
+    Should Not Be Empty    ${interval}
+    Should Match Regexp    ${interval}    ^[0-9]+(s|m|h)
```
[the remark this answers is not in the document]
pmtk: done

test/suites/c2cc/healthcheck.robot
```diff
@@ -0,0 +1,112 @@
...
+        ${expected_target}=    Set Variable    [${expected_ip}]:8080
+    ELSE
+        ${expected_target}=    Set Variable    ${expected_ip}:8080
+    END
+    Should Be Equal As Strings    ${stdout}    ${expected_target}    strip_spaces=True
+    ${interval}=    Oc On Cluster
+    ...    ${alias}
+    ...    oc get remoteclusters.microshift.io -l app.kubernetes.io/managed-by=c2cc-route-manager -o jsonpath='{.items[0].spec.probeInterval}'
+    Should Not Be Empty    ${interval}
+    Should Match Regexp    ${interval}    ^[0-9]+(s|m|h)$
```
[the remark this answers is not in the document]
pmtk: In https://github.com/openshift/microshift/pull/6727#discussion_r3297309656 you recommended to add the trailing `$`, now you want to remove it. Which one is it?

test/suites/c2cc/healthcheck.robot
```diff
@@ -0,0 +1,112 @@
...
+    ${expected_ip}=    Compute 11th IP    ${remote_svc_cidr}
+    ${stdout}=    Oc On Cluster
+    ...    ${alias}
+    ...    oc get remoteclusters.microshift.io -l app.kubernetes.io/managed-by=c2cc-route-manager -o jsonpath='{.items[0].spec.probeTarget}'
+    IF    ":" in """${expected_ip}"""
+        VAR    ${expected_target}=    [${expected_ip}]:8080
+    ELSE
+        VAR    ${expected_target}=    ${expected_ip}:8080
+    END
+    Should Be Equal As Strings    ${stdout}    ${expected_target}    strip_spaces=True
```
agullon: Instead of the If block and check for `${expected_target}` have you thought about using `Should Match Regexp`?
I'm fine with current approach

pmtk: Let's keep it - in case of error, this should be more readable

pkg/controllers/c2cc/controller.go
```diff
@@ -239,6 +258,9 @@ func (c *C2CCRouteManager) cleanupAll(ctx context.Context) {
 	if c.nftMgr != nil {
 		cleanups = append(cleanups, cleanable{"nftables", c.nftMgr.cleanup})
 	}
+	cleanups = append(cleanups, cleanable{"healthcheck-crd", func(ctx context.Context) error {
+		return assets.DeleteCRDs(ctx, healthcheckCRD, c.kubeconfig)
+	}})
```
pacevedom, line 263: Do we need a guard too for this one to keep the pattern consistent?

pkg/controllers/c2cc/controller.go
```diff
@@ -16,17 +18,21 @@ const (
...
 	kubeconfig string
 
-	kubeClient kubernetes.Interface
-	ovn        *ovnRouteManager
-	annotation *annotationManager
-	nftMgr     *nftablesManager
-	routes     *linuxRouteManager
-	svcRoutes  *serviceRouteManager
+	kubeClient       kubernetes.Interface
+	microshiftClient microshiftclient.MicroshiftV1alpha1Interface
```
pacevedom, line 29: Do we have the RBAC for this, according to the enhancement? `Write access restricted to the C2C controller's SA; read access to system:authenticated via a ClusterRole.`
This is using the admin kubeconfig so full access, but what about the probe pod? Or is this for a later PR?

pmtk: Probe pod (and its assets) will be in the next PR. Sneak peak: https://github.com/openshift/microshift/pull/6729/changes#diff-0b79572d55608d3cd7d777b94d19719df88a6d4cf61c5464b9906cb444dcd281

pkg/controllers/c2cc/controller.go
pmtk: But there's nothing to guard it with: assets.DeleteCRDs creates its own kube client

pkg/controllers/c2cc/controller.go
pacevedom: Just noticed that PR, disregard the comment then.

pmtk:
/hold
Need to rebase onto main and align with 3 clusters testing

pkg/controllers/c2cc/healthcheck.go
```diff
@@ -0,0 +1,107 @@
...
+	existingByName := make(map[string]*microshiftv1alpha1.RemoteCluster, len(existing.Items))
+	for i := range existing.Items {
+		existingByName[existing.Items[i].Name] = &existing.Items[i]
+	}
+
+	for name, want := range desired {
+		got, ok := existingByName[name]
+		if !ok {
+			if _, err := h.client.RemoteClusters().Create(ctx, want, metav1.CreateOptions{}); err != nil {
+				return fmt.Errorf("creating RemoteCluster %q: %w", name, err)
```
pacevedom: Minor, but if there is more than one remote and the first one fails, the subsequent ones have to wait. Should we use `errors.Join`? This also applies to the Update call below.

pmtk: We already have tests with 3 clusters, so let's address it straightaway

pacevedom (approved):
/lgtm

PR #6729, opened 2026-05-25
USHIFT-6797: C2CC Probe remote clusters

pmtk:
## Summary by CodeRabbit

* **New Features**
  * Cluster-to-cluster health-check system with a probe deployment and per-remote health CRs reporting Healthy/Unhealthy, last probe time, and errors.
* **Configuration**
  * New probeInterval setting (1s–5m, default 10s) to control probe frequency; probe service ClusterIP deterministically computed.
* **CLI**
  * Added hidden probe subcommand to run the probe server.
* **Installation / Runtime**
  * Manifests for namespace, RBAC, service account, service, and probe deployment included; CRD for RemoteCluster added.
* **Tooling**
  * Make target and script to generate CRD/artifacts.
* **Tests & QA**
  * End-to-end and unit tests validating probe behavior, CR lifecycle, and cluster IP computation.


assets/components/c2cc/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+        hostPath:
+          path: /usr/bin/microshift
+          type: File
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      priorityClassName: system-cluster-critical
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        operator: Exists
+        effect: NoSchedule
```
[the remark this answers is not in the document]
pmtk: Won't fix — MicroShift is single-node. Every other deployment (service-ca, OVN, CSI, DNS) uses the identical `nodeSelector: node-role.kubernetes.io/master` + NoSchedule toleration pattern. This is the standard convention, not a topology concern.

pkg/controllers/c2cc/probe.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+			if rc != nil {
+				pm.stopProbe(rc.Name)
+			}
+		},
+	})
+
+	factory.Start(ctx.Done())
+	factory.WaitForCacheSync(ctx.Done())
+	klog.Infof("Probe manager running, watching RemoteCluster CRs")
```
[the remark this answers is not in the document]
pmtk: Won't fix — `WaitForCacheSync(ctx.Done())` blocks until sync or context cancellation. If ctx is cancelled, the next line `<-ctx.Done()` exits immediately anyway — both paths converge on the same shutdown behavior.

pmtk:
Depends on https://github.com/openshift/microshift/pull/6727

pkg/controllers/c2cc/probe.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create microshift client: %w", err)
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprint(w, "ok")
+	})
+	server := &http.Server{Addr: ":8080", Handler: mux}
```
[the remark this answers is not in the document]
pmtk: Applied — fixed in d8d5e256a.

pkg/controllers/c2cc/probe.go
```diff
@@ -0,0 +1,228 @@
...
+			if !ok {
+				if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+					rc, _ = tombstone.Obj.(*microshiftv1alpha1.RemoteCluster)
+				}
+			}
+			if rc != nil {
+				pm.stopProbe(rc.Name)
+			}
+		},
+	})
```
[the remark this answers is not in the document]
pmtk: Applied — fixed in d8d5e256a.

pkg/controllers/c2cc/healthcheck.go
```diff
@@ -0,0 +1,107 @@
...
+
+		delete(existingByName, name)
+
+		if got.Spec.ProbeTarget == want.Spec.ProbeTarget && got.Spec.ProbeInterval == want.Spec.ProbeInterval {
+			continue
+		}
+
+		got.Spec = want.Spec
+		if _, err := h.client.RemoteClusters().Update(ctx, got, metav1.UpdateOptions{}); err != nil {
+			return fmt.Errorf("updating RemoteCluster %q: %w", name, err)
```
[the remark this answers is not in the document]
pmtk: Won't fix — reconcile loop runs every 2s and self-heals on the next cycle. Adding retry complexity for a transient 409 that resolves itself in one tick isn't justified.

pkg/controllers/c2cc/probe.go
```diff
@@ -0,0 +1,228 @@
...
+
+			if probeErr != nil {
+				consecutiveFailures++
+				klog.V(2).Infof("Probe %q failed (%d consecutive): %v", name, consecutiveFailures, probeErr)
+
+				if consecutiveFailures >= unhealthyThreshold {
+					status.State = "Unhealthy"
+				} else {
+					status.State = "Healthy"
+				}
```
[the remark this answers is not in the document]
pmtk: Won't fix — intentional hysteresis to prevent flapping. A single transient failure shouldn't mark a cluster unhealthy.

assets/components/c2cc/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+        ports:
+        - containerPort: 8080
+          name: probe
+          protocol: TCP
+        livenessProbe:
+          httpGet:
+            path: /
+            port: 8080
+          initialDelaySeconds: 10
+          periodSeconds: 10
```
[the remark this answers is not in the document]
pmtk: Won't fix — HTTP server starts listening synchronously before probe goroutines. Startup race is not realistic for an internal probe pod.

test/resources/c2cc.resource
```diff
@@ -283,6 +283,15 @@ Service Endpoints Should Exist
...
 
+Compute 11th IP
+    [Documentation]    Return the 11th host address in a CIDR (e.g. 10.43.0.0/16 -> 10.43.0.11).
+    [Arguments]    ${cidr}
+    VAR    ${cmd}=    import ipaddress; n=ipaddress.ip_network('${cidr}', strict=False); print(n[11])
+    ${result}=    Process.Run Process    python3    -c    ${cmd}
+    Should Be Equal As Integers    ${result.rc}    0
+    ${ip}=    Strip String    ${result.stdout}
+    RETURN    ${ip}
+
```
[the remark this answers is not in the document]
pmtk: Won't fix — service CIDRs are always /16 or /112 per config validation. A CIDR too small for 12 addresses is an impossible scenario in MicroShift.

assets/crd/microshift.io_remoteclusters.yaml
```diff
@@ -0,0 +1,85 @@
...
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              probeInterval:
+                default: 10s
+                description: Interval between probe attempts (e.g. "10s", "1m").
+                type: string
+              probeTarget:
```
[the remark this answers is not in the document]
pmtk: Won't fix — CRD is controller-managed, never user-edited. Validation is already enforced in the Go config layer.

Makefile
```diff
@@ -346,6 +346,11 @@ generate-config:
 verify-config: generate-config
 	./scripts/verify/verify-config.sh
 
+.PHONY: generate-crds
+generate-crds:
+	./scripts/fetch_tools.sh controller-gen && \
+	./scripts/generate-crds.sh
```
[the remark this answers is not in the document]
pmtk: Won't fix — build system scope creep, out of scope for this PR.

pkg/controllers/c2cc/deploy_probe.go
```diff
@@ -0,0 +1,73 @@
...
+	c2ccClusterRoleBinding = []string{"components/c2cc/clusterrolebinding.yaml"}
+	c2ccDeployment         = []string{"components/c2cc/deployment.yaml"}
+	c2ccService            = []string{"components/c2cc/service.yaml"}
+)
+
+func (c *C2CCRouteManager) deployProbe(ctx context.Context) error {
+	_, svcNet, err := net.ParseCIDR(c.cfg.Network.ServiceNetwork[0])
+	if err != nil {
+		return fmt.Errorf("failed to parse local service network: %w", err)
+	}
```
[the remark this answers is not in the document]
pmtk: Won't fix — ServiceNetwork is guaranteed non-empty after fillDefaults() and validation. No other caller in the codebase guards this.

pmtk:
Addressed CodeRabbit review-body findings:

- `deploy_probe.go:59` log verbosity nitpick: Applied — lowered to `klog.V(4)`, fixed in d8d5e256a.
- `probe.robot:123-130` extract `Compute 11th IP` to shared resource: Won't fix — this PR already does exactly that (commit e9c9fee1e moved it to `c2cc.resource`).

pkg/controllers/c2cc/probe.go
```diff
@@ -0,0 +1,244 @@
...
+			klog.Errorf("Failed to close probe response body: %v", err)
+		}
+	}()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed with unexpected status %d", resp.StatusCode)
+	}
+	return nil
+}
+
+func (pm *probeManager) updateStatus(ctx context.Context, name string, status microshiftv1alpha1.RemoteClusterStatus) error {
```
pacevedom, line 223: I know it should be impossible, but if there are 2 concurrent operations on the CR we might get a conflict failure? Should we enclose this with `retry.RetryOnConflict`?

pkg/controllers/c2cc/probe.go
```diff
@@ -0,0 +1,244 @@
...
+			if rc != nil {
+				pm.stopProbe(rc.Name)
+			}
+		},
+	}); err != nil {
+		return fmt.Errorf("failed to add RemoteCluster informer handlers: %w", err)
+	}
+
+	factory.Start(ctx.Done())
+	factory.WaitForCacheSync(ctx.Done())
```
pacevedom, line 98: Should we check the returned value here? If the cache sync fails we run the probe against potentially stale data.

pkg/controllers/c2cc/probe.go
```diff
@@ -0,0 +1,244 @@
...
+	go pm.runProbeLoop(probeCtx, rc.Name, rc.Spec.ProbeTarget, rc.Spec.ProbeInterval.Duration)
+}
+
+func (pm *probeManager) restartProbe(ctx context.Context, rc *microshiftv1alpha1.RemoteCluster) {
+	pm.mu.Lock()
+	if cancel, exists := pm.probes[rc.Name]; exists {
+		cancel()
+		delete(pm.probes, rc.Name)
+	}
+	pm.mu.Unlock()
```
agullon: should we replace this block with `stopProbe()`?

test/suites/c2cc/probe.robot
```diff
@@ -0,0 +1,125 @@
...
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+Test Tags           c2cc
+
+
+*** Variables ***
+${C2CC_NAMESPACE}       openshift-c2cc
+${PROBE_DEPLOYMENT}     c2cc-probe
+${PROBE_PORT}           8080
```
agullon: It looks this var is never used

test/suites/c2cc/probe.robot
```diff
@@ -0,0 +1,125 @@
...
+    ...    Verify RemoteCluster State    cluster-a    Healthy
+    Wait Until Keyword Succeeds    3m    10s
+    ...    Verify RemoteCluster State    cluster-b    Healthy
+    Wait Until Keyword Succeeds    3m    10s
+    ...    Verify RemoteCluster State    cluster-c    Healthy
+
+RemoteCluster Status Has LastProbeTime
+    [Documentation]    Verify that LastProbeTime is populated after probing starts.
+    FOR    ${alias}    IN    cluster-a    cluster-b    cluster-c
+        ${stdout}=    Oc On Cluster    ${alias}
```
agullon, line 56: **Suggestion: add a negative test for the Unhealthy transition**

The suite validates the happy path well (all clusters become `Healthy`), but there is no test for the `Unhealthy` state transition, which is the probe's core behavioral contract — the 3-consecutive-failure threshold defined in `probe.go:unhealthyThreshold`.

Without it, a regression that always reports `Healthy` regardless of probe results would pass all tests. The sibling `reconciliation.robot` suite already follows this fault-injection pattern extensively.

Note: scaling down the probe deployment on the remote cluster won't work here because the C2CC controller's `deployProbe()` self-heals it on every reconciliation loop. Instead, use a firewall rule to block probe traffic — the controller doesn't manage firewall rules on the probe port:

```robot
RemoteCluster Status Becomes Unhealthy When Probe Fails
    [Documentation]    Block probe traffic on cluster-b and verify cluster-a
    ...    reports Unhealthy for the corresponding RemoteCluster CR.
    ...    Also checks that the probe pod logs the failure and the
    ...    Errors field is populated in the CR status.
    # Block probe port on cluster-b so cluster-a's probes fail
    Command On Cluster    cluster-b
    ...    nft add rule inet filter input tcp dport 8080 drop
    # Wait for cluster-a to report Unhealthy (requires 3 consecutive failures)
    Wait Until Keyword Succeeds    3m    10s
    ...    Verify RemoteCluster State For Peer    cluster-a    ${CLUSTER_B_SVC_CIDR}    Unhealthy
    # Verify the Errors field is populated in the CR status
    ${errors}=    Oc On Cluster    cluster-a
    ...    oc get remoteclusters.microshift.io -o jsonpath='{.items[0].status.errors}'
    Should Not Be Empty    ${errors}
    # Verify the probe pod logged the consecutive failure
    ${logs}=    Oc On Cluster    cluster-a
    ...    oc logs deployment/${PROBE_DEPLOYMENT} -n ${C2CC_NAMESPACE} --tail=50
    Should Contain    ${logs}    failed
    Should Contain    ${logs}    consecutive
    [Teardown]    Command On Cluster    cluster-b
    ...    nft delete rule inet filter input tcp dport 8080 drop
```

The probe logs failures via `klog.V(2).Infof("Probe %q failed (%d consecutive): %v", ...)` inside the pod (not the systemd journal, since it runs as `microshift c2cc-probe` in the deployment). This would also exercise the `Errors` field which is currently untested.

test/suites/c2cc/probe.robot
```diff
@@ -0,0 +1,125 @@
...
+        ...    oc get remoteclusters.microshift.io -o jsonpath='{.items[0].status.lastProbeTime}'
+        Should Not Be Empty    ${stdout}
+    END
+
+RemoteCluster Status Has LastSuccessfulProbe
+    [Documentation]    Verify that LastSuccessfulProbe is populated when state is Healthy.
+    FOR    ${alias}    IN    cluster-a    cluster-b    cluster-c
+        ${stdout}=    Oc On Cluster    ${alias}
+        ...    oc get remoteclusters.microshift.io -o jsonpath='{.items[0].status.lastSuccessfulProbe}'
```
agullon:
```suggestion
        ...    oc get remoteclusters.microshift.io -o jsonpath='{.items[*].status.lastSuccessfulProbe}'
```
it won't change much but I think it's a better check.
Also the next `Should Not Be Empty` can be also improved to something like:

```
@{lines}=    Split To Lines    ${stdout}
${count}=    Get Length    ${lines}
Should Be True    ${count} >= 2    # Expected at least 2 CRs with lastProbeTime
FOR    ${t}    IN    @{lines}
    Should Not Be Empty    ${t}
END

File: test/suites/c2cc/probe.robot [hunk 18]
```diff
@@ -0,0 +1,125 @@
...
+    Wait Until Keyword Succeeds    3m    10s
+    ...    Verify RemoteCluster State    cluster-b    Healthy
+    Wait Until Keyword Succeeds    3m    10s
+    ...    Verify RemoteCluster State    cluster-c    Healthy
+
+RemoteCluster Status Has LastProbeTime
+    [Documentation]    Verify that LastProbeTime is populated after probing starts.
+    FOR    ${alias}    IN    cluster-a    cluster-b    cluster-c
+        ${stdout}=    Oc On Cluster    ${alias}
+        ...    oc get remoteclusters.microshift.io -o jsonpath='{.items[0].status.lastProbeTime}'
```
agullon: same comment as https://github.com/openshift/microshift/pull/6729/changes#r3332992731

test/suites/c2cc/probe.robot
pmtk: Good idea, I will take a look later.
In the mean time, do you think we should introduce a 4th state "Degraded" when just a single probing attempt fails?

agullon: not sure, I think this is a tricky question. Because, for me, it depends a lot on the `probeInterval`. Let me give an example:
 - interval == 1s, I'd marked as degraded if there are more than 5 failed attemps
 - interval == 1m, I don't want to wait 5 mins to marked as degradated.
 Not sure if I explain myself clear.

Regardless I don't think it's neccessary, user can always play with interval (increasing it) if the system is somewhat instable. But I'm happy if you add it.

pkg/controllers/c2cc/probe.go
pmtk: If the cache sync fails, then what should we do? Its docstring says:
```
WaitForCacheSync blocks until all started informers' caches were synced
```
so is stale cache possible? I would say at best no cache at all because this is start up

test/suites/c2cc/probe.robot
```diff
@@ -0,0 +1,204 @@
...
+    ...    oc get service ${PROBE_DEPLOYMENT} -n ${C2CC_NAMESPACE} -o jsonpath='{.spec.clusterIP}'
+    Should Be Equal As Strings    ${actual_ip}    ${expected_ip}    strip_spaces=True
+
+Verify RemoteCluster State
+    [Documentation]    Check that all RemoteCluster CRs on this cluster have the expected state.
+    [Arguments]    ${alias}    ${expected_state}
+    ${stdout}=    Oc On Cluster    ${alias}
+    ...    oc get remoteclusters.microshift.io -o jsonpath='{.items[*].status.state}'
+    Should Not Be Empty    ${stdout}
+    @{states}=    Split String    ${stdout}
```
agullon, line 152: **Suggestion:** `Verify RemoteCluster State` does not assert the expected number of CRs. The jsonpath `{.items[*].status.state}` returns fewer space-separated values if a CR is missing its `.status.state` entirely — the `FOR` loop iterates over fewer items and the test silently passes while claiming "all CRs are Healthy."

In a 3-cluster setup each cluster has 2 RemoteCluster CRs. Consider adding a count assertion after the `Split String`:

```robot
@{states}=    Split String    ${stdout}
${count}=    Get Length    ${states}
Should Be Equal As Integers    ${count}    2    Expected 2 RemoteCluster states, got ${count}
FOR    ${state}    IN    @{states}
    Should Be Equal As Strings    ${state}    ${expected_state}
END
```

The same gap applies to the `LastProbeTime` and `LastSuccessfulProbe` tests (lines 59 and 71) — those also iterate over `Split String` output without verifying the expected count.

pmtk: Thanks for catching that! Fixed

PR #6734, opened 2026-05-26
[release-4.17] USHIFT-7042: Enable EUS repos for RHEL 9.4 bootc test images

pacevedom:
The rhel-bootc:9.4 base image ships with repos pointing to RHEL 9 latest content instead of RHEL 9.4 EUS. This causes dnf to install greenboot-rs (0.16.x) from RHEL 9.8 instead of bash greenboot (0.15.x) from 9.4 EUS. Greenboot-rs sets RefuseManualStartStop=yes on the greenboot-healthcheck.service unit, breaking both the test framework's Restart Greenboot keyword and the rollback/reboot mechanism.

Lock releasever and enable EUS repositories before installing packages, matching the pattern used on release-4.20+ via rpm-repo-config.sh.


pacevedom:
/override ci/prow/ocp-full-conformance-rhel-eus

Unrelated failure, doesnt use the changes included in this PR.

PR #6735, opened 2026-05-26
[release-4.19] USHIFT-7042: Enable EUS repos for RHEL 9.4 and 9.6 bootc test images

pacevedom:
The bootc base images ship with repos pointing to RHEL 9 latest content instead of EUS. This causes dnf to install greenboot-rs (0.16.x) instead of bash greenboot (0.15.x). Greenboot-rs sets RefuseManualStartStop=yes on greenboot-healthcheck.service, breaking both the test framework's Restart Greenboot keyword and the rollback/reboot mechanism.

Lock releasever and enable EUS repositories before installing packages, matching the pattern used on release-4.20+ via rpm-repo-config.sh.


pacevedom:
/override ci/prow/e2e-aws-tests-arm

Unrelated error.

PR #6751, opened 2026-05-27
NO-ISSUE: Fix storage unit test panic

pmtk:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of configuration file handling to better manage edge cases with file notifications.
  * Enhanced validation for configuration files with clearer error reporting for invalid inputs.

### DIFF
--- a/pkg/components/storage.go
+++ b/pkg/components/storage.go
@@ -261,9 +261,14 @@ func loadCSIPluginConfig(ctx context.Context,
 					copyUserCfg(usrCfg, runtimeCfg, cancel)
 				}
 				if event.Has(fsnotify.Remove) {
-					klog.Warningf("lvmd config file %q was removed, this may be due to a reset user configuration or by mistake; "+
-						"now, the new config will be applied from inbuilt defaults", event.Name)
-					copyDefaultCfg(runtimeCfg, cancel)
+					if _, statErr := os.Stat(usrCfg); statErr != nil {
+						klog.Warningf("lvmd config file %q was removed, this may be due to a reset user configuration or by mistake; "+
+							"now, the new config will be applied from inbuilt defaults", event.Name)
+						copyDefaultCfg(runtimeCfg, cancel)
+					} else {
+						klog.Infof("lvmd config file %q received remove event but file still exists, re-applying user config", event.Name)
+						copyUserCfg(usrCfg, runtimeCfg, cancel)
+					}
 				}
 				if event.Has(fsnotify.Chmod) {
 					klog.Warningf("permissions were modified for %q, this may cause side-effects", event.Name)
--- a/pkg/components/storage_test.go
+++ b/pkg/components/storage_test.go
@@ -11,11 +11,11 @@ import (
 	"github.com/openshift/microshift/pkg/config/lvmd"
 )
 
-// runtimeCfgUpdateDelay is the delay to wait for the runtime config to be updated via the watch
-// in the tests. This is a bit of a hack to avoid having to wait for the watch to trigger.
-// In case of failures on loads, consider increasing this value.
-// In a normal I/O environment, this value should never be a problem.
-var runtimeCfgUpdateDelay = 50 * time.Millisecond
+const (
+	runtimeCfgPollInterval = 50 * time.Millisecond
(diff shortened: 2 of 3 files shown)


pmtk:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > pkg/config/lvmd/lvmd.go (1)> `40-46`: _⚠️ Potential issue_ | _🔴 Critical_
> > > **Fix `yaml.Unmarshal(buf, &l)` to `yaml.Unmarshal(buf, l)` and make the “empty file” check actually work**
> > > `l := new(Lvmd)` already yields a non-nil `*Lvmd`; passing `&l` gives `**Lvmd`, and the subsequent `if l == nil` check won’t reliably trigger on empty YAML, so empty lvmd files may be treated as successfully parsed. Use `yaml.Unmarshal(buf, l)` and detect empty input via `buf` (or validate required fields) rather than relying on the pointer being nilled by unmarshal.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@pkg/config/lvmd/lvmd.go` around lines 40 - 46, The current code passes &l to
> > > yaml.Unmarshal and then checks if l == nil, which is wrong because l is created
> > > via l := new(Lvmd) (a non-nil *Lvmd) and &l yields **Lvmd; change the Unmarshal
> > > call to yaml.Unmarshal(buf, l) to pass the *Lvmd, and replace the ineffective
> > > nil check with a real empty-file/validation check — for example, return an error
> > > if len(buf) == 0 or validate required fields on the Lvmd struct (e.g.,
> > > requiredName/ID fields) after unmarshalling; update error messages accordingly
> > > to reference the lvmd file path p and the failing validation.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In `@pkg/config/lvmd/lvmd.go`:
> - Around line 40-46: The current code passes &l to yaml.Unmarshal and then
> checks if l == nil, which is wrong because l is created via l := new(Lvmd) (a
> non-nil *Lvmd) and &l yields **Lvmd; change the Unmarshal call to
> yaml.Unmarshal(buf, l) to pass the *Lvmd, and replace the ineffective nil check
> with a real empty-file/validation check — for example, return an error if
> len(buf) == 0 or validate required fields on the Lvmd struct (e.g.,
> requiredName/ID fields) after unmarshalling; update error messages accordingly
> to reference the lvmd file path p and the failing validation.
> ```
> 
> ℹ️ Review info

The finding is incorrect. yaml.Unmarshal(buf, &l) where l is *Lvmd passes a **Lvmd, which allows the unmarshaller to set the pointer to nil on empty input. This is the intended behavior — the if l == nil check works correctly.

Empirical verification:
```go
package main

import (
      "fmt"
      "sigs.k8s.io/yaml"
)

type Lvmd struct {
      SocketName string `json:"socketName"`
}

func main() {
      l := new(Lvmd)
      fmt.Printf("before: l=%p, *l=%+v\n", l, *l)

      err := yaml.Unmarshal([]byte(""), &l)
      fmt.Printf("after empty unmarshal: l=%v, err=%v\n", l, err)

      l2 := new(Lvmd)
      err = yaml.Unmarshal([]byte("---\n"), &l2)
      fmt.Printf("after '---' unmarshal: l2=%v, err=%v\n", l2, err)

      l3 := new(Lvmd)
      err = yaml.Unmarshal(nil, &l3)
      fmt.Printf("after nil unmarshal: l3=%v, err=%v\n", l3, err)
}
```
Output:
```
before: l=0xc000012140, *l={SocketName:}
after empty unmarshal: l=<nil>, err=<nil>
after '---' unmarshal: l2=<nil>, err=<nil>
after nil unmarshal: l3=<nil>, err=<nil>
```
Changing to yaml.Unmarshal(buf, l) as suggested would break the nil check — the unmarshaller would write into the existing struct and l would never be nil.

PR #6767, opened 2026-05-29
[release-4.19] USHIFT-7124: Wait for existence of nvidia device plugin

pmtk:


scripts/ci-ai-model-serving/tests/04-verify-device-plugin.sh
```diff
@@ -2,6 +2,16 @@
 
 set -xeuo pipefail
 
+retries=30
+while [ ${retries} -gt 0 ] ; do
```
pacevedom, line 6: should we check for retries being 0 at the end of the loop to fail?

pacevedom (approved):
/lgtm

scripts/ci-ai-model-serving/tests/04-verify-device-plugin.sh
pmtk: It's decremented before the operation, so I think it's okay...?

PR #6783, opened 2026-06-01
[release-4.22] NO-ISSUE: Wait a little bit longer for existence of ConfigMap

pmtk:
We need to wait a little bit more because they manifests are applied in alphanumerical order, and the manifest we want to test might lose the lottery and end up at the very end.


pacevedom (approved):
/lgtm

PR #6789, opened 2026-06-02
PERFSCALE-2104: Enable and apply fieldalignment linter

sjug:
Enables the govet fieldalignment analyzer through the existing make verify-golangci path and applies the current baseline cleanup.

What changed:
- Enabled govet.fieldalignment in .golangci.yaml.
- Reordered internal/runtime structs that had no serialization, positional literal, reflection, unsafe, or binary-layout dependency.
- Kept field order for API, config, persisted JSON, startup JSON, OVSDB model, generated, and test/table-fixture shapes.
- Used narrow per-struct //nolint comments where source order is intentional and a file-level exclusion would hide future findings.

Why:
fieldalignment can reduce struct padding and pointer scan ranges, but it is a blunt analyzer. The useful CI value is catching future inefficient layouts in high-cardinality structs, not mechanically rewriting every struct in the tree. This PR fixes the safe internal cases and scopes exclusions so reviewers still get signal for new runtime structs.

Impact:
The structs changed here are mostly startup-time singletons or low-cardinality runtime objects, so direct memory/GC savings are expected to be small. The main benefit is the ongoing guardrail during review.

Verification:
- make verify-golangci
- go test ./pkg/admin/prerun ./pkg/servicemanager/startuprecorder ./pkg/servicemanager ./pkg/controllers/c2cc

## Summary by CodeRabbit

* **Chores**
  * Optimized struct field alignment across the codebase for improved memory efficiency and performance.
  * Enhanced code quality tooling configuration to enforce stricter linting standards across the project.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -73,6 +73,27 @@ linters:
           - errcheck
           - goconst
           - gosec
+      # fieldalignment is useful for runtime structs, but test fixtures are
+      # usually clearer in table-driven test order.
+      - path: '_test\.go'
+        text: 'fieldalignment:'
+        linters:
+          - govet
+      # Preserve Kubernetes API type conventions.
+      - path: 'pkg/apis/microshift/v1alpha1/types\.go'
+        text: 'fieldalignment:'
+        linters:
+          - govet
+      # Preserve user-facing configuration field order.
+      - path: 'pkg/config/.*\.go'
+        text: 'fieldalignment:'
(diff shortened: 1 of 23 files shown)


pacevedom:
/close

Most changes are to structs that are only created once and read only a handful of times. If we find a use case for a high number of requests that would benefit from this change we can take it again, but for now we lean towards dev readability.

PR #6763, opened 2026-05-29
USHIFT-6951: Add metrics exporters as optional MicroShift components

copejon:
Deliver metrics-server, kube-state-metrics, and node-exporter as optional MicroShift components packaged in a single `microshift-metrics` RPM. When installed alongside `microshift-observability`, the otel-collector automatically scrapes all three exporters via drop-in config.

### What's included

- **microshift-metrics RPM** — kustomize manifests for metrics-server, kube-state-metrics, and node-exporter with SCC RBAC, service-ca TLS, and health check registration
- **metrics-server kubelet client cert** — dedicated cert under the kube-apiserver-to-kubelet-signer provisioned as a Secret, avoiding hostPath SELinux issues
- **Auto-rebase integration** — CMO clone + manifest copy recipe in `handle_assets.py`, plus `update_metrics_images()` in `rebase.sh` to populate arch-specific overlays and release JSON from the OCP payload
- **Otel-collector drop-in** — `microshift-metrics.yaml` scrape configs for all three exporters using Kubernetes endpoints SD with service-CA TLS

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added optional metrics-server component for Kubernetes cluster metrics
  * Added kube-state-metrics for monitoring Kubernetes resource states
  * Added node-exporter for host and node-level metrics
  * Integrated OpenTelemetry Collector for metrics collection and export

* **Chores**
  * Added RPM packages for metrics components (x86_64 and aarch64)
  * Added RHEL 10 support in development environment tools
  * Updated deployment manifests and Kubernetes RBAC configurations for metrics workloads

PR #6808, opened 2026-06-05
USHIFT-6951: Add metrics-server as optional MicroShift component

copejon:
Add metrics-server as an optional MicroShift component deployed via kustomize manifests and packaged as `microshift-metrics-server` and `microshift-metrics-server-release-info` RPM sub-packages.

This is PR 1/3 splitting #6763 into independently-mergeable patches. The three PRs can merge in any order.
Siblings: #6809 (kube-state-metrics), #6810 (node-exporter)

### What's included

- **Kustomize manifests** (`assets/optional/metrics-server/`) — Deployment, APIService, RBAC, NetworkPolicy, PDB, service-ca TLS, and `openshift-monitoring` namespace
- **Kubelet client cert provisioning** — dedicated cert under kube-apiserver-to-kubelet-signer, applied as Secret + ConfigMap after the namespace exists (`pkg/cmd/metrics.go`)
- **Healthcheck registration** — `mergeWorkloads()` function and metrics-server map entry in optional workload paths
- **RPM sub-packages** — `%package metrics-server` and `%package metrics-server-release-info` with per-file installs
- **Otel-collector drop-in** — `otelcol.d/` directory structure and `microshift-metrics-server.yaml` scrape config; service file updated to load drop-in configs via bash wrapper
- **Auto-rebase integration** — `update_metrics_images()` in `rebase.sh` + `assets_metrics.yaml` for all three exporters (shared infrastructure lives in this PR)
- **Test infra** — RPM names added to `test/bin/common.sh`

## Release Notes

* **New Features**
  * Added metrics-server as an optional component, including namespace setup, RBAC, TLS assets, hardened deployment, metrics service exposure, and API registration.
  * Added OpenTelemetry Collector scraping for metrics-server via a new collector drop-in configuration.
* **Chores**
  * Added metrics-server optional RPM subpackages and packaging for manifests and release metadata.
  * Updated the MicroShift observability service command handling and improved optional workload merge behavior for checks.

### DIFF
--- a/assets/optional/metrics-server/00-namespace.yaml
+++ b/assets/optional/metrics-server/00-namespace.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-monitoring
+  labels:
+    name: openshift-monitoring
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
--- a/assets/optional/metrics-server/02-configmap-audit-profiles.yaml
+++ b/assets/optional/metrics-server/02-configmap-audit-profiles.yaml
@@ -1,0 +1,45 @@
+apiVersion: v1
+data:
+  metadata-profile.yaml: |-
+    "apiVersion": "audit.k8s.io/v1"
+    "kind": "Policy"
+    "metadata":
+      "name": "Metadata"
+    "omitStages":
+    - "RequestReceived"
+    "rules":
+    - "level": "Metadata"
+  none-profile.yaml: |-
+    "apiVersion": "audit.k8s.io/v1"
+    "kind": "Policy"
+    "metadata":
+      "name": "None"
+    "omitStages":
+    - "RequestReceived"
+    "rules":
+    - "level": "None"
+  request-profile.yaml: |-
+    "apiVersion": "audit.k8s.io/v1"
+    "kind": "Policy"
+    "metadata":
+      "name": "Request"
+    "omitStages":
+    - "RequestReceived"
+    "rules":
+    - "level": "Request"
+  requestresponse-profile.yaml: |-
+    "apiVersion": "audit.k8s.io/v1"
+    "kind": "Policy"
+    "metadata":
+      "name": "RequestResponse"
+    "omitStages":
+    - "RequestReceived"
+    "rules":
+    - "level": "RequestResponse"
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: cluster-monitoring-operator
+    app.kubernetes.io/part-of: openshift-monitoring
+  name: metrics-server-audit-profiles
+  namespace: openshift-monitoring
--- a/assets/optional/metrics-server/03-deployment.yaml
+++ b/assets/optional/metrics-server/03-deployment.yaml
@@ -1,0 +1,114 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics-server
+    app.kubernetes.io/managed-by: cluster-monitoring-operator
+    app.kubernetes.io/name: metrics-server
+    app.kubernetes.io/part-of: openshift-monitoring
+  name: metrics-server
+  namespace: openshift-monitoring
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: metrics-server
+      app.kubernetes.io/name: metrics-server
+      app.kubernetes.io/part-of: openshift-monitoring
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      annotations:
+        openshift.io/required-scc: restricted-v2
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+      labels:
+        app.kubernetes.io/component: metrics-server
+        app.kubernetes.io/name: metrics-server
+        app.kubernetes.io/part-of: openshift-monitoring
+    spec:
+      containers:
+        - args:
+            - --secure-port=10250
+            - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
+            - --kubelet-use-node-status-port
+            - --metric-resolution=15s
+            - --kubelet-certificate-authority=/etc/tls/kubelet-serving-ca-bundle/ca-bundle.crt
+            - --kubelet-client-certificate=/etc/tls/metrics-server-client-certs/tls.crt
+            - --kubelet-client-key=/etc/tls/metrics-server-client-certs/tls.key
+            - --tls-cert-file=/etc/tls/private/tls.crt
+            - --tls-private-key-file=/etc/tls/private/tls.key
+            - --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
+            - --shutdown-send-retry-after=true
+            - --shutdown-delay-duration=150s
+            - --disable-http2-serving=true
+          image: "quay.io/openshift/kube-metrics-server"
+          imagePullPolicy: IfNotPresent
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /livez
+              port: https
+              scheme: HTTPS
+            periodSeconds: 10
+          name: metrics-server
+          ports:
+            - containerPort: 10250
+              name: https
+              protocol: TCP
+          readinessProbe:
+            failureThreshold: 6
+            httpGet:
+              path: /readyz
+              port: https
+              scheme: HTTPS
+            initialDelaySeconds: 20
+            periodSeconds: 20
+          resources:
+            requests:
+              cpu: 1m
+              memory: 40Mi
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+            capabilities:
+              drop:
+                - ALL
+          terminationMessagePolicy: FallbackToLogsOnError
+          volumeMounts:
+            - mountPath: /etc/tls/private
+              name: secret-metrics-server-tls
+            - mountPath: /etc/tls/metrics-server-client-certs
+              name: secret-metrics-server-client-certs
+            - mountPath: /etc/tls/kubelet-serving-ca-bundle
+              name: configmap-kubelet-serving-ca-bundle
+            - mountPath: /etc/audit
+              name: metrics-server-audit-profiles
+              readOnly: true
+            - mountPath: /var/log/metrics-server
+              name: audit-log
+              readOnly: false
+      nodeSelector:
+        kubernetes.io/os: linux
+      priorityClassName: system-cluster-critical
+      serviceAccountName: metrics-server
+      terminationGracePeriodSeconds: 170
+      tolerations:
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/master
+      volumes:
+        - name: secret-metrics-server-client-certs
+          secret:
+            secretName: metrics-server-client-certs
+        - name: secret-metrics-server-tls
+          secret:
+            secretName: metrics-server-tls
+        - configMap:
+            name: kubelet-serving-ca-bundle
+          name: configmap-kubelet-serving-ca-bundle
+        - emptyDir: {}
+          name: audit-log
+        - configMap:
+            name: metrics-server-audit-profiles
+          name: metrics-server-audit-profiles
--- a/assets/optional/metrics-server/kustomization.yaml
+++ b/assets/optional/metrics-server/kustomization.yaml
@@ -1,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - 00-namespace.yaml
+  - 01-service-account.yaml
+  - 01-cluster-role.yaml
+  - 01-cluster-role-binding.yaml
+  - 01-cluster-role-binding-auth-delegator.yaml
+  - 01-role-binding-auth-reader.yaml
+  - 02-configmap-audit-profiles.yaml
+  - 03-deployment.yaml
+  - 04-service.yaml
+  - 04-api-service.yaml
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -868,6 +868,9 @@ fi
 # Use Git command to generate the log and replace the VERSION string
 # LANG=C git log --date="format:%a %b %d %Y" --pretty="tformat:* %cd %an <%ae> VERSION%n- %s%n" packaging/rpm/microshift.spec
 %changelog
+* Mon Jun 22 2026 Jonathan H. Cope <jcope@redhat.com> 5.0
+- Add metrics-server as optional rpm package
+
 * Tue Jan 20 2026 Pablo Acevedo Montserrat <pacevedo@redhat.com> 4.21.0
 - Add multus as dependency for sriov
 
--- a/packaging/rpm/microshift.spec
+++ b/packaging/rpm/microshift.spec
@@ -261,6 +261,25 @@ The microshift-cert-manager-release-info package provides release information fi
 release. These files contain the list of container image references used by Cert Manager
 and can be used to embed those images into osbuilder blueprints or bootc containerfiles.
 
+%package metrics-server
+Summary: Kubernetes metrics-server for MicroShift
+ExclusiveArch: x86_64 aarch64
+Requires: microshift = %{version}
+
+%description metrics-server
+The microshift-metrics-server package provides the metrics-server for MicroShift.
+Install this package to enable kubectl top and resource metrics via the Metrics API.
+
+%package metrics-server-release-info
+Summary: Release information for metrics-server for MicroShift
+BuildArch: noarch
+Requires: microshift-release-info = %{version}
+
+%description metrics-server-release-info
+The microshift-metrics-server-release-info package provides release information files for this
+release. These files contain the list of container image references used by the metrics-server
+and can be used to embed those images into osbuilder blueprints or bootc containerfiles.
+
 %package sriov
 Summary: SR-IOV Network Operator for MicroShift
 ExclusiveArch: x86_64 aarch64
@@ -599,6 +618,31 @@ cat assets/optional/cert-manager/manager/images-x86_64.yaml >> %{buildroot}/%{_p
 mkdir -p -m755 %{buildroot}%{_datadir}/microshift/release
 install -p -m644 assets/optional/cert-manager/release-cert-manager-{x86_64,aarch64}.json %{buildroot}%{_datadir}/microshift/release/
 
+# metrics-server
+install -d -m755 %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/00-namespace.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/01-service-account.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/01-cluster-role.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/01-cluster-role-binding.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/01-cluster-role-binding-auth-delegator.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/01-role-binding-auth-reader.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/02-configmap-audit-profiles.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/03-deployment.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/04-service.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/04-api-service.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+install -p -m644 assets/optional/metrics-server/kustomization.yaml %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+
+%ifarch %{arm} aarch64
+cat assets/optional/metrics-server/kustomization.aarch64.yaml >> %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server/kustomization.yaml
+%endif
+%ifarch x86_64
+cat assets/optional/metrics-server/kustomization.x86_64.yaml >> %{buildroot}/%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server/kustomization.yaml
+%endif
+
+# metrics-server-release-info
+mkdir -p -m755 %{buildroot}%{_datadir}/microshift/release
+install -p -m644 assets/optional/metrics-server/release-metrics-server-{x86_64,aarch64}.json %{buildroot}%{_datadir}/microshift/release/
+
 # sriov
 install -d -m755 %{buildroot}/%{_prefix}/lib/microshift/manifests.d/070-microshift-sriov
 install -d -m755 %{buildroot}/%{_prefix}/lib/microshift/manifests.d/070-microshift-sriov/crd
@@ -801,6 +845,13 @@ fi
 
 %files cert-manager-release-info
 %{_datadir}/microshift/release/release-cert-manager-{x86_64,aarch64}.json
+
+%files metrics-server
+%dir %{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server
+%{_prefix}/lib/microshift/manifests.d/080-microshift-metrics-server/*
+
+%files metrics-server-release-info
+%{_datadir}/microshift/release/release-metrics-server-{x86_64,aarch64}.json
 
 %files sriov
 %dir %{_prefix}/lib/microshift/manifests.d/070-microshift-sriov
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -155,6 +155,13 @@ func certSetup(cfg *config.Config) (*certchains.CertificateChains, error) {
 					Validity: alignValidity(cryptomaterial.ShortLivedCertificateValidity),
 				},
 				UserInfo: &user.DefaultInfo{Name: "system:kube-apiserver", Groups: []string{"kube-master"}},
+			}).WithClientCertificates(
+			&certchains.ClientCertificateSigningRequestInfo{
+				CSRMeta: certchains.CSRMeta{
+					Name:     "metrics-server-kubelet-client",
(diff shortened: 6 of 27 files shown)


copejon:
@coderabbit review

assets/optional/metrics-server/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - 00-namespace.yaml
+  - 01-service-account.yaml
+  - 01-cluster-role.yaml
+  - 01-cluster-role-binding.yaml
+  - 01-cluster-role-binding-auth-delegator.yaml
+  - 01-role-binding-auth-reader.yaml
+  - 02-configmap-audit-profiles.yaml
+  - 03-deployment.yaml
+  - 04-service.yaml
+  - 04-api-service.yaml
```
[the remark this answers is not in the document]
copejon: Stale: `network-policy-downstream.yaml` and `pod-disruption-budget.yaml` were intentionally deleted — PDB `minAvailable:1` blocks drains on single-node MicroShift, and the NetworkPolicy is unnecessary for this workload on MicroShift. The spec install lines were also removed in 82550e8fa.

scripts/auto-rebase/assets.yaml
```diff
@@ -301,6 +301,29 @@ assets:
...
+      - file: kustomization.yaml
+      - file: kustomization.x86_64.yaml
+        ignore: "gets generated during image rebase"
+      - file: kustomization.aarch64.yaml
+        ignore: "gets generated during image rebase"
+      - file: release-metrics-server-x86_64.json
+        ignore: "gets generated during image rebase"
+      - file: release-metrics-server-aarch64.json
+        ignore: "gets generated during image rebase"
+
```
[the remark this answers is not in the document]
copejon: Stale: these files were intentionally deleted — PDB `minAvailable:1` blocks drains on single-node MicroShift, and the NetworkPolicy is unnecessary for this workload on MicroShift. The spec install lines were also removed in 82550e8fa.

assets/optional/metrics-server/03-deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+          periodSeconds: 20
+        resources:
+          requests:
+            cpu: 1m
+            memory: 40Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+        terminationMessagePolicy: FallbackToLogsOnError
```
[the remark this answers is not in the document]
copejon: Fixed in df4382a4b.

pkg/components/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("creating clientset: %w", err)
+	}
+	const ns = "openshift-monitoring"
+	err = wait.PollUntilContextTimeout(ctx, 2*time.Second, 5*time.Minute, true, func(ctx context.Context) (bool, error) {
+		_, err := clientset.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+		if err == nil {
+			return true, nil
+		}
+		klog.V(2).Infof("Waiting for namespace %s to be created by kustomize", ns)
+		return false, nil
```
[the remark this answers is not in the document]
copejon: Fixed in 81264bbd3.

scripts/auto-rebase/assets_metrics.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - file: kustomization.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.x86_64.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.aarch64.yaml
+        ignore: "Provided by MicroShift"
+      - file: release-metrics-aarch64.json
+        ignore: "Provided by MicroShift"
+      - file: release-metrics-x86_64.json
+        ignore: "Provided by MicroShift"
```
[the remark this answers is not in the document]
copejon: Fixed in df4382a4b.

scripts/auto-rebase/rebase.sh
```diff
@@ -1111,6 +1112,95 @@ EOF
...
+            local image_names
+            image_names=$(grep -h 'image:' "${REPOROOT}/assets/optional/${component_dir}/"*.yaml 2>/dev/null \
+                | sed 's/.*image: *//; s/:.*//; s/@.*//' | sort -u)
+
+            for orig_image in ${image_names}; do
+                local release_tag="${METRICS_IMAGE_MAP[$orig_image]:-}"
+                if [[ -z "${release_tag}" ]]; then
+                    >&2 echo "WARNING: Unknown metrics image '${orig_image}' in ${component_dir}, skipping"
+                    continue
+                fi
```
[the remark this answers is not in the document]
copejon: Fixed in df4382a4b.

scripts/auto-rebase/rebase.sh
```diff
@@ -1111,6 +1112,102 @@ EOF
...
+            # Generate per-component release JSON
+            local json_key="${METRICS_COMPONENT_JSON_KEY[$component_dir]}"
+            local release_tag="${METRICS_EXPORTER_JSON_MAP[$json_key]}"
+            local new_image
+            new_image=$(jq -r ".references.spec.tags[] | select(.name == \"${release_tag}\") | .from.name" "${release_file}")
+            local new_digest="${new_image#*@}"
+            local redhat_name="${METRICS_REDHAT_REGISTRY_MAP[$json_key]}"
+            local component_release_json="${REPOROOT}/assets/optional/${component_dir}/release-${component_dir}-${arch}.json"
+            jq -n --arg base "$base_release" --arg img "${redhat_name}@${new_digest}" \
+                "{\"release\": {\"base\": \$base}, \"images\": {\"${json_key}\": \$img}}" > "${component_release_json}"
```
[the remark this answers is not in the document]
copejon: Fixed in 6ab794d24.

assets/optional/metrics-server/00-namespace.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-monitoring
+  labels:
+    name: openshift-monitoring
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
```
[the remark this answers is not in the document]
copejon: `privileged` is consistent with every other MicroShift system namespace (openshift-dns, openshift-ingress, openshift-ovn-kubernetes, etc.) and is required for metrics-server kubelet access.

copejon:
@coderabbit review

test/assets/observability/otel_config.yaml
```diff
@@ -59,6 +59,10 @@ exporters:
       enabled: true
   otlphttp/loki: # only for logs, exports the logs in the loki server
     endpoint: "http://{{LOKI_HOST}}:{{LOKI_PORT}}/otlp"
+  otlp:
+    endpoint: "localhost:4317"
+    tls:
+      insecure: true
```
agullon: why this block is needed if there are no tests updates? afaik this file is only used in the RF tests

copejon:
@coderabbit help

test/assets/observability/otel_config.yaml
copejon: This stub `otlp` exporter prevents the OpenTelemetry Collector from crashing when the observability RF test runs on systems with metrics components installed.

Each metrics component ships an otelcol drop-in config (`otelcol.d/microshift-metrics-*.yaml`) that defines a pipeline exporting to the `otlp` exporter. The production configs already define `otlp`, but this test config did not. When the RF test replaces the production config with this test config and restarts the service, the collector loads both:
1. This test config (previously lacked `otlp` exporter)
2. Any drop-in configs present on the system (which reference `otlp`)

Without the stub, the collector fails to start due to undefined exporter references.

The stub (`localhost:4317`, insecure, no backend) satisfies config validation without affecting test behavior. No new RF tests are needed — this isn't testing the metrics components, it's preventing existing observability tests from breaking when metrics components are installed.

agullon: I see, thanks for the explanation.

agullon:
I reviewed the 3 PRs and they look good to me. But at least some basic RF tests are missing.

pkg/cmd/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+
+	restCfg, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		return fmt.Errorf("building kubeconfig: %w", err)
+	}
+	clientset, err := kubernetes.NewForConfig(restCfg)
+	if err != nil {
+		return fmt.Errorf("creating clientset: %w", err)
+	}
+	const ns = "openshift-monitoring"
```
pmtk: Why `ns` is here, but `metricsServerManifestPath` outside the function?

pkg/cmd/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+const metricsServerManifestPath = "/usr/lib/microshift/manifests.d/080-microshift-metrics-server"
+
+func provisionMetricsServerCerts(ctx context.Context, cfg *config.Config) error {
```
pmtk: This looks like it might make a good generic function to create secrets with certs from the disk. Not big deal, just `metrics.go` seems a bit too specific to be in `cmd/`. So far I think it's mostly individual commands in each file

pkg/cmd/run.go
```diff
@@ -303,6 +303,18 @@ func RunMicroshift(cfg *config.Config) error {
 		// After MicroShift's core becomes ready, run the kustomizer (delete and/or apply manifests).
 		kustomize.NewKustomizer(cfg).RunStandalone(runCtx)
 
+		// Provision certs for optional components after kustomize creates their namespaces.
+		go func() {
+			defer func() {
+				if r := recover(); r != nil {
```
pmtk: Do we expect a panic here or is this just very defensive coding?

scripts/auto-rebase/assets.yaml
```diff
@@ -301,6 +301,29 @@ assets:
       - file: service.yaml
       - file: serviceaccount.yaml
 
+  - dir: optional/metrics-server/
```
pmtk: If we have `assets_metrics.yaml`, then I don't think we should have this entry in this file

copejon:
2 unrelated flakes resulted in failures

/test e2e-aws-tests 
/test e2e-aws-tests-bootc-periodic-el9

assets/optional/metrics-server/03-deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+          httpGet:
+            path: /readyz
+            port: https
+            scheme: HTTPS
+          initialDelaySeconds: 20
+          periodSeconds: 20
+        resources:
+          requests:
+            cpu: 1m
+            memory: 40Mi
```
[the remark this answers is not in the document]
copejon: This is necessary and also reflect the upstream CMO manifest exactly. No change needed.

pkg/components/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+
+func provisionMetricsServerCerts(ctx context.Context, cfg *config.Config) error {
+	exists, err := util.PathExists(metricsServerManifestPath)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		klog.V(2).Infof("Metrics-server manifests not found at %s, skipping cert provisioning", metricsServerManifestPath)
+		return nil
+	}
```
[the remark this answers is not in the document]
copejon: No change needed. If the component is installed, it is always enabled because microshift will deploy the manifests on (re)start, which is when the certs are generated.

pkg/cmd/run.go
copejon: Defensive, and overkill. Nothing the func can panic. Removed.

pkg/cmd/metrics.go
copejon: I've moved metrics.go to pkg/components and exported `provisionMetricsServerCerts`, which is now called in pkg/cmd/run.go. Of the 3 metrics components, `metrics-server` is special in that it requires a kubelet client cert in order to get cluster resource data. Hence the need to generate that cert just before applying the component's manifests.

pkg/cmd/metrics.go
copejon: both constants are now package-level in a grouped const block, matching the convention in certificateauthority.go. 

https://github.com/copejon/microshift/commit/2c6da6f226cfe8151ffb6234fe2f933924603bb1

pkg/cmd/metrics.go
copejon: https://github.com/copejon/microshift/commit/2c6da6f226cfe8151ffb6234fe2f933924603bb1

pkg/cmd/run.go
copejon: https://github.com/copejon/microshift/commit/2c6da6f226cfe8151ffb6234fe2f933924603bb1

scripts/auto-rebase/assets.yaml
copejon: Removed the duplicate block — assets_metrics.yaml is the sole tracker. https://github.com/copejon/microshift/commit/d61c8bed1ccfc9b2adbd70db3916af575feef389

copejon:
Ah, need to re-rebase. Handling that now

copejon (commented):
Code review (high effort, 7 angles, 1-vote verify). 5 findings survived verification.

pkg/components/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+
+	err = wait.PollUntilContextTimeout(ctx, 2*time.Second, 5*time.Minute, true, func(ctx context.Context) (bool, error) {
+		_, err := clientset.CoreV1().Namespaces().Get(ctx, metricsServerNamespace, metav1.GetOptions{})
+		if err == nil {
+			return true, nil
+		}
+		if !apierrors.IsNotFound(err) {
+			return false, fmt.Errorf("getting namespace %s: %w", metricsServerNamespace, err)
+		}
```
copejon, line 59: **Bug: Transient API errors abort namespace poll permanently**

`PollUntilContextTimeout` exits immediately when the condition returns a non-nil error. Any API error other than `NotFound` (e.g. connection refused, Unauthorized, request timeout) causes `fmt.Errorf` here to return a non-nil error, which terminates the poll loop. Since the caller goroutine in `run.go` has no retry, the Secret and ConfigMap are never created and the metrics-server pod stays in `ContainerCreating` indefinitely.

Fix: log non-NotFound errors and continue polling:
```suggestion
		if !apierrors.IsNotFound(err) {
			klog.V(2).Infof("Transient error checking namespace %s, will retry: %v", metricsServerNamespace, err)
			return false, nil
		}
```

pkg/cmd/run.go
```diff
@@ -303,6 +304,13 @@ func RunMicroshift(cfg *config.Config) error {
 		// After MicroShift's core becomes ready, run the kustomizer (delete and/or apply manifests).
 		kustomize.NewKustomizer(cfg).RunStandalone(runCtx)
 
+		// Provision certs for optional components after kustomize creates their namespaces.
+		go func() {
+			if err := components.ProvisionMetricsServerCerts(runCtx, cfg); err != nil {
+				klog.Errorf("Failed to provision metrics-server certs: %v", err)
```
copejon, line 311: **Bug: Fire-and-forget goroutine for deployment-blocking work**

`ProvisionMetricsServerCerts` is called in a goroutine that only logs errors. However, the metrics-server Deployment mounts `secret-metrics-server-client-certs` and `configmap-kubelet-serving-ca-bundle` as required volumes (no `optional: true`). If this function fails, the pod remains stuck in `ContainerCreating` indefinitely with no retry — the healthcheck times out with a misleading "deployment not ready" rather than "cert provisioning failed".

Consider adding retry logic around the `ProvisionMetricsServerCerts` call, or propagating the error.

pkg/components/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+	_, err = clientset.CoreV1().Secrets(metricsServerNamespace).Create(ctx, secret, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		_, err = clientset.CoreV1().Secrets(metricsServerNamespace).Update(ctx, secret, metav1.UpdateOptions{})
+	}
+	if err != nil {
+		return fmt.Errorf("applying metrics-server client cert secret: %w", err)
+	}
+
+	caPEM, err := os.ReadFile(cryptomaterial.KubeletClientCAPath(certsDir))
+	if err != nil {
```
copejon, line 107: **Concern: CA bundle for kubelet serving verification is overly broad**

`KubeletClientCAPath` returns a bundle containing ALL client cert signers (kube-control-plane-signer, kube-apiserver-to-kubelet-signer, admin-kubeconfig-signer, kubelet-signer, kube-csr-signer). This is consumed by `--kubelet-certificate-authority` to verify kubelet *serving* certs.

The kubelet serving cert is signed by `kube-csr-signer`, so verification works. But the bundle also trusts 3 unrelated client-auth CAs as valid kubelet serving cert signers. In MicroShift's single-node model the risk is low, but a narrower bundle (just `kubelet-signer` and its sub-CAs) would be more precise.

pkg/components/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+			},
+		},
+		Type: corev1.SecretTypeTLS,
+		Data: map[string][]byte{
+			"tls.crt": certPEM,
+			"tls.key": keyPEM,
+		},
+	}
+	_, err = clientset.CoreV1().Secrets(metricsServerNamespace).Create(ctx, secret, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
```
copejon: **Cleanup: Create-then-Update pattern lacks retry-on-conflict**

The hand-rolled `Create` / `if AlreadyExists then Update` pattern here (and on the ConfigMap at line 107) does not handle conflict errors on Update. If another controller modifies the resource between the failed Create and Update, the Update fails with `409 Conflict` and the function returns an error permanently.

The codebase has `pkg/assets/core.go` with `ApplySecretWithData` / `ApplyConfigMapWithData` using `resourceapply` with `retry.RetryOnConflict`. Those require an embedded asset template so they aren't drop-in, but wrapping the Update in `retry.RetryOnConflict` would make this more robust.

packaging/rpm/microshift.spec
```diff
@@ -236,6 +236,7 @@ and can be used to embed those images into osbuilder blueprints or bootc contain
 Summary: OpenTelemetry-Collector configured for MicroShift
 BuildArch: noarch
 Requires: microshift = %{version}
+Requires: microshift-metrics-server = %{version}
 Requires: opentelemetry-collector
```
copejon, line 239: **Design: observability hard-requires metrics-server, coupling independent features**

The otelcol.d drop-in pattern (`for f in .../otelcol.d/*.yaml; do [ -f "$f" ] && ...`) was designed for composable config. But this hard `Requires` forces metrics-server installation whenever observability is installed.

If `microshift-metrics-server.yaml` were shipped in the `microshift-metrics-server` RPM instead (only appears when metrics-server is installed), the `Requires` could be dropped and observability would be independently installable.

pkg/components/metrics.go
copejon: Fixed. Non-NotFound errors now log and continue polling (`return false, nil`) instead of aborting. https://github.com/copejon/microshift/blob/8aea5c27a131d3d033af73e3c480a4b98429dae7/pkg/components/metrics.go#L51-L53

pkg/cmd/run.go
copejon: Fixed. `ProvisionMetricsServerCerts` is now called synchronously with full error propagation — no goroutine. https://github.com/copejon/microshift/blob/8aea5c27a131d3d033af73e3c480a4b98429dae7/pkg/cmd/run.go#L308-L310

pkg/components/metrics.go
copejon: Fixed. Switched from `KubeletClientCAPath` (5 signers) to `KubeletServingCAPath` (only `kubelet-signer` + `kube-csr-signer`). https://github.com/copejon/microshift/blob/8aea5c27a131d3d033af73e3c480a4b98429dae7/pkg/components/metrics.go#L101

pkg/components/metrics.go
copejon: Fixed. Replaced hand-rolled Create/Update with `resourceapply.ApplySecret` and `resourceapply.ApplyConfigMap`, wrapped in `PollUntilContextTimeout` which retries any error including 409 Conflict. https://github.com/copejon/microshift/blob/8aea5c27a131d3d033af73e3c480a4b98429dae7/pkg/components/metrics.go#L89-L96

packaging/rpm/microshift.spec
copejon: Fixed. The otelcol drop-in is now owned by the `microshift-metrics-server` RPM, not observability. No `Requires:` from observability to metrics-server exists. https://github.com/copejon/microshift/blob/8aea5c27a131d3d033af73e3c480a4b98429dae7/packaging/rpm/microshift.spec#L859

pkg/cmd/init.go
```diff
@@ -364,6 +371,10 @@ func certSetup(cfg *config.Config) (*certchains.CertificateChains, error) {
 		[]string{"admin-kubeconfig-signer"},
 		[]string{"kubelet-signer"},
 		[]string{"kubelet-signer", "kube-csr-signer"},
+	).WithCABundle(
+		cryptomaterial.KubeletServingCAPath(certsDir),
+		[]string{"kubelet-signer"},
+		[]string{"kubelet-signer", "kube-csr-signer"},
```
pacevedom, line 377: wondering if this would fit in the previous CA bundle, there is only 1 difference and its just CAs.

pkg/components/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+const (
+	metricsServerManifestPath = "/usr/lib/microshift/manifests.d/080-microshift-metrics-server"
+	metricsServerNamespace    = "openshift-monitoring"
+)
+
+var metricsEventRecorder events.Recorder = events.NewLoggingEventRecorder("microshift-metrics-server", clock.RealClock{})
+
+func ProvisionMetricsServerCerts(ctx context.Context, cfg *config.Config) error {
```
pacevedom, line 34: Wondering if this function should be asynchronous. For an optional component it would block MicroShift readiness until its ready, and this is the only component doing this that I know of.

copejon: ah that's a good point. will fix

pkg/cmd/run.go
```diff
@@ -303,6 +304,11 @@ func RunMicroshift(cfg *config.Config) error {
 		// After MicroShift's core becomes ready, run the kustomizer (delete and/or apply manifests).
 		kustomize.NewKustomizer(cfg).RunStandalone(runCtx)
 
+		// Provision certs for optional components after kustomize creates their namespaces.
+		if err := components.ProvisionMetricsServerCerts(runCtx, cfg); err != nil {
+			return fmt.Errorf("failed to provision metrics-server certs: %w", err)
+		}
```
agullon: hi @copejon if I understand this block correctly, it blocks the go routine until the metrics server is installed and running. Also if there's any failure it reports microshift failed to start, right? Is this the desired behavior? It looks to me an overkill to report microshift failure if an extra component instalation fails.

pkg/components/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+		Data: map[string]string{
+			"ca-bundle.crt": string(caPEM),
+		},
+	}
+
+	err = wait.PollUntilContextTimeout(ctx, 2*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
+		_, _, err := resourceapply.ApplyConfigMap(ctx, clientset.CoreV1(), metricsEventRecorder, cm)
+		if err != nil {
+			return false, fmt.Errorf("applying kubelet serving CA configmap: %w", err)
```
agullon: hi @copejon, I'm comparing the error handeling of this ConfigMap apply function with the previous Secret apply function and I see few differences I don't quite understand, can you please review it?

- Secret in line 92 does not return error only log the error, but ConfigMap on line 122 returns the error.
- Secret in line 98 return error with `%w` and ConfigMap in line 127 return error with `%v`

copejon:
Scheduling failures blocking some of the tests from starting: `There are no nodes that your pod can schedule to`

/test e2e-aws-tests-bootc-el10
/test e2e-aws-tests-periodic-arm
/test ocp-full-conformance-serial-rhel-eus

assets/optional/metrics-server/02-configmap-audit-profiles.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - "level": "Request"
+  requestresponse-profile.yaml: |-
+    "apiVersion": "audit.k8s.io/v1"
+    "kind": "Policy"
+    "metadata":
+      "name": "RequestResponse"
+    "omitStages":
+    - "RequestReceived"
+    "rules":
+    - "level": "RequestResponse"
```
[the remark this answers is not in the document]
copejon: Won't fix — quoted keys are valid YAML; this is an upstream CMO manifest and changing it creates rebase drift.

packaging/observability/microshift-observability.service
```diff
@@ -8,7 +8,7 @@ ConditionPathExists=/var/lib/microshift/resources/observability-client/kubeconfi
 Environment=KUBECONFIG=/var/lib/microshift/resources/observability-client/kubeconfig
 Environment=K8S_NODE_NAME="%l"
 ExecStartPre=/usr/bin/mkdir -p /var/lib/microshift-observability
-ExecStart=/usr/bin/opentelemetry-collector --config=/etc/microshift/observability/opentelemetry-collector.yaml
+ExecStart=/bin/bash -c 'ARGS="--config=file:/etc/microshift/observability/opentelemetry-collector.yaml"; for f in /etc/microshift/observability/otelcol.d/*.yaml; do [ -f "$$f" ] && ARGS="$$ARGS --config=file:$$f"; done; exec /usr/bin/opentelemetry-collector $$ARGS'
```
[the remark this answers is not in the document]
copejon: Won't fix — file removed from this branch (otelcol decoupled).

scripts/auto-rebase/assets_metrics.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - file: 03-daemonset.yaml
+        ignore: "MicroShift removes metrics-client-ca volume/mount/arg (populated by CMO at runtime)"
+      - file: 04-service.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.x86_64.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.aarch64.yaml
+        ignore: "Provided by MicroShift"
```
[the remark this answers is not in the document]
copejon: Won't fix — assets_metrics.yaml no longer exists; assets restructured into assets_cluster_monitoring_operator.yaml.

scripts/auto-rebase/assets_metrics.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - file: 03-deployment.yaml
+        ignore: "MicroShift overrides: Recreate strategy, removes metrics-client-ca, image placeholders"
+      - file: 04-service.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.x86_64.yaml
+        ignore: "Provided by MicroShift"
+      - file: kustomization.aarch64.yaml
+        ignore: "Provided by MicroShift"
```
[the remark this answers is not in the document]
copejon: Won't fix — assets_metrics.yaml no longer exists; assets restructured into assets_cluster_monitoring_operator.yaml.

pkg/components/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+		Type: corev1.SecretTypeTLS,
+		Data: map[string][]byte{
+			"tls.crt": certPEM,
+			"tls.key": keyPEM,
+		},
+	}
+	_, err = clientset.CoreV1().Secrets(metricsServerNamespace).Create(ctx, secret, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		_, err = clientset.CoreV1().Secrets(metricsServerNamespace).Update(ctx, secret, metav1.UpdateOptions{})
+	}
```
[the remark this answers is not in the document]
copejon: Won't fix — code rewritten to use resourceapply.ApplySecret/ApplyConfigMap which handles resourceVersion internally.

copejon:
Unrelated flakes in peridiocs. The presubmit counterparts passed.

/test e2e-aws-tests-bootc-periodic-el9
/test e2e-aws-tests-bootc-periodic-arm-el10

copejon:
Unrelated dns flake

/test e2e-aws-tests-bootc-periodic-arm-el10

pkg/components/metrics.go
pmtk: Doesn't returning an error end the PollUntil* logic?

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ConditionWithContextFunc:
> ConditionWithContextFunc returns true if the condition is satisfied, or an error if the loop should be aborted.

PR #6875, opened 2026-06-15
USHIFT-6947: C2CC: Configurable routing table IDs

pmtk:
## Summary by CodeRabbit

* **New Features**
  * Added `clusterToCluster.routing` for C2CC Linux policy routing table IDs (`routeTableID`, `serviceRouteTableID`) with defaulting, validation (allowed range), and requirement that the two IDs differ.
* **Documentation**
  * Updated `config.yaml` examples and the configuration guide to include the new `routing` block.
  * Updated the published OpenAPI/JSON schema for the new required `routing` section.
* **Tests**
  * Added unit tests for routing defaulting, validation, and user override behavior.
  * Added Robot Framework suites verifying both default and custom routing table IDs are applied.

### DIFF
--- a/docs/user/howto_config.md
+++ b/docs/user/howto_config.md
@@ -40,6 +40,9 @@ clusterToCluster:
           domain: ""
           nextHop: ""
           serviceNetwork: []
+    routing:
+        routeTableID: 0
+        serviceRouteTableID: 0
 debugging:
     logLevel: ""
 dns:
@@ -206,6 +209,9 @@ clusterToCluster:
           domain: ""
           nextHop: ""
           serviceNetwork: []
+    routing:
+        routeTableID: 200
+        serviceRouteTableID: 201
 debugging:
     logLevel: Normal
 dns:
--- a/pkg/config/c2cc.go
+++ b/pkg/config/c2cc.go
@@ -26,9 +26,30 @@ type C2CCDNS struct {
 	CacheNegativeTTL *int `json:"cacheNegativeTTL,omitempty"`
 }
 
+type C2CCRouting struct {
+	// Linux policy routing table ID for direct routes to remote cluster CIDRs.
+	// The route protocol number is set to the same value.
+	// Must be between 1 and 252 (253-255 are reserved by the kernel).
+	// Must differ from serviceRouteTableID.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=252
+	// +kubebuilder:default=200
+	RouteTableID *int `json:"routeTableID,omitempty"`
+	// Linux policy routing table ID for service routes via the OVN management port.
+	// The route protocol number is set to the same value.
+	// Must be between 1 and 252 (253-255 are reserved by the kernel).
+	// Must differ from routeTableID.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=252
+	// +kubebuilder:default=201
+	ServiceRouteTableID *int `json:"serviceRouteTableID,omitempty"`
+}
+
 type C2CC struct {
 	// DNS cache settings for CoreDNS server blocks generated for remote clusters.
 	DNS C2CCDNS `json:"dns"`
+	// Linux policy routing table settings for C2CC routes.
+	Routing C2CCRouting `json:"routing"`
 	// List of remote clusters to establish connectivity with.
 	// C2CC is disabled when this list is empty.
 	RemoteClusters []RemoteCluster `json:"remoteClusters,omitempty"`
@@ -39,9 +60,11 @@ type C2CC struct {
 	ProbeInterval string `json:"probeInterval,omitempty"`
 
 	// Populated during validation with parsed network objects.
-	Resolved              []ResolvedRemoteCluster `json:"-"`
-	ResolvedAllCIDRs      []*net.IPNet            `json:"-"`
-	ResolvedProbeInterval time.Duration           `json:"-"`
+	Resolved                    []ResolvedRemoteCluster `json:"-"`
+	ResolvedAllCIDRs            []*net.IPNet            `json:"-"`
+	ResolvedProbeInterval       time.Duration           `json:"-"`
+	ResolvedRouteTableID        int                     `json:"-"`
+	ResolvedServiceRouteTableID int                     `json:"-"`
 }
 
 type RemoteCluster struct {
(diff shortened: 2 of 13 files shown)


docs/user/howto_config.md
```diff
@@ -40,6 +40,9 @@ clusterToCluster:
           domain: ""
           nextHop: ""
           serviceNetwork: []
+    routing:
+        routeTableID: 0
+        serviceRouteTableID: 0
```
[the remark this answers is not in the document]
pmtk: This is how all the values are set in this code block and we cannot change it. Below there's a snippet with defaults and it has correct 200 & 201 values

pkg/config/config.go
```diff
@@ -578,6 +585,7 @@ func (c *Config) updateComputedValues() error {
 	}
 
 	c.C2CC.stripEmptyRemoteClusters()
+	c.C2CC.resolveRoutingDefaults()
```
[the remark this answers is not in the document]
pmtk: Good point, but this is not the way to do it.
Also, it's kind of the intention: when C2CC gets disabled we want to clean up the resources and that also means the routes.

@agullon do you think we need to add "status: {Enabled,Disabled}" field to c2cc config to explicitly set the feature instead of relying on `len(remoteClusters)`?

pkg/config/c2cc_test.go
```diff
@@ -82,7 +92,7 @@ func mkC2CCConfig(c2cc C2CC) *Config {
 		Node: Node{
 			NodeIP: "10.100.0.1",
 		},
-		C2CC: withDNSDefaults(c2cc),
+		C2CC: withRoutingDefaults(withDNSDefaults(c2cc)),
 	}
```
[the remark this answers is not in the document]
pmtk: fixed

pkg/config/config.go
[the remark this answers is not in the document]
agullon: I don't see a real value for user adding enable/disable toggle. As far as I understand how current implementation works it'll be enough to remove all c2cc config and restart microshift.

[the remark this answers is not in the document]
pmtk: I agree, thx for input

docs/user/howto_config.md
agullon, line 45:
```suggestion
        routeTableID: 1
        serviceRouteTableID: 1
```
because min values are `1`, right?

pmtk: I don't know what this part of the .md is, but it has zeros everywhere.
And it's autogenerated, so we'd need to fix the generator

agullon: forget about this, this is just about the schema, `0` just means integer type. Other fields follow the same approach, example:
```
    ports:
        http: 0
        https: 0
```

pkg/config/c2cc.go
```diff
@@ -193,16 +216,58 @@ func parseAndValidateCIDR(cidr, field string, errs *[]error) *net.IPNet {
...
+	}
+	if c2cc.ResolvedServiceRouteTableID < 1 || c2cc.ResolvedServiceRouteTableID > 252 {
+		errs = append(errs, fmt.Errorf("routing.serviceRouteTableID must be between 1 and 252, got %d", c2cc.ResolvedServiceRouteTableID))
+	}
+	if c2cc.ResolvedRouteTableID == c2cc.ResolvedServiceRouteTableID {
+		errs = append(errs, fmt.Errorf("routing.routeTableID (%d) and routing.serviceRouteTableID (%d) must differ",
+			c2cc.ResolvedRouteTableID, c2cc.ResolvedServiceRouteTableID))
+	}
+	return errors.Join(errs...)
+}
```
agullon: Nit: The receiver `r *C2CCRouting` is never used in this method — everything reads from `c2cc.Resolved*`. Compare with `C2CCDNS.validate()` which correctly uses its receiver `d`.

Since this is really a C2CC-level concern (needs resolved values), consider making it a method on `*C2CC` instead:

```suggestion
func (c *C2CC) validateRouting() error {
	c.resolveRoutingDefaults()

	var errs []error
	if c.ResolvedRouteTableID < 1 || c.ResolvedRouteTableID > 252 {
		errs = append(errs, fmt.Errorf("routing.routeTableID must be between 1 and 252, got %d", c.ResolvedRouteTableID))
	}
	if c.ResolvedServiceRouteTableID < 1 || c.ResolvedServiceRouteTableID > 252 {
		errs = append(errs, fmt.Errorf("routing.serviceRouteTableID must be between 1 and 252, got %d", c.ResolvedServiceRouteTableID))
	}
	if c.ResolvedRouteTableID == c.ResolvedServiceRouteTableID {
		errs = append(errs, fmt.Errorf("routing.routeTableID (%d) and routing.serviceRouteTableID (%d) must differ",
			c.ResolvedRouteTableID, c.ResolvedServiceRouteTableID))
	}
	return errors.Join(errs...)
}
```

Then the call site becomes `c.validateRouting()` instead of `c.Routing.validate(c)`.

test/suites/c2cc/custom-route-tables.robot
```diff
@@ -0,0 +1,132 @@
...
+    [Documentation]    Verify that the old default tables (200/201) still contain routes
+    ...    from before the config change. MicroShift intentionally does not clean up
+    ...    C2CC state on exit - in case of a crash, pods keep running and C2CC
+    ...    connectivity stays online. Changing table IDs produces duplicated rules
+    ...    which are resolved by rebooting the system.
+    ${routes}=    Command On Cluster    cluster-a    ${IP_CMD} route show table 200
+    Should Contain    ${routes}    ${CLUSTER_B_POD_CIDR}
+    ${rules}=    Command On Cluster    cluster-a    ${IP_CMD} rule show
+    Should Contain    ${rules}    lookup 200
+
```
agullon, line 73: I think we should also check here for 201 table
```suggestion
      ${svc_routes}=    Command On Cluster    cluster-a    ${IP_CMD} route show table 201
      Should Contain    ${svc_routes}    ${CLUSTER_A_SVC_CIDR}
      Should Contain    ${rules}    lookup 201

```

test/suites/c2cc/custom-route-tables.robot
```diff
@@ -0,0 +1,132 @@
...
+    ...    rm -f /etc/microshift/config.d/${CUSTOM_ROUTING_DROPIN}.yaml
+
+Flush Custom Tables
+    [Documentation]    Remove routes and rules from custom tables to avoid interfering
+    ...    with subsequent tests that expect only tables 200/201.
+    Disruptive Command On Cluster    cluster-a
+    ...    ${IP_CMD} route flush table ${CUSTOM_ROUTE_TABLE}
+    Disruptive Command On Cluster    cluster-a
+    ...    ${IP_CMD} route flush table ${CUSTOM_SVC_ROUTE_TABLE}
+    ${rules}=    Disruptive Command On Cluster    cluster-a    ${IP_CMD} rule show
```
agullon: This is only a read command, right? Then I think it's safe to drop the `Disruptive`
```suggestion
    ${rules}=    Command On Cluster    cluster-a    ${IP_CMD} rule show
```

test/suites/c2cc/custom-route-tables.robot
```diff
@@ -0,0 +1,132 @@
...
+    Setup Kubeconfig
+    Register Local Cluster    cluster-a
+    Register Remote Cluster    cluster-b    ${HOST2_IP}    ${HOST2_SSH_PORT}    ${KUBECONFIG_B}
+    Apply Custom Routing Config
+    Restart And Wait For Healthy
+
+Teardown
+    [Documentation]    Remove custom routing drop-in, flush custom tables, restore defaults.
+    Remove Custom Routing Config
+    Flush Custom Tables
```
agullon: Should the flush be done after microshift restarts? If it's done before could be there any problem? I'm asking because `Remove Custom Routing Config` only removes the yaml but not change is done in microshift without a restart.

pmtk: Yeah, that's true. Thanks for catching that

PR #6930, opened 2026-06-24
USHIFT-6808: C2CC Disruption Tests (Restarting Services, toggling ethernet)

pmtk:
## Summary by CodeRabbit

* **New Features**
  * Added disruptive cluster-resilience tests that simulate VM, network, and service failures and verify recovery.
  * Added broader cross-cluster checks for connectivity, DNS, and cluster health after disruptions.
  * Added new bootc scenario runners for deterministic disruptive test execution.

* **Bug Fixes**
  * Improved cluster-state validation to handle the full set of clusters dynamically.
  * Strengthened recovery handling by restoring network access and re-establishing cluster connections during teardown.

PR #6943, opened 2026-06-25
USHIFT-6800: Add c2cc reboot tests

vimauro:
## Summary by CodeRabbit

* **Tests**
  * Added a C2CC reboot test suite covering single-cluster, dual, and three-cluster simultaneous reboot scenarios.
  * Improved reboot recovery by capturing per-cluster boot IDs, rebooting all targets concurrently, then reconnecting each cluster and confirming boot ID changes and Greenboot healthcheck completion.
  * Expanded post-reboot validation to include workloads readiness, C2CC connectivity (including source IP preservation), networking/routing and nftables/OVN behaviors, RemoteCluster health/probe timestamps, and cross-cluster DNS checks.
  * Generalized “remote clusters healthy” waiting to cover all configured remote clusters (and broadened disruptive test execution to the full C2CC suite).

### DIFF
--- a/test/resources/c2cc.resource
+++ b/test/resources/c2cc.resource
@@ -414,7 +414,7 @@ DNS Lookup Should Succeed
 
 Verify All RemoteClusters Healthy
     [Documentation]    Wait for all RemoteCluster CRs on all clusters to report Healthy.
-    FOR    ${alias}    IN    cluster-a    cluster-b    cluster-c
+    FOR    ${alias}    IN    @{ALL_CLUSTERS}
         Wait Until Keyword Succeeds    3m    10s
         ...    Verify RemoteCluster State    ${alias}    Healthy
     END
@@ -572,18 +572,6 @@ Reconnect To Cluster
     Deregister Remote Cluster    ${alias}
     Wait Until Keyword Succeeds    ${timeout}    15s
     ...    Register Remote Cluster    ${alias}    ${host}    ${ssh_port}    ${kubeconfig}
-
-Verify Cluster Is Healthy
-    [Documentation]    Verify MicroShift and all core workloads are ready.
-    [Arguments]    ${alias}
-    Command On Cluster    ${alias}    microshift healthcheck --timeout=300s
-
-Verify All RemoteClusters Healthy
-    [Documentation]    Wait for all RemoteCluster CRs on all clusters to report Healthy.
-    FOR    ${alias}    IN    @{ALL_CLUSTERS}
-        Wait Until Keyword Succeeds    3m    10s
-        ...    Verify RemoteCluster State    ${alias}    Healthy
-    END
 
 Reboot Clusters Simultaneously
     [Documentation]    Reboot one or more clusters and wait for all to come back
--- a/test/resources/c2cc.resource
+++ b/test/resources/c2cc.resource
@@ -578,7 +578,7 @@ Reboot Clusters Simultaneously
     ...    with a new boot ID and greenboot health check passed.
     ...    Issues reboot on all targets before waiting, so reboots overlap.
     [Arguments]    @{cluster_aliases}
-    VAR    &{boot_ids}=
+    VAR    &{boot_ids}=    &{EMPTY}
     FOR    ${alias}    IN    @{cluster_aliases}
         ${conn_id}=    Get From Dictionary    ${C2CC_SSH_IDS}    ${alias}
         SSHLibrary.Switch Connection    ${conn_id}
--- a/test/resources/c2cc.resource
+++ b/test/resources/c2cc.resource
@@ -572,3 +572,66 @@ Reconnect To Cluster
     Deregister Remote Cluster    ${alias}
     Wait Until Keyword Succeeds    ${timeout}    15s
     ...    Register Remote Cluster    ${alias}    ${host}    ${ssh_port}    ${kubeconfig}
+
+Verify Cluster Is Healthy
+    [Documentation]    Verify MicroShift and all core workloads are ready.
+    [Arguments]    ${alias}
+    Command On Cluster    ${alias}    microshift healthcheck --timeout=300s
+
+Verify All RemoteClusters Healthy
+    [Documentation]    Wait for all RemoteCluster CRs on all clusters to report Healthy.
+    FOR    ${alias}    IN    @{ALL_CLUSTERS}
+        Wait Until Keyword Succeeds    3m    10s
+        ...    Verify RemoteCluster State    ${alias}    Healthy
+    END
+
+Reboot Clusters Simultaneously
+    [Documentation]    Reboot one or more clusters and wait for all to come back
+    ...    with a new boot ID and greenboot health check passed.
+    [Arguments]    @{cluster_aliases}
+    &{boot_ids}=    Create Dictionary
+    FOR    ${alias}    IN    @{cluster_aliases}
+        ${conn_id}=    Get From Dictionary    ${C2CC_SSH_IDS}    ${alias}
+        SSHLibrary.Switch Connection    ${conn_id}
+        ${bootid}=    Get Current Boot Id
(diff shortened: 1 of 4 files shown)


test/suites/c2cc/reboot.robot
```diff
@@ -0,0 +1,146 @@
...
+...                 clusters simultaneously. After each reboot cycle, full-stack
+...                 verification confirms connectivity, infrastructure, health probes,
+...                 and DNS all recover.
+
+Resource            ../../resources/c2cc.resource
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+Test Tags           c2cc
```
agullon: I think we should add here `disruptive` tag to add these suite to Disruptive runs, same as the ones Patryk created on the other PR, [example](https://github.com/openshift/microshift/pull/6930/changes#diff-560abedf662a1e1e748c3297e9825a1234ee0497ca49a97af1dcb438af1a812dR15)

@pmtk wdyt?

test/suites/c2cc/reboot.robot
```diff
@@ -0,0 +1,146 @@
...
+    Teardown All Remote Clusters
+    Remove Kubeconfig
+
+Wait For Clusters Ready
+    [Documentation]    Wait for test pods and service endpoints to become ready
+    ...    after a reboot cycle.
+    Wait For Test Pods
+    Wait For Service Endpoints
+
+Verify Full C2CC Stack
```
agullon, line 71: @vimauro @pmtk I feel we are rewriting (almost) the same`Verify` keywords to check the cluster are ok on every PR, don't you?

This PR is ok as it is. I prefer to merge it now and later, on a follow up PR, do a refactor to group better these `Verify` checks. What do you think?

test/suites/c2cc/reboot.robot
pmtk: If they can fit into single scenario, then definitely. If not, it should be separate one.

Except I'd replace `c2cc` with `disruptive`. If `c2cc` is present, it would run within regular scenarios.

test/suites/c2cc/reboot.robot
vimauro: I think it makes sense. I've tried to reuse and share code as much as I could on my 2 PRs, but I think it make sense to move out the verification. I am happy to do some refactoring on a separate PR

test/suites/c2cc/reboot.robot
vimauro: Let me add it.
Do you think it would makes sense to add the `disruptive` tag also in the c2cc 10.2 upgrade tests ?

agullon: No, upgrade tests are not disruptive.
Disruptive means that we are creating a fault or failure on purpose, to check for errors reported and wait until good state is recover.
Upgrade is well documented path we expect clients do withput failures/issues.

test/resources/c2cc.resource
```diff
@@ -572,3 +572,55 @@ Reconnect To Cluster
...
+    END
+
+Cluster Should Be Rebooted
+    [Documentation]    Assert that a cluster has rebooted and greenboot health check has passed.
+    [Arguments]    ${alias}    ${old_bootid}
+    Reconnect Cluster    ${alias}
+    ${new_bootid}=    Get Current Boot Id
+    Should Not Be Equal    ${new_bootid}    ${old_bootid}
+    ${stdout}=    Command On Cluster    ${alias}
+    ...    systemctl show -p SubState greenboot-healthcheck.service --value
```
agullon, line 606: When greenboot enters the `failed` SubState (a oneshot that won't self-recover), this retry loop burns the full 10-minute timeout waiting for `exited` that will never come — and the timeout error message doesn't hint at the actual cause.

Consider adding an early-exit check for the `failed` state before the `exited` assertion:

```robot
    Should Not Be Equal As Strings    ${stdout}    failed    strip_spaces=True
    ...    msg=Greenboot healthcheck FAILED on ${alias}. Check greenboot logs.
    Should Be Equal As Strings    ${stdout}    exited    strip_spaces=True
```

This way a greenboot failure is caught immediately with a clear diagnostic instead of timing out after 10 minutes.

test/suites/c2cc/reboot.robot
```diff
@@ -0,0 +1,146 @@
...
+
+Test Tags           disruptive
+
+
+*** Test Cases ***
+Reboot Single Cluster
+    [Documentation]    Reboot cluster-a while cluster-b and cluster-c remain up.
+    ...    Verifies that the rebooted cluster re-establishes C2CC connectivity
+    ...    with both peers.
+    [Setup]    Ensure All Clusters Healthy
```
agullon:
```suggestion
    [Setup]    Verify All RemoteClusters Healthy
```
And remove `Ensure All Clusters Healthy` because for me does not make sense to define a keyword that calls another directly.

test/suites/c2cc/reboot.robot
```diff
@@ -0,0 +1,146 @@
...
+    Reboot Clusters Simultaneously    cluster-a
+    Wait For Clusters Ready
+    Verify Full C2CC Stack
+
+Reboot Two Clusters Simultaneously
+    [Documentation]    Reboot cluster-b and cluster-c at the same time.
+    ...    The surviving cluster-a must wait for both peers to recover.
+    ...    The two rebooted clusters must also reconnect with each other.
+    [Setup]    Ensure All Clusters Healthy
```
agullon:
```suggestion
    [Setup]    Verify All RemoteClusters Healthy
```
And remove `Ensure All Clusters Healthy` because for me does not make sense to define a keyword that calls another directly.

test/suites/c2cc/reboot.robot
```diff
@@ -0,0 +1,146 @@
...
+    Reboot Clusters Simultaneously    cluster-b    cluster-c
+    Wait For Clusters Ready
+    Verify Full C2CC Stack
+
+Reboot All Three Clusters Simultaneously
+    [Documentation]    Reboot all three clusters at once.
+    ...    Every cluster starts from scratch simultaneously — no running peer
+    ...    to reference. All must independently reconstruct C2CC state.
+    [Setup]    Ensure All Clusters Healthy
```
agullon:
```suggestion
    [Setup]    Verify All RemoteClusters Healthy
```
And remove `Ensure All Clusters Healthy` because for me does not make sense to define a keyword that calls another directly.

test/suites/c2cc/reboot.robot
```diff
@@ -0,0 +1,146 @@
...
+    Curl Remote Service Via DNS    cluster-a    cluster-b
+    Curl Remote Service Via DNS    cluster-a    cluster-c
+    Curl Remote Service Via DNS    cluster-b    cluster-a
+    Curl Remote Service Via DNS    cluster-b    cluster-c
+    Curl Remote Service Via DNS    cluster-c    cluster-a
+    Curl Remote Service Via DNS    cluster-c    cluster-b
+
+Ensure All Clusters Healthy
+    [Documentation]    Pre-condition: all clusters must have Healthy RemoteCluster CRs.
+    Verify All RemoteClusters Healthy
```
agullon:
```suggestion
```

test/suites/c2cc/reboot.robot
vimauro: nice catch, it's not needed anymore

agullon:
@vimauro unfortunatelly the tests you added were not exected.
The reason is we are only triggering `disruptive.robot` in https://github.com/openshift/microshift/blob/main/test/scenarios-bootc/c2cc/el102-src%40c2cc-disruptive.sh#L23 and https://github.com/openshift/microshift/blob/main/test/scenarios-bootc/c2cc/el98-src%40c2cc-disruptive.sh#L23

The fix you can add is to remove the `disruptive.robot` from both lines I shared with you.

PR #6944, opened 2026-06-25
USHIFT-6909: Add RF tests for custom DNS config

kasturinarra:
## Summary by CodeRabbit

* **New Features**
  * Added coverage for updating the DNS configuration by atomically replacing the custom Corefile and renaming it into place.
  * Verified that DNS reloads automatically, serves the updated hostname, and reflects the active config path without restarting the cluster.

* **Bug Fixes**
  * Improved cleanup of custom DNS test files by removing both the main config file and any temporary file left behind.

### DIFF
--- a/test/suites/standard2/dns-custom-config.robot
+++ b/test/suites/standard2/dns-custom-config.robot
@@ -68,6 +68,24 @@ Cluster Local Resolution With Custom Corefile
     Resolve Host From Pod    kubernetes.default.svc.cluster.local
     [Teardown]    Teardown Custom Corefile
 
+Atomic File Replacement Via Mv Updates ConfigMap
+    [Documentation]    Replace the custom Corefile atomically using mv
+    ...    (write temp file, then rename) and verify the ConfigMap is updated
+    ...    without restarting MicroShift. Also verifies show-config --mode
+    ...    effective reports the configured dns.configFile path.
+    [Setup]    Setup Custom Corefile With Hosts Entry
+    Resolve Host From Pod    ${HOSTNAME}
+    ${new_hostname}=    Generate Random HostName
+    ${corefile}=    Build Custom Corefile    ${new_hostname}    ${FAKE_LISTEN_IP}
+    Upload String To File    ${corefile}    ${CUSTOM_COREFILE_PATH}.tmp
+    Command Should Work    mv ${CUSTOM_COREFILE_PATH}.tmp ${CUSTOM_COREFILE_PATH}
+    Wait Until Keyword Succeeds    20x    5s
+    ...    ConfigMap Should Contain Hostname    ${new_hostname}
+    Resolve Host From Pod    ${new_hostname}
+    ${config}=    Show Config    effective
+    Should Be Equal As Strings    ${config.dns.configFile}    ${CUSTOM_COREFILE_PATH}
+    [Teardown]    Teardown Custom Corefile
+
 
 *** Keywords ***
 Build Custom Corefile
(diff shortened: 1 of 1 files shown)


test/suites/standard2/dns-custom-config.robot
```diff
@@ -68,6 +70,34 @@ Cluster Local Resolution With Custom Corefile
...
+    ...    without restarting MicroShift.
+    [Setup]    Setup Custom Corefile With Hosts Entry
+    Resolve Host From Pod    ${HOSTNAME}
+    ${new_hostname}=    Generate Random HostName
+    ${corefile}=    Build Custom Corefile    ${new_hostname}    ${FAKE_LISTEN_IP}
+    Upload String To File    ${corefile}    ${CUSTOM_COREFILE_PATH}.tmp
+    Command Should Work    mv ${CUSTOM_COREFILE_PATH}.tmp ${CUSTOM_COREFILE_PATH}
+    Wait Until Keyword Succeeds    20x    5s
+    ...    ConfigMap Should Contain Hostname    ${new_hostname}
+    Resolve Host From Pod    ${new_hostname}
```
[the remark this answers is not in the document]
kasturinarra: Build Custom Corefile generates a complete Corefile from the template with only the new hostname — it's not an append. When the ConfigMap is updated with the new file content, the old
  hostname is structurally absent. If ConfigMap Should Contain Hostname ${new_hostname} passes, it means the file watcher replaced the full ConfigMap content. There's no code path where both
  old and new hostnames could coexist in the ConfigMap.

  Adding a negative assertion (ConfigMap Should Not Contain Hostname ${HOSTNAME}) would be testing that the MicroShift file watcher does full-content replacement rather than append — that's
  a different concern from what the atomic mv test is validating, and it's already implicitly covered by how the template works.

test/suites/standard2/dns-custom-config.robot
```diff
@@ -68,6 +70,34 @@ Cluster Local Resolution With Custom Corefile
...
+    [Teardown]    Teardown Custom Corefile
+
+Multi Zone Custom Corefile Resolves Both Zones
```
pacevedom: Not sure about this one. The feature provides the possibility of swapping out the entire CoreDNS config file, and that is covered [here](https://github.com/openshift/microshift/blob/main/test/suites/standard2/dns-custom-config.robot#L32).
This test is loading a different Corefile and testing it works, but this was already proven by the test mentioned above. It looks like this is more of testing CoreDNS functionality than MicroShift capabilities?

test/suites/standard2/dns-custom-config.robot
```diff
@@ -68,6 +70,34 @@ Cluster Local Resolution With Custom Corefile
...
+Atomic File Replacement Via Mv Updates ConfigMap
+    [Documentation]    Replace the custom Corefile atomically using mv
+    ...    (write temp file, then rename) and verify the ConfigMap is updated
+    ...    without restarting MicroShift.
+    [Setup]    Setup Custom Corefile With Hosts Entry
+    Resolve Host From Pod    ${HOSTNAME}
+    ${new_hostname}=    Generate Random HostName
+    ${corefile}=    Build Custom Corefile    ${new_hostname}    ${FAKE_LISTEN_IP}
+    Upload String To File    ${corefile}    ${CUSTOM_COREFILE_PATH}.tmp
+    Command Should Work    mv ${CUSTOM_COREFILE_PATH}.tmp ${CUSTOM_COREFILE_PATH}
```
pacevedom, line 81: This test covers different semantics from fsnotify, so good addition!

test/suites/standard2/dns-custom-config.robot
kasturinarra: Thanks, removed the test case and moved show-config part to the other test i have.

test/suites/standard2/dns-custom-config.robot
```diff
@@ -124,6 +142,6 @@ Teardown Custom Corefile
 Remove Custom Corefile
     [Documentation]    Remove the custom Corefile from the host
     ${stdout}    ${stderr}    ${rc}=    Execute Command
-    ...    rm -f ${CUSTOM_COREFILE_PATH}
+    ...    rm -f ${CUSTOM_COREFILE_PATH} ${CUSTOM_COREFILE_PATH}.tmp
```
pacevedom, line 145: this might be a candidate for cleaning up by adding an argument to the keyword, but lets move forward for the sake of finishing testing on this one.

pacevedom (approved):
/lgtm

PR #6954, opened 2026-06-26
USHIFT-7187: C2CC Dual Stack support & tests

pmtk:
## Summary by CodeRabbit

* **New Features**
  * Remote cluster routing now accepts multiple `nextHop` entries, enabling separate IPv4/IPv6 gateways for dual-stack clusters.
  * Added dual-stack coverage for connectivity, infrastructure, cleanup, and recovery testing.
* **Bug Fixes**
  * Validation now rejects empty/invalid/duplicate family-specific `nextHop` entries and enforces gateway coverage per CIDR family.
  * Route and rule reconciliation now applies the correct gateway per IP family and skips routes when a family gateway is missing.
  * RemoteCluster identity is now derived from the primary next hop.
* **Documentation**
  * Updated OpenAPI schema and configuration examples to represent `nextHop` as an array.


test/scenarios-bootc/c2cc/el98-src@c2cc-dual-stack-v6.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# Sourced from scenario.sh and uses functions defined there.
+
+# shellcheck source=test/bin/c2cc_common.sh
+source "${SCRIPTDIR}/c2cc_common.sh"
```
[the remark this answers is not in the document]
pmtk: Scenario is not executed directly, but sourced by scenario.sh. All scenarios are doing the same thing.

agullon (commented):
## Review Findings

### 1. `get_host_ipv6()` can return empty string on failure silently

**`test/bin/c2cc_common.sh:35-43`**

Two issues in this function:

**Dead code:** `host_ip` is fetched on line 38 but never used — `run_command_on_vm` takes `${host}` (the hostname), not `${host_ip}`.

**Silent empty return:** If the VM has no global IPv6 address, `grep -oP` fails inside the remote shell, but the exit code is swallowed by the `| tail -1 | tr -d '\r'` pipeline. The function returns exit 0 with an empty string. The `|| return 1` guards at the call sites (lines 134-136, 283-284) never trigger.

The `configure_c2cc_hosts` path is safe (guarded by `-n` checks), but `c2cc_run_tests` would inject `--variable HOST2_IPV6:` (empty) into Robot Framework without error — which could cause confusing test behavior in dual-stack scenarios if the VM hasn't gotten its IPv6 address yet.

Suggested fix:
```bash
get_host_ipv6() {
    local host=$1
    local ipv6
    ipv6=$(run_command_on_vm "${host}" \
        "ip -6 addr show scope global | grep -oP '(?<=inet6 )([0-9a-f:]+)' | head -1" \
        | tail -1 | tr -d '\r')
    if [[ -z "${ipv6}" ]]; then
        echo "failed to get ${host} IPv6 address" >&2
        return 1
    fi
    echo "${ipv6}"
}
```

### 2. OVN and Linux route unit tests are single-stack only

**`pkg/controllers/c2cc/ovn_test.go`** and **`pkg/controllers/c2cc/routes.go`**

All OVN route tests use IPv4-only `NextHops` (`map[int]net.IP{2: net.ParseIP("192.168.1.10")}`). The new family-aware gateway selection logic:

```go
gw, ok := rc.NextHopForFamily(ipFamilyOf(cidr))
if !ok {
    continue
}
```

is untested at the unit level for dual-stack input. A test with both FAMILY_V4 and FAMILY_V6 entries in `NextHops` and dual-stack CIDRs would verify that each route gets the correct family-matched gateway. Similarly, `routes.go` has no direct unit tests for the family-matching logic.

Not blocking — the Robot Framework integration tests cover this end-to-end — but unit-level coverage would catch regressions earlier.

pmtk:
/retest
lol

cmd/generate-config/config/config-openapi-spec.json
```diff
@@ -232,8 +232,11 @@
                 "type": "string"
               },
               "nextHop": {
-                "description": "IP address of the remote cluster's node, used as next-hop for routing.",
-                "type": "string"
+                "description": "IP addresses of the remote cluster's node, used as next-hop for routing.\nAt most one IPv4 and one IPv6 address. Dual-stack clusters need both.",
```
agullon, line 235: If I understand this correctly, `At most one IPv4 and one IPv6 address` means it must be only one or none at all IPv4. Also it must be only one or none at all IPv6. 
I'm wondring that maybe it should be easier for users if we set 2 nextHop type String params instead: `nextHopIPv4` and `nextHopIPv6`.

Either way, should we add a check an report an error when more than one IPv4 or more than one IPv6 is set?

pmtk: I believe we already have it: https://github.com/openshift/microshift/pull/6954/changes#diff-33f015a27c6f16fd92c4faf87ea3650f37bfaeccaded03df44642adce44dad1dR200

test/scenarios-bootc/c2cc/el98-src@c2cc-dual-stack-v6.sh
```diff
@@ -0,0 +1,43 @@
...
+scenario_remove_vms() {
+    c2cc_remove_vms
+}
+
+scenario_run_tests() {
+    # shellcheck disable=SC2119
+    configure_c2cc_hosts
+    # IP_FAMILY=ipv6 so IP_CMD defaults to 'ip -6' for primary (IPv6) CIDRs.
+    # Dual-stack (IPv4) tests derive ip_cmd from CIDR content via IP Command For CIDR.
+    c2cc_run_tests "suites/c2cc/" "192.0.2.0/24" ipv6
```
agullon: why should not set the `foreign_cidr` as IPv6 here?

```suggestion
    c2cc_run_tests "suites/c2cc/" "2001:db8::/64" ipv6
```

test/scenarios-bootc/c2cc/el102-src@c2cc-dual-stack.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
agullon, line 1: why not adding 2 more scenarios?

- el102-src@c2cc-dual-stack-v6.sh
- el98-src@c2cc-dual-stack.sh

pmtk: Because I think it's a waste of resources - it's 6 more VMs but we already have 11 scenarios = 33 VMs.

test/scenarios-bootc/c2cc/el98-src@c2cc-dual-stack-v6.sh
pmtk: Yeah, but used value from the ipv6 scenario (/32 vs /64)

cmd/generate-config/config/config-openapi-spec.json
agullon: yes, you are right

test/scenarios-bootc/c2cc/el102-src@c2cc-dual-stack.sh
agullon: I see, they probably won't fit, it's ok, then. thanks

pkg/config/c2cc.go
```diff
@@ -68,8 +68,9 @@ type C2CC struct {
 }
 
 type RemoteCluster struct {
-	// IP address of the remote cluster's node, used as next-hop for routing.
-	NextHop string `json:"nextHop"`
+	// IP addresses of the remote cluster's node, used as next-hop for routing.
+	// At most one IPv4 and one IPv6 address. Dual-stack clusters need both.
+	NextHop []string `json:"nextHop"`
```
pacevedom, line 73: New slice is not covered by unit/RF tests to check emptiness or repeated ip families in the values.

pacevedom (commented):
What about dual stack support for health probe and DNS? These seem to be fixed to the ip family that comes in first in the configuration. Should we support having both too in these interfaces too? Is there any situation in which ipv4 may fail connectivity but ipv6 is able to do so?

test/bin/ci_phase_boot_and_test.sh
```diff
@@ -85,8 +85,17 @@ else
     progress=""
 fi
 
+# Limit amount of parallel scenarios for the C2CC jobs
+# to avoid over-provisioning the resources.
+jobs_arg=""
+if [[ "${SCENARIO_SOURCES}" =~ c2cc ]]; then
+    jobs_arg="-j ${C2CC_PARALLEL_JOBS:-8}"
```
agullon: why only 8? I think the limit is 30, so probably 10 will be ok.

pkg/config/c2cc.go
```diff
@@ -359,14 +410,16 @@ func validateRemoteCluster(
...
 
-	normalizedNextHop := res.NextHop.String()
-	if res.NextHop.Equal(nodeIP) || (nodeIPv6 != nil && res.NextHop.Equal(nodeIPv6)) {
-		errs = append(errs, fmt.Errorf("%s.nextHop %q must not equal the local node IP (routing loop)", label, normalizedNextHop))
-	}
-	if prev, ok := seenNextHops[normalizedNextHop]; ok {
-		errs = append(errs, fmt.Errorf("%s.nextHop %q duplicates remoteClusters[%d]", label, normalizedNextHop, prev))
-	} else {
-		seenNextHops[normalizedNextHop] = i
+	for _, hop := range res.NextHops {
```
pacevedom, line 412: Nit: if you need to care about error message ordering for tests (for example?), this is iterating over a map which is not ordered by definition, so its unpredictable.

pkg/controllers/c2cc/helpers_test.go
```diff
@@ -4,34 +4,59 @@ import (
...
+
+func parseNextHops(t *testing.T, hops []string) map[int]net.IP {
+	t.Helper()
+	m := make(map[int]net.IP, len(hops))
+	for _, h := range hops {
+		ip := net.ParseIP(h)
+		require.NotNil(t, ip, "invalid nextHop: %s", h)
+		family := 2 // FAMILY_V4
+		if ip.To4() == nil {
+			family = 10 // FAMILY_V6
```
pacevedom: Should we use the constants from netlink? It is in the rest of the code.

test/bin/ci_phase_boot_and_test.sh
pmtk: Right, 10 scenarios * 3 VMs still leaves some resources for the hypervisor

pacevedom (approved):
/lgtm
/hold

PR #6996, opened 2026-07-07
USHIFT-7342: Replace hardcoded sleep with state check before restart

pacevedom:
The Restart MicroShift keyword had an unconditional 10s sleep before every restart to avoid issuing restarts while the service was mid-transition. Replace it with an active check that polls systemd SubState for transitional states, passing instantly when the service is already stable and saving ~10s per restart.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the MicroShift restart flow by removing the fixed wait and instead waiting for the `microshift.service` state to fully settle, clearing any failed state, restarting, verifying both MicroShift and etcd PIDs changed, then refreshing kubeconfig and confirming readiness.
  * Adjusted `Wait For MicroShift Service` to use fewer retries and clarified that the timeout budget covers `TimeoutStartSec=4m`.
* **Tests**
  * Added a new check to ensure `microshift.service` is not transitioning (not activating, deactivating, or reloading) before proceeding with the restart.

### DIFF
--- a/test/resources/microshift-process.resource
+++ b/test/resources/microshift-process.resource
@@ -73,24 +73,74 @@ MicroShift Service Is Active
     Should Be Equal As Strings    ${state}    active
 
 Wait For MicroShift Service
-    [Documentation]    Wait for MicroShift service to be active
-    Wait Until Keyword Succeeds    50x    10s
+    [Documentation]    Wait for MicroShift service to be active.
+    ...    Budget covers TimeoutStartSec=4m (240s) with margin.
+    Wait Until Keyword Succeeds    30x    10s
     ...    MicroShift Service Is Active
 
-Restart MicroShift
-    [Documentation]    Restart the MicroShift service, waiting for any in-progress
-    ...    state transition to settle before issuing the restart.
-    Wait Until Keyword Succeeds    10x    5s
+Restart MicroShift    # robocop: off=LEN01
+    [Documentation]    Restart the MicroShift service safely, handling all systemd
+    ...    states including auto-restart loops after failures.
+    ...
+    ...    Background: microshift.service has Restart=always. After a startup
+    ...    failure (e.g. invalid config), systemd auto-restarts the process
+    ...    repeatedly. Each attempt fails and increments the rate-limit counter.
+    ...    Once StartLimitBurst (default 5) is exceeded within
+    ...    StartLimitIntervalSec (default 10s), systemd triggers start-limit-hit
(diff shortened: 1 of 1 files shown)


test/resources/microshift-process.resource
```diff
@@ -78,18 +78,31 @@ Wait For MicroShift Service
     ...    MicroShift Service Is Active
 
 Restart MicroShift
-    [Documentation]    Restart the MicroShift service with a delay to avoid too
-    ...    many start requests.
+    [Documentation]    Restart the MicroShift service, waiting for any in-progress
+    ...    state transition to settle before issuing the restart.
+    Wait Until Keyword Succeeds    10x    5s
```
Neilhamza: could 50s be too short? elsewhere in this same file "wait for MicroShift Service" allows up to 500

test/resources/microshift-process.resource
```diff
@@ -78,18 +78,31 @@ Wait For MicroShift Service
...
+    ...    MicroShift Service Is Not Transitioning
+    # Clear start-limit-hit so systemctl restart is not rejected after
+    # rapid failures (e.g. invalid config that was since removed).
+    Execute Command    systemctl reset-failed microshift.service    sudo=True
```
Neilhamza, line 134: now that reset-failed is introduced i wonder if transition check (wait) is redundant now

Neilhamza: we could also have a reutrn code check

Neilhamza (approved):

test/resources/microshift-process.resource
pacevedom: Bumped the timeout to 250s, as there is a 4m timeout in the service itself.

test/resources/microshift-process.resource
pacevedom: - reset-failed clears start-limit-hit so systemd accepts the restart command
- The transition check ensures a stable PID for correct restart detection

Without the transition check, if the service is activating and the PID changes mid-activation (crash + auto-restart), we'd capture a stale PID and the PID-change wait could pass without confirming our restart happened. reset-failed doesn't help with that.

pacevedom: Already covered. systemd.resource shows the Systemctl keyword checks RC and fails on non-zero:
```robot
  IF    ${rc} != 0    Systemctl Print Service Status And Logs    ${unit_name}
```
So it prints diagnostic logs and then asserts rc == 0.

PR #7024, opened 2026-07-09
USHIFT-7382: Add MTU boundary tests for C2CC and C2CC+IPsec

agullon:
## Summary

Add Robot Framework tests to exercise different MTU sizes for C2CC (Cluster-to-Cluster Connectivity) with and without IPsec, covering both IPv4 and IPv6.

### What this PR adds

**2 test suites** under `test/suites/c2cc/extra/`:

| Suite | Description | Test count |
|-------|-------------|-----------|
| `ipsec.robot` | IPsec E2E + DF-bit boundary tests at 1500 MTU | 10 |
| `mtu.robot` | Jumbo MTU (9000) for plain C2CC and C2CC+IPsec | 1 |

**12 new scenario scripts** (el98 + el102 × IPv4 + IPv6 × 3 suites):
- `c2cc-ipsec-ipv4/ipv6` — IPsec functional + DF-bit tests at 1500 MTU
- `c2cc-mtu-ipv4/ipv6` — jumbo frame (9000) C2CC without IPsec
- `c2cc-ipsec-mtu-ipv4/ipv6` — jumbo frame C2CC with IPsec (same `mtu.robot` with `IPSEC=true`)

### Test approach

Tests use `df-bit-send.py`, a Python3 script that sends UDP datagrams with the DF bit set using `IP_PMTUDISC_DO` (IPv4) / `IPV6_DONTFRAG` (IPv6). The script is piped via stdin to `nettest-pod` through `oc exec`. This validates MTU enforcement at the pod interface level without requiring `NET_RAW` capability — the `nettest-pod` runs as non-root under restricted PodSecurity.

The "should fail" assertions check for the specific `EMSGSIZE` error to distinguish genuine MTU rejection from infrastructure failures. Curl failure assertions also verify recognizable error modes (timeout/refused/unreachable).

All DF-bit tests exercise all 6 cluster pair directions (a→b, a→c, b→a, b→c, c→a, c→b). Boundary values are chosen to reject on both IPv4 (28B header overhead) and IPv6 (48B), so each test works correctly on both single-stack scenario variants.

### Unified jumbo MTU suite

`mtu.robot` is shared by both plain C2CC and IPsec scenarios via the `${IPSEC}` variable:
- **Without IPsec** (`c2cc-mtu` scenarios): validates DF-bit boundaries, large TCP transfers, and full C2CC connectivity
- **With IPsec** (`c2cc-ipsec-mtu` scenarios, `--variable IPSEC:true`): same tests plus ESP encapsulation verification

Pod MTU 9000 is set via kickstart `ovn.yaml` — OVN-K bakes the MTU into its database at initial creation and does not update it on restart, so the value must be present before MicroShift first starts.

### Jumbo MTU infrastructure

Setting jumbo MTU on bootc VMs required solving several layered problems:

1. **NIC MTU**: Three mechanisms set the guest NIC to 9000 before MicroShift starts:
   - NetworkManager `conf.d` (default ethernet MTU for auto-created connections)
   - NM dispatcher script (`ip link set` during connection activation, failures logged via `logger`)
   - systemd oneshot service (final safety net, script at `/etc/` because only `/etc/` and `/var/` persist across reboots on bootc)

2. **OVN pod MTU**: Set via kickstart `ovn.yaml` with MTU 9000 (= NIC MTU). The Geneve tunnel is local in single-node MicroShift, so no headroom subtraction is needed. C2CC VMs have no default route, so MicroShift's auto-detection falls back to 1500 without explicit configuration.

3. **Guest virtio-net MTU**: `launch_vm --network_mtu` propagates the MTU to QEMU's virtio-net driver via virt-install's `mtu.size` sub-option. A libvirt network's `<mtu>` element alone only affects the host-side bridge and tap devices.

4. **Jumbo libvirt networks**: `jumbo` (IPv4) and `jumbo-ipv6` (IPv6) networks with MTU 9000, created in `manage_hypervisor_config.sh` following the existing convention for network infrastructure.

5. **IPv6 setup**: `c2cc_setup_ipv6()` helper deduplicates CIDRs and mirror registry configuration across all IPv6 scenarios.

### MTU boundary values

| Scenario | Pod MTU | Pass boundary | Reject boundary |
|----------|---------|--------------|-----------------|
| IPsec at 1500 | ESP-adjusted PMTU | 1450B | 1472B |
| Jumbo (plain + IPsec) | 9000 (= NIC MTU) | 8952B (MTU-48) | 8973B (MTU-27) |

### CI infrastructure changes

- **Architecture-based scenario split**: Each arch runs one RHEL version to halve the scenario count (~12 per job instead of 23). The assignment rotates per commit (first hex digit of HEAD) so both combinations get exercised over time.

### Also included

- Moved `ipsec.robot` from `suites/c2cc/ipsec/` to `suites/c2cc/extra/`
- Renamed `c2cc-ipsec` → `c2cc-ipsec-ipv4` for consistency with new IPv6 variants
- IPv6 fixes in shared keywords: nftables `ip6 daddr`, curl bracket notation
- `Verify ESP Encapsulation On All Clusters` keyword shared between `mtu.robot` and `ipsec.robot`
- MTU-related keywords (DF-bit, pod MTU verification, large payload) in `c2cc.resource`

## Test plan

- [x] All 12 new scenarios pass on x86 CI (`e2e-aws-tests-bootc-c2cc`)
- [x] All 12 new scenarios pass on ARM CI (`e2e-aws-tests-bootc-c2cc-arm`)
- [x] Existing IPsec tests (10/10) pass on both architectures
- [x] IPv4 and IPv6 variants produce identical results
- [x] `verify-rf` (robocop check + format) passes
- [x] `verify-shell` (shellcheck) passes

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/test/assets/c2cc/df-bit-send.py
+++ b/test/assets/c2cc/df-bit-send.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+"""Send a UDP datagram with the DF (Don't Fragment) bit set.
+
+Uses IP_PMTUDISC_DO (IPv4) or IPV6_DONTFRAG (IPv6) so the kernel
+rejects datagrams that exceed the path MTU with EMSGSIZE instead of
+fragmenting them.  Auto-detects the address family from the
+destination address.
+
+Uses UDP instead of ICMP to avoid requiring NET_RAW capability.
+UDP overhead (IP+UDP) matches ICMP overhead (IP+ICMP): 28B for IPv4,
+48B for IPv6, so payload sizes are equivalent to ping -s values
+within each address family.
+
+Usage: python3 df-bit-send.py <dest_ip> <payload_size>
+Exit:  prints "OK" on success, raises OSError on rejection.
+"""
+
+import socket
+import sys
+
+IPPROTO_IPV6 = 41
+IPV6_DONTFRAG = 62
+IP_MTU_DISCOVER = 10
+IP_PMTUDISC_DO = 2
+
+dest = sys.argv[1]
+size = int(sys.argv[2])
+
+family = socket.AF_INET6 if ":" in dest else socket.AF_INET
+sock = socket.socket(family, socket.SOCK_DGRAM)
+
+if family == socket.AF_INET6:
+    sock.setsockopt(IPPROTO_IPV6, IPV6_DONTFRAG, 1)
+else:
+    sock.setsockopt(socket.IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_DO)
+
+try:
+    sock.sendto(b"A" * size, (dest, 9999))
+    print("OK")
+except OSError as e:
+    print(f"FAIL: {e}")
+finally:
+    sock.close()
--- a/test/bin/c2cc_common.sh
+++ b/test/bin/c2cc_common.sh
@@ -27,6 +27,36 @@ CLUSTER_C_SVC_CIDR_DUAL=""
 
 export TEST_RANDOMIZATION=suites
 
+# c2cc_setup_ipv6 overrides CIDRs and mirror registry for single-stack IPv6
+# scenarios. If a network name is provided, also sets VM_BRIDGE_IP and
+# WEB_SERVER_URL for that network (standard IPv6 scenarios). Jumbo IPv6
+# scenarios omit the network arg and set VM_BRIDGE_IP later, after creating
+# the jumbo-ipv6 network.
+# shellcheck disable=SC2120
+c2cc_setup_ipv6() {
+    local -r network="${1:-}"
+
+    # shellcheck disable=SC2034  # used by scenario.sh and kickstart templates
+    MIRROR_REGISTRY_URL="$(hostname):${MIRROR_REGISTRY_PORT}/microshift"
+
+    CLUSTER_A_POD_CIDR="fd01::/48"
+    CLUSTER_A_SVC_CIDR="fd02::/112"
+    CLUSTER_A_DOMAIN="cluster-a.remote"
+    CLUSTER_B_POD_CIDR="fd04::/48"
+    CLUSTER_B_SVC_CIDR="fd05::/112"
+    CLUSTER_B_DOMAIN="cluster-b.remote"
+    CLUSTER_C_POD_CIDR="fd07::/48"
+    CLUSTER_C_SVC_CIDR="fd08::/112"
+    CLUSTER_C_DOMAIN="cluster-c.remote"
+
+    if [[ -n "${network}" ]]; then
+        # shellcheck disable=SC2034  # used by scenario.sh
+        VM_BRIDGE_IP="$(get_vm_bridge_ip "${network}")"
+        # shellcheck disable=SC2034  # used by scenario.sh
+        WEB_SERVER_URL="http://[${VM_BRIDGE_IP}]:${WEB_SERVER_PORT}"
+    fi
+}
+
 get_host_ip() {
     local host=$1
     get_vm_property "${host}" ip || { echo "failed to get ${host} ip" >&2; return 1; }
@@ -188,11 +218,77 @@ ${network_yaml}IEOF
 EOF
 }
 
+# inject_kickstart_mtu configures jumbo MTU on the guest VM via kickstart.
+# The NIC and pod MTU must be set before MicroShift first starts — OVN-K
+# bakes the MTU into its database at initial creation and does not update
+# it on restart. Three mechanisms set the NIC MTU (NM conf.d, NM dispatcher,
+# systemd oneshot) and ovn.yaml sets the pod MTU.
+# Args: host mtu
+inject_kickstart_mtu() {
+    local -r host="${1}"
+    local -r mtu="${2}"
+
+    local -r ks_dir="${SCENARIO_INFO_DIR}/${SCENARIO}/vms/${host}"
+    cat >> "${ks_dir}/post-microshift.cfg" <<EOF
+# NM conf.d -- default ethernet MTU for auto-created connections
+mkdir -p /etc/NetworkManager/conf.d
+cat > /etc/NetworkManager/conf.d/99-jumbo-mtu.conf <<'NMCONF'
+[connection-ethernet-jumbo]
+match-device=type:ethernet
+ethernet.mtu=${mtu}
+NMCONF
+
+# NM dispatcher -- runs ip link set during connection activation
+mkdir -p /etc/NetworkManager/dispatcher.d
+cat > /etc/NetworkManager/dispatcher.d/99-jumbo-mtu <<'DISPEOF'
+#!/bin/bash
+[ "\$2" != "up" ] && exit 0
+ip link set dev "\$1" mtu ${mtu} 2>&1 || logger -t jumbo-mtu "Failed to set MTU ${mtu} on \$1"
+DISPEOF
+chmod 755 /etc/NetworkManager/dispatcher.d/99-jumbo-mtu
+restorecon -R /etc/NetworkManager/dispatcher.d/ 2>&1 || logger -t jumbo-mtu "restorecon failed for dispatcher.d"
+
+# Systemd service -- runs before microshift.service.
+# Script in /etc/ because only /etc/ and /var/ persist across reboots on bootc.
+cat > /etc/set-jumbo-mtu.sh <<'SCRIPTEOF'
+#!/bin/bash
+for dev in \$(ip -o link show type ether | awk -F: '{print \$2}' | tr -d ' '); do
+    ip link set dev "\$dev" mtu ${mtu}
+done
+SCRIPTEOF
+chmod 755 /etc/set-jumbo-mtu.sh
+restorecon /etc/set-jumbo-mtu.sh 2>&1 || logger -t jumbo-mtu "restorecon failed for set-jumbo-mtu.sh"
+
+cat > /etc/systemd/system/set-jumbo-mtu.service <<'SVCEOF'
+[Unit]
+Description=Set jumbo frame MTU on ethernet interfaces
+Before=microshift.service
+After=NetworkManager-wait-online.service
+
+[Service]
+Type=oneshot
+ExecStart=/etc/set-jumbo-mtu.sh
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
+SVCEOF
+mkdir -p /etc/systemd/system/multi-user.target.wants
+ln -sf /etc/systemd/system/set-jumbo-mtu.service /etc/systemd/system/multi-user.target.wants/set-jumbo-mtu.service
+
+# Set pod MTU to match NIC MTU. MicroShift auto-detection fails in C2CC
+# VMs (no default route), falling back to 1500 regardless of NIC MTU.
+mkdir -p /etc/microshift
+echo "mtu: ${mtu}" > /etc/microshift/ovn.yaml
+EOF
+}
+
 c2cc_create_vms() {
     local -r boot_commit_ref="${1}"
     local -r boot_blueprint="${2}"
     local -r network="${3:-default}"
     local -r ip_family="${4:-ipv4}"
+    local -r network_mtu="${5:-}"
 
     # Prepare kickstart for all hosts
     # prepare_kickstart args: vmname template commit_ref fips_enabled ipv6_only
@@ -223,9 +319,19 @@ c2cc_create_vms() {
         "${CLUSTER_C_POD_CIDR}" "${CLUSTER_C_SVC_CIDR}" \
         "${CLUSTER_C_POD_CIDR_DUAL}" "${CLUSTER_C_SVC_CIDR_DUAL}"
 
-    launch_vm "${boot_blueprint}" --vmname host1 --network "${network}"
-    launch_vm "${boot_blueprint}" --vmname host2 --network "${network}"
-    launch_vm "${boot_blueprint}" --vmname host3 --network "${network}"
+    # Set the NIC MTU via NetworkManager in the kickstart so the guest boots
+    # with the correct MTU before MicroShift starts.
+    local mtu_args=()
+    if [ -n "${network_mtu}" ]; then
+        mtu_args=(--network_mtu "${network_mtu}")
+        for host in host1 host2 host3; do
+            inject_kickstart_mtu "${host}" "${network_mtu}"
+        done
+    fi
+
+    launch_vm "${boot_blueprint}" --vmname host1 --network "${network}" "${mtu_args[@]}"
+    launch_vm "${boot_blueprint}" --vmname host2 --network "${network}" "${mtu_args[@]}"
+    launch_vm "${boot_blueprint}" --vmname host3 --network "${network}" "${mtu_args[@]}"
 }
 
 c2cc_remove_vms() {
@@ -238,6 +344,7 @@ c2cc_run_tests() {
     local -r suites_dir="${1}"
     local -r foreign_cidr="${2:-}"
     local -r ip_family="${3:-}"
+    local -r extra_vars="${4:-}"
 
     local foreign_cidr_var=""
     if [ -n "${foreign_cidr}" ]; then
@@ -317,6 +424,7 @@ c2cc_run_tests() {
         ${ip_family_var} \
         ${target_ref_var} \
         ${dual_cidr_vars} \
+        ${extra_vars} \
         --variable "BOOTC_REGISTRY:${MIRROR_REGISTRY_URL}" \
         "${suites_dir}"
 }
--- a/test/bin/ci_phase_boot_and_test.sh
+++ b/test/bin/ci_phase_boot_and_test.sh
@@ -85,16 +85,30 @@ else
     progress=""
 fi
 
-# Limit amount of parallel scenarios for the C2CC jobs to avoid
-# over-provisioning the resources: each C2CC scenario runs 3 VMs
-# (2 vCPUs / 4GiB each) and the c7g.metal instance has 64 cores / 128GiB.
-# Powering off the VMs of passed scenarios makes the job limit an
-# actual cap on the number of running VMs.
+# Each C2CC scenario runs 3 VMs (2 vCPUs / 4GiB each)
+# and the c7g.metal instance has 64 cores / 128GiB.
 jobs_arg=""
 scenario_action="create-and-run"
 if [[ "${SCENARIO_SOURCES}" =~ c2cc ]]; then
+    # Limit amount of parallel scenarios for the C2CC jobs
+    # to avoid over-provisioning the resources.
     jobs_arg="-j 8"
+
+    # Power off passed VMs so the job limit is an actual cap on running VMs.
     scenario_action="create-run-shutdown"
+
+    # Each arch runs one RHEL version to halve the scenario count.
+    # The assignment rotates per commit so both combos get coverage:
+    #   flip=0: x86 → el98,  ARM → el102
+    #   flip=1: x86 → el102, ARM → el98
+    flip=$(( 16#$(git rev-parse HEAD | cut -c1) % 2 ))
+    if [[ "$(uname -m)" == "x86_64" ]]; then
+        delete_ver=(el102 el98)
+    else
+        delete_ver=(el98 el102)
+    fi
+    echo "C2CC arch split: deleting ${delete_ver[flip]}-* (flip=${flip})"
+    find "${SCENARIOS_TO_RUN}" -name "${delete_ver[flip]}-*.sh" -delete
 elif [[ "${SCENARIO_SOURCES}" =~ .*releases.* ]]; then
     # Release scenarios have grown (e.g. the router scenarios were split
     # from 1 into 5) and no longer fit if every scenario's VMs stay up for
--- a/test/bin/scenario.sh
+++ b/test/bin/scenario.sh
@@ -804,6 +804,7 @@ EOF
 #           <boot_blueprint> \
 #           [--vmname <name>] \
 #           [--network <name>[,<name>...]] \
+#           [--network_mtu <size>] \
 #           [--vm_vcpus <vcpus>] \
 #           [--vm_memory <memory>] \
 #           [--vm_disksize <disksize>] \
@@ -820,6 +821,14 @@ EOF
 #   [--network <name>[,<name>...]]: A comma-separated list for the networks used
 #                                   when creating the VM. Each network entry will
 #                                   create a NIC and they are repeatable.
+#   [--network_mtu <size>]:         Sets the guest-visible MTU on every NIC via
+#                                   virt-install's mtu.size sub-option. Required
+#                                   in addition to a libvirt network's own <mtu>
+#                                   element — QEMU only negotiates a larger MTU
+#                                   with the guest's virtio-net driver when the
+#                                   domain's own <interface> XML requests it;
+#                                   the network-level MTU alone only affects the
+#                                   host-side bridge and tap devices.
 #   [--no_network]:                 Do not configure any network attachments (and
 #                                   therefore no NICs) for the VM.
 #   [--vm_vcpus <vcpus>]:           Number of vCPUs for the VM.
@@ -830,6 +839,7 @@ launch_vm() {
     # Set default values for the optional arguments
     local vmname="host1"
     local network="default"
+    local network_mtu=""
     local vm_memory=4096
     local vm_vcpus=2
     local vm_disksize=20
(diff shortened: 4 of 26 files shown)


test/suites/c2cc/ipsec/bigmtu.robot
```diff
@@ -0,0 +1,93 @@
...
+    ...    Skipped when pod MTU is 8900 (kernel caps sendto at pod MTU).
+    Skip If    '${EXPECTED_POD_MTU}' == '8900'    Pod MTU 8900 caps below this size
+    ${ip}=    Get Hello Pod IP    cluster-b
+    Ping With DF Bit Should Fail    cluster-a    ${ip}    8972
+
+Jumbo MTU Large TCP Transfer
+    [Documentation]    64KB curl POST through IPsec at jumbo MTU.
+    ...    TCP PMTUD handles segmentation at jumbo frame sizes.
+    ${ip}=    Get Hello Pod IP    cluster-b
+    Send Large Payload And Verify    cluster-a    ${ip}    65536
```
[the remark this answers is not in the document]
agullon: Applied — moved `Send Large Payload And Verify` from `ipsec.robot` local keywords to `ipsec.resource` so both suites can resolve it. Fixed in 0211fa48.

test/resources/ipsec.resource
```diff
@@ -126,3 +126,34 @@ Curl From Host Should Fail
...
+Ping With DF Bit Should Fail
+    [Documentation]    Send a UDP datagram with DF bit set. Asserts EMSGSIZE (Message too long).
+    ...    Proves the datagram exceeds the ESP-adjusted PMTU.
+    [Arguments]    ${alias}    ${dest_ip}    ${payload_size}
+    ${stdout}=    Oc On Cluster    ${alias}
+    ...    oc exec nettest-pod -n ${NAMESPACES}[${alias}] -- python3 -c 'import socket,sys;s=socket.socket(socket.AF_INET,socket.SOCK_DGRAM);s.setsockopt(socket.IPPROTO_IP,10,2);s.sendto(b"A"*${payload_size},(sys.argv[1],9999));print("OK")' ${dest_ip} 2>&1 || true
+    ...    allow_fail=${TRUE}
+    Should Not Contain    ${stdout}    OK
+    ...    DF-bit UDP send of ${payload_size}B should have been rejected but succeeded
+
```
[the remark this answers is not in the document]
agullon: Won't fix — the XFRM-adjusted PMTU is computed synchronously at route resolution time from the SA's crypto parameters and the outgoing interface MTU. In tunnel mode on a direct L2 path (libvirt bridge, no intermediate routers), the kernel knows the effective MTU before the first sendto(). There is no async PMTU discovery involved — EMSGSIZE is returned on the first call for oversized packets. A delivery check or retry would add complexity without addressing a real risk in this environment.

agullon:
Addressed CodeRabbit nitpick findings:

- `c2cc_common.sh:363-383` (unchecked `run_command_on_vm` returns): Won't fix — script runs with `set -e`, so `run_command_on_vm` failure already aborts. Same pattern as `configure_c2cc_host` which doesn't use `|| return 1` either.
- `nettest-pod.yaml:8-19` (missing `readOnlyRootFilesystem`, resource limits, `automountServiceAccountToken`): Won't fix — neither `curl-pod.yaml` nor `hello-microshift.yaml` uses any of these. Following the existing C2CC test pod pattern.
- `ipsec.resource:143-151` (docstring claims EMSGSIZE but only checks absence of "OK"): Won't fix — the "should pass" tests run first, so if Python/oc is broken, those fail before the "should fail" tests are reached. A false positive from an unrelated failure is caught upstream.

test/assets/c2cc/df-bit-send.py
```diff
@@ -0,0 +1,38 @@
...
+family = socket.AF_INET6 if ":" in dest else socket.AF_INET
+sock = socket.socket(family, socket.SOCK_DGRAM)
+
+if family == socket.AF_INET6:
+    sock.setsockopt(IPPROTO_IPV6, IPV6_DONTFRAG, 1)
+else:
+    sock.setsockopt(socket.IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_DO)
+
+sock.sendto(b"A" * size, (dest, 9999))
+print("OK")
```
pmtk: If the `sendto` fails, it won't reach OK and we'll get an exception?
Should we also close the socket? Just for correctness :)

test/suites/c2cc/extra/ipsec.robot
```diff
@@ -0,0 +1,178 @@
+*** Settings ***
```
pmtk, line 1: We still have the `test/suites/c2cc/ipsec/ipsec.robot` - should we delete the old file? I don't see it being used anywhere

test/bin/ci_phase_boot_and_test.sh
```diff
@@ -85,16 +85,41 @@ else
...
+    else
+        if [[ "${flip}" -eq 0 ]]; then
+            delete_ver=el98
+        else
+            delete_ver=el102
+        fi
+    fi
+
+    echo "C2CC arch split: deleting ${delete_ver}-* (flip=${flip})"
+    find "${SCENARIOS_TO_RUN}" -name "${delete_ver}-*.sh" -delete
```
pmtk: I must say, this is pretty neat idea

test/bin/c2cc_common.sh
```diff
@@ -188,11 +188,84 @@ ${network_yaml}IEOF
...
+cat > /etc/NetworkManager/dispatcher.d/99-jumbo-mtu <<'DISPEOF'
+#!/bin/bash
+[ "\$2" != "up" ] && exit 0
+ip link set dev "\$1" mtu ${mtu} 2>&1 || logger -t jumbo-mtu "Failed to set MTU ${mtu} on \$1"
+DISPEOF
+chmod 755 /etc/NetworkManager/dispatcher.d/99-jumbo-mtu
+restorecon -R /etc/NetworkManager/dispatcher.d/ 2>&1 || logger -t jumbo-mtu "restorecon failed for dispatcher.d"
+
+# Systemd service -- runs before microshift.service.
+# Script in /etc/ because /usr/ is read-only on bootc images.
```
pmtk: But is the system readonly during installation? I think it might not be yet, only after rebooting

test/bin/scenario.sh
```diff
@@ -1001,6 +1017,7 @@ launch_vm() {
         # to attach to the console. `unbuffer` provides the TTY.
         # shellcheck disable=SC2086
         if ${timeout_install} unbuffer sudo virt-install \
+            --check disk_size=off \
```
pmtk: Do we really need that already?

test/suites/c2cc/extra/mtu.robot
```diff
@@ -0,0 +1,53 @@
...
+    Ping DF Bit Between All Clusters    8400
+
+DF Bit 8872B Passes At Jumbo MTU
+    [Documentation]    8872B + 28B = 8900B, within the pod MTU (8922, set via ovn.yaml).
+    Ping DF Bit Between All Clusters    8872
+
+DF Bit 8850B Passes At Jumbo MTU
+    [Documentation]    Near pod MTU (8922, NIC MTU 9000 minus 78B headroom, set via
+    ...    ovn.yaml) — accepted on all cluster pairs.
+    Ping DF Bit Between All Clusters    8850
```
pmtk: These tests take 2 seconds each, but are they necessary? To me, it's either within the limit (under the max value), or it's not - testing three cases below the max is redundant I think.

test/suites/c2cc/extra/mtu.robot
```diff
@@ -0,0 +1,53 @@
...
+
+DF Bit 8895B Rejected Above Pod MTU
+    [Documentation]    8895B + 28B (IPv4) = 8923B > 8922B pod MTU. Rejected.
+    ...    Also 8895B + 48B (IPv6) = 8943B > 8922B.
+    Ping DF Bit Should Fail Between All Clusters    8895
+
+DF Bit 8972B Rejected Well Above Pod MTU
+    [Documentation]    8972B + 28B = 9000B, well above the pod MTU (8922B).
+    Ping DF Bit Should Fail Between All Clusters    8972
```
pmtk: Similar comment. I think "Above" and "Well Above" are redundant from "mathematical" point of view.
These things are well into production for many years, so I don't think we need to double check linux networking :)

test/suites/c2cc/extra/ipsec-mtu.robot
```diff
@@ -0,0 +1,140 @@
...
+    Verify XFRM Counters Incremented    cluster-b    ${baseline_b}
+    Verify XFRM Counters Incremented    cluster-c    ${baseline_c}
+
+# ── Phase 2 setup: restart validation ─────────────────────────────────────────
+
+Reconfigure Pod MTU To 8900
+    [Documentation]    Re-write pod MTU 8900 via ovn.yaml (same value as kickstart),
+    ...    restart MicroShift, and redeploy workloads. Validates the config
+    ...    survives a restart cycle with IPsec tunnel reestablishment.
+    Configure Pod MTU On All Clusters    8900
```
pmtk: If the `ovn.yaml` already had `mtu: 8900`, what does writing it again to the file gives us? It's the same value, nothing changes, so it's no-op, you could just restart microshift.
> Validates the config survives a restart cycle

I don't understand - the ovn.yaml file is not consumend on boot, so if you restart microshift or powercycle the host, it won't change and it will still contain the `mtu: 8900` set via kickstart.
Since it's the same value, I think ovnk won't reconfigure anything so I'm assuming the IPSec connection isn't dropped at all.
If we'd want to exercise that scenario (Pod MTU reconfiguration -> IPSec works) then maybe it we should change it to e.g. 8899?

test/suites/c2cc/extra/ipsec-mtu.robot
```diff
@@ -0,0 +1,140 @@
...
+    END
+
+Restart MicroShift On All Clusters
+    [Documentation]    Restart MicroShift on all clusters and wait for health + IPsec.
+    FOR    ${alias}    IN    @{ALL_CLUSTERS}
+        Command On Cluster    ${alias}    systemctl restart microshift
+    END
+    FOR    ${alias}    IN    @{ALL_CLUSTERS}
+        Wait Until Keyword Succeeds    5m    15s
+        ...    Command On Cluster    ${alias}    microshift healthcheck --timeout=30s
```
pmtk: I think that 30s timeout proves that "reconfiguration" from 8900 to 8900 does nothing. If the Pods' interfaces were to be reconfigured, the Pods would be restarted and 30s is very optimistic that all Pods will be ready in that time.

test/bin/ci_phase_boot_and_test.sh
agullon: thanks, we used this approach in Release Testing in Jenkins.
I'll try to add this idea to my shiftweek proposal refactor because most of the times arch is not relevant at all.

test/bin/scenario.sh
agullon: I'll drop it, this is not needed with current approach. I forgot to remove it.
This tells `virt-install` to not check if is enough space in the hypervisor for a new VM. I added it because there was a limit in hypervisor disk with more than 55 VMs. But we are not creating that many VMs anymore.

test/suites/c2cc/extra/mtu.robot
agullon: I agree, I'll remove all but `8872B` test.

test/suites/c2cc/extra/ipsec.robot
agullon: I agree, I'll remove it.

test/assets/c2cc/df-bit-send.py
agullon: I agree, I'll add a try:, except: and finally: blocks.

test/suites/c2cc/extra/mtu.robot
agullon: I agree, I'll remove 8972B test.

test/suites/c2cc/extra/ipsec-mtu.robot
agullon: yep, it was a mistake, I refactored all this robot file

test/suites/c2cc/extra/ipsec-mtu.robot
agullon: noted

test/bin/c2cc_common.sh
agullon: comment updated

test/assets/c2cc/df-bit-send.py
```diff
@@ -0,0 +1,44 @@
...
+    sock.setsockopt(socket.IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_DO)
+
+try:
+    sock.sendto(b"A" * size, (dest, 9999))
+    print("OK")
+except OSError as e:
+    print(f"FAIL: {e}")
+    sys.exit(1)
+finally:
+    sock.close()
```
pmtk: Do you think `finally` will execute or `sys.exit` will short circut the whole app and return early?

test/resources/ipsec.resource
```diff
@@ -115,17 +119,155 @@ Remove NFTables IPsec Enforcement Rules
...
+    FOR    ${alias}    IN    @{ALL_CLUSTERS}
+        ${mtu}=    Get Pod Interface MTU    ${alias}    nettest-pod    ${NAMESPACES}[${alias}]
+        Should Be Equal As Strings    ${mtu}    ${expected_mtu}
+        ...    ${alias}: pod MTU is ${mtu}, expected ${expected_mtu}
+    END
+
+Redeploy Test Workloads
+    [Documentation]    Delete and redeploy test workloads so pods pick up the new MTU.
+    Cleanup Test Workloads
+    Deploy Test Workloads
```
pmtk: Aren't these keywords in general more related to C2CC than ipsec?

test/suites/c2cc/extra/mtu.robot
```diff
@@ -0,0 +1,34 @@
...
+...                 Configures pod MTU to 9000 (matching NIC MTU), then validates
+...                 DF-bit acceptance and rejection at the boundary, large TCP
+...                 transfers, and full cross-cluster connectivity.
+...                 Requires the c2cc-mtu scenario.
+
+Resource            ../../../resources/microshift-process.resource
+Resource            ../../../resources/kubeconfig.resource
+Resource            ../../../resources/oc.resource
+Resource            ../../../resources/c2cc.resource
+Resource            ../../../resources/ipsec.resource
```
pmtk, line 15: Other side of having MTU related keywords in ipsec.resource

test/scenarios-bootc/c2cc/el102-src@c2cc-ipsec-mtu-ipv6.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Cluster B (host2): non-overlapping CIDRs
+CLUSTER_B_POD_CIDR="fd04::/48"
+CLUSTER_B_SVC_CIDR="fd05::/112"
+CLUSTER_B_DOMAIN="cluster-b.remote"
+
+# Cluster C (host3): non-overlapping CIDRs
+CLUSTER_C_POD_CIDR="fd07::/48"
+CLUSTER_C_SVC_CIDR="fd08::/112"
+CLUSTER_C_DOMAIN="cluster-c.remote"
```
pmtk: I'm wondering if we should move these variables to c2cc_common.sh and somehow "enable" them for ipv6 scenarios, because now they're duplicated in each ipv6 scenario

test/bin/c2cc_common.sh
```diff
@@ -223,9 +284,58 @@ c2cc_create_vms() {
...
+    bridge=$(sudo virsh net-info jumbo-ipv6 | grep '^Bridge:' | awk '{print $2}')
+    if [[ -z "${bridge}" ]]; then
+        echo "ERROR: Could not determine bridge name for jumbo-ipv6 network" >&2
+        return 1
+    fi
+    local ip
+    for ip in $(ip addr show "${bridge}" | grep "scope global" | awk '{print $2}'); do
+        sudo firewall-cmd --permanent --zone=trusted --add-source="${ip}"
+    done
+    sudo firewall-cmd --reload
```
pmtk: Not a big deal but wanted to point out: usually we create networks in the manage_hypervisor_config.sh.
Do you think we should move it there? If 2 scenarios (either two el98 or two el102) want to create that network, would we get race conditions?

test/bin/ci_phase_boot_and_test.sh
```diff
@@ -85,16 +85,41 @@ else
...
+        else
+            delete_ver=el98
+        fi
+    else
+        if [[ "${flip}" -eq 0 ]]; then
+            delete_ver=el98
+        else
+            delete_ver=el102
+        fi
+    fi
```
pmtk, line 109: Shorter version, use if you want - I don't mind what we have, just was wondering how to improve it
```
if [[ "$(uname -m)" == "x86_64" ]]; then
    vers=(el102 el98)
else
    vers=(el98 el102)
fi
delete_ver=${vers[flip]}
```

test/scenarios-bootc/c2cc/el102-src@c2cc-ipsec-ipv6.sh
```diff
@@ -0,0 +1,56 @@
...
+# Sets up 3 MicroShift clusters with C2CC and Libreswan IPsec (tunnel mode
+# protecting pod/service CIDRs) on a single-stack IPv6 network. Same test
+# suite as c2cc-ipsec, run over IPv6 instead of IPv4.
+
+# shellcheck source=test/bin/c2cc_common.sh
+source "${SCRIPTDIR}/c2cc_common.sh"
+
+# IPsec tests have ordering dependencies (setup verification must pass before
+# enforcement tests), so disable randomization.
+export TEST_RANDOMIZATION=none
```
pmtk: I think we could drop this, c2cc_common.sh has `export TEST_RANDOMIZATION=suites` by default, and since we run only 1 suite (single file ipsec.robot, not whole dir), this is redundant.

test/suites/c2cc/extra/ipsec-mtu.robot
```diff
@@ -0,0 +1,82 @@
...
+    Restart MicroShift And Wait For IPsec
+    Deploy Test Workloads
+    Verify Pod MTU On All Clusters    9000
+    Verify ESP Encapsulation On All Clusters
+    Ping DF Bit Between All Clusters    64
+    Ping DF Bit Between All Clusters    8952
+    Ping DF Bit Should Fail Between All Clusters    8973
+    Large Payload Between All Clusters    65536
+
+IPsec After Pod MTU Reconfiguration To 8900
```
pmtk: I need some insight into these two test cases. I see that what's different is that UDP datagram with DF is smaller and we check that the reconfiguration works.
But both tests exercise the the same "distance from the max" value? Do we need both? If yes, can we use RF templates?

test/suites/c2cc/extra/mtu.robot
```diff
@@ -0,0 +1,34 @@
...
+    ...    here so all MTU settings are explicit in the test.
+    Configure Pod MTU On All Clusters    9000
+    Restart MicroShift On All Clusters
+    Deploy Test Workloads
+    Verify Pod MTU On All Clusters    9000
+    Verify Full C2CC Connectivity
+    Ping DF Bit Between All Clusters    64
+    Ping DF Bit Between All Clusters    8952
+    Ping DF Bit Should Fail Between All Clusters    8973
+    Large Payload Between All Clusters    65536
```
pmtk: Very similar to ipsec-mtu.robot, but I don't have idea on common keywords without creating a mess, so we can leave that as it is

test/assets/c2cc/df-bit-send.py
agullon: good catch, I'll fix it

test/resources/ipsec.resource
agullon: yes, they are, I'll refactor it

test/suites/c2cc/extra/mtu.robot
agullon: also agree

test/scenarios-bootc/c2cc/el102-src@c2cc-ipsec-mtu-ipv6.sh
agullon: I'll do it

test/bin/c2cc_common.sh
agullon: thanks for this, it'll simplify the config a lot

test/bin/ci_phase_boot_and_test.sh
agullon: great idea, it feels it could be done even in only 1 line with some bash ugly command, anyway, I prefer your approch

test/scenarios-bootc/c2cc/el102-src@c2cc-ipsec-ipv6.sh
agullon: done

test/suites/c2cc/extra/ipsec-mtu.robot
agullon: you are right, not value at all having both
I'll go further and refactor it and merge both `ipsec-mtu.robot` and `mtu.robot` in only 1 robot file

test/suites/c2cc/extra/mtu.robot
agullon: I have an idea! please, review it ;)

test/resources/c2cc.resource
```diff
@@ -800,3 +802,111 @@ Verify Full C2CC Stack
     END
     Wait Until Keyword Succeeds    10m    10s    Verify C2CC Health Probes
     Wait Until Keyword Succeeds    10m    10s    Verify Full C2CC DNS
+
+Restart MicroShift On All Clusters
+    [Documentation]    Restart MicroShift on all clusters and wait for health.
+    FOR    ${alias}    IN    @{ALL_CLUSTERS}
+        Command On Cluster    ${alias}    systemctl restart microshift
```
pmtk: FYI `systemctl restart` is blocking, you could add `--no-block` to basically restart them in parallel

test/resources/c2cc.resource
```diff
@@ -800,3 +802,111 @@ Verify Full C2CC Stack
...
+    END
+    FOR    ${alias}    IN    @{ALL_CLUSTERS}
+        Wait Until Keyword Succeeds    5m    15s
+        ...    Command On Cluster    ${alias}    microshift healthcheck --timeout=30s
```
pmtk: I would prefer to use timeout=5m 
```suggestion
         Command On Cluster    ${alias}    microshift healthcheck --timeout=5m
```

test/scenarios-bootc/c2cc/el102-src@c2cc-ipsec-mtu-ipv6.sh
```diff
@@ -0,0 +1,35 @@
...
+# shellcheck disable=SC2119
+c2cc_setup_ipv6
+
+scenario_create_vms() {
+    # Set bridge IP after network creation for /etc/hosts resolution.
+    # Keep WEB_SERVER_URL at the default IPv4 — bootc kickstarts don't
+    # use REPLACE_OSTREE_SERVER_URL, and the hypervisor-side curl can't
+    # reach the IPv6 bridge due to libvirt's nftables filtering.
+    # shellcheck disable=SC2034  # used elsewhere
+    VM_BRIDGE_IP="$(get_vm_bridge_ip "jumbo-ipv6")"
```
pmtk: Is this different kind of network that we're creating in manage_hypervisor_config.sh?

agullon:
Applied — using `--no-block` for parallel restarts.

> FYI `systemctl restart` is blocking, you could add `--no-block` to basically restart them in parallel

test/resources/c2cc.resource
agullon: Applied — using `--no-block` for parallel restarts.

test/resources/c2cc.resource
agullon: Applied — using `--timeout=5m` now.

test/scenarios-bootc/c2cc/el102-src@c2cc-ipsec-mtu-ipv6.sh
agullon: Same network — just reading its bridge IP. Updated the stale comment.

PR #7027, opened 2026-07-09
USHIFT-7365: add dual-stack v4/v6 scenario symmetry to c2cc CI jobs

agullon:
## Summary
- Rename `*@c2cc-dual-stack` → `*@c2cc-dual-stack-v4` to make explicit that IPv4 is the primary stack
- Add `el102-src@c2cc-dual-stack-v6` scenario (IPv6 primary, EL10)
- Add `el98-src@c2cc-dual-stack-v4` scenario (IPv4 primary, EL9)

## Details

The c2cc CI jobs had asymmetric dual-stack coverage: el102 only had `c2cc-dual-stack` (IPv4 primary) and el98 only had `c2cc-dual-stack-v6` (IPv6 primary). This PR adds the missing variants so both RHEL versions test both dual-stack orderings.

### End state (13 scenarios)
| Scenario | EL10 | EL9 |
|---|---|---|
| c2cc (base) | ✅ | ✅ |
| c2cc-ipsec | ✅ | ✅ |
| c2cc-ipv6 | ✅ | ✅ |
| c2cc-dual-stack-v4 | ✅ (renamed) | ✅ (new) |
| c2cc-dual-stack-v6 | ✅ (new) | ✅ |
| c2cc-disruptive | ✅ | ✅ |
| c2cc-upgrade-ok | — | ✅ |

## Test plan
- [ ] Verify new scenario files match the structure of existing ones
- [ ] Run c2cc CI job to validate new scenarios boot and pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added dual-stack networking test scenarios for IPv4- and IPv6-primary configurations across three clusters.
  * Added automated virtual machine setup, cleanup, host configuration, and test execution for these scenarios.

* **Documentation**
  * Clarified IP command selection behavior for dual-stack testing, including IPv4 and IPv6 handling.

PR #7015, opened 2026-07-08
USHIFT-7346: C2CC upstream user documentation

pmtk:
## Summary by CodeRabbit

* **Documentation**
  * Added a new how-to guide for **Cluster-to-Cluster Connectivity (C2CC)**, including multi-cluster setup, remote health/latency verification, cross-cluster DNS behavior, cleanup steps, and operational considerations.
  * Expanded the **C2CC with IPsec** companion guide with concrete `ipsec`/XFRM verification examples, tighter traffic-capture expectations, and stronger nftables enforcement guidance with persistence and MTU sizing (including jumbo-frame scenarios).
  * Updated the user documentation index to link both new C2CC guides.

PR #7029, opened 2026-07-09
OCPBUGS-98219: Move RBAC manifests to the services that own them

pacevedom:
Context

  When the el98-lrel@multi-config-standard2 scenario was added to test TLS 1.3 configuration on ARM, it exposed a startup failure on restart: MicroShift would stop, apply a TLS config
  drop-in, start again, and crash with admission plugin "PodSecurity" failed to complete validation in 13s. The same failure was reported in OCPBUGS-33701 (closed as "Not a Bug") when
  changing memoryLimitMB on ARM.

  What's actually happening

  The SCC admission plugin has two sequential polling loops when validating a deployment create:
  - waitForReadyState() — up to 10s waiting for SCC informer cache sync (vendor/.../sccadmission/admission.go:596)
  - CreateProvidersFromConstraints() — up to 10s waiting for openshift.io/sa.scc.uid-range annotation on the target namespace (vendor/.../sccmatching/matcher.go:179)

  That annotation is set by namespace-security-allocation-controller, which runs inside cluster-policy-controller. But cluster-policy-controller depends on infrastructure-services-manager —
  the same service that's trying to create the deployments. This is a deadlock.

  On fresh boot it's masked: the target namespaces don't exist yet, so the deployment CREATE hits a client-side HTTP timeout (~1.4s), the handler retries, and by then
  cluster-policy-controller has started and annotated the namespace. On restart the namespaces already exist, so the CREATE goes straight to SCC admission, the 10s poll exhausts, and
  MicroShift exits.

  How the dependency got inverted

  cluster-policy-controller was originally introduced (3a28b7a096, July 2022) with only kube-apiserver as a dependency. Its RBAC manifests (6 ClusterRoles/ClusterRoleBindings for the
  namespace-security-allocation and PSA label syncer controllers) were placed in infrastructure-services-manager's applyDefaultRBACs alongside unrelated kube-controller-manager RBAC.

  Later (84ee03679a, September 2023), cluster-policy-controller was failing on startup because its RBAC didn't exist yet. Rather than moving the RBAC into the service that owns it, the fix
  was to add infrastructure-services-manager as a dependency — creating the circular dependency.

  What this PR does

  Move RBAC to owning services: applyDefaultRBACs was a grab bag applying RBAC for multiple services. Each service now applies its own RBAC in its Run method:
  - cluster-policy-controller applies its 6 manifests before signaling ready
  - kube-controller-manager applies its CSR approver RBAC alongside its namespaces
  - infrastructure-services-manager no longer applies RBAC for other services

  Add a real readiness check to cluster-policy-controller: the old code had close(ready) // todo — signaling ready before doing any work. Now it starts the controller in a goroutine and
  polls until the default namespace has the openshift.io/sa.scc.uid-range annotation, proving the namespace-security-allocation-controller is actually running.

  Add cluster-policy-controller as a dependency of infrastructure-services-manager: this guarantees the annotations exist before any deployment creates, regardless of CPU speed or system
  load. The dependency is safe because cluster-policy-controller now only depends on kube-apiserver — no cycle.

  Startup performance

  The hidden cost of the old design was always there — on every fresh boot, infrastructure-services-manager took ~22s because the SCC admission was silently polling for 10s on each
  deployment create waiting for annotations that hadn't been set yet. With the fix, it takes ~3.4s.

  Measured on same-arch x86 e2e-aws-tests, el98-src@standard-suite1 (two baselines from PRs #7015 and #7027 vs this PR):

| | Before | After |
|---|---|---|
| First boot | ~38-42s | ~22s |
| Restart | ~12s | ~12s |
| infrastructure-services-manager duration | ~22s | ~3.4s |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cluster policy controller startup by applying required RBAC on launch and waiting for default namespace security allocation to confirm readiness.
  * Updated cluster policy controller dependencies to require only the API server.
  * Refined kube-controller-manager initialization to apply OpenShift namespaces first, then the CSR approver role and binding, with cleaner error handling.
  * Updated infrastructure services startup to skip default RBAC application, apply user-critical PriorityClass, and include the cluster policy controller dependency.
  * Changed service registration order so the cluster policy controller starts before infrastructure services.


pkg/controllers/cluster-policy-controller.go
```diff
@@ -17,21 +17,30 @@ package controllers
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	klog "k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 )
 
 type ClusterPolicyController struct {
 	run        func(context.Context) error
+	applyRBAC  func(context.Context) error
```
copejon: why is `applyRbac` a func literal? It should be an unexported package func to reduce complexity.

pkg/controllers/cluster-policy-controller.go
```diff
@@ -17,21 +17,30 @@ package controllers
...
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	unstructuredv1 "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	klog "k8s.io/klog/v2"
 	"k8s.io/utils/clock"
```
copejon, line 39: Please refactor to match openshift import conventions

PR #7041, opened 2026-07-13
USHIFT-6973 Add CIS Level2 Smoke Testing

copejon:
Introduce a periodic CI scenario that installs MicroShift on a
CIS Level 2 hardened system and validates it functions correctly.

The test provisions a VM from the source image-installer ISO,
registers it with subscription-manager, installs OpenSCAP and
Ansible, applies CIS Level 2 hardening via ansible-role-rhel9-cis,
then runs four validation checks: OpenSCAP scan produces a report,
CIS failure count stays within threshold, all pods are running,
and a smoke test route is accessible.

New files:
- test/scenarios/periodics/el98-src@cis-lvl2.sh
- test/suites/cis/validate-cis-lvl2.robot
- test/assets/cis/cis-harden.yml
- test/assets/cis/cis-requirements.yml
- test/image-blueprints/layer2-presubmit/group1/rhel98-source.image-installer

The .image-installer marker enables build_images.sh to produce
rhel-9.8-microshift-source.iso, which is required for the liveimg
kickstart pattern (mutable RPM-based system needed for post-boot
CIS.

## Summary by CodeRabbit

* **New Features**
  * Added CIS Level 2 hardening validation for RHEL 9 and RHEL 10 environments.
  * Added checks for MicroShift functionality, pod readiness, networking, and route-based application access.
  * Added periodic scenarios for automated CIS testing on supported enterprise Linux releases.

* **Documentation**
  * Added a user guide covering CIS hardening considerations, required overrides, firewall configuration, and audit rule updates.
  * Added the guide to the user documentation index.


test/image-blueprints/layer2-presubmit/group1/rhel98-source.image-installer
ggiguash, line 1: We should avoid creation of another ISO just for this test. 
If we need RPM host, please follow the existing pattern of RPM scenarios

test/scenarios/periodics/el98-src@cis-lvl2.sh
```diff
@@ -0,0 +1,39 @@
...
+    run_command_on_vm host1 "sudo ansible-galaxy role install -r /tmp/cis-requirements.yml"
+
+    # Apply CIS Level 2 hardening
+    run_command_on_vm host1 "sudo ansible-playbook -c local /tmp/cis-harden.yml"
+
+    # Reboot to apply hardening changes and wait for the VM to come back
+    run_command_on_vm host1 "sudo reboot" || true
+    sleep 10
+    local -r ip=$(get_vm_property host1 ip)
+    wait_for_ssh "${ip}"
```
ggiguash: This is not reliable because we do not guarantee the host / ssh would go down in 10s.
Consider doing reboot in RF code where we have reliable sequences checking boot id.
Alternatively, check for some other scenarios for reboots.

test/assets/cis/cis-requirements-el9.yml
```diff
@@ -0,0 +1,7 @@
+roles:
+  - name: ansible-role-rhel9-cis
+    src: https://github.com/RedHatOfficial/ansible-role-rhel9-cis
```
ggiguash, line 3: Please, create another set of tests for RHEL 10.
https://github.com/RedHatOfficial/ansible-role-rhel10-cis

copejon: +1 note that this is going to require an additional iso. CIS hardening can't effectively be deployed against a bootc image. The blocker is that the playbook is written under the assumption of a mutable system and expects changes to persist across reboots. It wouldn't make sense to mutate the playbook (I tried) to fit the tests because it's likely to change over time. Another alternative I tried was to split the hardening process across build and runtime environs, but again that requires customizing the playbook, including reimplementing portions of it as in the scenario script. It's brittle, complex, and not maintainable.

pmtk: ~~el98 scenario uses rhel-9.8-microshift-source-isolated, but for el102 we already have test/image-blueprints-bootc/el10/layer3-periodic/group1/rhel102-bootc-source-isolated.containerfile - is it not enough?~~

I missed this is not bootc, why is not bootc? There is an RPM scenario in bootc scenarios.
When we move completely to el102, we won't support ostree anymore, so we'll have to do this work again

test/scenarios/periodics/el98-src@cis-lvl2.sh
```diff
@@ -0,0 +1,51 @@
...
+
+    # Apply CIS Level 2 hardening. The playbook is run detached because
+    # update-crypto-policies restarts sshd mid-run, killing the SSH session.
+    # The playbook itself uses -c local and survives the sshd restart.
+    run_command_on_vm host1 "nohup sudo ansible-playbook -c local /tmp/cis-harden.yml &>/tmp/cis-harden.log &"
+    local -r ip=$(get_vm_property host1 ip)
+    wait_for_ssh "${ip}"
+    local attempts=0
+    while ! run_command_on_vm host1 "grep -q 'PLAY RECAP' /tmp/cis-harden.log" 2>/dev/null ; do
+        (( attempts++ )) && (( attempts > 180 )) && { echo "Timed out waiting for CIS hardening"; return 1; }
```
pmtk: Not sure if we'll keep this procedure, but doesn't `{ return 1}` cause return from the whole function? It would be good to have some logs because otherwise we won't be able to debug that timeout.

test/suites/cis/validate-cis-lvl2.robot
```diff
@@ -0,0 +1,103 @@
...
+*** Test Cases ***
+OpenSCAP Scan Runs And Produces Report
+    [Documentation]    Run OpenSCAP scan with CIS Level 2 profile and verify it produces output files
+    Run CIS Level 2 Scan
+    Verify Remote File Exists With Sudo    ${OSCAP_REPORT_FILE}
+    Verify Remote File Exists With Sudo    ${OSCAP_RESULTS_FILE}
+
+CIS Failure Count Within Threshold
+    [Documentation]    Verify the number of CIS failures does not exceed the expected threshold
+    ${fail_count}=    Get CIS Failure Count
```
pmtk: This keyword depends on `OpenSCAP Scan Runs And Produces Report` running before `CIS Failure Count Within Threshold` but the scenario is missing `export TEST_RANDOMIZATION=none`

test/suites/cis/validate-cis-lvl2.robot
```diff
@@ -0,0 +1,103 @@
...
+    ...    return_rc=True
+    ...    return_stdout=True
+    ...    return_stderr=True
+    Should Be True    ${rc} == 0 or ${rc} == 2
+    ...    OpenSCAP scan failed with unexpected return code ${rc}: ${stderr}
+
+Get CIS Failure Count
+    [Documentation]    Parse the OpenSCAP results XML and count failures
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    grep -c '<result>fail</result>' ${OSCAP_RESULTS_FILE} || echo 0
```
pmtk: You specified `return_rc=True` but `|| echo 0` drops any errors. What if we cannot access the file? It will just pass as green

pmtk (commented):
Description seems outdated

> The .image-installer marker enables build_images.sh to produce
rhel-9.8-microshift-source.iso, which is required for the liveimg
kickstart pattern

I don't see .image-installer anywhere

docs/user/howto_cis_hardening.md
```diff
@@ -0,0 +1,117 @@
...
+
+## MicroShift's CIS Impact
+
+MicroShift introduces changes to a hardened system in three categories. These rules will fail on any system running Kubernetes containers and cannot be avoided without disabling core functionality.
+
+### IP Forwarding
+
+OVN-Kubernetes requires IP forwarding for pod networking. MicroShift enables `net.ipv4.ip_forward` and `net.ipv6.conf.all.forwarding` at startup.
+
+| Rule ID | OS |
```
ggiguash: Nit: In all the tables, instead of listing affected OS versions, can we create a column per supported OS and add `x` when a rule is applicable? This should improve readability

docs/user/howto_cis_hardening.md
```diff
@@ -0,0 +1,117 @@
...
+
+> Each variable is set to `false` to tell the role **not** to disable forwarding. This does not enable forwarding — MicroShift does that at startup.
+
+## Firewall Configuration
+
+CIS hardening enables `firewalld` and locks down all traffic. MicroShift requires several ports and trusted network sources. See [Firewall Configuration](howto_firewall.md) for the full list.
+
+At minimum, the pod and service networks must be trusted:
+
+```bash
```
ggiguash: Nit: If we're referencing the firewall configuration, let's say that it needs to be repeated as described in the referenced doc, instead of duplicating the settings here.

test/assets/cis/cis-harden-el9.yml
ggiguash, line 1: Let's add `-el9` suffix in the name here to avoid confusion.
We do this in CI job names for example - always add suffix without "assuming" the default OS.

docs/user/howto_cis_hardening.md
```diff
@@ -0,0 +1,117 @@
...
+sudo firewall-cmd --permanent --zone=trusted --add-source=10.42.0.0/16
+sudo firewall-cmd --permanent --zone=trusted --add-source=169.254.169.1
+sudo firewall-cmd --permanent --zone=trusted --add-source=fd01::/48
```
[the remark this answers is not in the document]
copejon: Applied — fixed in 26c1d781b.

test/scenarios-bootc/el10/periodics/el102-src@cis-lvl2.sh
```diff
@@ -0,0 +1,152 @@
...
+
+    if [[ -z "${rhocp}" ]] ; then
+        return
+    fi
+
+    if [[ "${rhocp}" =~ ^[0-9]{1,2}$ ]]; then
+        configure_cdn_repo \
+            "rhocp-${major}.${rhocp}" \
+            "Red Hat OpenShift ${major}.${rhocp} for RHEL 9" \
+            "https://cdn.redhat.com/content/dist/layered/rhel9/${arch}/rhocp/${major}.${rhocp}/os"
```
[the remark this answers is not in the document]
copejon: Applied — resolved arch on the VM via run_command_on_vm and pass explicitly to all callers. Fixed in 26c1d781b.

test/suites/cis/validate-cis-lvl2.robot
```diff
@@ -0,0 +1,103 @@
...
+    ...    sudo=False    return_rc=True    return_stdout=True    return_stderr=True
+    ${fail_count}=    Convert To Integer    ${stdout.strip()}
+    RETURN    ${fail_count}
```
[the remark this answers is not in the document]
copejon: Won't fix — keyword `Get CIS Failure Count` was deleted. The test was rewritten to use before/after delta scanning with Python ET parsing and `sudo=True`.

test/scenarios/periodics/el98-src@cis-lvl2.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
copejon: Won't fix — file `test/scenarios/periodics/el98-src@cis-lvl2.sh` was removed. Tests moved to `test/scenarios-bootc/`.

test/scenarios/periodics/el98-src@cis-lvl2.sh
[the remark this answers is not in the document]
copejon: Won't fix — stale. CIS hardening moved into the Robot suite; scenario script no longer runs playbooks.

test/scenarios/periodics/el98-src@cis-lvl2.sh
```diff
@@ -0,0 +1,51 @@
...
+    if run_command_on_vm host1 "grep -qE 'unreachable=[1-9]|failed=[1-9]' /tmp/cis-harden.log" 2>/dev/null ; then
+        run_command_on_vm host1 "cat /tmp/cis-harden.log" || true
+        echo "CIS hardening playbook failed"
+        return 1
+    fi
+
+    # Reboot to apply hardening changes and wait for the VM to come back
+    run_command_on_vm host1 "sudo reboot" || true
+    sleep 10
+    wait_for_ssh "${ip}"
```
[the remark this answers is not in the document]
copejon: Won't fix — stale. Reboots now handled by Robot Framework `Reboot And Reconnect` keyword with boot-ID change detection.

test/suites/cis/validate-cis-lvl2.robot
```diff
@@ -0,0 +1,201 @@
...
+    ...    sudo=True    return_rc=True    return_stdout=True    return_stderr=True
+    Should Be Equal As Integers    ${rc}    0    msg=Failed to set GRUB password: ${stderr}
+
+Regenerate Privileged Command Audit Rules
+    [Documentation]    Regenerate audit rules for setuid/setgid binaries after
+    ...    installing MicroShift RPMs. The initial rules from CIS hardening
+    ...    do not cover binaries added by the MicroShift packages.
+    Command Should Work
+    ...    bash -c "find / -xdev \\( -perm -4000 -o -perm -2000 \\) -type f 2>/dev/null | awk '{print \"-a always,exit -F path=\" \\$1 \" -F perm=x -F auid>=1000 -F auid!=unset -k privileged\"}' > /etc/audit/rules.d/99-privileged-post-install.rules"
+    Command Should Work    augenrules --load
```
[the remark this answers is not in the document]
copejon: Won't fix — keyword `Regenerate Privileged Command Audit Rules` was deleted entirely.

test/scenarios-bootc/el10/periodics/el102-src@cis-lvl2.sh
```diff
@@ -0,0 +1,154 @@
...
+
+# Sourced from scenario.sh and uses functions defined there.
+
+# CIS hardening runs inside the Robot suite and takes ~20 minutes
+TEST_EXECUTION_TIMEOUT=60m
+
+# The RPM-based image used to create the VM for this test does not
+# include MicroShift or greenboot, so tell the framework not to wait
+# for greenboot to finish when creating the VM.
+export SKIP_GREENBOOT=true
```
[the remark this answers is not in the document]
copejon: Won't fix — all 50+ scenario scripts in `test/scenarios-bootc/` use `#!/bin/bash` with no `pipefail`. These files are sourced by `scenario.sh`, not executed directly.

docs/user/howto_cis_hardening.md
```diff
@@ -0,0 +1,117 @@
...
+|:--------|:---|
+| `audit_rules_privileged_commands` | RHEL 9, 10 |
+
+This rule can be resolved by regenerating audit rules after installing MicroShift:
+
+```bash
+sudo find / -xdev \( -perm -4000 -o -perm -2000 \) -type f 2>/dev/null | \
+  awk '{print "-a always,exit -F path=" $1 " -F perm=x -F auid>=1000 -F auid!=unset -k privileged"}' | \
+  sudo tee /etc/audit/rules.d/99-privileged-post-install.rules > /dev/null
+sudo augenrules --load
```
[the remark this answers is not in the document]
copejon: Won't fix — RHEL 9 and 10 standard layout has `/usr` on the root filesystem. `-xdev` only excludes pseudo-filesystems (`/proc`, `/sys`, `/dev`), not `/usr`.

docs/user/howto_cis_hardening.md
```diff
@@ -0,0 +1,117 @@
...
+
+CIS hardening enables `firewalld` and locks down all traffic. MicroShift requires several ports and trusted network sources. See [Firewall Configuration](howto_firewall.md) for the full list.
+
+At minimum, the pod and service networks must be trusted:
+
+```bash
+sudo firewall-cmd --permanent --zone=trusted --add-source=10.42.0.0/16
+sudo firewall-cmd --permanent --zone=trusted --add-source=169.254.169.1
+sudo firewall-cmd --permanent --zone=trusted --add-source=fd01::/48
+sudo firewall-cmd --permanent --zone=public --add-port=6443/tcp
```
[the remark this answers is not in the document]
copejon: Won't fix — `howto_firewall.md` itself shows 6443/tcp on the public zone. This is the established pattern; users can restrict further per their security requirements.

test/suites/cis/validate-cis-lvl2.robot
```diff
@@ -0,0 +1,218 @@
...
+    [Documentation]    Parse an OSCAP results XML file and return the list of
+    ...    failing rule IDs (short names without the SSG prefix).
+    [Arguments]    ${results_file}
+    Verify Remote File Exists With Sudo    ${results_file}
+    ${stdout}    ${stderr}    ${rc}=    Execute Command
+    ...    python3 -c "import xml.etree.ElementTree as ET; ns={'x':'http://checklists.nist.gov/xccdf/1.2'}; tree=ET.parse('${results_file}'); rules=[rr.get('idref').replace('xccdf_org.ssgproject.content_rule_','') for rr in tree.findall('.//x:rule-result',ns) if rr.find('x:result',ns).text=='fail']; print('\\n'.join(sorted(rules)))"
+    ...    sudo=True    return_rc=True    return_stdout=True    return_stderr=True
+    Should Be Equal As Integers    ${rc}    0    msg=Failed to parse results: ${stderr}
+    @{rules}=    Split String    ${stdout.strip()}    \n
+    RETURN    ${rules}
```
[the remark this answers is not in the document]
copejon: Won't fix — the empty-string edge case is benign: both scans produce `['']` when zero failures, so the set difference cancels out. The pathological case (post has fewer failures than baseline) is unreachable since MicroShift always introduces failures.

copejon:
Addressed CodeRabbit review-body nitpick findings:

- **chmod needs `sudo=True`** (validate-cis-lvl2.robot:92): Won't fix — `Command Should Work` defaults to `sudo_mode=True`.
- **Extract shared `write_repo_file` helper** (el102-src@cis-lvl2.sh): Won't fix — the mktemp/tee/copy pattern is used identically across dozens of existing scenario scripts. Adding abstraction would diverge from established conventions.

test/assets/cis/cis-harden-el9.yml
copejon: Done — renamed both `cis-harden.yml` → `cis-harden-el9.yml` and `cis-requirements.yml` → `cis-requirements-el9.yml`, updated Robot defaults to match.

ggiguash:
Overriding failed tests due to CI infra issues to save cycles
/lgtm
/override ci/prow/e2e-aws-tests-bootc-periodic-arm-el9

PR #7257, opened 2026-08-27
USHIFT-7488: scripts: prefer locally built RPMs over product repos in configure-vm.sh

mkowalski:
## What

Set `priority=1` on the `microshift-local` repository used by `configure-vm.sh` so dnf always installs the locally built RPMs instead of any `microshift` RPMs published in enabled product repositories.

## Why

All `release-4.21` conformance CI jobs (periodics and presubmits) have been permafailing with OVN never initializing (`failed to get ofport of -ex`, OCPBUGS-105440) even after the ovn-kubernetes fix merged (openshift/ovn-kubernetes#3413) and is present in nightlies since `4.21.0-0.nightly-2026-08-24-221323`.

Investigation of the failing job sosreports showed why: CI builds RPMs from the checked-out source into `_output/rpmbuild` and installs them via `--repofrompath`, but dnf resolved `microshift` from `rhocp-4.21-for-rhel-9-x86_64-rpms` instead, because the released NEVRA (`4.21.29-202608170915.p0.gbc6f556`) sorts higher than the locally built nightly version. Install transaction from a failing run of pull-ci-openshift-microshift-release-4.21-ocp-full-conformance-rhel-eus:

```
microshift  x86_64  4.21.29-202608170915.p0.gbc6f556.assembly.4.21.29.el9  rhocp-4.21-for-rhel-9-x86_64-rpms
```

That released RPM pins the pre-fix `ovn-kubernetes-microshift` image (built 2026-08-12 from `cf07889`, 44 commits behind `release-4.21`, missing the OCPBUGS-105440 fix), so the tested nodes always hit the already-fixed bug. This is also why the rebase PR #7242 cannot pass CI despite pinning a fixed image.

With repository priority 1 on the local repo, dnf prefers packages from it regardless of version ordering, so CI tests what it built.

## How was this tested

`bash -n` and `shellcheck` on the modified script (no new findings). Behavioral verification will come from CI on this PR: the conformance jobs should now install the source-built RPMs (which pin the fixed ovn-kubernetes image from the rebase base) — not verified locally.

---
This PR description was generated using AI. Please verify before acting on it.
Assisted-By: Claude Fable 5

PR #7258, opened 2026-08-27
NO-JIRA: scripts: prefer locally built RPMs over product repos in configure-vm.sh

mkowalski:
## What

Set `priority=1` on the `microshift-local` repository used by `configure-vm.sh` so dnf always installs the locally built RPMs instead of any `microshift` RPMs published in enabled product repositories.

Same change as https://github.com/openshift/microshift/pull/7257 (release-4.21), applied to this branch since the script is identical here and equally exposed.

## Why

On release-4.21, all conformance CI jobs permafailed because dnf resolved `microshift` from `rhocp-4.21-for-rhel-9-x86_64-rpms` (released NEVRA `4.21.29-...` sorts higher than the locally built nightly version) instead of the source-built RPMs in the `--repofrompath` local repo. The released RPM pins an outdated ovn-kubernetes image (missing the OCPBUGS-105440 fix), so CI kept hitting an already-fixed bug. See #7257 for the full investigation.

The same silent version-ordering hazard exists on every branch once a matching product repo publishes a higher-versioned microshift RPM. With repository priority 1 on the local repo, dnf prefers packages from it regardless of version ordering, so CI tests what it built.

## How was this tested

`bash -n` and `shellcheck` on the modified script (no new findings). Behavioral verification comes from CI on this PR; not verified locally.

---
This PR description was generated using AI. Please verify before acting on it.
Assisted-By: Claude Fable 5

## Summary by CodeRabbit

* **Bug Fixes**
  * Local MicroShift installations now prioritize the locally generated repository over published repositories.
  * Repository GPG checking remains disabled for local development environments.

### DIFF
--- a/scripts/devenv-builder/configure-vm.sh
+++ b/scripts/devenv-builder/configure-vm.sh
@@ -337,10 +337,15 @@ function build_and_install_microshift_rpms() {
         "${DNF_RETRY}" "localinstall" "$(find ~/microshift/_output/rpmbuild/RPMS -type f -name "*.rpm" $(printf ' -not -name *%s* ' ${SKIPPED_RPMS}))"
     else
         createrepo "${HOME}/microshift/_output/rpmbuild"
+        # Set the highest priority on the local repository to ensure that the
+        # locally built RPMs are preferred over any RPMs published in product
+        # repositories (e.g. rhocp-4.y), whose release versions may sort higher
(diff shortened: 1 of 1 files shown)


pacevedom:
Replaced by https://github.com/openshift/microshift/pull/7262
/close

PR #7270, opened 2026-08-27
[release-4.18] USHIFT-7451: replace nginx-unprivileged with busybox in test assets

eslutsky:
This is an manual cherry-pick of https://github.com/openshift/microshift/pull/7199

The docker.io/nginxinc/nginx-unprivileged:latest image is a multi-arch manifest list with 14 platform variants. When mirroring with `skopeo copy --all --preserve-digests`, Quay v3.11.7 rejects one of the platform manifests as "manifest invalid", causing the mirror step to fail and aborting CI jobs (e.g. e2e-aws-tests-arm).

These test pods only run `sleep 1d` and do not use any nginx functionality, so replace the image with quay.io/microshift/busybox:1.36 which is already mirrored as a single-arch image. This eliminates the problematic multi-arch manifest list from the mirror set entirely.

### DIFF
--- a/docs/contributor/storage/default_csi_plugin.md
+++ b/docs/contributor/storage/default_csi_plugin.md
@@ -161,7 +161,7 @@ spec:
     - sh
     - -c
     - sleep 1d
-    image: nginxinc/nginx-unprivileged:latest
+    image: quay.io/microshift/busybox:1.36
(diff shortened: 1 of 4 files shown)


pacevedom (approved):
/lgtm

PR #7305, opened 2026-09-02
NO-ISSUE: Use --wait=false for namespace deletion in RF teardowns

agullon:
## Summary

- Use `--wait=false` in the `Remove Namespace` RF keyword so `oc delete namespace` returns immediately instead of blocking until all finalizers are processed.

**Root cause:** On ARM with dual-stack, OVN reconciliation after a network config change (dual-stack → single-stack) saturates the CPU. When suite teardown fires `oc delete namespace` during this window, the namespace garbage collection — competing with OVN for CPU — exceeds the 5-minute RF process timeout in `Run With Kubeconfig`. The killed `oc` process returns rc=1, the `Should Be Equal As Integers` assertion fails, and all passing tests are retroactively marked as suite teardown failures.

**Fix:** `--wait=false` marks the namespace for deletion and returns immediately. The API server continues cleanup asynchronously. This is safe because:
- All 13 callers of `Remove Namespace` use it in teardown contexts where no subsequent test depends on the namespace being fully gone
- Test VMs are torn down after the suite completes, so any in-progress namespace deletion is irrelevant

**Observed in:** `e2e-aws-tests-bootc-release-arm-el9`, scenario `el98-lrel@dual-stack-configuration1` — all 4 dual-stack tests passed but were retroactively marked failed due to suite teardown timeout.

## Test plan

- [x] Verify no caller of `Remove Namespace` depends on synchronous namespace deletion
- [ ] Confirm dual-stack suite teardown completes without timeout on ARM

## Summary by CodeRabbit

* **Bug Fixes**
  * Namespace deletion now returns without waiting for the namespace removal process to complete.

### DIFF
--- a/test/resources/kubeconfig.resource
+++ b/test/resources/kubeconfig.resource
@@ -94,4 +94,4 @@ Create Random Namespace
 Remove Namespace
     [Documentation]    Removes the given namespace.
     [Arguments]    ${ns}
-    Run With Kubeconfig    oc delete namespace ${ns}
+    Run With Kubeconfig    oc delete namespace ${ns} --wait=false


agullon:
Consolidated into #7326

PR #7337, opened 2026-09-07
OCPEDGE-2973: Add kubelet image credential provider configuration

Neilhamza:
## What

Adds two optional keys under the `kubelet:` config section:

```yaml
kubelet:
  imageCredentialProviderConfigPath: /etc/microshift/credential-providers.yaml
  imageCredentialProviderBinDir: /usr/libexec/microshift/credential-providers
```

These are **kubelet flags, not KubeletConfiguration fields**. MicroShift reads them out of the schemaless `kubelet:` map, validates them at startup, sets them on the embedded kubelet, and filters them out of the generated `KubeletConfiguration`. Everything else under `kubelet:` still passes through, and `show-config` reports the keys exactly as the user wrote them.

**Validation (first failure wins):** both keys required together and absolute; config path resolves to a regular file or directory, bin dir to a directory; the two must differ; every path component (and, for directories, every entry) must be root-owned and not group/other-writable; canonical paths are handed to kubelet. The provider config is then pre-checked with **kubelet's own strict decoder** — non-empty config dir, each file decodes (`v1`/`v1beta1`/`v1alpha1`, unknown fields rejected), no duplicate provider names, each `providers[].name` resolves to an executable in the bin dir. This is why MicroShift parses the config: kubelet `os.Exit(1)`s on a bad provider config after startup, so we surface it as an ordinary config error instead. Error messages name the key that needs fixing.

Design: openshift/enhancements#2089.

## Tests

- `pkg/config/kubelet_test.go`: key reading, `KubeletPassthrough` (drops exactly the two keys), and the full validation + trusted-path + structural table (real temp files/symlinks/FIFO; ownership injected so the suite runs non-root).
- `pkg/node/kubelet_test.go`: reserved keys stripped from the generated `KubeletConfiguration`; flags set to canonical values.
- `test/suites/standard2/kubelet-credential-provider.robot`: happy path, single-key, missing bin dir, world-writable bin dir, missing/duplicate/unresolved provider, empty config dir — each failure asserts one specific error and recovers.

`generate-config`/`verify-config`, `go build`, `go test`, `golangci-lint`, `verify-rf` all pass.

## Validation

End-to-end on a real RHEL 9.6 host against **real Amazon ECR** (upstream `ecr-credential-provider`): configured log line with canonical paths, reserved keys absent from the generated config, real pod pull with **no** `imagePullSecrets`, and every failure mode above rejected with the expected message. Full matrix in the [validation comment](https://github.com/openshift/microshift/pull/7337#issuecomment-5711616049).

## Review

All comments across three review rounds are addressed with a commit or a documented reason (see the inline threads). Key round-2 changes: read the keys directly (no map iteration), `os.Stat`+exec-bit instead of `exec.LookPath`, duplicate-name rejection, raw/canonical field split behind an accessor, lazy codec (`sync.OnceValue`), single ancestor walk, `entry.IsDir()` symlink-to-dir parity, and the RF suite moved to `standard2/` with one specific regex per case.

## Notes

- `show-config --mode effective` now validates the credential-provider paths (consistent with `dns.go` already stat-ing files) — an invalid path makes it error rather than print.
- Extended-ACL and `EACCES` "run as root" checks were dropped as redundant/unreachable after host validation (mode bits already reflect the ACL mask; all callers run as root).
- User-facing walkthrough (ECR/GCR/ACR, cache tuning) lives in OSDOCS ([OCPEDGE-2976](https://redhat.atlassian.net/browse/OCPEDGE-2976)); `docs/user/howto_config.md` has a short pointer plus the SELinux `bin_t` placement rule.


pkg/config/kubelet.go
```diff
@@ -0,0 +1,282 @@
...
+	return out
+}
+
+// ConfiguredKubeletCredentialProviderPaths returns the raw, user-configured
+// (non-canonical) values of the two credential-provider keys. It exists so
+// pkg/node can report the original paths in diagnostics without needing to know
+// the reserved key names, which stay private to this package.
+func (c *Config) ConfiguredKubeletCredentialProviderPaths() (configPath, binDir string) {
+	configPath, _ = kubeletStringValue(c.Kubelet, kubeletImageCredentialProviderConfigPathKey)
+	binDir, _ = kubeletStringValue(c.Kubelet, kubeletImageCredentialProviderBinDirKey)
```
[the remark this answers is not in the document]
Neilhamza: Removed the accessor entirely; the configured values are available from `show-config`, so the log line now records only the canonical paths kubelet receives.

Neilhamza:
## Manual end-to-end validation on a RHEL 9.6 host

Validated this branch (commit `0a3907e69`) on a real single-node MicroShift host.
The base cluster was first installed from **upstream `main`** and proven healthy
(node Ready, all pods Running), then the branch RPMs were swapped in — so the
backward-compatibility baseline below is measured against a known-good install.
No AWS/ECR — a local password-protected TLS registry (`registry:2` on the node
loopback `127.0.0.1:5000`, self-signed cert, htpasswd auth) plus a mock credential
provider returning static credentials for that registry.

### The core proof: the pull outcome flips solely with what the provider returns

With **no `imagePullSecrets`** on the pod or any `default` service account, and **no
stored registry credentials anywhere** CRI-O or podman reads (verified: `auth.json`,
`/var/lib/kubelet/config.json`, `/etc/crio/openshift-pull-secret` all clean), the only
thing that can authenticate a pull is the credential provider:

| Provider returns | Pull result |
|---|---|
| correct password | `Successfully pulled image` — pod Running |
| **wrong password** | `authentication required` → **ImagePullBackOff** |
| restored password | `Successfully pulled image` — pod Running (recovery) |

The registry itself was confirmed to reject unauthenticated pulls before any test ran,
so a "pull succeeded" cannot be a false pass from an open registry.

### Feature behavior (baseline + configured)

- **Backward compatible:** with neither key set, MicroShift starts, all pods Running,
  and no credential-provider line is logged (feature inactive).
- **Configured:** the exact QE log line appears with the canonical (symlink-resolved)
  paths — `Kubelet image credential provider configured configPath="…" binDir="…"`.
- **Provider registered without error:** after the configured line, kubelet registered
  the provider with no error (no `plugin binary … did not exist`); node Ready. This
  closes the "configured ≠ usable" distinction — the provider is actually loaded, not
  just accepted at config time.
- **Reserved keys excluded from generated KubeletConfiguration:**
  `grep -c imageCredentialProvider` on the generated kubelet config returns **0**.
- **`show-config --mode effective`** echoes both keys exactly as configured.
- **SELinux:** the bin dir and mock binary carry `bin_t`, matching the documented
  placement rule (`/usr/libexec`).

### Credential caching (cacheDuration: 1m)

Mock instrumented to count invocations; kubelet credential cache emptied via restart:

| Event | Provider invocations |
|---|---|
| before any pull | 0 |
| pull #1 (cache empty) | 1  — provider invoked |
| pull #2 (within 1-min window) | 1  — **cache served, provider not re-invoked** |
| pull #3 (past expiry, ~106s later) | 2  — cache expired, provider re-invoked |

### Robot Framework suite (`test/suites/configuration1/kubelet-credential-provider.robot`)

Ran the committed suite from the branch against the host: **5 passed, 0 failed.**
The three negative cases were confirmed to catch MicroShift genuinely failing to start
(service enters failed/auto-restart) with the specific validation error in the journal:
- only one key set → `must be set together`
- missing bin dir → `imageCredentialProviderBinDir … does not exist`
- world-writable bin dir → `must be owned by root and not writable by group or others`

### Note for QE ([OCPEDGE-2974](https://redhat.atlassian.net/browse/OCPEDGE-2974))

On a host that is **not** a harness-provisioned VM, the RF suite needs the host's
address in `apiServer.subjectAltNames` so `Setup Kubeconfig` can fetch a kubeconfig
keyed to that address. This was added as a temporary drop-in for the run and removed
during cleanup. The credential-provider tests themselves make no API calls; this is
purely a harness-connectivity accommodation.

### Cleanup

Registry, images, mock provider, all drop-ins, and the mock log removed; MicroShift
restarted to a clean dormant state (all 7 pods Running, no keys in effective config).
The branch build remains installed. No changes were made to the git branch during
validation.

Neilhamza:
## Full edge-case validation against real Amazon ECR

Ran the complete A–G edge-case matrix on a real RHEL 9.6 host (SELinux Enforcing)
against real Amazon ECR with the upstream `ecr-credential-provider` v1.36.1 using
the instance role — build `86190e241` (this PR, rebased on main). Every failure
case recovered to a healthy 7-pod cluster before the next.

After the host run, three checks were removed as redundant/unreachable (ACL —
mask is visible in mode bits; EACCES branch — all callers are root; SELinux
static-creds note — not reproducible). Net effect is a smaller diff.

### Results
- **Happy paths (A1–A7):** dormant when keys absent / empty; configured line with
  canonical paths; reserved keys excluded from the generated KubeletConfiguration;
  real ECR pull with **no imagePullSecrets**; directory config path and symlinked
  bin dir both resolve correctly.
- **Structural path validation (B1–B7):** all seven reject with the expected
  message (missing/relative/nonexistent paths, non-dir bin dir, FIFO config,
  config↔binDir collision).
- **Trusted-path ownership (C1–C7):** every ownership/permission violation
  (writable dir/binary/config, non-root owner, non-root symlink target, writable
  ancestor) fails closed with the ownership message; ownership error correctly
  wins over structural error.
- **Provider-config decode / os.Exit guard (D1–D13):** missing binary (full joined
  path reported), empty dir, malformed YAML, wrong kind, unregistered apiVersion,
  empty providers, `/` in name, non-exec binary — all become clean config errors.
  JSON, multi-file directories, and **v1 / v1beta1 / v1alpha1** all accepted.
  **D12 (key result):** an unknown field (`matchImage`) is rejected by the strict
  decoder at config load and **never reaches kubelet's `os.Exit(1)`** — the
  original registration-crash finding is closed.
- **Cache/token lifecycle on real ECR (F):** tokens rotate per call (distinct
  SHA-256, username `AWS`, provider-returned `cacheDuration: 6h`); one exec-plugin
  invocation serves two in-window pulls; a restart clears the in-memory provider
  cache; ECR-pulled workloads survive a restart with no re-pull/re-auth.
- **Committed Robot suite on the real host:** 7 / 8 pass. The one failure is the
  ACL case — see below.

### One check to remove: the extended-ACL rule is redundant
The trusted-path rule has a dedicated extended-POSIX-ACL check
(`aclForTrust`/`checkNoExtendedACL`) with the rationale that "mode bits do not
reveal ACL write grants." **That rationale is false.** When an extended ACL is
present, Linux reports the ACL *mask* in the object's group permission bits, so any
ACL entry with effective write access makes the group-write bit visible and the
existing mode check (`mode&0o022`) rejects it first. Confirmed on the host:

```
chmod 0755 d; setfacl -m u:nobody:rwx d   -> stat 0775 (drwxrwxr-x)  # mode check catches it
chmod 0755 d; setfacl -m u:nobody:r-x d   -> stat 0755 (drwxr-xr-x)  # grants no write, harmless
```

The only ACL that reaches the dedicated check is one that grants **no** write
(e.g. `r-x`), which is harmless. So the mode-bit check was already sufficient
against ACL write grants; the ACL check adds nothing. This is also why the Robot
case "Extended ACL On Bin Directory Prevents Start" fails: it grants `rwx` and
asserts the "must not have an extended ACL" message, but the mode check fires first
with the ownership message. MicroShift still refuses to start in that case — the
security outcome is correct, only the check that fires (and the message) differ.

**Follow-up:** drop `aclForTrust` + `checkNoExtendedACL` (and the now-unused
`unix` import), the ACL unit tests and the "no ACL passes" test, and the Robot ACL
case (suite back to 7); replace the ACL rationale in the enhancement with the
correct mask-reflection explanation. One fewer thing to maintain, and a doc
sentence that's actually true.

### Two smaller findings (documented, non-blocking)
- The `EACCES` "run as root" branch in `decodeCredentialProviderNames` is
  unreachable via `show-config`: `show-config` already refuses non-root with
  "command requires root privileges" before reading any config file. The branch's
  stated rationale (non-root show-config vs a 0600 file) does not hold.
- The static-credentials SELinux concern (creds under `/root/.aws`,
  `admin_home_t`) did not reproduce on this build: `kubelet_t` read `/root/.aws`
  with zero AVC denials. The `/etc/microshift` placement remains reasonable
  portability guidance but isn't required to avoid an AVC here.

Full per-case evidence and the raw logs are attached to the validation artifacts.
AWS writes during the entire run: exactly one (the test image push).

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    Command Should Work    rm -rf ${CP_BIN_DIR} ${CP_CONFIG_FILE}
+    Restart MicroShift
+    Remove Kubeconfig
+    Logout MicroShift Host
+
+Restart MicroShift With Cursor
+    [Documentation]    Record the journal cursor, then restart MicroShift
+    ${cursor}=    Get Journal Cursor
+    VAR    ${CURSOR}=    ${cursor}    scope=TEST
+    Restart MicroShift
```
kasturinarra, line 158: No other file in the test suites uses VAR with scope=. For consistency, these should be, can we do the same ?
  Set Test Variable    \${CURSOR}    ${cursor}

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+
+Valid Configuration Applies Kubelet Flags
+    [Documentation]    With both keys set, MicroShift starts, logs the configured paths, reports them in
+    ...    show-config, and keeps them out of the generated KubeletConfiguration
+    [Setup]    Apply Credential Provider Config    ${CP_VALID}
+    Pattern Should Appear In Log Output    ${CURSOR}    ${CP_CONFIGURED_LOG}
+    ${config}=    Show Config    effective
+    Should Be Equal As Strings    ${config.kubelet.imageCredentialProviderConfigPath}    ${CP_CONFIG_FILE}
+    Should Be Equal As Strings    ${config.kubelet.imageCredentialProviderBinDir}    ${CP_BIN_DIR}
+    Command Should Fail    grep -q imageCredentialProvider ${KUBELET_GENERATED_CONFIG}
```
kasturinarra, line 88: Is this assertion necessary? The test already validates that the configuration is applied correctly (log output + show-config)

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    [Teardown]    Remove Credential Provider Config
+
+Missing Bin Directory Prevents Start
+    [Documentation]    MicroShift fails to start when the bin directory does not exist
+    [Setup]    Apply Invalid Credential Provider Config    ${CP_MISSING_BIN_DIR}
+    Pattern Should Appear In Log Output    ${CURSOR}    imageCredentialProviderBinDir
+    Pattern Should Appear In Log Output    ${CURSOR}    does not exist
```
kasturinarra, line 95: The problem is these are disconnected — it checks that "imageCredentialProviderBinDir" appears somewhere in the logs and "does not exist" appears somewhere in the logs, but doesn't verify
  they're in the same log line. Any unrelated "does not exist" message would satisfy the second check.

  A more robust approach would be to check for a single, more specific pattern that captures the full error message would be good.

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    [Teardown]    Run Keywords    Remove Credential Provider Config    AND    Restart MicroShift
+
+Only One Key Prevents Start
+    [Documentation]    MicroShift fails to start when only one of the two keys is set
+    [Setup]    Apply Invalid Credential Provider Config    ${CP_ONLY_CONFIG_PATH}
+    Pattern Should Appear In Log Output    ${CURSOR}    must be set together
```
kasturinarra, line 101: A more robust approach would be to check for a single, more specific pattern that captures the full error message would be good.

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    ...    AND    Remove Credential Provider Config
+    ...    AND    Restart MicroShift
+
+Missing Provider Binary Prevents Start
+    [Documentation]    MicroShift fails to start when a provider names a binary absent from the bin directory.
+    ...    Validation fails before kubelet is configured, so the "configured" line must not appear.
+    [Setup]    Run Keywords    Upload String To File    ${CP_BAD_PROVIDER_CONFIG}    ${CP_CONFIG_FILE}
+    ...    AND    Apply Invalid Credential Provider Config    ${CP_VALID}
+    Pattern Should Appear In Log Output    ${CURSOR}    no executable at
+    Pattern Should Appear In Log Output    ${CURSOR}    no-such-provider
```
kasturinarra, line 119: A more robust approach would be to check for a single, more specific pattern that captures the full error message would be good.

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    Pattern Should Not Appear In Log Output    ${CURSOR}    ${CP_CONFIGURED_LOG}
+    [Teardown]    Run Keywords    Upload String To File    ${CP_PROVIDER_CONFIG}    ${CP_CONFIG_FILE}
+    ...    AND    Remove Credential Provider Config
+    ...    AND    Restart MicroShift
+
+Empty Configuration Directory Prevents Start
+    [Documentation]    MicroShift fails to start when the configuration directory holds no config files
+    [Setup]    Run Keywords    Command Should Work    install -d -o root -g root -m 0755 ${CP_CONFIG_DIR}
+    ...    AND    Apply Invalid Credential Provider Config    ${CP_EMPTY_DIR_CONFIG}
+    Pattern Should Appear In Log Output    ${CURSOR}    contains no .json, .yaml, or .yml
```
kasturinarra, line 129: someone reading the test can't tell what the expected error actually is. It would be clearer to match the full error message

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    [Teardown]    Run Keywords    Command Should Work    rm -rf ${CP_CONFIG_DIR}
```
kasturinarra, line 130: if the test fails before reaching its teardown the directory survives. Adding a defensive rm -rf for it in the suite Teardown would make cleanup more resilient

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
+*** Settings ***
```
kasturinarra, line 1: Instead of adding it in a new suite, let us add it to standarad2/kubelet-credential-provider.robot ? 
@pacevedom any obections ?

test/suites/configuration1/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+
+Resource            ../../resources/common.resource
+Resource            ../../resources/microshift-config.resource
+Resource            ../../resources/microshift-process.resource
+Library             ../../resources/journalctl.py
+
+Suite Setup         Setup
+Suite Teardown      Teardown
+
+Test Tags           slow    restart
```
kasturinarra, line 12:
```suggestion
Test Tags           restart    slow
```

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+
+// KubeletPassthrough returns a copy of the kubelet map with the MicroShift-owned
+// keys removed, so only genuine KubeletConfiguration settings are written to the
+// generated kubelet config file. A nil map returns nil.
+func (c *Config) KubeletPassthrough() map[string]any {
+	if c.Kubelet == nil {
+		return nil
+	}
+	out := make(map[string]any, len(c.Kubelet))
+	for k, v := range c.Kubelet {
```
qJkee, line 120: c.Kubelet already a map. Why we need to iterate over all values when we already know the keys?

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+
+	// Check that both keys are absolute before touching the filesystem.
+	for _, p := range paths {
+		if !filepath.IsAbs(p.value) {
+			return fmt.Errorf("kubelet.%s (%q) must be an absolute path", p.key, p.value)
+		}
+	}
+
+	canonical := make([]string, len(paths))
+	for i, p := range paths {
```
qJkee, line 177: why we need 2 separate loops for this? We can merge them

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+
+	return canonical, nil
+}
+
+// validateDirEntries applies the trusted-path rule to every entry in dir.
+// Symlinked entries are resolved and the full rule, including the target's
+// ancestors, is applied to the target.
+func validateDirEntries(dir string) error {
+	entries, err := os.ReadDir(dir)
```
qJkee: will filepath like `a/b/c/d/e/........n100` resolved correctly?
Maybe better to use `filepath.Walk`?

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+// validateTrustedChain walks every component of the canonical (already
+// symlink-resolved) path from / to the final object and requires each to be
+// owned by root and not writable by group or others.
+func validateTrustedChain(canonical string) error {
+	for _, component := range trustedPathComponents(canonical) {
+		uid, mode, err := statForTrust(component)
+		if err != nil {
+			return err
+		}
+		if uid != 0 || mode&0o022 != 0 {
```
qJkee: what is `0o022`? Can you move this to named const

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+			// Kubelet joins the bin dir and the provider name directly; a name
+			// containing a separator would escape the bin dir, so reject it.
+			if strings.Contains(name, "/") {
+				return prefix(fmt.Errorf("provider name %q must not contain \"/\"", name))
+			}
+			// Report the joined path, never exec.LookPath's return value: on
+			// error LookPath returns an empty string, which is the upstream
+			// defect that prints "plugin binary executable  did not exist".
+			joined := filepath.Join(canonicalBinDir, name)
+			if _, err := exec.LookPath(joined); err != nil {
```
qJkee, line 332: What is the reason for checking if binary is available via `exec.LookPath`? This checks if bin available in PATH, which might be not the case for us

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+
+// validateCredentialProviderStructure verifies the structural conditions that
+// would otherwise make kubelet call os.Exit(1) at provider registration: a
+// configuration directory with no configuration files, a file that does not
+// decode as a CredentialProviderConfig, a file that declares no providers, and a
+// provider name that does not resolve to an executable in the bin directory. It
+// does not replicate kubelet's semantic validation. configKey is the configured
+// value, used only in messages; canonicalConfigPath and canonicalBinDir are the
+// symlink-resolved paths the checks operate on.
+func validateCredentialProviderStructure(configKey, canonicalConfigPath, canonicalBinDir string) error {
```
qJkee, line 306: Should this function check for duplicates too?

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+// generated KubeletConfiguration.
+var kubeletReservedKeys = []string{
+	kubeletImageCredentialProviderConfigPathKey,
+	kubeletImageCredentialProviderBinDirKey,
+}
+
+// statForTrust returns the owning uid and mode of an already symlink-resolved
+// path. It is a package-level variable so tests can exercise the trusted-path
+// ownership rules without running as root.
+var statForTrust = func(path string) (uid uint32, mode os.FileMode, err error) {
```
qJkee, line 66: why this function declared as a var?

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+// to read the credential provider configuration: see the scheme setup in
+// vendor/k8s.io/kubernetes/pkg/credentialprovider/plugin/plugin.go (lines
+// 69-70, 135-138) and decode() in the sibling config.go. Strict decoding
+// rejects unknown fields, and all three API versions kubelet accepts
+// (v1alpha1, v1beta1, v1 of kubelet.config.k8s.io) are registered together with
+// the internal type and its conversions. Building the decoder from the same
+// vendored packages keeps this structural check from diverging from the kubelet
+// compiled into the same binary; a lenient decoder would let a typo'd field
+// through to the os.Exit at registration.
+var credentialProviderCodecs = func() serializer.CodecFactory {
```
qJkee, line 32: This runs at package init, and pkg/config is imported by cmd/microshift/main.go and
  cmd/generate-config — so every process start pays for four AddToScheme calls (the internal
  kubeletconfig type plus v1alpha1/v1beta1/v1, with all their generated conversion and defaulting
  functions), even though the codec is only ever touched when a credential-provider config is
  actually present.

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+// configuration directory with no configuration files, a file that does not
+// decode as a CredentialProviderConfig, a file that declares no providers, and a
+// provider name that does not resolve to an executable in the bin directory. It
+// does not replicate kubelet's semantic validation. configKey is the configured
+// value, used only in messages; canonicalConfigPath and canonicalBinDir are the
+// symlink-resolved paths the checks operate on.
+func validateCredentialProviderStructure(configKey, canonicalConfigPath, canonicalBinDir string) error {
+	prefix := func(err error) error {
+		return fmt.Errorf("error validating kubelet.%s (%q): %w",
+			kubeletImageCredentialProviderConfigPathKey, configKey, err)
```
qJkee, line 309: This function checks for binaries and configs, but logs only `kubeletImageCredentialProviderConfigPathKey`

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+	// (empty config directory, undecodable config, unresolvable provider name)
+	// into ordinary fail-fast configuration errors. configPath (the configured
+	// value) is used only for the error prefix; the filesystem work uses the
+	// canonical paths.
+	if err := validateCredentialProviderStructure(configPath, canonical[0], canonical[1]); err != nil {
+		return err
+	}
+
+	// Store canonical paths only once both keys have passed validation.
+	for i, p := range paths {
```
qJkee, line 206: This mutates Config from inside validate().readKubeletCredentialProviderKeys re-reads the raw strings out of c.Kubelet (config.go:600, inside updateComputedValues), so any later updateComputedValues() call resets these two fields from canonical back to the user-written values — and pkg/node then hands kubelet unresolved symlinked paths, with no error anywhere.

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+			if isNotExistErr(err) {
+				return nil, fmt.Errorf("configuration file %q does not exist (dangling symlink)", entryPath)
+			}
+			return nil, err
+		}
+		info, err := os.Lstat(resolved)
+		if err != nil {
+			return nil, err
+		}
+		if info.IsDir() {
```
qJkee, line 386: Wrong object: this is `info.IsDir()` (the resolved symlink target), but kubelet checks `entry.IsDir()` (the DirEntry). For `b.yaml -> /some/dir` we skip it; kubelet doesn't (DirEntry.IsDir() is false for a symlink), so it os.ReadFile's a directory and hits the os.Exit(1) this function is here to prevent. Use `entry.IsDir()`.

pkg/config/kubelet.go
```diff
@@ -0,0 +1,459 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// validateTrustedChain walks every component of the canonical (already
+// symlink-resolved) path from / to the final object and requires each to be
+// owned by root and not writable by group or others.
+func validateTrustedChain(canonical string) error {
```
qJkee, line 285: Called once per directory entry, and each call re-walks from `/` — so `/`, `/usr`, `/usr/local`...

qJkee (changes requested):

Neilhamza:
### Round-2 smoke validation — fresh EC2 (RHEL 9.6) + real Amazon ECR

Validated the round-2 commit on a fresh host: base built from `upstream/main`, branch RPMs swapped
in, real ECR (us-west-2) via the instance role. **Product behavior: correct in all 9 product cases.**
One test-only assertion defect surfaced in the RF suite and is fixed in `b449d6c3a` (details below).

**Round-2 changes exercised, all correct:**

- **Raw vs canonical split + accessor** — symlinked bin dir resolves to its canonical target for kubelet while `show-config` still reports the raw user path; survives a second `updateComputedValues()`.
- **Real ECR pull** — pod pulled `…/microshift-cred-provider-test:v1` via the instance role with **no** `imagePullSecrets` and nothing added to the CRI-O pull secret.
- **exec-bit provider check** (`os.Stat`+regular+`0o111`) — a missing binary and a `chmod 0644` binary both fail before kubelet, with the message attributed to `imageCredentialProviderBinDir`:

  ```
  error validating kubelet.imageCredentialProviderBinDir ("/usr/libexec/microshift/credential-providers"): provider "no-such-provider" (declared in "/etc/microshift/credential-providers.yaml") has no executable at "/usr/libexec/microshift/credential-providers/no-such-provider"
  ```

- **Duplicate provider name across files** rejected:

  ```
  provider "ecr-credential-provider" is declared more than once (in ".../a.yaml" and ".../b.yaml")
  ```

- **Symlink-to-directory parity** — a `x.yaml` symlink pointing at a directory is rejected rather than silently skipped, and no kubelet CRI-registration failure is reached:

  ```
  configuration file "/etc/microshift" is not a regular file
  ```

  A **real** subdirectory named `x.yaml` is still skipped (kubelet `DirEntry.IsDir()` parity) and startup succeeds.

- **Strict decoder** rejects an unknown field (`providers[0].matchImage`) at config load; empty-dir and single-key cases also error as expected.
- **Lazy codec** — `microshift version` ~0.08s, `show-config --mode default` ~0.10s (sub-second).

**One test-only fix — RF suite `standard2/kubelet-credential-provider.robot` (fixed in `b449d6c3a`).**
The smoke run caught a broken assertion in `Missing Provider Binary Prevents Start`: it failed
deterministically even though the product logs the error correctly (S4/S5 above). The regex wrapped
the provider name with a single-char `.` on each side (`provider .no-such-provider.`), but klog
escapes the inner quotes of `err="…"` as `\"` (two chars), so the journal reads
`provider \"no-such-provider\"` — the single `.` can't span `\"`. Verified on the real journal: the
old pattern → **0** matches, the new `.+` pattern → **10**. It was the only failure-case regex that
wrapped a klog-escaped quoted token this way, which is why the other three passed. Fixed by loosening
those two `.` to `.+` in `b449d6c3a`; the product was already correct. (Rebased onto current `main`
in the same push.)

test/suites/standard2/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    Command Should Work    rm -rf ${CP_BIN_DIR} ${CP_CONFIG_FILE}
+    Restart MicroShift
+    Remove Kubeconfig
+    Logout MicroShift Host
+
+Restart MicroShift With Cursor
+    [Documentation]    Record the journal cursor, then restart MicroShift
+    ${cursor}=    Get Journal Cursor
+    VAR    ${CURSOR}=    ${cursor}    scope=TEST
+    Restart MicroShift
```
Neilhamza: Good point on consistency, but `Set Test Variable` is flagged by this repo's robocop config (DEPR05 — deprecated in favour of `VAR ... scope=TEST`), so `make verify-rf` fails if I switch to it. I kept the `VAR ... scope=TEST` form for that reason. Happy to change repo-wide if we'd rather relax that rule.

test/suites/standard2/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+
+Valid Configuration Applies Kubelet Flags
+    [Documentation]    With both keys set, MicroShift starts, logs the configured paths, reports them in
+    ...    show-config, and keeps them out of the generated KubeletConfiguration
+    [Setup]    Apply Credential Provider Config    ${CP_VALID}
+    Pattern Should Appear In Log Output    ${CURSOR}    ${CP_CONFIGURED_LOG}
+    ${config}=    Show Config    effective
+    Should Be Equal As Strings    ${config.kubelet.imageCredentialProviderConfigPath}    ${CP_CONFIG_FILE}
+    Should Be Equal As Strings    ${config.kubelet.imageCredentialProviderBinDir}    ${CP_BIN_DIR}
+    Command Should Fail    grep -q imageCredentialProvider ${KUBELET_GENERATED_CONFIG}
```
Neilhamza: It covers a different property than the log line / show-config. Those prove the two keys were accepted and surface back to the user; this assertion (`grep -q imageCredentialProvider` against the generated KubeletConfiguration → Should Fail) is the guarantee that they're **stripped** from the generated KubeletConfiguration and handed to kubelet as flags rather than config fields — the whole point of the feature. If a future change let them fall through into the KubeletConfiguration, only this check would catch it. Kept for that reason.

test/suites/configuration1/kubelet-credential-provider.robot
Neilhamza: Fixed — each failure case now asserts a single specific regex against one log line instead of two disconnected `Pattern Should Appear` checks.

One subtlety a smoke run on a real host caught and is worth flagging: klog renders the error as `err="..."` and escapes the inner quotes as `\"` (two characters), so the journal actually reads `provider \"no-such-provider\"`. A single `.` can't span `\"`, so the provider-binary pattern needed `.+` around the name — fixed in `b449d6c3a`. Verified against a real host journal (old pattern → 0 matches, new → the logged line).

test/suites/configuration1/kubelet-credential-provider.robot
Neilhamza: Done — all failure cases now use one specific full-message regex per case (see the detailed reply on the related thread), including the klog `\"`-escaping fix in `b449d6c3a`.

test/suites/configuration1/kubelet-credential-provider.robot
Neilhamza: Done — switched to a single specific full-message pattern per failure case (see the detailed reply on the related thread).

test/suites/configuration1/kubelet-credential-provider.robot
Neilhamza: Agreed — each failure case now matches the full, specific error message, so the expected error is readable directly from the test.

test/suites/standard2/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
...
+    [Teardown]    Run Keywords    Upload String To File    ${CP_PROVIDER_CONFIG}    ${CP_CONFIG_FILE}
+    ...    AND    Remove Credential Provider Config
+    ...    AND    Restart MicroShift
+
+Empty Configuration Directory Prevents Start
+    [Documentation]    MicroShift fails to start when the configuration directory holds no config files
+    [Setup]    Run Keywords    Command Should Work    install -d -o root -g root -m 0755 ${CP_CONFIG_DIR}
+    ...    AND    Apply Invalid Credential Provider Config    ${CP_EMPTY_DIR_CONFIG}
+    Pattern Should Appear In Log Output    ${CURSOR}    contains no .json, .yaml, or .yml
+    [Teardown]    Run Keywords    Command Should Work    rm -rf ${CP_CONFIG_DIR}
```
Neilhamza: Added — the suite Teardown now does a defensive `rm -rf ${CP_BIN_DIR} ${CP_CONFIG_FILE} ${CP_CONFIG_DIR}` so a mid-test failure can't leave fixtures behind, in addition to the per-test teardown.

test/suites/standard2/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,180 @@
+*** Settings ***
```
Neilhamza: Done — moved into `test/suites/standard2/kubelet-credential-provider.robot` via `git mv` (sibling of `dns-custom-config.robot`); the separate suite is gone.

test/suites/configuration1/kubelet-credential-provider.robot
Neilhamza: Applied — `Test Tags    restart    slow`.

pkg/config/kubelet.go
Neilhamza: Right — no longer iterating. It reads the two keys directly out of the `c.Kubelet` map by name via `kubeletStringValue(c.Kubelet, key)`; no loop over all entries.

pkg/config/kubelet.go
Neilhamza: Merged — the absolute-path check and the canonicalization are now a single loop over the two entries.

pkg/config/kubelet.go
Neilhamza: Good catch — replaced `exec.LookPath` with an explicit `os.Stat` on `filepath.Join(binDir, name)` requiring a regular file with an execute bit (`Mode().IsRegular() && Mode()&0o111 != 0`). No `$PATH` search at all now.

pkg/config/kubelet.go
Neilhamza: Added — duplicate provider names are now rejected across all config files (and twice within one file): `provider %q is declared more than once (in %q and %q)`, naming both files.

pkg/config/kubelet.go
Neilhamza: Removed the package-level `var`. Ownership lookup is now a field injected into the trusted-path checker (`newTrustChecker(ownershipFn)`); production passes `lstatOwnership`, tests inject a fake — no mutable package global.

pkg/config/kubelet.go
Neilhamza: Fixed — the codec is now built lazily via `sync.OnceValue`, so it's constructed only on first use (i.e. when a credential-provider config is actually present). Processes that never touch the feature no longer pay the four `AddToScheme` calls at init.

pkg/config/kubelet.go
Neilhamza: Addressed — the message now names the key that actually needs fixing: provider-binary problems (missing / non-executable binary) are attributed to `imageCredentialProviderBinDir` and name the file the provider was declared in, while config-file problems keep the `imageCredentialProviderConfigPath` prefix.

pkg/config/kubelet.go
Neilhamza: Fixed — raw and canonical are now separate fields. The map reader writes only the raw fields; validation writes only the canonical fields (idempotently). `pkg/node` reads them through a new `KubeletImageCredentialProviderPaths()` accessor, so a later `updateComputedValues()` can no longer revert a validated path back to the user-written form.

pkg/config/kubelet.go
Neilhamza: Fixed — `collectCredentialProviderConfigFiles` now skips an entry only when `entry.IsDir()` (the DirEntry) is true, matching kubelet. A symlink named `x.yaml` pointing at a directory is no longer skipped; it's rejected as "not a regular file" before it can reach kubelet's `os.ReadFile` / `os.Exit(1)`. Added a unit test for the symlink-to-dir case.

pkg/config/kubelet.go
Neilhamza: Fixed — the trusted-path checker memoizes verified components, so a shared ancestor (every entry under one bin dir shares `/`, `/usr`, …) is stat'd once per validation instead of once per entry. A unit test asserts the stat count.

test/suites/standard2/kubelet-credential-provider.robot
kasturinarra: got it, make sense to me then.

test/suites/standard2/kubelet-credential-provider.robot
kasturinarra: Agree it's worth testing, but this is asserting an internal implementation detail (flag vs. config field) rather than end-to-end behavior. A unit test on the config generation would cover this more precisely and run faster.

test/suites/standard2/kubelet-credential-provider.robot
```diff
@@ -0,0 +1,188 @@
...
+    [Setup]    Apply Invalid Credential Provider Config    ${CP_MISSING_BIN_DIR}
+    Pattern Should Appear In Log Output    ${CURSOR}    imageCredentialProviderBinDir.*file or directory does not exist
+    [Teardown]    Run Keywords    Remove Credential Provider Config    AND    Restart MicroShift
+
+Only One Key Prevents Start
+    [Documentation]    MicroShift fails to start when only one of the two keys is set
+    [Setup]    Apply Invalid Credential Provider Config    ${CP_ONLY_CONFIG_PATH}
+    Pattern Should Appear In Log Output
+    ...    ${CURSOR}
+    ...    imageCredentialProviderConfigPath and kubelet.imageCredentialProviderBinDir must be set together
```
kasturinarra, line 103: should this be .* instead of . or is this expected ?

Neilhamza: Expected  that's the literal . in kubelet.imageCredentialProviderBinDir, and an unescaped . matches it. Not the \" case from the provider assertion (two chars, hence .* there). Escaped it to kubelet\. for strictness.

test/suites/standard2/kubelet-credential-provider.robot
Neilhamza: Fair — it's an implementation property, already pinned by `Test_GenerateConfig` in `pkg/node/kubelet_test.go`. Removed from the suite in `38c3ebd32`; the log line and `show-config` are the user-visible checks.

test/suites/standard2/kubelet-credential-provider.robot
Neilhamza: For the record, escaped in `38c3ebd32` — the pattern is now `kubelet\.imageCredentialProviderBinDir must be set together`.

pkg/config/kubelet.go
```diff
@@ -0,0 +1,534 @@
...
+
+// validateCredentialProviderStructure verifies the structural conditions that
+// would otherwise make kubelet call os.Exit(1) at provider registration: a
+// configuration directory with no configuration files, a file that does not
+// decode as a CredentialProviderConfig, a file that declares no providers, a
+// provider name declared more than once, and a provider name that does not
+// resolve to an executable in the bin directory. It does not replicate kubelet's
+// semantic validation. configKey and binDirKey are the configured values, used
+// only in messages; the checks operate on the symlink-resolved paths.
+func validateCredentialProviderStructure(configKey, binDirKey, canonicalConfigPath, canonicalBinDir string) error {
```
pacevedom: There is a gap in this function, it does not validate the same required fields as kubelet does. For example, a configuration without apiVersion, matchImages or defaultCacheDuration passes this validation but kubelet wont start properly, making MicroShift crash.
The validation function is not exported in kubelet, so maybe we need to mimic the behavior of the required fields?

pkg/config/kubelet.go
```diff
@@ -0,0 +1,534 @@
...
+		}
+	}
+
+	return canonical, nil
+}
+
+// validateDirEntries applies the trusted-path rule to every entry in dir.
+// Symlinked entries are resolved and the full rule, including the target's
+// ancestors, is applied to the target.
+func validateDirEntries(dir string, checker *trustChecker) error {
```
pacevedom: There is a caveat in this function. If the dir contains anything that is not owned by root then MicroShift wont start, when that is not really an issue. Validation should only care about json, yaml and yml files because that is what the `readCredentialProviderConfig` consumes, ignoring everything else.
Can we filter that out here too to mimic the behavior from the kubelet, as that function is not exported either?

pkg/config/kubelet.go
Neilhamza: Agreed — that was the residual I'd documented rather than closed. Everything kubelet's `validateCredentialProviderConfig` relies on is exported (`ParseSchemelessURL`, `IsQualifiedName`, the `SchemeGroupVersion` and cache-type constants), so `validateCredentialProviderSemantics` now mirrors it rule for rule, same order and messages, with a note to re-check at each rebase. The only rule not mirrored is the `KubeletServiceAccountTokenForCredentialProviders` gate check on `tokenAttributes`, since the gate's value at registration comes from the kubelet passthrough; that's called out in the code. Unit tests cover each rule including your example (a provider with only `name` now fails with the three `Required value` errors), plus a new RF case for a missing `matchImages`. 7dd710c60.

pkg/config/kubelet.go
Neilhamza: Yes, that was stricter than needed. It now checks only what kubelet consumes: in a config directory the non-directory `.json`/`.yaml`/`.yml` entries (one shared predicate `isCredentialProviderConfigEntry`, used by both the trusted-path walk and the file collection so they can't diverge), and for the bin dir the declared provider binaries themselves instead of every entry. Both directories still have to be root-owned and not group/other-writable, and each declared binary is symlink-resolved and checked at its target with ancestors. `howto_config.md` updated to match. 7dd710c60.

Neilhamza:
Round-4 review addressed in `7dd710c60` (branch rebased on `main`). Both of @pacevedom's comments on `pkg/config/kubelet.go` are fixed:

- **Required-field validation gap:** `validateCredentialProviderSemantics` now mirrors kubelet's unexported `validateCredentialProviderConfig` rule for rule (same order and messages), so a config missing `apiVersion`/`matchImages`/`defaultCacheDuration` (or with an invalid name, unparseable match image, or bad `tokenAttributes`) is rejected up front instead of making kubelet `os.Exit(1)`. The only rule not mirrored is the `KubeletServiceAccountTokenForCredentialProviders` gate check, which depends on the kubelet passthrough at registration — documented in the code.
- **Over-strict directory validation:** trusted-path checks now cover only what kubelet consumes — `.json`/`.yaml`/`.yml` files in a config directory (shared predicate with the file collection) and the declared provider binaries in the bin dir, rather than every entry. Both directories are still required root-owned and non-writable.

Unit tests add one case per semantic rule plus the only-`name` example from the review; a new RF case covers a config missing `matchImages`. Docs and the regenerated config are updated. Re-requested review from @pacevedom, @qJkee and @kasturinarra.

pacevedom (approved):
/lgtm

PR #7327, opened 2026-09-04
[release-5.0] NO-ISSUE: Test and script fixes found during 5.0.0-rc.0 testing

agullon:
## Summary

Backport of #7326 to `release-5.0`. A consolidation of test-harness and CI fixes discovered while running the 5.0.0-rc.0 release test suites. All changes are test/CI-only (no product code): they address CI flakes and failures — resource exhaustion, teardown cascades, log-scan false positives, timeout limits, and asynchronous-resource races.

Parent PR (main): openshift/microshift#7326 — the commits here are identical to that PR, rebased onto `release-5.0`.

## Changes

- **Namespace teardowns**: `Remove Namespace` now deletes with `allow_fail=True` (matching the existing repo idiom) and polls until the namespace is actually gone — asserting a real `NotFound` — so a slow deletion killed at the 300s process timeout no longer cascades into a suite-teardown failure, while still guaranteeing the name is free for reuse.
- **Release scenario timeouts**: centralize `GREENBOOT_TIMEOUT=1200` and `TEST_EXECUTION_TIMEOUT=60m` in `ci_phase_boot_and_test.sh` for release scenarios and remove the now-redundant per-scenario overrides.
- **ginkgo VM disk**: restore `--vm_disksize 30` for the ginkgo scenario, whose storage specs exhaust the topolvm VG at the 20GB default.
- **log-scan cert-manager race**: allowlist the benign, transient cert-manager ServiceAccount "forbidden" startup race in the standard2 Log Scan test (a scoped, namespace-specific regex; genuine `forbidden` regressions still fail), and clarify the double-boot test with a `Boot And Scan Journal` keyword.
- **journald rate limiting**: disable journald rate limiting for the logging suite (drop-in wrapped in `bash -c` for correct sudo redirection, with `mkdir -p` for self-containment) so high-verbosity log output isn't dropped.
- **hostname test**: use the reserved `.example` TLD instead of `.local` to avoid mDNS interference.
- **flaky async-resource waits**: `statefulset-pvc` waits for the StatefulSet pod to exist before the readiness check, and `ai-model-serving-online` retries the kserve `ServingRuntime` apply until the validating webhook has endpoints.

## Testing

- Robot Framework suites lint clean (robocop check + format).
- Python helper (`journalctl.py`) passes flake8; shell scripts pass shellcheck.
- Validated against the 5.0.0-rc.0 release jobs where these failures were originally observed (arm/x86 el9/el10 release scenarios).

PR #7326, opened 2026-09-04
NO-ISSUE: Test and script fixes found during 5.0.0-rc.0 testing

agullon:
## Summary

A consolidation of test-harness and CI fixes discovered while running the 5.0.0-rc.0 release test suites. All changes are test/CI-only (no product code): they address CI flakes and failures — resource exhaustion, teardown cascades, log-scan false positives, timeout limits, and asynchronous-resource races.

## Changes

- **Namespace teardowns**: `Remove Namespace` now deletes with `allow_fail=True` (matching the existing repo idiom) and polls until the namespace is actually gone — asserting a real `NotFound` — so a slow deletion killed at the 300s process timeout no longer cascades into a suite-teardown failure, while still guaranteeing the name is free for reuse.
- **Release scenario timeouts**: centralize `GREENBOOT_TIMEOUT=1200` and `TEST_EXECUTION_TIMEOUT=60m` in `ci_phase_boot_and_test.sh` for release scenarios and remove the now-redundant per-scenario overrides.
- **ginkgo VM disk**: restore `--vm_disksize 30` for the ginkgo scenario, whose storage specs exhaust the topolvm VG at the 20GB default.
- **log-scan cert-manager race**: allowlist the benign, transient cert-manager ServiceAccount "forbidden" startup race in the standard2 Log Scan test (a scoped, namespace-specific regex; genuine `forbidden` regressions still fail), and clarify the double-boot test with a `Boot And Scan Journal` keyword.
- **journald rate limiting**: disable journald rate limiting for the logging suite (drop-in wrapped in `bash -c` for correct sudo redirection, with `mkdir -p` for self-containment) so high-verbosity log output isn't dropped.
- **hostname test**: use the reserved `.example` TLD instead of `.local` to avoid mDNS interference.
- **flaky async-resource waits**: `statefulset-pvc` waits for the StatefulSet pod to exist before the readiness check, and `ai-model-serving-online` retries the kserve `ServingRuntime` apply until the validating webhook has endpoints.

## Testing

- Robot Framework suites lint clean (robocop check + format).
- Python helper (`journalctl.py`) passes flake8; shell scripts pass shellcheck.
- Validated against the 5.0.0-rc.0 release jobs where these failures were originally observed (arm/x86 el9/el10 release scenarios).

PR #7363, opened 2026-09-10
NO-ISSUE: align statefulset/log-scan tests with the main-branch versions

agullon:
## Summary

The test/script consolidation PR merged to `release-5.0` (#7327) before two follow-up refinements landed on its `main`-branch equivalent (#7326). This brings those two refinements to `release-5.0` so the branches stay consistent. Test-only.

## Changes

- **`otp-workloads/statefulset-pvc.robot`** — wait for the StatefulSet via `microshift healthcheck --namespace ${NAMESPACE} --statefulsets hello-statefulset --timeout 5m` instead of the `oc wait` on a named pod (`Wait Until Resource Exists` + `Named Pod Should Be Ready`). Cleaner and uses the product healthcheck directly (`--statefulsets` is supported on release-5.0, `pkg/cmd/healthcheck.go`). Renames `${POD_NAME}` → `${STATEFULSET_NAME}`.
- **`standard2/log-scan.robot`** — rename `Boot And Scan Journal` → `Service Startup And Scan Journal` and `Scan Boot Journal` → `Scan Service Startup Journal` (plus wording), since the test restarts the MicroShift service rather than rebooting the host.

## Testing

- robocop check + format clean on both suites.

## Notes

Backport of the delta between #7326 (main) and the already-merged #7327 (release-5.0).

### DIFF
--- a/test/suites/otp-workloads/statefulset-pvc.robot
+++ b/test/suites/otp-workloads/statefulset-pvc.robot
@@ -13,8 +13,8 @@ Suite Teardown      Teardown Suite
 
 *** Variables ***
 ${STATEFULSET_YAML}     ./assets/otp-workloads/statefulset-pvc.yaml
+${STATEFULSET_NAME}     hello-statefulset
 ${PVC_NAME}             www-hello-statefulset-0
-${POD_NAME}             hello-statefulset-0
 
 
 *** Test Cases ***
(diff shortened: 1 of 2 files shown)


pmtk:
a lot of tests are failing because of the ovnk stale flows bug, we need https://github.com/openshift/microshift/pull/7401 to unblock

PR #7364, opened 2026-09-10
USHIFT-7432: Fix OVN multinode networking regression from removal of --sb-address flag

eslutsky:
## Summary

Fixes broken OVN multinode pod networking introduced by PR #6912 (June 2026), which removed \`--nb-address\`/\`--sb-address\` from \`ovnkube --init-node\` during a nightly rebase. Without those flags, worker nodes fall back to local unix sockets that don't exist on workers, causing each node to form an isolated OVN RAFT cluster — no geneve tunnels form and all cross-node pod traffic is silently dropped.

## Root cause

\`commit 8a0f4f23eb\` removed deprecated CLI flags that were the sole mechanism for pointing worker nodes at the primary's NB/SB databases:

\`\`\`diff
-  --nb-address "tcp:PRIMARY_IP:9641"
-  --sb-address "tcp:PRIMARY_IP:9642"
\`\`\`

The upstream OVN-K design intends the config file (\`[OvnNorth]\`/\`[OvnSouth]\` stanzas in \`ovnkube.conf\`) to replace these flags. This PR implements that design in MicroShift.

**Note**: the \`ovnkube\` binary in the current container image does not yet parse \`address=\` from the \`[OvnNorth]\`/\`[OvnSouth]\` INI sections correctly (gcfg error: \`can't store data at section "OvnSouth", variable "address"\`). A fix is needed upstream in OVN-K to restore \`Address string \\\`gcfg:"address"\\\`\` in \`OvnAuthConfig\`, so the config-file mechanism actually works end-to-end for workers.

### 1. Label primary node / restrict master DaemonSet

Both nodes carry \`node-role.kubernetes.io/master\`, so \`ovnkube-master\` (which includes the full SBDB/NBDB/northd stack) was scheduled on every node. Each node started its own isolated OVN RAFT cluster with no knowledge of the other.

- \`pkg/node/kubelet.go\`: primary gets label \`node.microshift.io/role=primary\` at kubelet startup (detected by absence of bootstrap kubeconfig)
- \`assets/components/ovn/multi-node/master/daemonset.yaml\`: nodeSelector changed to \`node.microshift.io/role: primary\`

### 2. Add SBDB/NBDB addresses to \`ovnkube.conf\` and guard worker writes

- \`assets/components/ovn/common/configmap.yaml\`: adds \`[OvnNorth]\`/\`[OvnSouth]\` stanzas with the primary's TCP NB/SB addresses in multinode mode — the replacement for the removed CLI flags
- \`pkg/components/networking.go\`:
  - Worker nodes only deploy \`node/daemonset.yaml\`; primary deploys master + node (workers no longer run a local SBDB/NBDB isolated from the primary)
  - Only the primary writes the \`ovnkube-config\` ConfigMap — a worker writing it would overwrite the correct primary IP with its own
  - Passes \`MultiNodeEnabled\` render param so the configmap template can conditionally emit the stanzas

## Test plan

- [ ] Verify \`ovnkube-master\` pod runs only on the primary node in multinode mode
- [ ] Verify geneve tunnels form between all nodes: \`ovs-vsctl show | grep geneve\`
- [ ] Verify cross-node pod communication: ping between pods on different nodes
- [ ] Run \`el98-src@cncf-conformance\` scenario

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pkg/node/kubelet.go
+++ b/pkg/node/kubelet.go
@@ -89,6 +89,9 @@ func (s *KubeletServer) configure(cfg *config.Config) {
 	kubeletFlags.NodeLabels["node-role.kubernetes.io/worker"] = ""
 	kubeletFlags.NodeLabels["node.openshift.io/os_id"] = osID
 	kubeletFlags.NodeLabels["node.kubernetes.io/instance-type"] = "rhde"
+	if !cfg.BootstrapKubeConfigExists() {
+		kubeletFlags.NodeLabels["node.microshift.io/role"] = "primary"
+	}
 
 	kubeletConfig, err := loadConfigFile(filepath.Join(config.DataDir, "/resources/kubelet/config/config.yaml"))
 
--- a/assets/components/ovn/common/configmap.yaml
+++ b/assets/components/ovn/common/configmap.yaml
@@ -34,3 +34,11 @@ data:
     election-lease-duration=137
     election-renew-deadline=107
     election-retry-period=26
+{{- if .MultiNodeEnabled}}
+
+    [OvnNorth]
+    address=tcp:{{.NodeIP}}:{{.OVN_NB_PORT}}
+
+    [OvnSouth]
+    address=tcp:{{.NodeIP}}:{{.OVN_SB_PORT}}
+{{- end}}
--- a/assets/components/ovn/multi-node/master/daemonset.yaml
+++ b/assets/components/ovn/multi-node/master/daemonset.yaml
@@ -141,6 +141,19 @@ spec:
             --db-nb-cluster-local-proto=tcp \
             --no-monitor"
 
+          # On worker nodes the configmap contains [OvnNorth] address= pointing to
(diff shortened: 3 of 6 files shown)


pkg/node/kubelet.go
```diff
@@ -89,6 +89,9 @@ func (s *KubeletServer) configure(cfg *config.Config) {
 	kubeletFlags.NodeLabels["node-role.kubernetes.io/worker"] = ""
 	kubeletFlags.NodeLabels["node.openshift.io/os_id"] = osID
 	kubeletFlags.NodeLabels["node.kubernetes.io/instance-type"] = "rhde"
+	if !cfg.BootstrapKubeConfigExists() {
```
pacevedom, line 92: So should we gate this with the multinode flag too so we avoid the label appearing in single node deployments? This way we do not change the semantics of labels from prior versions.

